PR #10, opened 2019-07-15
scheduler stateful restart

yangwwei:


yangwwei:
hi @wilfred-s 

Thanks for helping to review this. Your comments make sense to me. Respectively, please see

>We already have a schedulerevent handling setup (for update and delete partition) which is not linked to current state machine. Why disconnect the two?

Good point, I just modified the patch to make it align. Will explain the detail next.

> Then there are two other events in the package schedulerevent already that do not affect the scheduler but are scheduling events, we need to think about that as it confusing, might make more sense to rename

I've done a bit more than just renaming. The reason behind this is, put stateMachine under `scheduler.go` was not enough. The stateMachine should manage yunikorn-core state that covers all services, which includes `cache`, `scheduler` and `rmProxy`. You'll see in my 2nd commit, I moved statemachine to the top-level package (same as rest services). With this change, all services should be aware of which state currently scheduler is, in order to do corresponding actions. E.g scheduler should not handle new requests before recovery is done.

>How does the recovery and the scheduler state hook into the shims and their registration or unregistration? I would expect an exposed state that would allow shims to register (beside a running state). Both shim and core in recovery at the same time does not make sense. Especially looking forward and making sure this works with multiple shims.

please refer to the [design doc](https://github.com/cloudera/yunikorn-core/blob/master/docs/design/resilience.md), see shim state chart. In recovery mode, after shim registers itself with yunikorn-core, it then starts to `recovery` too. In this mode, it doesn't accept new app or allocate new pods, it just collects all allocated apps/allocations and send them back to core. Once core is fully recovered, then shim can exit the `recovering` state and get back to work. Do you still think having them both in recovery mode doesn't make sense?

>If nodes trigger the recovery of applications etc as per the doc, what does the recovery of the scheduler do? The config is already loaded

Scheduler needs to collect all allocated pods info, and deduct partition/queue resources. it's kind of replay of previous allocations.

> the flag for recovery is not using the same naming convention as the other flags used.

Fixed

> in the cache we have defined the events, states and machine in one file. that keeps things easy to maintain and in sync.

This is because those are just events, no states. But with adding statemachine, we have both events and states. You don't want to put states into a file like `events.go`.

> It is not clear what we do on stop.

This is now mostly used in the test case, to determine if the scheduler is stopped. Used in `statemachine_test.go`

wilfred-s:
The main issue I have at this point is that this does not work if I have more than 1 shim. I cannot pull the core back into a recovery mode when I already have a shim that is scheduling.

These are two of the case we do not handle and instead of redesigning this we at least need to be able to handle this:
1) multi shim on startup
* state of the core is: _new_
* shim one registers: scheduler goes to _recovering_
* shim two registers: scheduler is _recovering_, core cannot go to same state again (panic ?)
* shim two finishes recovery success event comes in and takes the scheduler to _running_

2) additional shim while running
* state of the core is: _running_, shim one is running
* shim two registers: scheduler is _running_, core cannot go to previous state again (panic ?)
* shim two is linked to the scheduler in _running_ state

We should also be able to handle in the core states an unregister of a shim. I cannot understand how we would handle the "Lifecycle of RM-Scheduler communication" as defined in the scheduler interface.

In all cases we have a problem. I think we need a different approach in which we redefine the states for the core and how the shims influence that state.

sunilgovind:
regarding state management, could we discuss in a weekly call.

yangwwei:
I am currently investigating to just have states in the shim layer, and scheduler core just remains have no notion of scheduler state. This seems to be a simpler design.

yangwwei:
Discard this PR, and move to #20

PR #19, opened 2019-07-29
Example updates for volumes (EBS added)

wilfred-s:
New demo video added for volume scheduling. 
Changes are also added to the k8shim to support these examples

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -127,10 +127,13 @@ kubectl create -f examples/predicates/pod-anti-affinity-example.yaml
 This deployment ensures 2 pods cannot be co-located together on same node.
 If this yaml is deployed on 1 node cluster, expect 1 pod to be started and the other pod should stay in a pending state.
 
-### A volume example
-There are two examples with volumes available. The NFS example does not work on docker desktop and requires [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/).
+### Volume examples
+There are three examples with volumes available. The NFS example does not work on docker desktop and requires [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/). 
+The EBS volume requires a kubernetes cluster running on AWS (EKS). 
 
-CAUTION: Both examples will generate an unending stream of data in a file called `dates.txt` on the mounted volume. This could cause a disk to fill up and execution time should be limited. 
+CAUTION: All examples will generate an unending stream of data in a file called `dates.txt` on the mounted volume. This could cause a disk to fill up and execution time should be limited. 
+
+#### Local volume
 * create the local volume and volume claim
 ```
 kubectl create -f examples/volume/local-pv.yaml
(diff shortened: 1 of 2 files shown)


docs/user-guide.md
```diff
@@ -139,3 +142,42 @@ kubectl create -f examples/volume/local-pv.yaml
 ```
 kubectl create -f examples/volume/pod-local.yaml
 ```
+
+#### NFS volume
+* create the NFS server
```
yangwwei: can we add a link to `nfs-server.yaml` in k8s-shim project?

docs/user-guide.md
```diff
@@ -139,3 +142,42 @@ kubectl create -f examples/volume/local-pv.yaml
...
+```
+kubectl create -f nfs-server.yaml
+```
+* get the IP address for the NFS server and update the pod yaml
+```
+kubectl get services
```
yangwwei: can we put some copy-to-run commands here?
for the update, we can say replace the server name of `nfs-volume` in `pod-nfs.yaml` (also add a link here) with the nfs-server address got from upon command.

docs/user-guide.md
```diff
@@ -139,3 +142,42 @@ kubectl create -f examples/volume/local-pv.yaml
...
+```
+aws ec2 create-volume --volume-type gp2 --size 10 --availability-zone us-west-1
+```
+The `VolumeId` is part of the returned information of the create command.
+
+* create the pod that uses a direct volume reference:
+```
+kubectl create -f pod-ebs-direct.yaml
+```
+* create the volume (pv) and a pod that uses a pvc to claim the existing volume:
```
yangwwei: volume -> persistent volume
a pvc -> persistent volume claim (pvc)

it will also be good if we can link to Kubernetes PV/PVC document for them

docs/user-guide.md
```diff
@@ -139,3 +142,42 @@ kubectl create -f examples/volume/local-pv.yaml
...
+```
+kubectl create -f ebs-pv.yaml
+kubectl create -f pod-ebs-exist.yaml
+```
+* create a storage class to allow auto provisioning and create the pod that uses the mechanism:
```
yangwwei: I suppose this is referring to `dynamical provisioning`? Shall we rename "auto" to "dynamical" if this is the case.

yangwwei:
Looks like you have more detailed docs at https://github.com/cloudera/yunikorn-k8shim/commit/b20dad54997f687cd03250f4fd0f79a2571dd714#diff-0f2249ec0884d373546ccf0a3a2d6d12. Maybe you can just link to that doc here instead of duplicating those details. Thx

docs/user-guide.md
wilfred-s: updated

docs/user-guide.md
wilfred-s: updated

docs/user-guide.md
wilfred-s: updated

wilfred-s: updated

PR #20, opened 2019-07-29
recover scheduler state from previous allocated allocations

yangwwei:
A couple of changes since last time

- removed `recovering` state from yunikorn-core
- let the core scheduler stay silly without maintaining a state, it just handles requests whatever a shim sends
- we'll add the `recovering` state in shim, so it only sends existing allocations to core under this state, new pending pods will be waiting after it exits the recovering state (recovery done).

### DIFF
--- a/pkg/cache/cluster_info.go
+++ b/pkg/cache/cluster_info.go
@@ -247,6 +247,7 @@ func (m *ClusterInfo) processNodeUpdate(request *si.UpdateRequest) {
     if len(request.NewSchedulableNodes) > 0 {
         acceptedNodes := make([]*si.AcceptedNode, 0)
         rejectedNodes := make([]*si.RejectedNode, 0)
+        existingAllocations := make([]*si.Allocation, 0)
         for _, node := range request.NewSchedulableNodes {
             nodeInfo, err := NewNodeInfo(node)
             if err != nil {
@@ -265,6 +266,7 @@ func (m *ClusterInfo) processNodeUpdate(request *si.UpdateRequest) {
                         zap.String("nodeId", node.NodeId),
                         zap.String("partition", nodeInfo.Partition))
                     acceptedNodes = append(acceptedNodes, &si.AcceptedNode{NodeId: node.NodeId})
+                    existingAllocations = append(existingAllocations, node.ExistingAllocations...)
                     continue
                 } else {
                     errorMessage := fmt.Sprintf("Failure while adding new node, rejected the node, error=%s", err)
@@ -286,6 +288,12 @@ func (m *ClusterInfo) processNodeUpdate(request *si.UpdateRequest) {
             RMId:          request.RmId,
             AcceptedNodes: acceptedNodes,
             RejectedNodes: rejectedNodes,
+        })
+
+        // notify the scheduler to recover existing allocations
+        m.EventHandlers.SchedulerEventHandler.HandleEvent(&schedulerevent.SchedulerAllocationUpdatesEvent{
+            ExistingAllocations: existingAllocations,
+            RMId:                request.RmId,
         })
     }
 }
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -227,7 +227,59 @@ func (m *Scheduler) processAllocationReleaseByAllocationKey(allocationAsksToRele
     }
 }
 
+func (m *Scheduler) recoverExistingAllocations(existingAllocations []*si.Allocation, rmId string) {
+    // the recovering of existing allocations looks like a replay of the scheduling process,
+    // in general, a scheduling process takes following steps
+    //  1) add scheduling info to schedulers cache, including app, pending requests etc
+    //  2) Look order partition -> queue -> app
+    //  3) For pending asks of a app, scheduler runs scheduling logic to propose a node
+    //  4) Process the proposal generated in step 3, update data in cache
+    // the recovery is repeating all steps except step 2
+    for _, alloc := range existingAllocations {
+        log.Logger.Info("recovering allocations for app",
+            zap.String("applicationId", alloc.ApplicationId),
+            zap.String("nodeId", alloc.NodeId),
+            zap.String("queueName", alloc.QueueName),
+            zap.String("partition", alloc.Partition),
+            zap.String("allocationId", alloc.Uuid))
+
+        // add scheduling asks
+        schedulingAsk := ConvertFromAllocation(alloc, rmId)
+        if err := m.updateSchedulingRequest(schedulingAsk); err != nil {
+            log.Logger.Warn("failed...", zap.Error(err))
+        }
+
+        // handle allocation proposals
+        if err := m.updateSchedulingRequestPendingAskByDelta(&commonevents.AllocationProposal{
+            NodeId:            alloc.NodeId,
+            ApplicationId:     alloc.ApplicationId,
+            QueueName:         alloc.QueueName,
+            AllocatedResource: resources.NewResourceFromProto(alloc.ResourcePerAlloc),
+            AllocationKey:     alloc.AllocationKey,
+            Tags:              alloc.AllocationTags,
+            Priority:          alloc.Priority,
+            PartitionName:     common.GetNormalizedPartitionName(alloc.Partition, rmId),
+        }, -1); err != nil {
+            log.Logger.Error("failed to increase pending ask",
+                zap.Error(err))
+        }
+    }
+}
+
 func (m *Scheduler) processAllocationUpdateEvent(ev *schedulerevent.SchedulerAllocationUpdatesEvent) {
+    if len(ev.ExistingAllocations) > 0 {
+        // in recovery mode, we only expect existing allocations being reported
+        if len(ev.NewAsks) > 0 || len(ev.RejectedAllocations) > 0 || ev.ToReleases != nil{
+            log.Logger.Warn("illegal SchedulerAllocationUpdatesEvent," +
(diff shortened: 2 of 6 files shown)


yangwwei:
this is supposed to fix #10

yangwwei:
@wangdatan , @wilfred-s please help to review.
This is core side changes to recover states from existing allocations, based on discussions, I am trying to keep scheduler-core logic as simple as possible. Moving recovering state to shim. Note, recovering state is required because we need to get scheduler fully recovered before scheduling new requests, otherwise the result will be messed up. More detailed design can be found at https://github.com/cloudera/yunikorn-core/blob/master/docs/design/resilience.md. Thx

pkg/scheduler/scheduler.go
```diff
@@ -227,7 +227,59 @@ func (m *Scheduler) processAllocationReleaseByAllocationKey(allocationAsksToRele
...
+        log.Logger.Info("recovering allocations for app",
+            zap.String("applicationId", alloc.ApplicationId),
+            zap.String("nodeId", alloc.NodeId),
+            zap.String("queueName", alloc.QueueName),
+            zap.String("partition", alloc.Partition),
+            zap.String("allocationId", alloc.Uuid))
+
+        // add scheduling asks
+        schedulingAsk := ConvertFromAllocation(alloc, rmId)
+        if err := m.updateSchedulingRequest(schedulingAsk); err != nil {
```
wilfred-s: How do we know that the application is already added to the cache?
I don't see the application being added to the cache on recovery

pkg/cache/cluster_info.go
```diff
@@ -265,6 +266,7 @@ func (m *ClusterInfo) processNodeUpdate(request *si.UpdateRequest) {
                         zap.String("nodeId", node.NodeId),
                         zap.String("partition", nodeInfo.Partition))
                     acceptedNodes = append(acceptedNodes, &si.AcceptedNode{NodeId: node.NodeId})
+                    existingAllocations = append(existingAllocations, node.ExistingAllocations...)
```
wilfred-s: I have one question which I can not find in the changes: how does this interact with the node registration? When a node registers it also provides a list of existing allocations which get added to the PartitionInfo in the cache when we call partition.addNewNode 6 lines above here through this:
'''
    // Add allocations that exist on the node when added
    if len(existingAllocations) > 0 {
        log.Logger.Info("add existing allocations", ...
  ...
'''
Since we are replaying the allocation already should we still have this?
How do we handle a node re-registration outside of the shim recovery which is what we do now as far as I am aware.

pkg/scheduler/tests/scheduler_recovery_test.go
```diff
@@ -0,0 +1,398 @@
...
+	"github.com/cloudera/yunikorn-core/pkg/common/configs"
+	"github.com/cloudera/yunikorn-core/pkg/common/resources"
+	"github.com/cloudera/yunikorn-core/pkg/entrypoint"
+	"github.com/cloudera/yunikorn-scheduler-interface/lib/go/si"
+	"gotest.tools/assert"
+	"testing"
+)
+
+// Test basic interactions from rm proxy to cache and to scheduler.
+func TestSchedulerRecovery(t *testing.T) {
```
wilfred-s: This a good case test. Would be good to test the bad scenarios too.
what happens if a partition, queue or app does not exist ?

pkg/scheduler/scheduler.go
yangwwei: It relies on shim to provide info of existing applications. When shim is in recovery mode, it scans all existing apps, and simply send to scheduler-core as new applications, e.g

```
&si.UpdateRequest{
    ....,
    NewApplications: []*si.AddApplicationRequest{
			{
				ApplicationId: "app-1",
				QueueName:     "root.a",
				PartitionName: "",
			},
		},
}
```
when core gets these info, it adds application to cache just like a fresh submitted app.

pkg/scheduler/tests/scheduler_recovery_test.go
yangwwei: This is a good suggestion, let me see how to add some negative tests. Thank you.

pkg/cache/cluster_info.go
yangwwei: The code you pasted upon was the place where we handle existing allocations in the cluster-level cache, e.g `PartitionInfo`, `QueueInfo`. The reason to collect `existingAllocations` here is to send this info to the scheduler so the scheduler can update its cache, aka `SchedulingQueue`, `SchedulingApplication` etc.

yangwwei: For node re-registration, I don't think we allow that.
```
 if node := pi.nodes[node.NodeId]; node != nil {
        return fmt.Errorf("partition %s has an existing node %s, node name must be unique", pi.Name, node.NodeId)
    }
```
if shim sends 2 same node as NewNodeInfo in UpdateRequest, the 2nd one will be rejected.

wilfred-s: My concern is that we can have a loop back: update in one place (cache or scheduler) which generates an event to update the other side which we already did update. I am not afraid about the same node registering twice we have that case covered.

As long as we're sure that there is no feedback loop or we handle it correctly by breaking the loop then I am OK. Just not 100% sure that we have coded that defensively.

pkg/scheduler/scheduler.go
wilfred-s: Can I propose the following then:
- Add to the design that we make sure the shim sends the application first (full scan).
- The queue for events must not have any new application events waiting when we start processing the allocations and  or node additions. If we do there is a big chance of failures.

wangdatan:
Thanks @yangwwei for the patch and design doc.

Spent some time today to think about the problems we need to solve. 

I think general solution of the design seems correct, however there're some comments about the proposed workflows:  

1) For YuniKorn-shim, it should be always recover state from etcd. I think we should not have a launch flag to indicate it is recovery or not. 
2) So there should have no direct arc from Registered to Running state.

wangdatan:
Also, not directly related to recovery, I think we haven't well-considered multi-scheduler co-existing in the picture. 

There're 3 kinds of combinations: 

1) YuniKorn is the only scheduler in the cluster. 
2) Multiple schedulers in the cluster, YuniKorn is one of them, but YuniKorn doesn't share node with other schedulers.  
3) Multiple schedulers in the cluster, YuniKorn is one of them, and YuniKorn shares node with other schedulers.

For #1, it is already considered as done. 

For #2, the problem is, all nodes will be added to YuniKorn no matter if it is share-able with other schedulers or not. I think shim layer should handle this. Maybe what we can do is introduce a special taint like "YUNIKORN_EXCLUSIVE" and use that to whitelist nodes which will be used by YuniKorn. And shim only sends such nodes to YuniKorn. 

For #3, it is a bit tricky because race condition could happen when YuniKorn assign an allocation and bind to a node, and at the same time, another scheduler did the same thing. (Question here, will bind operation fail by K8s if issues like reserved resource of a node exceed total_available happens?). There're two solutions: 

a. Report an allocated resource on the node if there's an allocation from another scheduler. Con of the solution is YuniKorn is hard to get resource accounting correct. (Such as which queue the used resource should be accounted to). 
b. Report a change of total node resource. Con of the solution is this can lead to (rapid) changes of total cluster resource in YuniKorn, cause confusions to end-users.  

So I prefer to only do #2 for now, seems like low-hanging fruit and fills an important need.

yangwwei:
@wangdatan  thanks for the comments.

> For YuniKorn-shim, it should be always recover state from etcd. I think we should not have a launch flag to indicate it is recovery or not.

I think we can do that. Just need to confirm if the list node or pod API returns a complete list of nodes and pods on this cluster. No matter what's the current state. E.g node is lost. One thing bothering me was how we know when the recovery is finished.

>I think we haven't well-considered multi-scheduler co-existing in the picture.

I suggest moving the problem of how to support co-locate with other schedulers to another thread. Kubelet has safeguards before binding a pod to a node, that runs a bunch of predicates, similar to the allocation process. Node resource must be checked. For #2 you mentioned above, I like the idea of using taints, that sounds like a user-side-zero-effort approach. Just need a [admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#extendedresourcetoleration) might get this work.

pkg/scheduler/scheduler.go
yangwwei: for your proposed points
> Add to the design that we make sure the shim sends the application first (full scan)

Yes, we'll make sure shim sends existing apps and allocations first. We'll do a full node-scan with node lister api. I'll add this to the design.

> The queue for events must not have any new application events waiting when we start processing the allocations and or node additions. If we don't there is a big chance of failures.

We will also ensure this from the shim layer.

yangwwei:
@wilfred-s does this PR look OK now?  do you want to approve this PR for me?

pkg/scheduler/scheduler.go
wilfred-s: OK, will wait for the shim changes to confirm this
with that pre-requisite I am OK

pkg/scheduler/tests/scheduler_recovery_test.go
wilfred-s: newly added tests cover this

pkg/cache/cluster_info.go
wilfred-s: Based on off-line conversations there should not be a loop as it breaks on the fact that an ID can only be added once and the loop will stop.

wilfred-s:
resolves #10 on the core side

PR #24, opened 2019-08-07
add UT to cover the case of recovering apps before nodes

yangwwei:
During recovery, we need to send apps first before nodes. This UT is to add coverage for that.

### DIFF
--- a/pkg/scheduler/tests/scheduler_recovery_test.go
+++ b/pkg/scheduler/tests/scheduler_recovery_test.go
@@ -668,3 +668,64 @@ partitions:
 	assert.Equal(t, app01.ApplicationInfo.ApplicationId, "app-1")
 	assert.Equal(t, app01.ApplicationInfo.QueueName, "root.a")
 }
+
+// test scheduler recovery that only registers nodes and apps
+func TestAppRecoveryAlone(t *testing.T) {
+	serviceContext := entrypoint.StartAllServicesWithManualScheduler()
+	proxy := serviceContext.RMProxy
+
+	// Register RM
+	configData := `
+partitions:
+  -
+    name: default
+    queues:
+      - name: root
+        queues:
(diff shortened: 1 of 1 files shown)


pkg/scheduler/tests/scheduler_recovery_test.go
```diff
@@ -667,4 +667,65 @@ partitions:
 		GetSchedulingApplication("app-1", "[rm:123]default")
 	assert.Equal(t, app01.ApplicationInfo.ApplicationId, "app-1")
 	assert.Equal(t, app01.ApplicationInfo.QueueName, "root.a")
+}
+
+// test scheduler recovery that only registers nodes and apps
```
wilfred-s: This should be app only

yangwwei: Yeah, let me fix it. Thx

pkg/scheduler/tests/scheduler_recovery_test.go
```diff
@@ -667,4 +667,65 @@ partitions:
...
+		t.Error(err.Error())
+	}
+
+	// Register apps alone
+	err = proxy.Update(&si.UpdateRequest{
+		NewApplications: []*si.AddApplicationRequest{
+			{
+				ApplicationId: "app-1",
+				QueueName:     "root.a",
+				PartitionName: "",
```
sunilgovind: Do we need to specify resource request also here, such as memory or cpu ?

yangwwei: Nope, this is not needed for registering an app per scheduler-interface. See more at https://github.com/cloudera/yunikorn-scheduler-interface/blob/d992f6689cd7a4b81f750f7b1a57f6cc7c677857/si.proto#L239-L243

PR #28, opened 2019-08-13
application state is accepted after scheduler recovery

yangwwei:
During recovery, when we start to processing existing allocations, we should transit app's state from Accepted to Running, this was missed in the initial patch for recovery, this PR is to fix this issue.

### DIFF
--- a/pkg/cache/partition_info.go
+++ b/pkg/cache/partition_info.go
@@ -169,18 +169,6 @@ func (pi *PartitionInfo) addNewNode(node *NodeInfo, existingAllocations []*si.Al
             zap.String("nodeId", node.NodeId),
             zap.Int("existingAllocations", len(existingAllocations)))
         for _, alloc := range existingAllocations {
-            // if this is to start processing existing allocations for an app,
-            // transit app's state from Accepted to Running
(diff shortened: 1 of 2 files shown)


pkg/cache/partition_info.go
```diff
@@ -169,6 +169,18 @@ func (pi *PartitionInfo) addNewNode(node *NodeInfo, existingAllocations []*si.Al
             zap.String("nodeId", node.NodeId),
             zap.Int("existingAllocations", len(existingAllocations)))
         for _, alloc := range existingAllocations {
+            // if this is to start processing existing allocations for an app,
+            // transit app's state from Accepted to Running
+            if app, ok := pi.applications[alloc.ApplicationId]; ok {
```
wilfred-s: Shouldn't this happen after we have successfully added the allocation(s)?
We check the outcome of the allocation and reject the node if the allocation fails. I think we should only move the apps to running when the node addition has succeeded just before we update the metric that the node is active.

yangwwei: We can do that. I wasn't having a strong preference about the time point, what you said makes sense, I just revised the logic a bit to do that. Thx

PR #38, opened 2019-08-26
Remove strings.ReplaceAll for Go 1.11 compatibility

### LINKED ISSUE #37: Build error on go 1.11
Getting the below error while building yunikorn-core on go 1.11.

```
../yunikorn-core/pkg/scheduler/placement/rule.go:112:9: undefined: strings.ReplaceAll
note: module requires Go 1.12
make: *** [build_image] Error 2
```

Looks like strings.ReplaceAll is only available in Go 1.12 .  The docs mention go 1.11+ can be used. So we can either,
* Change the doc to mention only Go 1.12+
(or)
* Change ReplaceAll to Replace to make it work in go 1.11 .

```
-       return strings.ReplaceAll(name, cache.DOT, cache.DotReplace)
+       return strings.Replace(name, cache.DOT, cache.DotReplace, -1)
```

wilfred-s:
Replace the new strings.ReplaceAll() method call with the old strings.Replace() call.

ReplaceAll was introduced in G0 1.12 as most of the calls use a -1 anyway.

Closes #37

### DIFF
--- a/pkg/scheduler/placement/rule.go
+++ b/pkg/scheduler/placement/rule.go
@@ -109,5 +109,5 @@ func normalise(name string) string {
 
 // Replace all dots in the generated queue name before making it a fully qualified name.
 func replaceDot(name string) string {
-	return strings.ReplaceAll(name, cache.DOT, cache.DotReplace)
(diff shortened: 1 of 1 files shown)


yangwwei:
this resolves #37

PR #40, opened 2019-08-28
ACL and Filter use user regexp when they should use group regexp

wilfred-s:
In the ACL and filter when we check the groups we use the less restrictive user regexp.
It does not break behaviour but we should be using the correct regexp to make sure the configuration is correctly checked.

### DIFF
--- a/pkg/common/security/acl.go
+++ b/pkg/common/security/acl.go
@@ -30,7 +30,10 @@ const (
     Space     = " "
 )
 
-var UserNameRegExp = regexp.MustCompile("^[a-z_]([a-z0-9_-]{0,31}|[a-z0-9_-]{0,30}\\$)$")
+// User and group regexp, must allow at least what we allow in the config checks
+// See configs.UserNameRegExp and configs.GroupRegExp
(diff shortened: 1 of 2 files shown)


pkg/common/security/acl.go
```diff
@@ -30,7 +30,10 @@ const (
     Space     = " "
 )
 
-var UserNameRegExp = regexp.MustCompile("^[a-z_]([a-z0-9_-]{0,31}|[a-z0-9_-]{0,30}\\$)$")
+// User and group regexp, must allow at least what we allow in the config checks
+// See configs.UserNameRegExp and configs.GroupRegExp
+var UserNameRegExp = regexp.MustCompile("^[_a-zA-Z][a-zA-Z0-9_.@-]*[$]?$")
+var GroupRegExp = regexp.MustCompile("^[_a-zA-Z][a-zA-Z0-9_-]*$")
```
yangwwei: I am wondering why `UserNameRegExp` and `GroupRegExp ` instead of `userNameRegExp`/`groupRegExp`, since this regex is only used in this file.

wilfred-s: They should have been private.
Fixed both of them

PR #42, opened 2019-08-28
add UT to test scheduling behaviors when exceeds queue capacity

yangwwei:
add a UT to cover the following case
- send requests that exceed total queue capacity
- verify capacity gets fully used, and left some requests pending
- release some existing allocations and that can give room for pending requests

### DIFF
--- a/pkg/scheduler/tests/scheduler_smoke_test.go
+++ b/pkg/scheduler/tests/scheduler_smoke_test.go
@@ -939,13 +939,37 @@ partitions:
     waitForAcceptedApplications(mockRM, "app-added-2", 1000)
 }
 
-func TestSchedulingOverCapacity(t *testing.T) {
-    serviceContext := entrypoint.StartAllServicesWithManualScheduler()
-    defer serviceContext.StopAll()
-    proxy := serviceContext.RMProxy
-
-    // Register RM
-    configData := `
+func TestSchedulingOverMaxCapacity(t *testing.T) {
+    var parameters = []struct {
+        name        string
+        configData  string
+        leafQueue   string
+        askMemory   int64
+        askCPU      int64
+        numOfAsk    int32
+    }{
+        {"scheduleOverParentQueueCapacity",
+            `
+partitions:
+  -
+    name: default
+    queues:
+      -
+        name: root
+        submitacl: "*"
+        queues:
+          -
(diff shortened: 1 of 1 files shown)


pkg/scheduler/tests/scheduler_smoke_test.go
```diff
@@ -936,3 +938,136 @@ partitions:
...
+            {
+                NodeId: "node-1:1234",
+                Attributes: map[string]string{
+                    "si.io/hostname": "node-1",
+                    "si.io/rackname": "rack-1",
+                },
+                SchedulableResource: &si.Resource{
+                    Resources: map[string]*si.Quantity{
+                        "memory": {Value: 100},
+                        "vcore":  {Value: 10},
```
wilfred-s: To make sure that we do not get limited by the node we should do one of two things:
- decrease the queue size to 50MB/5vcore and ask for 70/7
or
- increase the node to 150MB/15vcores and leave the ask

In the case we have now even if we regress in the queue area the node size will prevent the test from failing.

pkg/scheduler/tests/scheduler_smoke_test.go
```diff
@@ -936,3 +938,136 @@ partitions:
...
+        submitacl: "*"
+        queues:
+          -
+            name: default
+            properties:
+              application.sort.policy: fair
+            resources:
+              guaranteed:
+                memory: 100
+                vcore: 10
```
wilfred-s: Can we minimise the config as much as possible?
We do not use the properties or guaranteed resources in the test.

pkg/scheduler/tests/scheduler_smoke_test.go
yangwwei: Done

pkg/scheduler/tests/scheduler_smoke_test.go
yangwwei: Yep, we can do that. Fixed in the commit-2.

yangwwei:
> Can we extend the test to also cover the case that the max of a parent is set instead of the queue the application is running in directly.
> This indirect limit should also be enforced and I don't think we smoke test that.

I have done that in the new commit, leverages table driven tests to cover both scenarios. Pls take a look.

PR #44, opened 2019-08-30
Add User limit objects to the queue in the config

wilfred-s:
- cleanup and doc of config
- extend validator to check user limit object on partition and queue
- extend config tests to include user limits

### DIFF
--- a/config/limit.yaml
+++ b/config/limit.yaml
@@ -3,11 +3,16 @@ partitions:
     queues:
       - name: root
         limits:
-          - users: john,sue
-              maxapplications: 10
-              maxresources: {memory: 10000, vcore: 10}
-          - users: bob
-              maxapplications: 100
+          - limit:
+            users:
+            - john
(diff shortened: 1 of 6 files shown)


wilfred-s:
Major update based on the discussions we have had earlier:
- rebase on top of the current head
- move to a generic limit object
- use same yaml structure as for filters

Documentation update is pending and should be added before commit

yangwwei:
Hi @wilfred-s 

The changes in this PR looks good to me, as long as it declares general user/group limits, I am fine to merge it. But the only thing I think we add to our document, that is to say, we are not recognizing users/groups in the scheduler yet, that is something in our roadmap. I don't want to give users the impression they can actually use these now, this feature can be complete only when we build the shim to send correct user/group info.

wilfred-s:
Added comment to the documentation to state that shim work is outstanding for functionality to work correctly.

PR #46, opened 2019-09-03
Scheduler core logging picks up configured options from zap global logger

yangwwei:
During testing, found earlier this was not working, this is because the logger in the core was initiated in the `init()` function, which will be loaded before shim initiates its logger (where it loads configurations from CLI). The fix is to remove `init()` function, and use a lazy init with sync.once instead.


sunilgovind:
looks good to me. pushing this.

PR #50, opened 2019-09-18
Fix potential deadlock in SchedulingQueue

TaoYang526:
There may be a potential deadlock when:
 (1) scheduling thread holds the lock of parent queue and waits for read lock of child queue (calling stack: Scheduler#singleStepSchedule -> Scheduler#findAllocationAsks -> Scheduler#findNextAllocationAskCandidate -> Scheduler#sortSubqueuesFromQueue -> SchedulingQueue#GetPendingResource) 
and meanwhile
 (2) scheduler event handler holds the lock of child queue and waits for read lock of parent queue (calling stack: Scheduler#handleSchedulerEvent -> Scheduler#processAllocationUpdateEvent -> Scheduler#updateSchedulingRequest -> SchedulingQueue#IncPendingResource)

I think the proper sequence of nested lock should be from parent queue to child queue, the contrary way should be corrected.

Deadlock stacks:
image

### DIFF
--- a/pkg/scheduler/scheduling_queue.go
+++ b/pkg/scheduler/scheduling_queue.go
@@ -121,26 +121,26 @@ func (sq *SchedulingQueue) updateSchedulingQueueInfo(info map[string]*cache.Queu
 // Update pending resource of this queue
 func (sq *SchedulingQueue) IncPendingResource(delta *resources.Resource) {
     sq.lock.Lock()
-    defer sq.lock.Unlock()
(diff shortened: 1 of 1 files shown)


yangwwei:
Nice catch @TaoYang526 
I will get this merged once we received your ICLA. Thanks!

yangwwei:
Thanks @TaoYang526 for getting this fixed!

TaoYang526:
Thanks @yangwwei for the review and commit.

Issue #52, opened 2019-09-23
bin-packing scheduling for yunikorn


sunilgovind:

yangwwei:
We also need a design doc for this

sunilgovind:
Sure @yangwwei , uploading same sooner

PR #55, opened 2019-09-26
initial version of bin-packing

sunilgovind:
This is an initial version

- separate thread to compute nodes
- api to fetch a node which is best fit
- hot refresh is done

Todo:
- UT are there in another patch, but with some refactoring which i did, i cant run that now. I will upload a new patch fixing same
- Improving api layer for hot refresh


yangwwei:
hi @sunilgovind 
For my comment
> For bin-packing, can we support a configuration, e.g bin-size or something similar? To control the usage of each node doesn't exceed a certain threshold unless there is no more available nodes?
we can leave it for now. lets get a basic working version first. Does that make sense to you?

config/queues.yaml
```diff
@@ -1,5 +1,8 @@
 partitions:
   - name: default
+    globalpolicies:
+      - name: myschedulingpolicy
+        policy: binpacking
```
wangdatan: Should we make the global_policy to be more extendable? I'm wondering why we need a name field here? Should it be just a list of policy with per-policy parameters?

wangdatan: And how abt call it "node_sort_policy" to make it to be more specific?

pkg/cache/partition_info.go
```diff
@@ -93,6 +94,18 @@ func NewPartitionInfo(partition configs.PartitionConfig, rmId string, info *Clus
...
+    for _, policy := range partition.GlobalPolicies {
+        switch policy.Policy {
+        case configs.SchedulingBinPackingPolicy:
+            p.globalSchedulingPolicy = NewSchedulingPolicy(policy)
+            log.Logger().Info("SchedulingBinPackingPolicy policy is set.")
+        default:
+            log.Logger().Info("No default scheduling policy is set.")
+        }
+    }
+
```
wangdatan: We need to define what is "default scheduling policy", to me it is ambiguous to say there's no "default scheduling policy". There should be always one policy, such as "round-robin".

pkg/cache/scheduling_policy.go
```diff
@@ -0,0 +1,45 @@
...
+func NewSchedulingPolicy(policy configs.GlobalPolicy) (*SchedulingPolicy) {
+	sp := &SchedulingPolicy{Name: strings.ToLower(policy.Name),
+		PolicyType: policy.Policy,
+	}
+
+	log.Logger().Debug("new scheduling policy added",
+		zap.String("policy name", sp.Name),
+		zap.String("type", sp.PolicyType))
+	return sp
+}
```
wangdatan: To me fair and bin-packing are not in the same level, just like I mentioned above, it should be node sorting policy. 

Fairness should be app/user/queue sorting policy. And round-robin/utilization/bin-packing are node sorting policy

wangdatan: And I think we should have only one node sorting policy per partition, not multiple ..

pkg/scheduler/allocator.go
```diff
@@ -214,6 +211,56 @@ func (m *Scheduler) tryBatchAllocation(partition string, candidates []*Schedulin
     return allocations, failedAsks
 }
 
+// TODO: convert this as an interface.
```
wangdatan: Yeah make sense, this should be part of node_sorting_policy, it should return an iterator of sorted node list.

pkg/scheduler/allocator.go
```diff
@@ -214,6 +211,56 @@ func (m *Scheduler) tryBatchAllocation(partition string, candidates []*Schedulin
...
+    }
+
+    schedulingNodeList := make([]*SchedulingNode, len(nodeList))
+    for idx, v := range nodeList {
+        schedulingNodeList[idx] = NewSchedulingNode(v)
+    }
+
+    return schedulingNodeList
+}
+
```
wangdatan: The `getNodeList` is the old logic, I think we should try to avoid a full copy of node list and creation of scheduling node after that.

pkg/scheduler/allocator.go
```diff
@@ -214,6 +211,56 @@ func (m *Scheduler) tryBatchAllocation(partition string, candidates []*Schedulin
...
+        // once we have a node which can fit the allocation ask, send back for scheduling
+        if node.CheckAndAllocateResource(candidate.AllocatedResource, false /* preemptionPhase */) {
+            if log.IsDebugEnabled() {
+                log.Logger().Debug("Selected node:",
+                    zap.Any("id", node))
+            }
+            return append(make([]*SchedulingNode, 0, 1), node)
+        }
+    }
```
wangdatan: Why handle checkAllocationConditition, and check Allocate resource here? We have logic in the `allocate` already. I think we should not let the "node sorting policy" to handle allocate or not based on a specific resource request. 

The node sort policy should just sort the node, it may filter out some invalid node (such as 0 available capacity).         And I think we should not create a copy of node here. 

Ideally, we should have a state kept in node-sort-policy, when new node updated, new pod allocated, etc. it will be notified instead of sorting every time.

wangdatan:
I will be fine with not adding the stateful node sorting mechanism (like what we have in YARN) now. But I still think we should make the node sorting policy interface clearly defined.

wangdatan:
Also, I'm wondering if we can get rid of the on-demand construction of scheduling node, it could cause lots of perf issues. To me, we can keep a reference to node_info inside scheduling_node. I remember I considered it before but somehow I chose an easy path here. 

Maybe we should do that in the later patches, just want to point it out.

yangwwei:
Hi @sunilgovind 

Please take a look at `allocatior#regularAllocate`, this part accesses from a random point in the sorted node list, would it be conflicting the bin-packing logic?

pkg/cache/partition_info.go
yangwwei: +1

pkg/cache/scheduling_policy.go
yangwwei: I think what Sunil mentioned here FAIR is the reverse pattern of bin-packing. Sorting nodes and put less utilized nodes in the beginning. Not the queue ordering policy fair when we order apps. Is that correct @sunilgovind ?

config/queues.yaml
yangwwei: The arbitrary name is quite confusing to me.

sunilgovind: i see, i was more less thinking that we can choose a custom policy injection later, so a name will be helpful for user to define any polices per partition, global or queue level. so they can refer the policy once defined at the top level, just referring name alone later.

node_sort_policy can be of any sort. we can sort by descending , or packing model,  may be looking at some different score per node etc. 

so do u think it should be like
      - policy: node_sort_policy
           type: roundrobin/binpacking
?

pkg/cache/partition_info.go
sunilgovind: yes. agree.

pkg/cache/scheduling_policy.go
sunilgovind: yes you are are @yangwwei

yangwwei:
Another comment, does it make sense to move`scheduling_policy.go` to some other package other than `pkg/cache` package?

sunilgovind:
@yangwwei @wangdatan  pls help to check the latest changes.
I redid the random sort of nodes in default case in a different way in SortNodes. pls check.

yangwwei:
@sunilgovind  pls help to check our latest comments. Thanks

pkg/common/configs/config.go
```diff
@@ -108,6 +109,12 @@ type User struct {
     MaxApplications uint64            `yaml:",omitempty" json:",omitempty"`
 }
 
+// Global Node Sorting Policy section
+// - type: different type of policies supported (binpacking, fifo, random etc)
+type NodeSortingPolicy struct {
+    Type string
```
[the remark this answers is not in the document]
sunilgovind:
```threshold=80
```
Thats one of the options we are expecting for sure immediate and further more

pkg/common/configs/configvalidator.go
```diff
@@ -189,6 +194,29 @@ func checkUserDefinition(partition *PartitionConfig) error {
     return nil
 }
 
+// Check for global policy
+func checkNodeSortingPolicy(partition *PartitionConfig) error {
+
+    // get the policy
+    rule := partition.NodeSortPolicy
```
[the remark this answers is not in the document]
sunilgovind: This is changed now in latest patch.
Ack

pkg/scheduler/allocator.go
```diff
@@ -91,10 +91,11 @@ func (m *Scheduler) singleStepSchedule(nAlloc int, preemptionParam *preemptionPa
 
 func (m *Scheduler) regularAllocate(nodes []*SchedulingNode, candidate *SchedulingAllocationAsk) *SchedulingAllocation {
     nNodes := len(nodes)
-    startIdx := rand.Intn(nNodes)
+    if log.IsDebugEnabled() {
```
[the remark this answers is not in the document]
sunilgovind: ack

sunilgovind:
Comments are addressed now. Updating patch
cc @yangwwei

pkg/scheduler/tests/scheduler_smoke_test.go
```diff
@@ -1212,6 +1212,154 @@ partitions:
...
+    schedulerQueueRoot := scheduler.GetClusterSchedulingContext().GetSchedulingQueue("root", "[rm:123]default")
+
+    // Check scheduling queue a
+    schedulerQueueA := scheduler.GetClusterSchedulingContext().GetSchedulingQueue("root.a", "[rm:123]default")
+
+    // Check scheduling queue b
+    schedulerQueueB := scheduler.GetClusterSchedulingContext().GetSchedulingQueue("root.b", "[rm:123]default")
+
+    if nil != err {
+        t.Error(err.Error())
```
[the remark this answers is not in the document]
sunilgovind: I agree, most of the tests are doing this already.
I think we need to check

pkg/scheduler/sorters.go
```diff
@@ -72,17 +72,16 @@ func SortNodes(nodes []*SchedulingNode, sortType SortType) {
...
+            return resources.CompFairnessRatioAssumesUnitPartition(l.CachedAvailableResource, r.CachedAvailableResource) > 0
         })
+        // Do an in-place shuffle with out creating a new-array
+        random := rand.New(rand.NewSource(time.Now().Unix()))
+        for len(nodes) > 0 {
+            n := len(nodes)
+            randIndex := random.Intn(n)
+            nodes[n-1], nodes[randIndex] = nodes[randIndex], nodes[n-1]
+            nodes = nodes[:n-1]
+        }
```
yangwwei: This loop ends when nodes length is zero
After `SortNodes` is called,  the nodes become empty. this doesn't seem to be the correct behavior for a sorting method.

pkg/scheduler/sorters.go
```diff
@@ -72,17 +72,16 @@ func SortNodes(nodes []*SchedulingNode, sortType SortType) {
...
-            // Do an in-place shuffle with out creating a new-array
-            random := rand.New(rand.NewSource(time.Now().Unix()))
-            for len(nodes) > 0 {
-                n := len(nodes)
-                randIndex := random.Intn(n)
-                nodes[n-1], nodes[randIndex] = nodes[randIndex], nodes[n-1]
-                nodes = nodes[:n-1]
-            }
-            return ret
+            return resources.CompFairnessRatioAssumesUnitPartition(l.CachedAvailableResource, r.CachedAvailableResource) > 0
```
yangwwei: Since nodes will be shuffled later, why we need to sort them in the first place?

pkg/common/configs/configvalidator.go
```diff
@@ -18,6 +18,7 @@ package configs
 
 import (
     "fmt"
+    "github.com/cloudera/yunikorn-core/pkg/cache"
```
yangwwei: This seems to introduce cyclic dependency, it could not build.

pkg/common/configs/config.go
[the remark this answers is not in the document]
wilfred-s: OK, that warrants an object to make future expansion possible.

config/queues.yaml
```diff
@@ -1,8 +1,7 @@
 partitions:
   - name: default
-    globalpolicies:
-      - name: myschedulingpolicy
-        policy: binpacking
+    nodesortpolicy:
```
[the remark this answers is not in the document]
sunilgovind: NodeSortingDefault is removed.

>> can we rename nodesortpolicy to schedulingpolicy?
in earlier discussion we thought of keeping as nodesortpolicy, this was one of wangdas comment

pkg/scheduler/sorters.go
```diff
@@ -71,7 +72,16 @@ func SortNodes(nodes []*SchedulingNode, sortType SortType) {
...
+            ret := resources.CompFairnessRatioAssumesUnitPartition(l.CachedAvailableResource, r.CachedAvailableResource) > 0
+            // Do an in-place shuffle with out creating a new-array
+            random := rand.New(rand.NewSource(time.Now().Unix()))
+            for len(nodes) > 0 {
+                n := len(nodes)
+                randIndex := random.Intn(n)
+                nodes[n-1], nodes[randIndex] = nodes[randIndex], nodes[n-1]
+                nodes = nodes[:n-1]
+            }
+            return ret
```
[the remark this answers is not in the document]
sunilgovind: this is no longer needed as we have an iterator.

pkg/cache/nodesorting_policy.go
```diff
@@ -17,32 +17,60 @@ limitations under the License.
...
-	PolicyType string
+	PolicyType SortingPolicy
+}
+
+type SortingPolicy int
+
+const (
+	BinPackingPolicy = iota
+	FairnessPolicy
+	Default
```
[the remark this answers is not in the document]
sunilgovind: default is removed

pkg/scheduler/sorters.go
sunilgovind: the same logic is implemented in iterator.

pkg/scheduler/sorters.go
sunilgovind: as per new impl., its more optimized. pls check iterator code.

pkg/common/configs/configvalidator.go
sunilgovind: removed all cyclic dependencies for policies.

pkg/common/configs/configvalidator.go
```diff
@@ -189,6 +190,22 @@ func checkUserDefinition(partition *PartitionConfig) error {
...
+    // get the policy
+    policy := partition.NodeSortPolicy
+
+    log.Logger().Debug("global node sorting policies:",
+        zap.String("policy type", policy.Type))
+
+    // Defined polices.
+    configuredNodeSortingPolicy, error := cache.FromString(policy.Type)
+
+    log.Logger().Info("Node sorting policy:", zap.Any("policy name", policy.Type), zap.Any(", value", configuredNodeSortingPolicy))
```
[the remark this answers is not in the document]
sunilgovind: handled in my latest patch

sunilgovind:
@yangwwei  @wilfred-s 
- all comments are handled.
- tested in real env
- test cases are passing

pls help to check.

pkg/scheduler/allocator.go
sunilgovind: done

config/queues.yaml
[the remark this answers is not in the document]
wilfred-s: no rename needed and default is removed in the latest version

pkg/scheduler/allocator.go
wilfred-s: iterators are implemented and passed around

pkg/scheduler/allocator.go
```diff
@@ -18,13 +18,13 @@ package scheduler
 
 import (
     "context"
+    "github.com/cloudera/yunikorn-core/pkg/cache"
```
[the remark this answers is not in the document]
wilfred-s: cyclic import is fixed.
follow up issue to be logged to clean up where and how we store the policy in the objects

pkg/scheduler/tests/scheduler_smoke_test.go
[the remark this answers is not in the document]
wilfred-s: fixed via new commit

wilfred-s:
This fixes #52

PR #59, opened 2019-10-08
Cleanup pending resource update and app removal

wilfred-s:
This is a follow up for #50.

The cleanup should first update the parent queues and then the current queue. This partially reverses the change from #50. It does keep in place the locking fix which is the real problem that needed to be fixed.

It also fixes the TODO that was in the decrement. We should never go negative on the pending resource decrement. Replacing Sub with SubEliminateNegative. The same change is also added to the removal of the scheduling app.


yangwwei:
Hi @wilfred-s apologies that get to this one so late.
For the patch, use `SubEliminateNegative` makes sense to me. However, what's the difference between updating parents first comparing to updating the child first. I don't see any difference as the dec/inc functions are no-error calls. Are you seeing any issues that lead to this change?

wilfred-s:
First point is consistency: when we update all other queue usages we recurse up and update from the root down to the leaf. We follow the same pattern for all the queue updates. With the lock as part of the unwind of the recursion we also prevent a possible inconsistency when we have concurrent updates from multiple leaf queues.

I just saw that we have a problem in `RemoveQueue` with the current locking after #50. We have performed the checks and can remove the queue we now unlock. While we release the lock we could add a new application to the queue which is being removed. We should not release the lock while we remove the queue as this causes inconsistencies: an app gets linked to a removed queue...

I also thought about not using `SubEliminateNegative` and returning an error or logging it so we don't hide the inconsistent state on decrement. I have not done that as yet because we *can* and do call `IncPendingResource` with a negative delta which really turns it into a decrement. I was going to log a new issue around that but I think we need to tackle that now.

I am going to update the request to clean this up.

wilfred-s:
The updates as discussed before:
- Fix queue lock on removal
- Add new sub error on negative call and use in pending resources
- Remove parent queue lock to get child list (use locked function)

This 3rd change was not used before but it removes the parent queue lock from the ask_finder. this reduces the time the lock is held on the parent queue to a minimum. While the child queues are processed the parent queue is now no longer locked.
@TaoYang526 this is an enhanced version of the change in #50

pkg/scheduler/asks_finder.go
```diff
@@ -64,15 +64,12 @@ func (m *Scheduler) findAllocationAsks(partitionTotalResource *resources.Resourc
 // sorting type for the parent queue.
 // Stopped queues will be filtered out at a later stage.
 func sortSubqueuesFromQueue(parentQueue *SchedulingQueue) []*SchedulingQueue {
-    parentQueue.lock.RLock()
-    defer parentQueue.lock.RUnlock()
```
yangwwei: Doubt about removing this lock.
It might be possible during the same time when the scheduler doing the queues sorting, another allocation added/removed to a leaf queue, which invalidates the sorting result, isn't it?

wilfred-s: Sorting is just one step and the sort can be invalidated already. 
- Getting a copy of the child list is still locked
- After we sort the queue list is passed back. At that point in time we already have released the lock and allocations will occur.
- the child queues which are sorted were never locked.

Sorting also happens on two field that are not lock protected: ProposingResource for the scheduling side and the cached GuaranteedResource field. Both are accessed outside locks. So locking of the parent queue could only have had an indirect effect preventing modifications on the ProposingResource but never on the cached GuaranteedResource.

Based on all that I removed the lock

yangwwei: I was quite surprised that the accessing of queue resources were not lock protected, I just went over the related code again. Now I see why.  We are not using real multi-thread allocating. In tryBatchAllocation, we can possibly launch a few threads, but we wait on them to return proposals. So sort queue and modify queue resources are not happening concurrently. I think we are safe to remove this lock now.

But we may need to revise this model sometime later. Usually, we run allocation threads concurrently and produce proposals and the main scheduler thread just consumes them by either committing or rejecting them.

wilfred-s: I agree when we start moving towards multithreaded allocations we need to go over the whole locking with a fine-tooth comb. We might get a different set of locks when we get to that.

wilfred-s:
Based on the last comment I don't see anything holding this back.
Merging this later today if there are no further comments

yangwwei:
Yes, this should be good.
Merging this now. Thanks @wilfred-s

TaoYang526:
Sorry for replying so late. I just quickly go through this PR, IIUC, maybe there's still a risk of deadlock when self-check goroutine of PartitionManager holds child queue's read lock (calling SchedulingQueue#RemoveQueue) and try to get parent queue's write lock, and meanwhile another goroutine may holds parent queue's write lock and try to get child queue's write lock (calling SchedulingQueue#updateSchedulingQueueInfo for example). Could you please check if this may happen? @wilfred-s  @yangwwei

wilfred-s:
That cannot happen and should never happen. The scheduling queue can only be removed if the queue is not used in scheduling anymore. This means that either the queue is a managed queue that was removed from the configuration (not active anymore) or an unmanaged queue that is empty and has nothing left.
There is no way that the update of the queue can be called in that case.

With the changes made around the updates we walk up a queue tree without lock. While we unwind the recursion we make the change at a level using a lock. When the parent is done we lock the child only. That guarantees that we never have more than 1 lock at the same time except in really special cases which we have control over.

TaoYang526:
Thanks @wilfred-s for the clarification, it's clear now. It's my fault to update lock in SchedulingQueue#RemoveQueue without further understanding, thanks for fixing it.

PR #60, opened 2019-10-22
add some unit tests to verify the correctness of queue fairness

yangwwei:
Some simple UT to cover fairness testing.

### DIFF
--- a/pkg/scheduler/sorters_test.go
+++ b/pkg/scheduler/sorters_test.go
@@ -25,41 +25,41 @@ import (
 // verify queue ordering is working
 func TestSortQueues(t *testing.T) {
 	root, err := createRootQueue()
-	root.CachedQueueInfo.GuaranteedResource = resources.NewResourceFromMap(
-		map[string]resources.Quantity{"memory": 1000, "vcore": 1000})
 	if err != nil {
 		t.Fatalf("failed to create basic root queue: %v", err)
 	}
+	root.CachedQueueInfo.GuaranteedResource = resources.NewResourceFromMap(
+		map[string]resources.Quantity{"memory": 1000, "vcore": 1000})
 
 	q0, err := createManagedQueue(root, "q0", false)
+	if err != nil {
+		t.Fatalf("failed to create leaf queue: %v", err)
+	}
 	q0.CachedQueueInfo.GuaranteedResource = resources.NewResourceFromMap(
 		map[string]resources.Quantity{"memory": 500, "vcore": 500})
 	q0.ProposingResource = resources.NewResourceFromMap(map[string]resources.Quantity{
 		"memory": resources.Quantity(300),
 		"vcore" : resources.Quantity(300)})
+
+	q1, err := createManagedQueue(root, "q1", false)
 	if err != nil {
 		t.Fatalf("failed to create leaf queue: %v", err)
 	}
-
-	q1, err := createManagedQueue(root, "q1", false)
 	q1.CachedQueueInfo.GuaranteedResource = resources.NewResourceFromMap(
 		map[string]resources.Quantity{"memory": 300, "vcore": 300})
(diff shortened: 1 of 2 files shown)


pkg/scheduler/sorters_test.go
```diff
@@ -0,0 +1,114 @@
...
+	q1.ProposingResource = resources.NewResourceFromMap(map[string]resources.Quantity{"memory": 100, "vcore": 100})
+	SortQueue(queues, FairSortPolicy)
+	assert.Equal(t, len(queues), 3)
+	assert.Equal(t, queues[0].Name, "q2")
+	assert.Equal(t, queues[1].Name, "q1")
+	assert.Equal(t, queues[2].Name, "q0")
+}
+
+
+func newQueueInfo(name string, parent *cache.QueueInfo,
```
wilfred-s: Can we leverage the methods that we have in the scheduling_queue_tests for creating the queues:
`func createRootQueue()`
`func createManagedQueue(parentQI *SchedulingQueue, name string, parent bool)`
Any changes in the queue structs will then only impact one location.

Setting guaranteed and proposed resources can be simply achieved by updating the returned queues.

pkg/scheduler/sorters_test.go
```diff
@@ -0,0 +1,114 @@
...
+	q0     := newSchedulingQueue("q0", q0info, qRoot,300, 300)
+	q1     := newSchedulingQueue("q1", q1info, qRoot,200, 200)
+	q2     := newSchedulingQueue("q2", q2info, qRoot, 100, 100)
+
+	queues := []*SchedulingQueue{q0, q1, q2}
+	SortQueue(queues, FairSortPolicy)
+	assert.Equal(t, len(queues), 3)
+	assert.Equal(t, queues[0].Name, "q2")
+	assert.Equal(t, queues[1].Name, "q1")
+	assert.Equal(t, queues[2].Name, "q0")
```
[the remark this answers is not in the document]
yangwwei: That's right, let me fix this.

pkg/scheduler/sorters_test.go
yangwwei: I've updated the latest patch to do so, thanks for the suggestion.

pkg/scheduler/sorters_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestSortQueues(t *testing.T) {
+	// define queue info without guaranteed resources
+	root   := newQueueInfo("root", nil, 1000, 1000, 1000, 1000)
+	q0info := newQueueInfo("q0", root, 0, 0, 500, 500)
+	q1info := newQueueInfo("q1", root, 0, 0, 200, 200)
+	q2info := newQueueInfo("q2", root, 0, 0, 300, 300)
```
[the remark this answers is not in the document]
yangwwei: I've added ut to verify this case, thanks for pointing this out.

pkg/scheduler/sorters_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+// verify queue ordering is working
+func TestSortQueues(t *testing.T) {
+	root, err := createRootQueue()
+	root.CachedQueueInfo.GuaranteedResource = resources.NewResourceFromMap(
+		map[string]resources.Quantity{"memory": 1000, "vcore": 1000})
+	if err != nil {
```
wilfred-s: should swap the err check and the use of the returned queue as it could be nil (throughout the file)

yangwwei:
Really sorry, I totally forgot this issue being left over.
@wilfred-s please take a look at the update, it fixes the things you mentioned in the comment, that's a good catch, thanks!
It did introduce a change to go.mod, that was done automatically, which removes some unused dependencies, I think it is OK to include.

PR #64, opened 2019-11-06
queue-usage-ratio should be compared separately when sorting queues

TaoYang526:
For example:
queue1: allocated-resource:50, guaranteed-resource:100, usage-ratio=50/100=0.5
queue2: allocated-resource:100, guaranteed-resource:300, usage-ratio=0.33
After sorting queues, queue2 should be placed before queue1, but current result is opposite.

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -369,6 +369,18 @@ func CompUsageRatio(left, right, total *Resource) int {
     return compareShares(lshares, rshares)
 }
 
+// Calculate share for left of total and right of total separately.
+// This returns the same value as compareShares does:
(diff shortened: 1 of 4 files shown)


yangwwei:
I think this fix makes sense to me. Earlier our testing was using the same queue capacity so probably that's the reason why the problem did not reveal. The fix is good.
But looking at the code, I start to wonder why we even need `CompUsageRatio`, for the places calling this function, it can be replaced by `CompUsageShares`, isn't it? @wilfred-s , please also take a look at this.

yangwwei:
Let me merge the change for now. @wilfred-s , if you have any concerns, please let us know. We can always revisit this.

PR #65, opened 2019-11-12
Revert "Support unconfirmed resource for nodes to improve scheduling fairness"

wilfred-s:
Reverts cloudera/yunikorn-core#63

After a lot of discussion we have decided to revert the change for now as we have found issues that cause the cache and the scheduler to go out of sync. The change also breaks #57 as it removes the scheduling node immediately without keeping track of the usage and the links to apps etc.


yangwwei:
Yep, we'll continue the discussion and come up with a better solution in the next few days. cc @TaoYang526  for awareness.

TaoYang526:
Thanks @wilfred-s @yangwwei for this feedback, #63 indeed has a problem that SchedulingNode#cachedAvailableResources may go out of sync when releasing the allocations, sorry for updating that in a hurry without further thought. For the removal of scheduling nodes, I have thought scheduling nodes may only keep schedulable nodes so that some process like sorting and iterating maybe more efficient, I should have got this out for discussion before and will discuss some unclear thoughts at first next time. I would like to participate this discussion and hope to get a  perfect solution later. Thanks.

PR #66, opened 2019-11-15
Refactor scheduling node iterator and sorters

wilfred-s:
A mixture of naming inconsistencies and shared code is cleaned up in the
iterator. The sorter has been updated to be consistent over the different
object types. Unit tests for the iterator and sorters are added.

Flow on changes for naming changes in allocator and pre-emption.

The FairSortingIterator did not work as expected: it would never start at a random point in the list. The iterator is left but nodes after sorting are now always iterated over via the DefaultIterator


pkg/scheduler/node_iterator.go
```diff
@@ -0,0 +1,106 @@
...
+	it := &DefaultNodeIterator{}
+	it.nodes = schedulerNodes
+	it.size = len(schedulerNodes)
+	return it
+}
+
+// Random iterator, wraps the base iterator
+// Iterates over the list from a random starting position in the list.
+// The iterator automatically wraps at the end of the list.
+type RandomNodeIterator struct {
```
yangwwei: The name is a bit confusing. RandomNodeIterator to me sounds like it picks up a random node every time. But the actual, it is a round-robin starting from a random point. Can we call this `RoundRobinNodeIterator`?

pkg/scheduler/node_iterator_test.go
```diff
@@ -0,0 +1,172 @@
+package scheduler
```
yangwwei: missing license header

wilfred-s: oops

pkg/scheduler/node_iterator.go
wilfred-s: ah yes that is much better

yangwwei:
hi @wilfred-s good to go, feel free to merge it by yourself, thanks!

wilfred-s:
Thank you @yangwwei  for the review

Issue #69, opened 2019-11-21
Add throughput test for YuniKorn-core


wangdatan:
We need a unittest-like throughput tool for yunikorn-core for easier troubleshooting.

yangwwei:
Resolved by #70

PR #70, opened 2019-11-21
Add new throughput tests (#69)

wangdatan:
This is for Issue #69

### DIFF
--- a/pkg/scheduler/tests/schedulertestutils.go
+++ b/pkg/scheduler/tests/schedulertestutils.go
@@ -150,6 +150,26 @@ func waitForRejectedApplications(m *MockRMCallbackHandler, appId string, timeout
     }
 }
 
+// Wait till reaches minimum number of accepted Nodes, return duration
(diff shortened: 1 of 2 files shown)


pkg/scheduler/tests/schedulertestutils.go
```diff
@@ -150,6 +150,26 @@ func waitForRejectedApplications(m *MockRMCallbackHandler, appId string, timeout
     }
 }
 
+// Wait till reaches minimum number of accepted Nodes, return duration
+func timeTillMinNumberOfAcceptedNodes(m *MockRMCallbackHandler, minNodes int) time.Duration {
```
yangwwei: We need a proper timeout here, in case something wrong we cannot get expected nodes/allocations in time.
Can we replace this with `utils.WaitFor`?

wangdatan: @yangwwei Done!

PR #71, opened 2019-11-21
Improve scheduler perf test

TaoYang526:
Glad to see scheduler perf test is added, it is useful for us to know scheduling abilities quickly and debug the scheduling perf locally, I have some suggestions about this:
1. Now this test doesn't contain verification about correctness and will last for a few seconds, we have to wait more time to execute 'make test', I suppose to mark this test as non-short and default run 'go test -short'
2. Too many DEBUG logs will be printed in console when running this test, we can change log level dynamically to avoid this.
3. totalAllocations in the test should be 100000 instead of 10000, here I think it's better to adjust the test scale and don't have to wait all allocations are done. For waitForMinAllocations in schedulertestutils.go, we can replace the sleep condition 'allocLen != nAlloc' with 'allocLen < nAlloc' to end this test in advance.

@wangdatan @yangwwei , could you please help to review this init PR, hope to hear your thoughts and improve it later,  Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ build: commands
 .PHONY: test
 test:
 	@echo "running unit tests"
-	go test ./... -cover $(RACE) -tags deadlock
+	go test ./... -cover $(RACE) -tags deadlock -short
 	go vet $(REPO)...
 
 # Simple clean of generated files only (no local cleanup).
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ build: commands
 .PHONY: test
 test:
 	@echo "running unit tests"
-	go test ./... -cover $(RACE) -tags deadlock -short
+	go test ./... -cover $(RACE) -tags deadlock
 	go vet $(REPO)...
 
 # Simple clean of generated files only (no local cleanup).
--- a/pkg/scheduler/tests/scheduler_perf_test.go
+++ b/pkg/scheduler/tests/scheduler_perf_test.go
@@ -19,12 +19,19 @@ package tests
 import (
     "github.com/cloudera/yunikorn-core/pkg/common/configs"
     "github.com/cloudera/yunikorn-core/pkg/entrypoint"
+    "github.com/cloudera/yunikorn-core/pkg/log"
     "github.com/cloudera/yunikorn-scheduler-interface/lib/go/si"
+    "go.uber.org/zap"
+    "strconv"
     "testing"
     "time"
 )
 
 func TestSchedulerThroughput5KNodes(t *testing.T) {
+    if testing.Short() {
+        t.Skip("skipping performance testing in short mode")
+    }
+    log.InitAndSetLevel(zap.WarnLevel)
     // Start all tests
     serviceContext := entrypoint.StartAllServices()
     defer serviceContext.StopAll()
@@ -89,7 +96,7 @@ partitions:
     // register 5000 nodes
     totalNode := 5000
     for i:=0; i < totalNode; i++ {
-        nodeName := "node-" + string(i)
+        nodeName := "node-" + strconv.Itoa(i)
         node := &si.NewNodeInfo{
             NodeId:
             nodeName + ":1234",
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -62,7 +62,7 @@ build: commands
 .PHONY: test
 test:
 	@echo "running unit tests"
-	go test ./... -cover $(RACE) -tags deadlock
+	go test ./... -cover $(RACE) -tags deadlock -short
```
wilfred-s: I do not agree with this, benchmark tests should be moved to use `go bench`
We should file a separate issue to adjust the build process to allow benchmarks to be created.

Until we have done that I think we need to run as is and not add the `-short` option

pkg/scheduler/tests/scheduler_perf_test.go
```diff
@@ -19,12 +19,19 @@ package tests
...
+    "strconv"
     "testing"
     "time"
 )
 
 func TestSchedulerThroughput5KNodes(t *testing.T) {
+    if testing.Short() {
+        t.Skip("skipping performance testing in short mode")
+    }
+    log.InitAndSetLevel(zap.WarnLevel)
```
wilfred-s: When we mark this as a benchmark and correctly implement the test we should not need it.
As a temporary workaround I can understand that moving to the standard level (Info) would be more in line to get a view of how the scheduler behaves. Unless the standard is Warn for most deployments.

Makefile
TaoYang526: Yes, this test looks more like a benchmark test, can I update it to be a benchmark test in this PR?

pkg/scheduler/tests/scheduler_perf_test.go
TaoYang526: Make sense to me, thanks.

Makefile
wilfred-s: You can do that, it does require a bit more change but I am OK with that if you want to take that on.
We don't need the benchmark in the makefile as a separate target just yet, we can run them manually.

TaoYang526:
@wilfred-s, I have updated this PR in last commit, please help to review the last commit, thanks.
Updates:
(1) rollback Makefile via removing '-short'
(2) change log level in perf test from 'Warn' to 'Info'
(3) improve TestSchedulerThroughput5KNodes to be benchmark tests with different conditions: request 10000 pods when the number of nodes is 500,1000,2000 or 5000.

Now we can run benchmark tests manually and get benchmark results like this:
```
BenchmarkScheduling/500Nodes/10000Pods-8         	   10000	    314560 ns/op
--- BENCH: BenchmarkScheduling/500Nodes/10000Pods-8
    scheduler_perf_test.go:119: Total time to add 500 node 39.024449ms, 12812.480709 per second
    scheduler_perf_test.go:168: Total time to allocate 10000 containers in 3.145724095s, 3178.918334 per second
--
BenchmarkScheduling/1000Nodes/10000Pods-8        	   10000	    822558 ns/op
--- BENCH: BenchmarkScheduling/1000Nodes/10000Pods-8
    scheduler_perf_test.go:119: Total time to add 1000 node 122.329034ms, 8174.674215 per second
    scheduler_perf_test.go:168: Total time to allocate 10000 containers in 8.225802284s, 1215.686890 per second
--
BenchmarkScheduling/2000Nodes/10000Pods-8        	   10000	   2332072 ns/op
--- BENCH: BenchmarkScheduling/2000Nodes/10000Pods-8
    scheduler_perf_test.go:119: Total time to add 2000 node 215.944315ms, 9261.646920 per second
    scheduler_perf_test.go:168: Total time to allocate 10000 containers in 23.320862587s, 428.800606 per second
--
BenchmarkScheduling/5000Nodes/10000Pods-8        	   10000	   7331169 ns/op
--- BENCH: BenchmarkScheduling/5000Nodes/10000Pods-8
    scheduler_perf_test.go:119: Total time to add 5000 node 780.679261ms, 6404.678912 per second
    scheduler_perf_test.go:168: Total time to allocate 10000 containers in 1m13.311881002s, 136.403539 per second
```

yangwwei:
This looks much nicer now, thanks for getting this done @TaoYang526 .

yangwwei:
I think we need to update the Developer guide accordingly to introduce how to run benchmark test and what does the result means, this can be a follow-up task.

TaoYang526:
Thanks @wilfred-s @yangwwei for the review and commit.

Issue #62, opened 2019-11-05
TestConfigWatcherExpiration UT fails intermittently


TaoYang526:
=== RUN   TestConfigWatcherExpiration
2019-11-05T18:20:48.643+0800       DEBUG  configs/configwatcher.go:92 configuration file unchanged
2019-11-05T18:20:49.646+0800       DEBUG  configs/configwatcher.go:92 configuration file unchanged
2019-11-05T18:20:50.647+0800       DEBUG  configs/configwatcher.go:92 configuration file unchanged
2019-11-05T18:20:51.651+0800       DEBUG  configs/configwatcher.go:92 configuration file unchanged
--- FAIL: TestConfigWatcherExpiration (5.01s)
    configwatcher_test.go:133: assertion failed: expression is false: waitForStopped(cw) == nil
FAIL

Referenced code:
```
// if configWatcher is not running, kick-off running it
// if configWatcher is already running, this is a noop
func (cw *ConfigWatcher) Run() {
	select {
	case cw.soloChan <- 0:
		ticker := time.NewTicker(1 * time.Second)
		quit := make(chan bool)
		go func() {
			for {
				select {
				case <-ticker.C:
					if !cw.runOnce() {
						<-cw.soloChan
						return
					}
				case <-quit:
					<-cw.soloChan
					ticker.Stop()
					return
				}
			}
		}()

		time.AfterFunc(cw.expireTime, func() {
			quit <- true
		})
	default:
		log.Logger().Info("config watcher is already running")
	}
}
```

The cause of this failure is that golang select statement will randomly choose a case when multiple cases can run, we can't guarantee that it can quit in 5 seconds which is the stop timeout defined in TestConfigWatcherExpiration.
I think perhaps we can add a new select statement with default case ahead of this select statement then move the quit case inside, so that it always check quit case first and will go forward to tick if not satisfied, in this way, it can quit after one tick interval at most after expired.
New code like this:
```
		go func() {
			for {
				select {
				case <-quit:
					<-cw.soloChan
					ticker.Stop()
					return
				default:
				}
				select {
				case <-ticker.C:
					if !cw.runOnce() {
						<-cw.soloChan
						return
					}
				}
			}
		}()
```
Looking forward to your feedbacks, thanks!

yangwwei:
Hi @TaoYang526 

It makes sense, can you submit a patch for this? Thanks a lot!

PR #72, opened 2019-11-21
Make sure config watcher can be stopped in time

TaoYang526:
This is a trivial improvement for config watcher to make sure it can be stopped in time and can solve  #62 (TestConfigWatcherExpiration UT fails intermittently).

### DIFF
--- a/pkg/common/configs/configwatcher.go
+++ b/pkg/common/configs/configwatcher.go
@@ -117,6 +117,7 @@ func (cw *ConfigWatcher) Run() {
 					ticker.Stop()
 					return
 				default:
+					break
 				}
 				select {
 				case <-ticker.C:
@@ -124,6 +125,10 @@ func (cw *ConfigWatcher) Run() {
 						<-cw.soloChan
 						return
 					}
+				case <-quit:
+					<-cw.soloChan
+					ticker.Stop()
+					return
 				}
 			}
 		}()
--- a/pkg/common/configs/configwatcher.go
+++ b/pkg/common/configs/configwatcher.go
@@ -112,15 +112,18 @@ func (cw *ConfigWatcher) Run() {
 		go func() {
 			for {
 				select {
+				case <-quit:
+					<-cw.soloChan
+					ticker.Stop()
+					return
+				default:
+				}
+				select {
 				case <-ticker.C:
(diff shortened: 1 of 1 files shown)


pkg/common/configs/configwatcher.go
```diff
@@ -111,16 +111,19 @@ func (cw *ConfigWatcher) Run() {
 		quit := make(chan bool)
 		go func() {
 			for {
+				select {
+				case <-quit:
+					<-cw.soloChan
+					ticker.Stop()
+					return
+				default:
```
yangwwei: Can we add a `break` for the default clause?
I guess they work the same way with or without break, but just want to make it an explicit jump out of select if that makes sense.

pkg/common/configs/configwatcher.go
```diff
@@ -111,16 +111,19 @@ func (cw *ConfigWatcher) Run() {
...
 				select {
 				case <-ticker.C:
 					if !cw.runOnce() {
 						<-cw.soloChan
 						return
 					}
-				case <-quit:
-					<-cw.soloChan
-					ticker.Stop()
-					return
```
wilfred-s: This case should not be removed. If you are in this select we could wait for the ticker to expire (1 sec) before we exit to get triggered in the select above.
This means that if the quit channel is checked, no quit, we move into this select and wait. If the quit is triggered just after we go into this 2nd select we wait 1 sec. A repeated run of the test shows that not having the select adds about 10 seconds on 100 runs. In other words it occurs 1 in every 10 cases.
Setting up a priority select should always follow the double select pattern etc

pkg/common/configs/configwatcher.go
wilfred-s: agree for clarity, not for functionality

TaoYang526: Thanks for the review, make sense to me.

pkg/common/configs/configwatcher.go
TaoYang526: Thanks for the detailed analysis, got it.

wilfred-s:
this resolves #62

wilfred-s:
Thank you @TaoYang526 for the contribution

PR #77, opened 2019-12-06
Format metric names of node usage to avoid validation failures

TaoYang526:
Names of node resource on K8s may contain some invalid characters for prometheus, for example: 'group/name' or 'group.name', we may encounter validation failures when registering node usage metrics. 
This PR proposes to Format metric names via replacing invalid characters(not in alpha-beta, numeric, '_' and ':' characters) with '\_' and add the prefix '\_' when metric name starts with a number.

### DIFF
--- a/pkg/metrics/init.go
+++ b/pkg/metrics/init.go
@@ -133,7 +133,9 @@ func GetQueueMetrics(name string) CoreQueueMetrics {
 	}
 }
 
-func FormatMetricName(metricName string) string {
+// Format metric name based on the definition of metric name in prometheus, as per
+// https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+func formatMetricName(metricName string) string {
 	if len(metricName) == 0 {
 		return metricName
 	}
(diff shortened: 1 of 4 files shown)


pkg/metrics/init.go
```diff
@@ -129,4 +131,24 @@ func GetQueueMetrics(name string) CoreQueueMetrics {
 		m.queues[name] = queueMetrics
 		return queueMetrics
 	}
+}
+
+func FormatMetricName(metricName string) string {
```
yangwwei: Can protect this function since this is supposed to be only called within the package
---> formatMetricsName

pkg/metrics/init.go
```diff
@@ -129,4 +131,24 @@ func GetQueueMetrics(name string) CoreQueueMetrics {
 		m.queues[name] = queueMetrics
 		return queueMetrics
 	}
+}
+
```
yangwwei: Can we add some comments here, saying this is according to https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels.

pkg/metrics/init.go
TaoYang526: Done

pkg/metrics/init.go
TaoYang526: Done

PR #80, opened 2020-01-03
Linter integration

wilfred-s:
This is a major update of the code consisting of 3 pull requests. One PR for each of the 3 repositories involved: core, k8shim and scheduler-interface.

The changes are all linked to the new usage of `golangci-lint` and issues found by those linter checks:
* gofmt
* goimports (multiple commits)
* gosimple and unused
* ID and other capitalisations (multiple commits)
* error, static, shadowing and simple lint checkers
* gosec
* copyright date update
* documentation and build integration

There are still issues open but this is a good base to start with.


yangwwei:
Hi @wilfred-s 

If I understand this correctly, we don't need to worry about the [license](https://github.com/golangci/golangci-lint/blob/master/LICENSE)
 of golangci-int, correct? Seems like we only need to use its library.

Second, this PR almost touches every file, how can we prevent it from messing up the commit history? Once committed, this will cause a lot of difficulties for us to track historical changes.

Another thing, there seem to be some other changes other than correcting format, such as

```
var (
   -endpoint = flag.String("endpoint", "unix://tmp/yunikorn.sock", "YuniKorn endpoint")
   +	endpoint = flag.String("endpoint", "tcp://localhost:3333", "YuniKorn endpoint")
)
```
how many of these are non-formatting changes?

wilfred-s:
Correct we do not have to worry about the license. It is used like a compiler or any other tool. We do not use the source or integrate any of the source into the project. It is not even used as a library.

To check out the commits, they have all been tagged with what they change:
- formatting fix: for the format and layout changes
- lint fix: for all issues found by the linter(s)
- copyright: just the copyright update
- lint: adding the documentation and lint config

Overall the lint fixes are probably one third to a half of all changes.
For example the change you noticed came out of the `unused` check. The `endpoint` variable was not used and is now properly hooked into the flag.

docs/coding-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+The wiki provides a good collection of comments from code reviews.
+Most of the comments will be checked in the automated checks described below.
+
+When using an IDE, like GoLand or Visual Studio Code, use the builtin options.
+Most IDEs will provide an extensive list of checks or formatting options that help formatting and point out code issues. 
+
+## Automated checks
+Not all code will be written using an IDE.
+Even between contributors the settings might not be the same in all installs.
+To help keep code formatted consistently a [lint](https://en.wikipedia.org/wiki/Lint_(software) tool is part of the code approval.
```
yangwwei: Are we missing a ")" for the link of lint here?

wilfred-s: yes it is a typo, fixed that and some other typos in the last commit

wilfred-s:
This PR is required for cloudera/yunikorn-k8shim#71 and uses cloudera/yunikorn-scheduler-interface#5

Issue #74, opened 2019-12-02
Overhaul of allocation, reservation and preemption logic


wangdatan:
**Why this issue?**

- The previous approach for allocation and preemption: 

```
Allocation --------------------------->
Preemption -> ComputeIdealAllocation ->  SingleStepSchedule. 
  
  SingleStepSchedule:
     -> FindAllocationAsks (which find a list of allocation requests)
     -> TryBatchAllocation (and pass in the allocation requests).
        -> allocate:
           -> (If it is preemption): CrossQueuePreemptionAllocate
             -> trySurgicalPreemptionOnNode 
           -> (If it is non-preemption): RegularAllocate. 

        -> Then use a complex algorithm [1] to blacklist requests 
```

What does the algorithm [1] looks like? 

```
    for failedAllocationKey := range failedToAllocationKeys {
        if preemptionParam.crossQueuePreemption {
            preemptionParam.blacklistedRequest[failedAllocationKey] = true
        } else {
            curWaitValue := m.waitTillNextTry[failedAllocationKey]
            if curWaitValue == 0 {
                curWaitValue = 2
            } else if curWaitValue < (1 << 20) {
                // Increase missed value if it is less than 2^20 (TODO need do some experiments about this value)
                curWaitValue <<= 1
            }
            m.waitTillNextTry[failedAllocationKey] = curWaitValue
        }
    }
```

After thought, I don't think it is correct, let's say there're 3 requests in a queue sorted by priority is R1 > R2 > R3.

```
1st time: Allocate order: R1 > R2 > R3. 
Assume, there's no resource available on the cluster to satisfy R1, R2.
So scheduler allocate 1 space for R3, and waitValue (see Algorithm [1]) of R1, R2 became 2 

2nd time: Allocation for R1 and R2 will be skipped, scheduler directly goes to R3 without looking at R1 and R2.
```

**Proposed Solution:**

Go to similar solution of YARN CapacityScheduler/FairScheduler introduce resource reservation on node to preserve allocation order. 

New function invoking structure considering allocation and preemption. 

```
SingleStepAllocation: 
   -> TryAllocate()

      for q in ordered-queues: 
         for app in ordered-apps:
            for req in app.requests: 
               if allocate(req) != nil: // allocated or reserved something 
                  break-the-nested-for-loop 

allocate(req): 
   // Very similar to existing regularAllocate, 
   // but reserve req on node if cannot allocate 
   // because of available space.


Preemption:
[...]

PR #81, opened 2020-01-03
Overhaul of reservation/allocation logic of scheduler

wangdatan:
This is PR for #74

Please note that this is still a working-in-progress patch, but the majority of the structure should be completed. What have done so far: 

1) Large restructure of scheduler logic, optimization of lock, etc. 
2) Reservation logic for large container with unit test.
3) Unreserve logic when request doesn't need or cannot be allocated. 

Expected regression from master: 
1) Multiple allocation per batch now disabled (which is not compatible with reservation logic). Instead, I would prefer to fix the lazy-sorting logic.

Plan to finish with the same PR: 
1) More tests to make sure allocation works. 
2) Lazy preemption logic with reservation.
3) Move reservation to a better node logic.

Plan to finish with a different PR: 
1) Sorting-related improvement.


yangwwei:
Hi @wangdatan 

Thanks for working on this, I have done a quick look at the core logic changes, and come up with a couple of high-level questions

**About reservation logic**
Currently, we are memorizing the best-node for the reservation. When the cluster is heavily utilized, this will cause efficiency problems, because it needs to iterate every node and calculate the corresponding score for each allocation. Also, I did not see the reserved resource is being counted during calculating the score for reservation, which can be an important factor.

Seems like we allow to reserve multiple requests per node. Do we need a stop sign for reservations? That tells the scheduler when to stop reserving resources on a node.

When we sort nodes by available resources, reserved resources are not counted. This means when we reserve some more resources on a node, the sorting algo is unaware of this and it will still put the node in the front of the list. 

**Internal structure**
This approach, the reservation/unreservation logic is internally managed inside of scheduling context, we are maintaining reserved requests in schedulingNode, schedulingApp, etc. This will create some limitations to support further advanced stuff. Such as the auto-scaling we were discussing. If we can pass the reservation state to k8s-shim, we can do lazy-pod-binding (reserve a pod on an un-exist node and wait for auto-scaler to provision the node). Do you think it is possible that we can follow the logic of generating allocation-proposal and commit to the cache? Just add another kind of allocation-proposal RESERVE. 

**About request priority**
I noticed this patch adds some logic to handle request priority, I feel this creates complexity. Not saying the code but to the users. How can we expect users to set reasonable priorities for their requests before understanding how this works.

**About recovery**
I feel we could not do anything in recovery for reserved requests. During restart, all these states will be lost. But as long as we can recover pending requests, that should be fine, we can just do reservations again once the scheduler is recovered. This is something we need to test out and make sure nothing breaks.

wangdatan:
Thanks @yangwwei, 

>  Currently, we are memorizing the best-node for the reservation. When the cluster is heavily utilized, this will cause efficiency problems, because it needs to iterate every node and calculate the corresponding score for each allocation.

It is actually done within the allocation call, since we already doing iterate the nodes and try to allocate on all of them, the additional logic of reservation-score should be relatively cheap to do. (Now the bottleneck is sorting nodes).

> Also, I did not see the reserved resource is being counted during calculating the score for reservation, which can be an important factor.
And, 
> Seems like we allow to reserve multiple requests per node. Do we need a stop sign for reservations? 

For current reservation (reservation during allocation) logic, we only allow one node reserved by at most one request. Multiple request reservation per node may happen only on preemption, rebalance of resources in the cluster, or autoscaling (scale-up). So we will take care of multiple resource request reservation on one node later. Currently I just added the logic to make sure multiple resource request can be easier added later.

> When we sort nodes by available resources, reserved resources are not counted.

This is a good point, will update it to make sure reserved nodes came last. 

>  When do unreservation at the beginning of each scheduling iteration, I think we can handle multiple allocations at a time, can we?

Did you mean handle multiple unreservation at a time? Yes we did. For multiple reservation or allocation, currently it is not yet supported. As we chatted before, we can wait for the out-of-band node sorting patch done to bring the performance back to ideal.

> Do you think it is possible that we can follow the logic of generating allocation-proposal and commit to the cache?
This is a good point, and I'm still exploring it. So far, my proposal is to strictly keep reservation logic and state inside scheduler, it should not visible by an outsider of component like cache/k8s. Once we export reservation logic to the outsider, I'm a bit worried we have to sync reservation logics like regular allocation, which prevent we do fast swapping reservation, drop reservation, etc. (we have to wait for an outsider component to confirm reservation-related changes, etc.). 
For autoscaling, I think we should be well-handled it. My current rough thought is to have a separate goroutine (like the preemption one), to add inferno node to the cluster, and do reservation based on that. There might be some issues we will encounter, but I didn't see any unsolvable issues so far.

> I noticed this patch adds some logic to handle request priority, I feel this creates complexity. Not saying the code but to the users.
I'm not sure if I understand your concern or not. It should be the same AskProto.Priority, which is same as k8s's pod priority. Application should have relative importance of its requests, user doesn't require understand but application framework need (like Spark/Flink or existing K8s deployment).  

For recovery: 
> I feel we could not do anything in recovery for reserved requests. During restart, all these states will be lost. 
Yes, this is by-design and scheduler can do reservation based on scheduler state. It is possible the new reservation result is different from the previous one, but since we strictly keep it internal to scheduler, I think it is fine. Sure we need to do more tests for that.

yangwwei:
> Once we export reservation logic to the outsider, I'm a bit worried we have to sync reservation logic like regular allocation, which prevents we do fast swapping reservation, drop reservation, etc. (we have to wait for an outsider component to confirm reservation-related changes, etc.).

There are pros and cons. The pros are
 - more consistent handling style
 - increase the usability and debuggability, users will be easier to see what's going on under the hood. E.g my pod is still pending, but it has reserved some resources on some node.

yangwwei:
There is another issue about unreservation, it looks like we only look at the nodes that have been reserved for `tryAllocateFromReservationRequest`. This could be a problem. If we reserve a request on a node where runs some long-running pods occupying the resources, then it is likely the request will keep waiting forever. Can we try the reserved node first, if not fit, then try all rest available nodes?

yangwwei:
**allocator.go**

1) line 272 - 274: redundant check

2) few naming confusions
- partitionCtx.HandleReservationProposal
- partitionCtx.HandleAllocationProposal
I would not call them proposals, in the earlier context, `AllocationProposal` is defined so the scheduler can submit proposals and cache layer decides to accept or reject them. Here, it is more like the internal handlings inside of the scheduler. I suggest to rename them to  something like: HandleReservedRequest/HandleUnreservedRequest/HandleAllocatedRequest.

3) It's a bit confusing when reading the code related to resources.
Can we organize the code like:
for `schediling_application`, `scheduling_node`, hiding the methods such as `DecAllocatingResource`, `incAllocatingResource`, `decreaseAllocatingResource`, make them all internal use within the struct. And expose some functions such as `addAllocation`, `addReservation`, `releaseAllocation`, `releaseReservation` for both app and node. So we can create some common methods in partitionContext, that can be reused in `allocator#handleSchedulingAllocation` and `scheduler#confirmOrRejectSingleAllocationProposal`. 

**scheduling_application.go**

1) line 258: I think it is possible when a request is reserved, but then got removed from `SchedulingApplication#Requests`. Then this line may cause nil pointer:

```
numToDrop := m.allocKeyToNumReservedRequests[allocKey] - int(m.Requests.requests[allocKey].PendingRepeatAsk)
```
can we add an extra check for the existence of `allocKey` in the map?

2) In `allocateForOneRequest`, my understanding is you want to calculate a score for each node when a request cannot fit. And then among these unfit nodes, find the best one to reserve. However, bestScore never updates which can cause updating the nodes all the time.

3) line 127: replace to `zap.Error`

4) line 183: unhandled error

5) `IncreaseAllocatingAndParentQueuesTestOnly`: this is marked for testing only, but I am curious why not just call `m.queue.IncAllocatingResourceFromTheQueueAndParents`? This is confusing.

**scheduling_partition.go**

1) line 317, 352 : this logging may cause nil pointer. We can leverage `zap.Error` to print the info, also shouldn't this be ERROR level?
```
    log.Logger().Info("Failed to handle reservation cancellation",
                zap.Error(err))
```

2)

**scheduling_node.go**

1) `fitInGapScore`: 
Currently, the score only is calculated based on resourceInShort/totalAsk, but I think it makes more sense to use formula `resourceInShort/nodeTotalResource`. Because when a node has more total resources available, it usually means the node has more chance to release more resources than smaller nodes. 

For example, let's say we have an ask {memory: 50}, node A has available {availableMemory: 10, totalMemory: 100}; node B has {availableMemory: 10, totalMemory: 200}. Using current scorer,  both nodes get score=1; but use proposed formula, scoreA=0.4, scoreB=0.2. which means B is a better fit.

2) `CheckAllocateConditions`
When check conditions, it calls `checkPredicateForRequest`, this will call some essential checks for the pod, such as `PodFitsResources(pod, meta, nodeInfo)`. This will fail when a node has no available resource for the pod. This is a conflict with reservation. Should we skip this? (The cost is the reservation might be invalidated during allocationOnReservation... not sure if there is a better way to handle this).

**requests.go**

1) line 44, 52, missing space between m and *AppSchedulingRequest

2) for `sortedRequestsByPriority []*SchedulingAllocationAsk`, this seems to be better replaced by a priority queue, e.g use container/heap.

**scheduling_allocation.go**
line 30: rename `Unreserve` to `Unreservation` or rename other 3 consts for consistency.

### Other

I found some issues while reviewing the patch.
RemoveApplications in UpdateRequest is not handled properly. We need to create a TODO item for this, including handle the removal of reserved requests for a removed application. This handles the scenario, e.g kubectl delete -f xxx.yaml. This is essential when we start to support some K8s operators.

### Nits

**scheduling_node.go**
171: redundant type float
228: Only allow -> only allow to

**scheduling_application.go**
line  363: incomplete comment

visibility issues
- `scheduling_application.go`
   TryAllocateFromReservationRequests -> tryAllocateFromReservationRequests
- `scheduling_partition.go`
   AddNewReservation -> addNewReservation
   GetReservationAppListClone -> getReservationAppListClone
   HandleUnreservedRequest -> handleUnreservedRequest 
   HandlePendingResourceDecreaseForQueues -> handlePendingResourceDecreaseForQueues
- `scheduling_queue.go`
   IncPendingResource -> incPendingResource 
   DecPendingResourceFromTheQueueAndParents -> decPendingResourceFromTheQueueAndParents
   AddSchedulingApplication -> addSchedulingApplication
   RemoveSchedulingApplication -> removeSchedulingApplication

- `scheduling_node_test.go`
   There are some incorrect comments 

function receivers should be consistent, we need to correct `SchedulingApplication`, `SchedulingNode`, `SchedulingQueue`,

pkg/scheduler/scheduler.go
```diff
@@ -141,21 +122,38 @@ func (m *Scheduler) updateSchedulingRequest(schedulingAsk *SchedulingAllocationA
     return err
 }
 
-func (m *Scheduler) updateSchedulingRequestPendingAskByDelta(allocProposal *commonevents.AllocationProposal, deltaPendingAsk int32) error {
-    m.lock.Lock()
-    defer m.lock.Unlock()
+// Confirm or reject allocation proposal, when allocationAdded > 0, it is confirm; otherwise it is reject.
+// Accept value of allocationAdded is either +1 or -1
+// This will update node, app, and queues
+func (m *Scheduler) confirmOrRejectSingleAllocationProposal(allocProposal *commonevents.AllocationProposal, allocationAdded int32) error {
```
wilfred-s: Can we split this in 3 methods:
```
confirmAllocationProposal(allocProposal *commonevents.AllocationProposal) error {
    return processAllocationProposal(allocProposal, true)
}
rejectAllocationProposal(allocProposal *commonevents.AllocationProposal) error {
    return processAllocationProposal(allocProposal, false)
}
processAllocationProposal(allocProposal *commonevents.AllocationProposal, confirm bool) error {
.....
}
```
That makes it clear whet we are doing.

pkg/scheduler/scheduler.go
```diff
@@ -315,26 +308,24 @@ func (m *Scheduler) processAllocationUpdateEvent(ev *schedulerevent.SchedulerAll
...
 
-    // Allocations events cannot contain accepted and rejected allocations at the same time.
-    // There should never be more than one accepted allocation in the list
-    // See cluster_info.processAllocationProposalEvent()
     if len(ev.AcceptedAllocations) > 0 {
-        alloc := ev.AcceptedAllocations[0]
-        // decrease the outstanding allocation resources
-        m.clusterSchedulingContext.updateSchedulingNodeAlloc(alloc)
+        for _, alloc := range ev.AcceptedAllocations {
+            // decrease the outstanding allocation resources
```
wilfred-s: There is no change on the event generation side so this change does not make sense, the event can still only contain 1 accepted allocation.

wangdatan:
@yangwwei 

Thanks for commenting: 

- There're a few suggestions to use ERROR instead of DEBUG or INFO, I didn't change that because they're not error (they could cause by change of scheduling state in multi-thread env).
- I think I have resolved all other lints as much as I could.

> It's a bit confusing when reading the code related to resources. Can we organize the code like: .. 

This part seems like a lot of work, can we do it separately? 

> CheckAllocateConditions
When check conditions, it calls checkPredicateForRequest

This sounds like a big issue, should we skip PodFitsResources for all scenarios? Since we have checked inside YuniKorn already.

> RemoveApplications in UpdateRequest is not handled properly. We need to create a TODO item for this, including handle the removal of reserved requests for a removed application. 

I think this is handled. See TestReservationRemoveAppThenRemoveNode

wangdatan:
@wilfred-s  thanks for adding your comments, mostly make sense to me. A few thoughts: 

For following two I will check it in later patches: 

> please check grammar in comments (sentences sometimes do not make sense)
commented out variables or code should be removed instead
> files are missing a newline at the end (shown by a little red sign in the GitHubUI) please add them

For comments regarding to unsafe pointer, now I changed to atomic.value, please let me know what you think. 

> The shim (RM) has signalled that the node should be removed. Why would we need to then inform the RM which allocations need to be removed?

To me this is important since there're running containers on this node. This will make allocation removal to be consistent no matter if it is released by application or removed when node removed. YARN did the same too (when NM removed, app get noficiation of both node removal and container completed events).

>  Using an atomic means that the status can be changed while we schedule as it is no longer in a lock. Atomicity does not guarantee, is any ordering of observability of values, it just guarantees correctness of the memory value.

Updated to mutex. 

> Adding a whole package for one call to IntMin is a bit much, cleaner solution would be to add the min function locally to the code that needs it.

Removed math.go

> ReservedSchedulingRequest Related questions 

I think I have updated, please check. 

> Why do we store the whole SchedulingNode here? 

We still need to use the node reference, otherwise, get a node reference inside an application will be hard (requires lock app and then lock partition). Please let me know what you think. 

> Can we split this in 3 methods:
I end up with split this to 2 :) 

> There is no change on the event generation side so this change does not make sense, the event can still only contain 1 accepted allocation.

Yeah, i knew but i tend to make it to be more generic so once we add multiple allocations we don't have to change the code. I can change it back also since I don't have strong opinion for that.

wangdatan:
Please find the latest PR and branch.

yangwwei:
Move consolidated PR to #89

PR #82, opened 2020-01-17
During recovery existing allocations should inherits queue name from the application

yangwwei:
When placement rule is enabled, job/task's queue name may or may not equal to queueName in pod's spec. For both fresh start and recovery case, we should enforce following rules
- app queue name is determined by the Placement Rule Manager
- task queue name is inherited from app

this way we can handle targetting queue consistently. Otherwise, we saw some failures like:

2020-01-17T19:10:40.329Z WARN cache/cluster_info.go:338 Failure while adding new node, node rejected with error queue does not exist or is not a leaf queue root.default

causing recovery timeout.


yangwwei:
The fix has a UT attached to reproduce the issue. @wilfred-s , please help to review this fix.

PR #88, opened 2020-02-07
[YUNIKORN-10] add pluggable app management design doc

yangwwei:
doc preview: https://github.com/apache/incubator-yunikorn-core/blob/design/docs/design/pluggable-app-management.md
code changes are in https://github.com/apache/incubator-yunikorn-k8shim/pull/72


wilfred-s:
When I was reading the doc I had found that `The Problem` should have been first in the text, explain what problem we are trying to fix. 
* Describing the problem
The current way we schedule and group an application is based on a label on the pod. This generic way works for any type of workload. It does however give us a limited information on the lifecycle and application. On the kubernetes side operators have been introduced to provide more detail on the application and help scheduling. We cannot use them currently and want to add that functionality.
* Introduce the operator
The current first part of the doc: CRDs etc
Should we mention Flink as one of the operators too?
There is a need to fit/compare the current generic way into the operator: what is the difference just the CRDs?
* Design
Keep as is. No comments on that part

yangwwei:
hi @wilfred-s 

I have updated the doc, pls take another look. Thx

yangwwei:
hi @wilfred-s 
does this look good now?

PR #92, opened 2020-02-28
fix a scheduler deadlock 

yangwwei:
looks like we are having a deadlock situation when
1. scheduler is running the `trySchedule` loop
2. scheduler is trying to recover some existing allocations, calling `scheduling_application#recoverOnNode`

in the 1st call, it always tries to acquire the queue's lock, then app's lock;
in the 2nd call, it tries to acquire the app's lock then queue's lock

then it's easy to hit deadlock when they are running concurrently.

### DIFF
--- a/pkg/cache/node_info.go
+++ b/pkg/cache/node_info.go
@@ -103,6 +103,12 @@ func (ni *NodeInfo) GetAvailableResource() *resources.Resource {
 	return ni.availableResource.Clone()
 }
 
+func (ni *NodeInfo) GetCapacity() *resources.Resource {
+	ni.lock.RLock()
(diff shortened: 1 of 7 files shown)


yangwwei:
Since we are using the fine-grained locking mechanism, I am suggesting to fix this issue with the changes in this PR. The idea is to avoid calling `sa.updateAskRepeatInternal()` in `recoverOnNode()`. I was first trying to move the app lock inside of  `updateAskRepeatInternal()` but that doesn't work because this "internal" call is assumed to be app-lock-free.

wilfred-s:
This has been superseded by YUNIKORN-6

PR #96, opened 2020-03-04
[YUNIKORN-1] Support app/task priority aware scheduling

TaoYang526:
This is an initial implementation based on the latest design doc in YUNIKORN-1

@yangwwei @wilfred-s could you please help to review this PR? Thanks.


yangwwei:
Hi @TaoYang526 

Thanks for the patch, I like the patch does a good abstraction fo requests, so it only creates small footprints on core-scheduling code path. However, the implementation of `SortableLinkedMap` is a bit over-complex to me. I am thinking if we can simplify this. A few high-level comments:

1. please create `common/maps` for map related files and the `SortableLinkedMap` should have its own file. The layout can look like below

```
  common/
      maps/
          map.go
          sortablelinkedmap.go
          sortablelinkedmap_test.go
```
For `SortableLinkedMap`

The following func seems like a `compareFunc()`, the name of `isPreFunc()` is a bit confusing.
```
isPreFunc func(i, j interface{}) bool
```

2. I feel the `SortableLinkedMap` can be simplified with the following options

2.1

TaoYang526:
Thanks @yangwwei for the review and comments.
Replay about the comments
1. LGTM
2.1 matchFunc helps to locate the first pending request so that we can quickly find out the top priority group and the first pending request in a priority group, otherwise we have to iterate all requests to find pending request for every cycle.
2.2 isFixedOrder can be used for different scenarios, for now the priority of request can't be modified so that we can set isFixedOrder to true to avoid reordering this request in the map, but if we support updating the priority in the future, we can set isFixedOrder to false. But it's fine for me to move it and not support resorting when updating the entry.
3 We have lots of operations need to get request by allocationKey, if not keep all requests in map, we have to find it from priority groups, perhaps it's fine to keep references in several places for better performance?

TaoYang526:
Addressed comments from @yangwwei  and @wangdatan.
Updates in the latest commit:
(1) Remove resort & isFixedOrder logic to simplify SortableLinkedMap.
(2) Merge PriorityFifo and Fifo policies.
(3) Support validating configured app-sort-policy.
(4) Update UTs.

TaoYang526:
@yangwwei  @wilfred-s 
To confirm that the performance won't regress unexpectedly, I have added benchmark tests for apps sorting and pending requests iteration in latest PR.
For apps sorting, it tests the performance of sorting pending apps with different policies, test results in my local environment like this (unit of consuming-time is µs):

  |   | 10 apps | 100 apps | 1000 apps
-- | -- | -- | -- | --
Before | Fair | 8.9 | 118.9 | 1129.0
  | Fifo | 3.9 | 44.2 | 657.8
After | Fair | 8.6 | 125.2 | 1203.0
  | Fifo | 4.7 | 81.2 | 1483.0

There is  little change for Fair policy, and there's not too much change for Fifo policy, which is expected since now Fifo has more things to be considered.

For pending requests iteration, it tests get and iterate 10,000 sorted pending requests with different number of priorities,  expect the performance is only related to the number of pending requests, won't be affected by other factors such as the number of priorities and pending percentage, and should have better performance than before, test results in my local environment like this (unit of consuming-time is µs):

  | 1 priority | 2 priorities | 5 priorities | 10 priorities
-- | -- | -- | -- | --
Before | 883.4 | 1642.5 | 2774.3 | 3246.9
After | 288.7 | 279.9 | 286.3 | 292.3

According to the results, it will improve a lot than before, at least 3 times enhancement and more priorities leads to more enhancement. Moreover, only priority is considered for sorting before, if it supports sorting by createTime as in this PR, the consuming-time will be much more(about 6ms for 10,000 requests).

wilfred-s:
Moving to draft for now, we'll have another good look when the flaky tests are fixed and we have travis running

wilfred-s:
The core scheduler design has completely changed since this PR started. We need to start from scratch on this one.

PR #97, opened 2020-03-06
Fix flaky unit tests on master branch

yangwwei:


wilfred-s:
The fact that the tests now pass shows that there is a timing issue and that the code is not really broken.
Not sure how to track this down.

yangwwei:
I had some time yesterday so I created this PR to narrow down the UT issue.
In our smoke test, TestBasicScheduler, TestReservation stuff can fail intermittently. I tried to run these tests locally lots of times, I can sometimes reproduce. E.g

`go test -count=20 -run TestBasicScheduler`

By increasing the count number, I see more weird things happening. So far, not many progress.. We've been such smoke tests for a while, there must be something changed after https://github.com/apache/incubator-yunikorn-core/pull/89.

PR #102, opened 2020-03-15
[YUNIKORN-34] Add document to explain benchmark result.

yangwwei:
Add doc for benchmark


yangwwei:
Hi @TaoYang526 

Can you please help to take a look at this?
I spent some time to summarize the results in this doc. the purpose is for other people's reference.
Please help to check and let me know your comments, thanks!

TaoYang526:
Thanks @yangwwei for this effort.
This doc is very clear and easy to understand, just a small place need to be updated in the Scheduler Throughput section: "launching 50,000 pods with heterogeneous resource requests", 
 the prerequisite of this test is that all requests can be satisfied so that heterogeneous resource requests seem unnecessary,  even without heterogeneous requests, there may be different cases for this test, such as one app, many apps in a queue, many apps among several queues, and we just tested the simplest case that only have one app with 50,000 requests, and the results in other scenarios may be different but should be not much (since the most time-consuming phase is node sorting). So I think we can just leave "launching 50,000 pods" without detailed description, thoughts?

yangwwei:
Hi @TaoYang526 

Thanks for the comments, please see my updates in the new commit.

PR #103, opened 2020-03-17
[YUNIKORN-28] Support validating conf via REST API

TaoYang526:
As discussed in https://github.com/apache/incubator-yunikorn-k8shim/pull/81, we need to validate conf via REST API in scheduler core.


pkg/webservice/dao/validate_info.go
```diff
@@ -0,0 +1,24 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package dao
+
+type ValidateConfResponse struct {
+    Allowed bool
+    Error   string
```
yangwwei: Rename `Error` to `Reason`, error usually means an error type, reason can be a string.
Can we also add the annotations: e.g `json:"allowed"`, for serialization/deserilization

pkg/webservice/handlers.go
```diff
@@ -118,6 +120,24 @@ func GetNodesInfo(w http.ResponseWriter, r *http.Request) {
...
+	}
+	var result dao.ValidateConfResponse
+	if err != nil {
+		result.Allowed = false
+		result.Error = err.Error()
+	} else {
+		result.Allowed = true
+	}
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		panic(err)
```
yangwwei: Hmmm, we should not have any `panic` here.
Can we replace all panic to `http.Error()`?

pkg/webservice/dao/validate_info.go
TaoYang526: make sense. Thanks.

pkg/webservice/handlers.go
TaoYang526: Here is just like other handers, I'm not sure is it intentional, may be panic can be caught by the framework of rest server?

yangwwei: Sorry I was not quite clear. I know, this is a historic issue, not a new issue with this PR.
We should not have any panic in our code, if something really goes wrong, we can send the error via `http.Error()`.

TaoYang526: @yangwwei You are right, http.Error is the better way to handle this. I have tested panic and http.Error: (1) When panic happened, the scheduler process didn't stop (the connection is running in a new go-routine and I think the panic must be handled), and client got error like this: `curl: (52) Empty reply from server`; (2) When I replaced it with http.Error like this: `http.Error(w, err.Error(), http.StatusInternalServerError)`, client can got the error message in the body of response but status code is 200, cased by we have updated the status code before handling this request in handlers.go#writeHeaders, it can be fixed when I removed the last line: `w.WriteHeader(http.StatusOK)`.
Now all related places have been corrected, please take a look again. Thanks.

yangwwei: Oh, that's a bug. Thanks for finding this out.
The fix looks all good to me. Approving the PR.

yangwwei:
LGTM, merging this now.

PR #107, opened 2020-03-24
[YUNIKORN-30] fix flaky unit test

wilfred-s:
Production code fixes for incorrect logging only
Clean up of smoke test logic and unused code
Fix lint testing to use head ref from git


pkg/scheduler/scheduler.go
```diff
@@ -548,6 +548,8 @@ func (s *Scheduler) GetClusterSchedulingContext() *ClusterSchedulingContext {
 // Visible by tests
 func (s *Scheduler) MultiStepSchedule(nAlloc int) {
 	for i := 0; i < nAlloc; i++ {
+		log.Logger().Debug("Scheduler manual stepping",
+			zap.Int("count", i))
```
yangwwei: Can we remove this?
Even this is DEBUG, it would print too many lines of logs.

PR #108, opened 2020-03-24
[YUNIKORN-47] Support node update action

yangwwei:
We need to support the live updates of node capacity.
There are multiple use cases for this:
1) users could reconfigure kubelet with new capacity (e.g reserve more resources for system usage). See more in https://kubernetes.io/docs/tasks/administer-cluster/reconfigure-kubelet/#reconfiguring-the-kubelet-on-a-running-node-in-your-cluster, and https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/. We need to update our node capacity accordingly.
2) use this API to dynamically change node capacity to exclude resources allocated by other schedulers (see more in https://issues.apache.org/jira/browse/YUNIKORN-16)


pkg/cache/node_info.go
```diff
@@ -109,6 +112,40 @@ func (ni *NodeInfo) GetCapacity() *resources.Resource {
...
+	ni.availableResource.SubFrom(ni.occupiedResource)
+}
+
+func (ni *NodeInfo) GetOccupiedResource() *resources.Resource {
+	ni.lock.RLock()
+	defer ni.lock.RUnlock()
+	return ni.occupiedResource.Clone()
+}
+
+func (ni *NodeInfo) SetOccupiedResource(occupiedResource *resources.Resource) {
```
wilfred-s: Should not be exported

pkg/cache/node_info.go
```diff
@@ -109,6 +112,40 @@ func (ni *NodeInfo) GetCapacity() *resources.Resource {
 	return ni.totalResource.Clone()
 }
 
+func (ni *NodeInfo) SetCapacity(newCapacity *resources.Resource) {
```
wilfred-s: Should not be exported

pkg/cache/node_info.go
```diff
@@ -109,6 +112,40 @@ func (ni *NodeInfo) GetCapacity() *resources.Resource {
...
+	ni.lock.Lock()
+	defer ni.lock.Unlock()
+	if resources.Equals(ni.totalResource, newCapacity) {
+		log.Logger().Info("skip updating capacity, not changed")
```
wilfred-s: Info seems to high to log this, Debug is better

pkg/cache/node_info.go
```diff
@@ -109,6 +112,40 @@ func (ni *NodeInfo) GetCapacity() *resources.Resource {
...
+	ni.lock.RLock()
+	defer ni.lock.RUnlock()
+	return ni.occupiedResource.Clone()
+}
+
+func (ni *NodeInfo) SetOccupiedResource(occupiedResource *resources.Resource) {
+	ni.lock.Lock()
+	defer ni.lock.Unlock()
+	if resources.Equals(ni.occupiedResource, occupiedResource) {
+		log.Logger().Info("skip updating occupiedResource, not changed")
```
wilfred-s: Info seems to high to log this, Debug is better

pkg/cache/node_info.go
```diff
@@ -109,6 +112,40 @@ func (ni *NodeInfo) GetCapacity() *resources.Resource {
...
+	defer ni.lock.Unlock()
+	if resources.Equals(ni.totalResource, newCapacity) {
+		log.Logger().Info("skip updating capacity, not changed")
+		return
+	}
+
+	ni.totalResource = newCapacity
+	ni.availableResource = ni.totalResource.Clone()
+	ni.availableResource.SubFrom(ni.allocatedResource)
+	ni.availableResource.SubFrom(ni.occupiedResource)
```
wilfred-s: Move to a separate function `setAvailabaleResource()` and call from:
* SetCapacity()
* SetOccupiedResource()
* NewNodeInfo()

Keeps the logic the same everywhere.

pkg/cache/node_info.go
```diff
@@ -56,6 +58,7 @@ func NewNodeInfo(proto *si.NewNodeInfo) *NodeInfo {
 		NodeID:            proto.NodeID,
 		totalResource:     resources.NewResourceFromProto(proto.SchedulableResource),
 		allocatedResource: resources.NewResource(),
+		occupiedResource:  resources.NewResource(),
```
wilfred-s: Are you sure that this is always zero when we register a new node or could we have an allocation already?

pkg/cache/partition_info.go
```diff
@@ -477,12 +477,17 @@ func (pi *PartitionInfo) releaseAllocationsForApplication(toRelease *commonevent
 	for _, alloc := range allocationsToRelease {
 		// remove allocation from node
 		node := pi.nodes[alloc.AllocationProto.NodeID]
+		// node could be nil if when release happens, the node is already
+		// removed from the cluster. in this case, we need to make sure
+		// resources are released correctly. otherwise these resources will
+		// be leaked.
```
wilfred-s: This cannot happen, and is impossible.
When a node is removed the allocations are released by calling `removeNodeAllocations()` at that point they are passed back and should have been released. The queue will get out of sync if we do this twice.

pkg/scheduler/scheduler.go
```diff
@@ -199,10 +199,12 @@ func (s *Scheduler) processAllocationReleaseByAllocationKey(allocationAsksToRele
 				log.Logger().Info("release allocation",
 					zap.String("allocation", toRelease.Allocationkey),
 					zap.String("appID", toRelease.ApplicationID),
-					zap.String("message", toRelease.Message))
+					zap.String("message", toRelease.Message),
+					zap.Int("ReservedToRelease", reservedAsks))
```
wilfred-s: `ReservedToRelease` should be `reservedAskReleased`

pkg/scheduler/scheduler.go
```diff
@@ -199,10 +199,12 @@ func (s *Scheduler) processAllocationReleaseByAllocationKey(allocationAsksToRele
...
 				// update the partition if the asks were reserved (clean up)
 				if reservedAsks != 0 {
-					s.clusterSchedulingContext.getPartition(toRelease.PartitionName).unReserveUpdate(toRelease.ApplicationID, reservedAsks)
+					s.clusterSchedulingContext.getPartition(toRelease.PartitionName).
+						unReserveUpdate(toRelease.ApplicationID, reservedAsks)
```
wilfred-s: we do not limit line length I don't think we should split this

pkg/scheduler/tests/scheduler_operation_test.go
```diff
@@ -0,0 +1,514 @@
...
+                vcore: 20
+`
+	// Start all tests
+	ms := &mockScheduler{}
+	defer ms.Stop()
+
+	err := ms.Init(configData, false)
+	if err != nil {
+		t.Fatalf("RegisterResourceManager failed: %v", err)
+	}
```
wilfred-s: We should use `assert.NilError(t, err, "RegisterResourceManager failed")`
I started to move all smoke tests to use this assert

pkg/scheduler/tests/scheduler_operation_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+	// Check queues of cache and scheduler.
+	partitionInfo := ms.clusterInfo.GetPartition("[rm:123]default")
+	assert.Assert(t, nil == partitionInfo.Root.GetMaxResource(), "partition info max resource nil")
```
wilfred-s: no Yoda expressions: partiotion.. == nil

pkg/scheduler/tests/scheduler_operation_test.go
```diff
@@ -0,0 +1,514 @@
...
+	waitForPendingAppResource(t, schedulingApp, 20, 1000)
+
+	// Register a node
+	err = ms.proxy.Update(&si.UpdateRequest{
+		NewSchedulableNodes: []*si.NewNodeInfo{
+			{
+				NodeID: "node-1:1234",
+				Attributes: map[string]string{
+					"si.io/hostname": "node-1",
+					"si.io/rackname": "rack-1",
```
wilfred-s: unused in tests should be removed and made an empty map

pkg/scheduler/tests/scheduler_utils.go
```diff
@@ -103,6 +103,20 @@ func waitForNodesAllocatedResource(t *testing.T, cache *cache.ClusterInfo, parti
...
+	var totalNodeResource resources.Quantity
+	err := common.WaitFor(10*time.Millisecond, time.Duration(timeoutMs)*time.Millisecond, func() bool {
+		totalNodeResource = 0
+		for _, nodeID := range nodeIDs {
+			totalNodeResource += cache.GetPartition(partitionName).GetNode(nodeID).GetAvailableResource().Resources[resources.MEMORY]
+		}
+		return totalNodeResource == availableMemory
+	})
+	if err != nil {
+		t.Fatalf("Failed to wait for available resource %v and node %v, called from: %s", availableMemory, nodeIDs, caller())
```
wilfred-s: use `assert.NilError(t, err, "...")

pkg/scheduler/scheduling_partition.go
```diff
@@ -308,6 +308,19 @@ func (psc *partitionSchedulingContext) addSchedulingNode(info *cache.NodeInfo) {
...
+func (psc *partitionSchedulingContext) updateSchedulingNode(info *cache.NodeInfo) {
+	if info == nil {
+		return
+	}
+
+	psc.Lock()
+	defer psc.Unlock()
+	// check consistency and reset to make sure it is consistent again
+	if schedulingNode, ok := psc.nodes[info.NodeID]; ok {
+		schedulingNode.updateNodeInfo(info)
```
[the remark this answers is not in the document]
yangwwei: The comment was misleading. I have removed it.
I added a WARN log when the node is not there but we attempt to do an update.

pkg/scheduler/scheduling_node.go
```diff
@@ -74,13 +74,28 @@ func (sn *SchedulingNode) GetReservations() []string {
...
+	sn.RLock()
+	defer sn.RUnlock()
 	return sn.nodeInfo.GetAllocatedResource()
 }
 
+// expose this for tests
+func (sn *SchedulingNode) GetAvailableResourceForTest() *resources.Resource {
+	return sn.getAvailableResource()
+}
+
```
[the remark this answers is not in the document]
yangwwei: Yes, we can do that. This simplifies the contract.
I have renamed `getAvailableResource()` to `GetAvailableResource ()`

pkg/cache/node_info.go
yangwwei: Done

pkg/cache/node_info.go
yangwwei: Done

pkg/cache/node_info.go
yangwwei: Done

pkg/cache/node_info.go
yangwwei: done

pkg/cache/node_info.go
yangwwei: Make sense to me.
I moved to code to:

```
// refresh node available resource based on the latest allocated and occupied resources.
// this call assumes the caller already acquires the lock.
func (ni *NodeInfo) refreshAvailableResource() {
	ni.availableResource = ni.totalResource.Clone()
	ni.availableResource.SubFrom(ni.allocatedResource)
	ni.availableResource.SubFrom(ni.occupiedResource)
}
```

pkg/cache/node_info.go
yangwwei: The existing allocations for non-yunikorn-scheduled pods will be reported through UPDATE call. Not part of existing allocations.

pkg/scheduler/scheduler.go
yangwwei: Done

pkg/scheduler/scheduler.go
yangwwei: reverted the change.

pkg/scheduler/tests/scheduler_operation_test.go
yangwwei: Ah, I did not know this short cut. Let me use this for the tests added

pkg/scheduler/tests/scheduler_operation_test.go
yangwwei: done

pkg/scheduler/tests/scheduler_utils.go
yangwwei: yeah, but we might need another cleanup JIRA for this. Given all checks in this file are using a similar way.

pkg/scheduler/tests/scheduler_operation_test.go
yangwwei: fixed

yangwwei:
Hi @wilfred-s  I have made the changes, please check.
The UT failure seems not to relate to my change, as it was failing at https://github.com/apache/incubator-yunikorn-core/pull/110 too, that was just some images update.
We might need to look at this separately.

pkg/cache/node_info.go
wilfred-s: perfect that is what I wanted to know

pkg/cache/partition_info.go
```diff
@@ -477,12 +477,17 @@ func (pi *PartitionInfo) releaseAllocationsForApplication(toRelease *commonevent
...
-				zap.Any("allocation", alloc))
-			continue
+			log.Logger().Info("node is not found or the allocation is not found",
+				zap.String("node", alloc.AllocationProto.NodeID),
+				zap.String("allocationID", alloc.AllocationProto.UUID))
+		} else {
+			node.RemoveAllocation(alloc.AllocationProto.UUID)
 		}
-		node.RemoveAllocation(alloc.AllocationProto.UUID)
 		totalReleasedResource.AddTo(alloc.AllocatedResource)
```
wilfred-s: This needs to be reverted too. If we have a nil node we need to continue. It is almost an error case but we can only update if we have a node and an alloc.
The queue will get out of sync if we remove the alloc while the node is nil. The nil node is a result of the queue already being updated.

pkg/scheduler/tests/scheduler_utils.go
wilfred-s: filed follow up jira: YUNIKORN-53

wilfred-s:
Unit tests passed on second run. Not sure what happened but even a simple partition create failed in one test causing one of the failures. This is not the test code but an environment issue.

pkg/cache/partition_info.go
yangwwei: Sorry, I missed this one earlier. Just done in the latest commit.

wilfred-s:
changes committed closing PR

PR #109, opened 2020-03-25
[YUNIKORN-14] Add rest API to retrieve app/container history info

adamantal:
This PR only handles the server (core) part of this issue. The UI related changes will be handled in [YUNIKORN-8](https://issues.apache.org/jira/browse/YUNIKORN-8).

The initial implementation consists of creating a `HistoricalPartitionInfo` object which stores some metrics (currently the number of application and containers). This is updated in 10 minutes by `HistoricalPartitionInfoUpdater`

If there's need the poll time can be configurable, right now it is just hardcoded to 10 minutes, and we will keep 1 day of history (144 items).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.9.1
 	github.com/satori/go.uuid v1.2.0
-	github.com/stretchr/testify v1.5.1
+	github.com/stretchr/testify v1.5.1 //indirect
 	go.uber.org/atomic v1.5.1 // indirect
 	go.uber.org/multierr v1.4.0 // indirect
 	go.uber.org/zap v1.13.0
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/apache/incubator-yunikorn-scheduler-interface v0.0.0-20200327234544-149aaa3d8e48
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.3.5 // indirect
 	github.com/gorilla/mux v1.7.3
 	github.com/kr/text v0.2.0 // indirect
 	github.com/looplab/fsm v0.1.0
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.9.1
 	github.com/satori/go.uuid v1.2.0
-	github.com/stretchr/testify v1.5.1 // indirect
+	github.com/stretchr/testify v1.5.1
 	go.uber.org/atomic v1.5.1 // indirect
 	go.uber.org/multierr v1.4.0 // indirect
 	go.uber.org/zap v1.13.0
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/apache/incubator-yunikorn-scheduler-interface v0.0.0-20200327234544-149aaa3d8e48
-	github.com/golang/protobuf v1.3.5 // indirect
+	github.com/golang/protobuf v1.3.5
 	github.com/gorilla/mux v1.7.3
 	github.com/kr/text v0.2.0 // indirect
 	github.com/looplab/fsm v0.1.0
@@ -13,6 +13,7 @@ require (
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/prometheus/client_golang v1.4.0
+	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.9.1
 	github.com/satori/go.uuid v1.2.0
 	github.com/stretchr/testify v1.5.1 // indirect
--- a/pkg/entrypoint/entrypoint.go
+++ b/pkg/entrypoint/entrypoint.go
@@ -22,6 +22,8 @@ import (
 	"github.com/apache/incubator-yunikorn-core/pkg/cache"
 	"github.com/apache/incubator-yunikorn-core/pkg/handler"
 	"github.com/apache/incubator-yunikorn-core/pkg/log"
+	"github.com/apache/incubator-yunikorn-core/pkg/metrics"
+	"github.com/apache/incubator-yunikorn-core/pkg/metrics/history"
 	"github.com/apache/incubator-yunikorn-core/pkg/rmproxy"
 	"github.com/apache/incubator-yunikorn-core/pkg/scheduler"
 	"github.com/apache/incubator-yunikorn-core/pkg/webservice"
@@ -57,6 +59,9 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
 	scheduler := scheduler.NewScheduler(cache)
 	proxy := rmproxy.NewRMProxy()
 
+	imHistory := history.NewInternalMetricsHistory(1440)
+	metricsCollector := metrics.NewInternalMetricsCollector(imHistory)
+
 	eventHandler := handler.EventHandlers{
 		CacheEventHandler:     cache,
 		SchedulerEventHandler: scheduler,
@@ -68,6 +73,7 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
 	cache.StartService(eventHandler)
 	scheduler.StartService(eventHandler, opts.manualScheduleFlag)
 	proxy.StartService(eventHandler)
+	metricsCollector.StartService()
 
 	context := &ServiceContext{
 		RMProxy:   proxy,
@@ -77,7 +83,7 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
 
 	if opts.startWebAppFlag {
 		log.Logger().Info("ServiceContext start web application service")
-		webapp := webservice.NewWebApp(cache)
+		webapp := webservice.NewWebApp(cache, imHistory)
 		webapp.StartWebApp()
 		context.WebApp = webapp
 	}
--- a/pkg/entrypoint/entrypoint.go
+++ b/pkg/entrypoint/entrypoint.go
@@ -33,6 +33,7 @@ import (
 type StartupOptions struct {
 	manualScheduleFlag bool
 	startWebAppFlag    bool
+	metricsHistorySize int
 }
 
 func StartAllServices() *ServiceContext {
@@ -41,6 +42,7 @@ func StartAllServices() *ServiceContext {
 		StartupOptions{
 			manualScheduleFlag: false,
 			startWebAppFlag:    true,
+			metricsHistorySize: 1440,
 		})
 }
 
@@ -51,6 +53,7 @@ func StartAllServicesWithManualScheduler() *ServiceContext {
 		StartupOptions{
 			manualScheduleFlag: true,
 			startWebAppFlag:    false,
+			metricsHistorySize: 0,
 		})
 }
 
@@ -58,9 +61,6 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
 	cache := cache.NewClusterInfo()
 	scheduler := scheduler.NewScheduler(cache)
 	proxy := rmproxy.NewRMProxy()
-
-	imHistory := history.NewInternalMetricsHistory(1440)
-	metricsCollector := metrics.NewInternalMetricsCollector(imHistory)
 
 	eventHandler := handler.EventHandlers{
 		CacheEventHandler:     cache,
@@ -73,12 +73,17 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
 	cache.StartService(eventHandler)
 	scheduler.StartService(eventHandler, opts.manualScheduleFlag)
 	proxy.StartService(eventHandler)
-	metricsCollector.StartService()
 
 	context := &ServiceContext{
 		RMProxy:   proxy,
 		Cache:     cache,
 		Scheduler: scheduler,
+	}
+
+	imHistory := history.NewInternalMetricsHistory(opts.metricsHistorySize)
+	if opts.metricsHistorySize != 0 {
+		metricsCollector := metrics.NewInternalMetricsCollector(imHistory)
+		metricsCollector.StartService()
 	}
 
 	if opts.startWebAppFlag {
--- a/pkg/entrypoint/entrypoint.go
+++ b/pkg/entrypoint/entrypoint.go
@@ -80,8 +80,9 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
 		Scheduler: scheduler,
 	}
 
-	imHistory := history.NewInternalMetricsHistory(opts.metricsHistorySize)
+	var imHistory *history.InternalMetricsHistory
 	if opts.metricsHistorySize != 0 {
+		imHistory = history.NewInternalMetricsHistory(opts.metricsHistorySize)
 		metricsCollector := metrics.NewInternalMetricsCollector(imHistory)
 		metricsCollector.StartService()
 	}
--- a/pkg/metrics/history/internal_metrics.go
+++ b/pkg/metrics/history/internal_metrics.go
@@ -27,17 +27,16 @@ import (
 // for the web UI's front page.
 // For more detailed metrics collection use Prometheus.
 type InternalMetricsHistory struct {
-	records []*MetricsRecord
+	records []*metricsRecord
 	limit   int
 
 	// internal implementation of limited array
 	pointer int
-	full    bool
 
 	sync.RWMutex
 }
 
-type MetricsRecord struct {
+type metricsRecord struct {
 	Timestamp         time.Time
 	TotalApplications int
 	TotalContainers   int
@@ -45,7 +44,7 @@ type MetricsRecord struct {
 
 func NewInternalMetricsHistory(limit int) *InternalMetricsHistory {
 	return &InternalMetricsHistory{
-		records: make([]*MetricsRecord, limit),
+		records: make([]*metricsRecord, limit),
 		limit:   limit,
 	}
 }
@@ -54,7 +53,7 @@ func (h *InternalMetricsHistory) Store(totalApplications, totalContainers int) {
 	h.Lock()
 	defer h.Unlock()
 
-	h.records[h.pointer] = &MetricsRecord{
+	h.records[h.pointer] = &metricsRecord{
 		time.Now(),
 		totalApplications,
 		totalContainers,
@@ -63,32 +62,19 @@ func (h *InternalMetricsHistory) Store(totalApplications, totalContainers int) {
 	h.pointer++
 	if h.pointer == h.limit {
 		h.pointer = 0
-		h.full = true
 	}
 }
 
-// the return of this function is ordered by the time of addition
-func (h *InternalMetricsHistory) GetRecords() []*MetricsRecord {
+// contract: the non-nil values are ordered by the time of addition
+// may contains nil values, those should be handled (filtered) on the caller's side
+func (h *InternalMetricsHistory) GetRecords() []*metricsRecord {
 	h.RLock()
 	defer h.RUnlock()
 
-	returnRecords := make([]*MetricsRecord, 0)
-	if h.full {
-		pointer := h.pointer
-		for i := 0; i < h.limit; i++ {
-			returnRecords = append(returnRecords, h.records[pointer])
-			pointer++
-			if pointer == h.limit {
-				pointer = 0
-			}
-		}
-	} else {
-		for _, record := range h.records {
-			if record != nil {
-				returnRecords = append(returnRecords, record)
-			}
-		}
-	}
+	returnRecords := make([]*metricsRecord, h.limit-h.pointer)
+	copy(returnRecords, h.records[h.pointer:])
+	returnRecords = append(returnRecords, h.records[:h.pointer]...)
+
 	return returnRecords
 }
 
--- a/pkg/metrics/history/internal_metrics.go
+++ b/pkg/metrics/history/internal_metrics.go
@@ -1,0 +1,74 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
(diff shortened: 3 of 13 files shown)


pkg/webservice/handlers.go
```diff
@@ -235,3 +236,57 @@ func getNodeJSON(nodeInfo *cache.NodeInfo) *dao.NodeDAOInfo {
 		Schedulable: nodeInfo.IsSchedulable(),
 	}
 }
+
+func GetApplicationHistory(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+
+	var result []*dao.ApplicationHistoryDAOInfo
+	lists := gClusterInfo.ListPartitions()
+	for _, k := range lists {
```
yangwwei: right now it reports back per partition
I don't think front-end expect this. we should aggregate this to the cluster level. Can we change the `HistoricPartitionInfo` to `HistoricClusterInfo`?

pkg/cache/history/history_info.go
```diff
@@ -0,0 +1,56 @@
...
+package history
+
+import (
+	"time"
+)
+
+var LIMIT = 144 // 10 minutes in 1 day
+
+type HistoricalPartitionInfo struct {
+	records []*PartitionInfoRecord
```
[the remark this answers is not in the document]
adamantal: Added a simple `sync.Mutex` for this.

pkg/cache/history/history_info.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+type PartitionInfoRecord struct {
+	Timestamp         time.Time
+	TotalApplications int
+	TotalContainers   int
+}
+
+func (hpInfo *HistoricalPartitionInfo) Store(totalApplications, totalContainers int) {
```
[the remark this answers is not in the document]
adamantal: Done (for `HistoricalClusterInfo`)

pkg/cache/history/history_info.go
```diff
@@ -0,0 +1,56 @@
...
+ limitations under the License.
+*/
+
+package history
+
+import (
+	"time"
+)
+
+var LIMIT = 144 // 10 minutes in 1 day
```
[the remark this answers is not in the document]
adamantal: Done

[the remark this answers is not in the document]
adamantal: Modified the constants accordingly

pkg/cache/history_collector.go
```diff
@@ -0,0 +1,76 @@
...
+package cache
+
+import (
+	"time"
+
+	"github.com/apache/incubator-yunikorn-core/pkg/cache/history"
+	"github.com/apache/incubator-yunikorn-core/pkg/log"
+)
+
+var tickerDefault = 10 * time.Minute
```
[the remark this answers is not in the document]
adamantal: Done

pkg/cache/history_collector.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (hpInfoUpdater *HistoricalPartitionInfoUpdater) Start() {
+	go func() {
+		for {
+			select {
+			case <-hpInfoUpdater.finished:
+				return
+			case <-hpInfoUpdater.ticker.C:
+				log.Logger().Info("Adding current status to historical partition data")
```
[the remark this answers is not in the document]
adamantal: Done

pkg/cache/partition_info.go
```diff
@@ -103,6 +106,11 @@ func newPartitionInfo(partition configs.PartitionConfig, rmID string, info *Clus
 	// TODO get the resolver from the config
 	p.userGroupCache = security.GetUserGroupCache("")
 
+	hpInfo := &history.HistoricalPartitionInfo{}
+	p.history = hpInfo
+	p.historyUpdater = NewHistoricalPartitionInfoUpdater(p, hpInfo)
+	p.historyUpdater.Start()
+
```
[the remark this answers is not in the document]
adamantal: I only moved the `Start` call to `StartService`, but I kept the construction of the `HistoricalParitionInfoUpdater` object in the constructor.

pkg/webservice/handlers.go
adamantal: Agreed. Modified this and other classes in the PR accordingly.

adamantal:
Strangely the test passes on my local - I didn't touch `scheduler_recovery_test.go`, so I think it is probably unrelated.

adamantal:
Tested the patch locally with the following results:
```
AAntal-MBP15:yunikorn-core adamantal$ kubectl logs yunikorn-scheduler-76ff98dff4-lb77q yunikorn-scheduler-k8s | grep "Adding current status"
2020-03-26T15:35:10.911Z	DEBUG	cache/history_collector.go:61	Adding current status to historical partition data
2020-03-26T15:36:10.912Z	DEBUG	cache/history_collector.go:61	Adding current status to historical partition data
2020-03-26T15:37:10.912Z	DEBUG	cache/history_collector.go:61	Adding current status to historical partition data
```
So as expected, we are collecting the metrics in every minute.

I also checked the rest API endpoint, and it seems like there's a bug. I missed a "/" from the beginning of the REST endpoints, so they are not accessible. I will fix it, test it locally and update the PR.

adamantal:
Fixed the bug and tried out locally.

Sample output of `localhost:9080/ws/v1/history/containers`:
```json
[
    {
        "timestamp": 1585238263802408000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585238323803183000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585238383803268000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585238443804046000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585238503806253000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585238563807452000,
        "totalContainers": "0"
    }
]
```

adamantal:
Created a brand new architecture according to @wilfred-s comment on the jira.

Checked on my local with the `make run` command on k8shim, also starting 3 sleep pods using the `deployments/examples/sleep/sleeppods.yaml` file.
Example output of the `(localhost:9080)/ws/v1/history/containers` endpoint:

```json
[
    {
        "timestamp": 1585579879950325000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585579939947939000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585579999942027000,
        "totalContainers": "0"
    },
    {
        "timestamp": 1585580059941905000,
        "totalContainers": "3"
    }
]
```

yangwwei:
Thanks @adamantal !
@wilfred-s , can you help to review this change?

adamantal:
Rebased due to conflict in the `go.mod` file (YUNIKORN-63).

pkg/webservice/handlers.go
```diff
@@ -236,3 +236,39 @@ func getNodeJSON(nodeInfo *cache.NodeInfo) *dao.NodeDAOInfo {
 		Schedulable: nodeInfo.IsSchedulable(),
 	}
 }
+
+func GetApplicationHistory(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+
+	var result []*dao.ApplicationHistoryDAOInfo
+	records := imHistory.GetRecords()
+	for _, record := range records {
```
wilfred-s: Need to check the impact of the earlier mentioned unlimited growth solution on this iteration loop

pkg/metrics/metrics_collector_test.go
```diff
@@ -0,0 +1,65 @@
...
+	metricsCollector.StartService()
+
+	go func() {
+		metrics := GetSchedulerMetrics()
+		i := 0
+		for i < 3 {
+			metrics.IncTotalApplicationsRunning()
+			metrics.AddAllocatedContainers(2)
+			i += 1
+			time.Sleep(400 * time.Millisecond)
```
wilfred-s: Do we really need to sleep between each update? Extends the test run unneeded

pkg/metrics/metrics_collector_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"time"
+
+	"gotest.tools/assert"
+
+	"github.com/apache/incubator-yunikorn-core/pkg/metrics/history"
+)
+
+func TestHistoricalPartitionInfoUpdater(t *testing.T) {
+	metricsHistory := history.NewInternalMetricsHistory(3)
+	setInternalMetricsCollectorTickerForTest(500 * time.Millisecond)
```
wilfred-s: Can we shorten this to something like 10 msec?

pkg/metrics/metrics_collector.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+		}
+	}()
+}
+
+func (u *InternalMetricsCollector) Stop() {
+	u.stopped <- true
+}
+
+func setInternalMetricsCollectorTickerForTest(newDefault time.Duration) {
```
wilfred-s: Use a comment like `visible for test only` or something like it
We should not use _test_ in the function name unless it is really a test.

pkg/metrics/metrics_collector.go
```diff
@@ -0,0 +1,89 @@
...
+	tickerDefault = newDefault
+}
```
wilfred-s: missing newline

pkg/metrics/init.go
```diff
@@ -54,6 +56,7 @@ type CoreSchedulerMetrics interface {
 	// Metrics Ops related to ScheduledAllocationSuccesses
 	IncAllocatedContainer()
 	AddAllocatedContainers(value int)
+	getAllocatedContainers() *prometheus.Counter
```
wilfred-s: can we not simply return the primitive we need?
Hide all the prometheus logic inside one file, see other review comments on how.

adamantal: Makes sense, fixed

pkg/metrics/metrics_collector_test.go
adamantal: I wanted to wait for at least 2 cycles. Do you think it is still an issue after ticker has been configured to 10msec (from comment below)?

pkg/metrics/history/internal_metrics.go
```diff
@@ -0,0 +1,74 @@
...
+		records: make([]*MetricsRecord, 0),
+		limit:   limit,
+	}
+}
+
+func (h *InternalMetricsHistory) Store(totalApplications, totalContainers int) {
+	h.mutex.Lock()
+	defer h.mutex.Unlock()
+
+	h.records = append(h.records,
```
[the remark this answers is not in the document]
adamantal: Nice catch. I actually got through that post during my Go tutorial, but didn't realize that I made a mistake here until now.
I started with a containers/list implementation, but got into problems, because it was not typesafe and had to case everywhere where it is used.

Implemented a circular, sized array using the pointer and the limit variables. If there's any official typesafe implementation in Golang, or some popular package to carry out this, please let me know and I'll use that. I think this is the most light implementation that we could use.

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/apache/incubator-yunikorn-scheduler-interface v0.0.0-20200327234544-149aaa3d8e48
-	github.com/golang/protobuf v1.3.5 // indirect
```
sunilgovind: could we avoid this change

pkg/metrics/history/internal_metrics.go
```diff
@@ -0,0 +1,99 @@
...
+		for _, record := range h.records {
+			if record != nil {
+				returnRecords = append(returnRecords, record)
+			}
+		}
+	}
+	return returnRecords
+}
+
+func (h *InternalMetricsHistory) GetLimit() int {
```
sunilgovind: Limit is set only one time when system is booted up, could you please confirm on that.
If that's the case, do we need a mutex here?

pkg/metrics/metrics_collector.go
```diff
@@ -0,0 +1,81 @@
...
+				log.Logger().Debug("Adding current status to historical partition data")
+
+				totalAppsRunning, err := m.scheduler.getTotalApplicationsRunning()
+				if err != nil {
+					log.Logger().Warn("Could not encode metric.", zap.Error(err))
+					continue
+				}
+				totalContainersRunning, err := m.scheduler.getAllocatedContainers()
+				if err != nil {
+					log.Logger().Warn("Could not encode metric.", zap.Error(err))
```
sunilgovind: cud we add metric name also here?

pkg/webservice/handlers.go
```diff
@@ -236,3 +236,39 @@ func getNodeJSON(nodeInfo *cache.NodeInfo) *dao.NodeDAOInfo {
...
+		element := &dao.ApplicationHistoryDAOInfo{
+			Timestamp:         record.Timestamp.UnixNano(),
+			TotalApplications: strconv.Itoa(record.TotalApplications),
```
sunilgovind: in general, please check spacing. is it tab or spaces?
code allignment?

pkg/webservice/handlers.go
```diff
@@ -236,3 +236,39 @@ func getNodeJSON(nodeInfo *cache.NodeInfo) *dao.NodeDAOInfo {
...
+
+func GetContainerHistory(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+
+	var result []*dao.ContainerHistoryDAOInfo
+	records := imHistory.GetRecords()
+	for _, record := range records {
+		element := &dao.ContainerHistoryDAOInfo{
+			Timestamp:       record.Timestamp.UnixNano(),
+			TotalContainers: strconv.Itoa(record.TotalContainers),
```
sunilgovind: are we having only TotalContainers/Apps returned in an array when we query /ws/v1/history/containers ?
Cud u pls share a sample response?

pkg/metrics/history/internal_metrics.go
```diff
@@ -0,0 +1,99 @@
...
+	limit   int
+
+	// internal implementation of limited array
+	pointer int
+	full    bool
+
+	sync.RWMutex
+}
+
+type MetricsRecord struct {
```
wilfred-s, line 40: Do we need to export this?
IDE says there is no need to export.

pkg/metrics/metrics_collector.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"go.uber.org/zap"
+
+	"github.com/apache/incubator-yunikorn-core/pkg/log"
+	"github.com/apache/incubator-yunikorn-core/pkg/metrics/history"
+)
+
+var tickerDefault = 1 * time.Minute
+
+type InternalMetricsCollector struct {
```
wilfred-s, line 32: Do we need to export?
IDE says that it is not needed

go.mod
wilfred-s: a `go mod tidy` adds it back again on my system so I think we need to follow that and remove this from the change

pkg/webservice/handlers.go
adamantal: What is the intended usage of the `go fmt` command?

I have the following error message running it from the project's root:
```
can't load package: package github.com/apache/incubator-yunikorn-core: unknown import path "github.com/apache/incubator-yunikorn-core": cannot find module providing package github.com/apache/incubator-yunikorn-core
```

pkg/entrypoint/entrypoint.go
```diff
@@ -73,14 +73,19 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
...
-	metricsCollector.StartService()
 
 	context := &ServiceContext{
 		RMProxy:   proxy,
 		Cache:     cache,
 		Scheduler: scheduler,
 	}
 
+	imHistory := history.NewInternalMetricsHistory(opts.metricsHistorySize)
+	if opts.metricsHistorySize != 0 {
```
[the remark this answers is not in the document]
adamantal: Agreed. Fixed

pkg/metrics/history/internal_metrics.go
adamantal: It is used in `webservice/handlers.go`, though the name `MetricsRecord` is not explicitly used. I fix it then.

pkg/metrics/history/internal_metrics.go
```diff
@@ -40,35 +45,55 @@ type MetricsRecord struct {
...
+				pointer = 0
+			}
+		}
+	} else {
+		for _, record := range h.records {
+			if record != nil {
+				returnRecords = append(returnRecords, record)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
adamantal: Indeed, I fixed accordingly.

I'd be more comfortable with filtering the nil values here, assuming we could have other caller's of this function, but right now it is ok to do it in the webapp.

pkg/metrics/metrics_collector_test.go
wilfred-s: with a shorter sleep I am OK with the test and the wait we put in.

pkg/metrics/history/internal_metrics.go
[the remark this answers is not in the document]
wilfred-s: new impl looks good

pkg/webservice/handlers.go
adamantal: Could you check whether the fixing commit is good? I hope I get your first item right.

pkg/metrics/scheduler.go
```diff
@@ -200,6 +201,14 @@ func (m *SchedulerMetrics) AddAllocatedContainers(value int) {
 	m.allocatedContainers.Add(float64(value))
 }
 
+func (m *SchedulerMetrics) getAllocatedContainers() (value int, err error) {
```
wilfred-s, line 204: We should not mix named and unnamed returns.
We have standardised on unnamed everywhere else.

pkg/metrics/scheduler.go
```diff
@@ -265,6 +274,12 @@ func (m *SchedulerMetrics) SetTotalApplicationsRunning(value int) {
 	m.totalApplicationsRunning.Set(float64(value))
 }
 
+func (m *SchedulerMetrics) getTotalApplicationsRunning() (int, error) {
+	metric := &dto.Metric{}
+	err := m.totalApplicationsRunning.Write(metric)
+	return int(*metric.Gauge.Value), err
+}
```
wilfred-s, line 281: See `getAllocatedContainers` same nil pointer here, solution is same, don't use named returns

pkg/webservice/handlers.go
```diff
@@ -236,3 +236,49 @@ func getNodeJSON(nodeInfo *cache.NodeInfo) *dao.NodeDAOInfo {
...
 }
+
+func GetApplicationHistory(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+
+	if imHistory != nil {
+		var result []*dao.ApplicationHistoryDAOInfo
+		records := imHistory.GetRecords()
+		for _, record := range records {
+			if record != nil {
```
wilfred-s, line 247: Can we have records that are not nil after the first nil record? I think not, but please check.
If we cannot why not use a break to jump out of the loop:
```
if record == nil {
	break
}
```

pkg/webservice/handlers.go
```diff
@@ -236,3 +236,49 @@ func getNodeJSON(nodeInfo *cache.NodeInfo) *dao.NodeDAOInfo {
 		Schedulable: nodeInfo.IsSchedulable(),
 	}
 }
+
+func GetApplicationHistory(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+
+	if imHistory != nil {
```
wilfred-s, line 243: use unindented flow instead of an else clause:
```
if imHistory == nil {
	http.Error...
	return
}
var result ....
```

pkg/webservice/handlers.go
```diff
@@ -236,3 +236,49 @@ func getNodeJSON(nodeInfo *cache.NodeInfo) *dao.NodeDAOInfo {
...
+		}
+		if err := json.NewEncoder(w).Encode(result); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
+	} else {
+		http.Error(w, "Internal metrics collection is not enabled.", http.StatusInternalServerError)
+	}
+}
+
+func GetContainerHistory(w http.ResponseWriter, r *http.Request) {
```
wilfred-s, line 263: same comments as per `GetApplicationHistory`

pkg/metrics/metrics_collector_test.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}()
+
+	time.Sleep(11 * time.Millisecond)
+	metricsCollector.Stop()
+
+	records := metricsHistory.GetRecords()
+	assert.Equal(t, 3, len(records), "Expected exactly 2 history records")
+	for i, record := range records {
+		if i == 0 {
```
wilfred-s, line 52: use a switch statement or just direct `records[0]` etc

pkg/metrics/metrics_collector_test.go
```diff
@@ -0,0 +1,62 @@
...
+			i += 1
+			time.Sleep(4 * time.Millisecond)
+		}
+	}()
+
+	time.Sleep(11 * time.Millisecond)
+	metricsCollector.Stop()
+
+	records := metricsHistory.GetRecords()
+	assert.Equal(t, 3, len(records), "Expected exactly 2 history records")
```
wilfred-s, line 50: nit: text not updated to 3

adamantal: Silly of me, thanks for reminding!

pkg/webservice/handlers.go
adamantal: Fixed this, but I think we're relying too much on the actual implementation of the `internalMetricsHistory`, and it makes the code too fragile if it is modified in the future.

pkg/metrics/scheduler.go
adamantal: Thanks, didn't know that. Is it standardised on a per file or a per package base?

pkg/metrics/metrics_collector.go
```diff
@@ -0,0 +1,81 @@
...
+			select {
+			case <-u.stopped:
+				return
+			case <-u.ticker.C:
+				log.Logger().Debug("Adding current status to historical partition data")
+
+				totalAppsRunning, err := m.scheduler.getTotalApplicationsRunning()
+				if err != nil {
+					log.Logger().Warn("Could not encode totalApplications metric.", zap.Error(err))
+					continue
```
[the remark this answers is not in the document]
adamantal: I used -1, added a comment on the class.

yangwwei:
Looks good to me, thanks @wilfred-s , @adamantal !
I think we can merge this. Only one minor thing, we might want to check with the front-end. I saw when we are not able to get app/container count, it is set to `-1`.  We need to make sure the front-end supports this value.

wilfred-s:
Thank you Weiwei, I'll log a jira against the webapp to make sure we can handle the -1 value.

PR #120, opened 2020-04-01
[YUNIKORN-64] fix zap config creation

wilfred-s:
Zap logger used the default DEV config instead of our own, changing to our config to get full control.
Cleanup of the init code of the logger including test.
This reverts changes from: YUNIKORN-62

### DIFF
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -38,14 +38,14 @@ func Logger() *zap.Logger {
 			// is running as a deployment mode, or running with another non-go code
 			// shim. In this case, we need to create our own logger.
 			// TODO support log options when a global logger is not there
-			c := zap.NewDevelopmentConfig()
-			config = &c
-			newLogger, err := config.Build()
-			// this should really not happen so just write to stdout
+			config = createConfig()
+			var err error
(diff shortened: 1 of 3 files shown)


wilfred-s:
The test written only works if we have a clean go engine to execute it. If we have executed tests before the logger will have been initialised causing the test to fail. Need to re-think how we can test this consistently.

yangwwei:
Changes look good. My understanding is, as long as we have fixed the setup issues in k8shim side, we are good. As the core side logging is reusing the configs from k8shim.

These changes are to ensure when the core is not running with our k8shim, we still have a good zap logging setup. Correct?

wilfred-s:
> Changes look good. My understanding is, as long as we have fixed the setup issues in k8shim side, we are good. As the core side logging is reusing the configs from k8shim.

yes for production cases we are good.

> These changes are to ensure when the core is not running with our k8shim, we still have a good zap logging setup. Correct?

yes and also during tests.  We need to make sure the tests run a configuration as close to what we do in production. If we would have had that before we would not have seen the panic on a nil in production.

yangwwei:
All make sense. +1.
Merging now.

PR #121, opened 2020-04-02
[YUNIKORN-73] Temporarily disable the badge on github.

yangwwei:
Let's reenable this once we get a stable UT env.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # Apache YuniKorn (incubating) - A Universal Scheduler
 
-![Green Build](https://github.com/apache/incubator-yunikorn-core/workflows/Green%20Build/badge.svg)
-
 <img src="https://raw.githubusercontent.com/apache/incubator-yunikorn-core/master/docs/images/logo/yunikorn-logo-blue.png" width="200">
 
 ----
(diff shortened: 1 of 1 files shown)


adamantal:
What is the agreement on the follow-up items? (We have to re-enable this some time). We should either add TODO to do the code, or file a follow-up jira targeting 0.9 or later.

README.md
```diff
@@ -1,11 +1,5 @@
 # Apache YuniKorn (incubating) - A Universal Scheduler
 
-![Green Build](https://github.com/apache/incubator-yunikorn-core/workflows/Green%20Build/badge.svg)
-
-<img src="https://raw.githubusercontent.com/apache/incubator-yunikorn-core/master/docs/images/logo/yunikorn-logo-blue.png" width="200">
-
-----
```
wilfred-s, line 7: Do we really want to remove the logo too?

wilfred-s:
> What is the agreement on the follow-up items? (We have to re-enable this some time). We should either add TODO to do the code, or file a follow-up jira targeting 0.9 or later.

This is purely the badge that is getting removed, not the action itself. The action will still run but the badge does not update as expected even after passed runs. So for now we remove it until we get time to sort it all out.

README.md
yangwwei: Apologies, certainly not intentional.
Just correct the patch, we just need to remove the badge for now.

PR #122, opened 2020-04-03
[YUNIKORN-75] REST API for changing log level

adamantal:
Exposed REST API for changing the `Logger` object. Currently only supports the log level.

### DIFF
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -20,7 +20,9 @@ package webservice
 
 import (
     "encoding/json"
+    "github.com/apache/incubator-yunikorn-core/pkg/log"
     "github.com/apache/incubator-yunikorn-core/pkg/webservice/dao"
+    "go.uber.org/zap/zapcore"
     "gotest.tools/assert"
     "net/http"
     "strings"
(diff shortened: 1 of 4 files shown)


adamantal:
This is currently a work in progress patch. Will update and clear out the PR (do not review yet).

adamantal:
I found the `zap` `Logger`an utterly not user friendly package.

My proposal: wrap the original core in a new implementation where we override some level-related function to take control over the log level. I could not find any cases where this could be manually settable or pluggable for an existing `Logger` - only in cases where the user created the `Logger` object itself from config or manually. I created another Logger based on the existing logger and overwrote its relevant functions.

If you can think of any better solution, I'm open to suggestions.

wilfred-s:
We already have a really simple way to set the level. Check the function `InitAndSetLevel` in the logger code. Setting the level is as changing the value setting it in the active zap config: `config.Level.SetLevel(level)`. That uses an already created level. Setting the level based on a string: `config.Level.UnmarshalText( text )` This just needs to be wrapped into one function: `func SetLevel(newLvl string) err` and called from the REST service.

adamantal:
> We already have a really simple way to set the level. Check the function `InitAndSetLevel` in the logger code. Setting the level is as changing the value setting it in the active zap config: `config.Level.SetLevel(level)`. That uses an already created level. Setting the level based on a string: `config.Level.UnmarshalText( text )` This just needs to be wrapped into one function: `func SetLevel(newLvl string) err` and called from the REST service.

Though the unit tests were passing as me, just like yours in `logger_test.go`, but running integration test was failing on my local. As far as I can tell, `logger.go#L75`returns false:
```go
if logger = zap.L(); isNopLogger(logger) {
```
I assumed we already have some default logger initialized at that point.
After that we skip the config generation, saving it to a global variable, and calling `config.Level.SetLevel(level)` in `logger.go#L127` inside `InitAndSetLevel` will not take any effect.

It might be related to how `make run` works in the k8shim. Probably worth trying it out in a real cluster, but I haven't done that before.

pkg/webservice/routes.go
```diff
@@ -98,6 +98,14 @@ var routes = Routes{
...
 	},
 
+	// endpoint to set attributes of the logger
+	Route{
+		"Log",
+		"POST",
+		"/ws/v1/log",
+		ConfigureLogger,
+	},
+
```
sunilgovind, line 108: Its better have GET method also know what the set log level in YuniKorn.
Usually it ll be like GET the level, and then set it with desired value.
Thoughts. @wilfred-s @adamantal

pkg/webservice/handlers_test.go
```diff
@@ -20,11 +20,16 @@ package webservice
...
 import (
     "encoding/json"
-    "github.com/apache/incubator-yunikorn-core/pkg/webservice/dao"
-    "gotest.tools/assert"
     "net/http"
     "strings"
     "testing"
+
+    "gotest.tools/assert"
+
```
sunilgovind, line 28: could we pls remove some unwanted lines here

yangwwei:
Our runtime model is having 1 single binary that built k8shim and core together, initially, everything starts with the shim, it initiates the logging things; then in the core side, it simply reuses the global logger set by shim: https://github.com/apache/incubator-yunikorn-k8shim/blob/dbd74df39403f53b00415210f3410395fb2ad04b/pkg/log/logger.go#L78.

This task requires us to do the opposite, the rest API server is living in the scheduler-core. what we need here is the ability to modify logging level at the core side and at the same time this change can be picked up by k8shim. Not sure if the global logger can work this way, this needs to test out.

pkg/webservice/routes.go
wilfred-s: yes both should be implemented for completeness: 
- GET to retrieve
- PUT to update

POST is for creating a new object PUT is for updating an existing object.
See https://restfulapi.net/rest-put-vs-post/

wilfred-s:
Moving design discussion to jira

adamantal:
Wrote a bigger comment on the jira, please express your opinion there.

wilfred-s:
Closing this PR
We're keeping the jira open but have run into some implementation issues that we need to figure out before we can implement this.

Thank you @adamantal for all the work done

PR #123, opened 2020-04-05
[YUNIKORN-77] De-dup event queue methods inside yunikorn-core, and improve logs

wangdatan:
@wilfred-s , @yangwwei can you help to review?

### DIFF
--- a/pkg/common/eventqueue.go
+++ b/pkg/common/eventqueue.go
@@ -1,0 +1,70 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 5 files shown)


pkg/common/eventqueue.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	if len(q.eventQueue)%1000 == 0 && len(q.eventQueue) > 5000 {
+		log.Logger().Warn("#events are waiting in the queue",
+			zap.String("EventQueueName", q.name),
+			zap.Int("currentQueueSize", len(q.eventQueue)))
+	}
+}
+
+func (q *EventQueue) Pop() interface{} {
```
adamantal, line 60: Though it's quite obvious by looking at the implementation, could you add a comment that it is a blocking call?

pkg/common/eventqueue.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"reflect"
+
+	"github.com/apache/incubator-yunikorn-core/pkg/log"
+	"go.uber.org/zap"
+)
+
+type EventQueue struct {
+	name       string
+	eventQueue chan interface{}
```
adamantal, line 30: Is allowing anything the best approach here? Shouldn't we use e.g. `stringer` as event?

PR #125, opened 2020-04-06
[YUNIKORN-79] Internal metrics are unavailable through REST API

adamantal:
The assumption about the internal data structure was not correct. Let's disable the optimization and also add a return statement before we got to the "Internal Server Error" line.

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -265,22 +265,23 @@ func GetContainerHistory(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 
 	if imHistory != nil {
-		var result []*dao.ContainerHistoryDAOInfo
-		records := imHistory.GetRecords()
-		for _, record := range records {
-			if record == nil {
-				continue
-			}
-			element := &dao.ContainerHistoryDAOInfo{
-				Timestamp:       record.Timestamp.UnixNano(),
-				TotalContainers: strconv.Itoa(record.TotalContainers),
-			}
-			result = append(result, element)
-		}
-		if err := json.NewEncoder(w).Encode(result); err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		}
+		http.Error(w, "Internal metrics collection is not enabled.", http.StatusInternalServerError)
 		return
 	}
-	http.Error(w, "Internal metrics collection is not enabled.", http.StatusInternalServerError)
+	var result []*dao.ContainerHistoryDAOInfo
+	records := imHistory.GetRecords()
+	for _, record := range records {
+		if record == nil {
+			continue
+		}
+		element := &dao.ContainerHistoryDAOInfo{
+			Timestamp:       record.Timestamp.UnixNano(),
+			TotalContainers: strconv.Itoa(record.TotalContainers),
+		}
+		result = append(result, element)
+	}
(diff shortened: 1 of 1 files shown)


pkg/webservice/handlers.go
```diff
@@ -245,7 +245,7 @@ func GetApplicationHistory(w http.ResponseWriter, r *http.Request) {
 		records := imHistory.GetRecords()
 		for _, record := range records {
 			if record == nil {
-				break
+				continue
```
wilfred-s, line 248: The break is correct, it pushes you out of the for loop. Which means you are done collecting records.

The issue is in the return of the function. Leverage the "least indent flow" and early returns to code:
```
	if imHistory == nil {
		http.Error(w, "Internal metrics collection is not enabled.", http.StatusInternalServerError)
		return
	}
	var result []*dao....
	records := ....
	....
}
```

adamantal: No, actually not.

Let's say we have `[<2,3>, <3,4>, nil, nil]` with `pointer=2` in its internal data structure. In the `GetRecords` function first the class copies its array between indexes pointer and the end (`[nil, nil]`), then appends the rest which is the array between 0 to the pointer index (`[<2,3>, <3,4>]`). This results in the following array: `[nil, nil, <2,3>, <3,4>]`. When in the `handler.go#GetApplicationHistory` 's for loop we detect the first nil item in the array (which is the item #0), we break the for loop immediately which is not desired, since we have non-nil entries after that, that we haven't serialized yet.

My original suggestion was to filter the nils in `InternalMetricsCollector` level, because it's more straightforward what to do there. The simplest solution here though to just do this simple fix.

pkg/webservice/handlers.go
```diff
@@ -241,23 +241,24 @@ func GetApplicationHistory(w http.ResponseWriter, r *http.Request) {
...
-			}
-			result = append(result, element)
+		http.Error(w, "Internal metrics collection is not enabled.", http.StatusInternalServerError)
+		return
+	}
+	var result []*dao.ApplicationHistoryDAOInfo
+	records := imHistory.GetRecords()
+	for _, record := range records {
+		if record == nil {
+			continue
```
wilfred-s, line 251: We still should break here: internal filling of the array stops when we have run out of stats to collect. If we cannot convert we set it to -1. So the only way the record can be nil is if we have never written to it and that means all follow up values are nil.

pkg/webservice/handlers.go
```diff
@@ -279,6 +280,7 @@ func GetContainerHistory(w http.ResponseWriter, r *http.Request) {
 		if err := json.NewEncoder(w).Encode(result); err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 		}
+		return
```
wilfred-s, line 283: This should be the equivalent of what we have for application history

pkg/webservice/handlers.go
adamantal: As I wrote in my comment [above](https://github.com/apache/incubator-yunikorn-core/pull/125#discussion_r404605856) in more detail, the array returned by the `GetRecords()` may have `nil` values **at the beginning**, but have non-nil values later.

pkg/webservice/handlers.go
adamantal: Sorry, I missed it. Fixed.

PR #127, opened 2020-04-09
[YUNIKORN-87] Add predicate check for reservation

wilfred-s:
Before creating the reservation the candidate node is checked for a
predicate fix. A failed predicate check prevents the reservation.
Strict enforcement of the reservation number to the number of ask
repeats.

DO NOT COMMIT
go.mod changes must be added waiting for apache/incubator-yunikorn-scheduler-interface#11


wilfred-s:
Last push was to update references to the SI only, merging with that change included

PR #128, opened 2020-04-10
[YUNIKORN-90] Correct the order of input parameters for strings.HasPrefix

TaoYang526:


yangwwei:
Good catch. Looks like this is not covered by UT. @TaoYang526  can you please add UT for this? Thanks

TaoYang526:
Thanks @yangwwei for the review. 
Already added UT for methods used in production in utils.go.

TaoYang526:
Thanks @yangwwei

PR #130, opened 2020-04-17
[YUNIKORN-98] Fix metrics enabled check

wilfred-s:
The change of the handler flow for metrics enablement in YUNIKORN-79
turned off metrics in all cases. Ths change fixes the check.
Added unit tests for the history handlers and refactored the test code.
Partial cleanup of exported function and lint fixes


wilfred-s:
Merging as this is currently a blocker for 0.8 and a shim fix

PR #134, opened 2020-04-29
[YUNIKORN-93] Queue maximum capacity cleanup

kingamarton:


pkg/cache/partition_info.go
```diff
@@ -715,55 +715,46 @@ func checkAndSetResource(resource *resources.Resource) string {
...
+	if usedCapacity == nil {
+		return resources.NewResource()
+	}
+	absResource := make(map[string]resources.Quantity)
+	for resourceName, availableResource := range capacity.Resources {
+		var absResValue float32
+		if availableResource != 0 {
+			if usedResource, ok := usedCapacity.Resources[resourceName]; ok {
+				absResValue = float32(usedResource) / float32(availableResource) * 100
+			}
```
wilfred-s, line 734: Float divisions are 0 safe in go: it will return either +INF or -INF based on the signs.
We need to make sure that we handle all this over and underflow safe. If used = maxInt64, available is 10 then the abs value will overflow an Int64 on conversion...

If `availableResource` is 0 we set no usage. That seems counter intuitive to me.
The case that there is no capacity set but there is usage we skip it, again not logical.

We need to come up with a way to describe those cases:
- no capacity set, but usage is a 0
- capacity set to 0, usage is larger than 0

pkg/cache/partition_info.go
```diff
@@ -715,55 +715,46 @@ func checkAndSetResource(resource *resources.Resource) string {
...
-		infos = append(infos, queue)
+		absResource[resourceName] = resources.Quantity(absResValue)
 	}
+	return resources.NewResourceFromMap(absResource)
+}
 
-	return infos
+func (qi *QueueInfo) GetQueueInfos() dao.QueueDAOInfo {
+	qi.RLock()
+	defer qi.RUnlock()
```
wilfred-s, line 743: Now that this is part of the QueueInfo it should be moved to the QueueInfo file.
I know it is allowed to do this but maintenance will become really difficult when we leave it here.

NOTE: tests should also be moved to the correct file and if no tests are available we need a test for this.

pkg/cache/partition_info_test.go
```diff
@@ -772,3 +772,46 @@ partitions:
 	assert.Equal(t, len(m), 2)
 	assert.Assert(t, reflect.DeepEqual(m["memory"], []int{1, 1, 0, 0, 0, 0, 0, 0, 1, 0}))
 }
+
+func TestCalculateAbsUsedCapacity(t *testing.T) {
```
wilfred-s, line 776: See above this belongs in `Resource` and we need to test all cases:
* nil, nil
* zero, nil
* set res, nil
* set res, zero
* set res, set res
* partial set res, set res
* set res, partial set

and overflow checks
using `t.log` is good during development but we have seen each log entry tagged as a failure when there is another failures in any test in the file when we run the test in a github workflow. Please remove them for now

pkg/common/configs/configvalidator.go
```diff
@@ -76,6 +76,9 @@ func checkResource(res map[string]string) (int64, error) {
 	var totalres int64
 	for _, val := range res {
 		rescount, err := strconv.ParseInt(val, 10, 64)
+		if rescount < 0 {
+			return 0, fmt.Errorf("not valid negative resource found")
+		}
```
wilfred-s, line 81: Nit: Logical would be to first check for a parse error and then the value set after a parse error

pkg/cache/partition_info.go
kingamarton: - no capacity set, but usage is a 0: 
if no max capacity is set but we have resource usage I would return an empty resource, because we cannot calculate the absolute resource usage, since we don’t know what would mean 100%, so the web ui will show n/a

- capacity set to 0, usage is larger than 0: in my opinion this should not be a valid case. Since I am calculating the absolute usage according to the max capacity, what is a hard limit in my understanding. If we allow for the usage to exceed the max capacity, this limit will totally loose its sense. Are there cases when this is allowed?

pkg/cache/partition_info.go
```diff
@@ -715,55 +715,46 @@ func checkAndSetResource(resource *resources.Resource) string {
...
-		queue.Capacities = dao.QueueCapacity{
-			Capacity:        checkAndSetResource(child.GetGuaranteedResource()),
-			MaxCapacity:     checkAndSetResource(child.GetMaxResource()),
-			UsedCapacity:    checkAndSetResource(child.GetAllocatedResource()),
-			AbsUsedCapacity: "20",
+func calculateAbsUsedCapacity(capacity *resources.Resource, usedCapacity *resources.Resource) *resources.Resource {
+	if capacity == nil {
+		return nil
+	}
+	if usedCapacity == nil {
```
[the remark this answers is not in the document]
kingamarton: I don't think it is a good idea to handle zero and nil in the same way. In case of nil I would return an empty resource. In case of all zeros I would do the calculations and will result in maxInt for zero capacity and zero resource for zero usage.

pkg/cache/partition_info.go
kingamarton: In the new changes et I used the following approach:
- in case of missing value for a resource type indifferently if it is the used or max capacity, I skipped the calculation
- capacity set to 0, usage is larger than 0: I made the calculation and returned maxInt
Please let me know if you have some better ideas for this cases.

pkg/cache/partition_info.go
kingamarton: Moved it + also added a small unit test for it

pkg/cache/queue_info_test.go
```diff
@@ -327,3 +329,82 @@ func TestMaxResource(t *testing.T) {
 		t.Errorf("root max setting not picked up by parent queue expected %v, got %v", res, parent.GetMaxResource())
 	}
 }
+
+func TestGetQueueInfos(t *testing.T) {
+	root, err := createRootQueue()
+	if err != nil {
+		t.Fatalf("failed to create basic root queue: %v", err)
+	}
+	rootMax, err := resources.NewResourceFromConf(map[string]string{"memory": "2048", "vcores": "10"})
```
wilfred-s, line 338: We're shadowing `err` here

pkg/cache/queue_info_test.go
```diff
@@ -327,3 +329,82 @@ func TestMaxResource(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create configuration: %v", err)
+	}
+	root.setMaxResource(rootMax)
+
+	var parent *QueueInfo
+	parentUsed, err := resources.NewResourceFromConf(map[string]string{"memory": "1012", "vcores": "2"})
```
wilfred-s, line 345: Move the declaration of `parent` closer to its usage and fix `err` shadowing

pkg/cache/queue_info_test.go
```diff
@@ -327,3 +329,82 @@ func TestMaxResource(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("failed to create queue: %v", err)
+	}
+	err = parent.IncAllocatedResource(parentUsed, false)
+	if err != nil {
+		t.Fatalf("failed to increment allocated resource: %v", err)
+	}
+
+	var child1 *QueueInfo
+	var child1used, _ = resources.NewResourceFromConf(map[string]string{"memory": "1012", "vcores": "2"})
```
wilfred-s, line 359: Why are we ignoring `err` here? Does not line up with the rest of the test.

pkg/common/resources/resources_test.go
```diff
@@ -1245,3 +1245,65 @@ func TestFitInScore(t *testing.T) {
 	fit = NewResourceFromMap(map[string]Quantity{"first": 1, "second": 1})
 	assert.Equal(t, res.FitInScore(fit), 2.0, "FitInScore on resource with multiple negative quantities failed")
 }
+func TestCalculateAbsUsedCapacity(t *testing.T) {
```
wilfred-s, line 1248: Missing newline before function start.

pkg/common/configs/configvalidator.go
```diff
@@ -79,6 +79,9 @@ func checkResource(res map[string]string) (int64, error) {
 		if err != nil {
 			return 0, fmt.Errorf("resource parsing failed: %v", err)
 		}
+		if rescount < 0 {
+			return 0, fmt.Errorf("not valid negative resource found")
```
wilfred-s, line 83: Can we add the negative value to the message for easier tracing: 
`"resource value invalid, cannot be negative: %d", rescount`

go.mod
```diff
@@ -35,7 +35,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.9.1
 	github.com/satori/go.uuid v1.2.0
-	github.com/stretchr/testify v1.5.1 // indirect
+	github.com/stretchr/testify v1.5.1
```
wilfred-s, line 38: This change should not be there. I do not see it as an import in the change anywhere.

kingamarton: you are right. I removed the usage because lint showed it as blacklisted, and forgot to remove from here as well. I will remove it

PR #135, opened 2020-04-29
[YUNIKORN-53] Use asserts for nil error checks in test code

adamantal:
For yunikorn-core component


adamantal:
Unit test passed on my local

pkg/cache/partition_info_test.go
```diff
@@ -39,6 +39,12 @@ partitions:
         - name: default
 `
 
+func createDefaultPartitionInfo(t *testing.T) *PartitionInfo {
```
yangwwei, line 42: I think this way, we will lose the line of number where it failed.
There are multiple places calling `createDefaultPartitionInfo`, any of them failed, the output will only contain the line number in this function, then we are kind of lose tracking of which place causes this failure.
Can we do not pass `testing.T` in the parameter, let this function return an error and move the `assert.NilError` out of this helper function.

pkg/cache/queue_info_test.go
```diff
@@ -24,17 +24,20 @@ import (
 
 	"github.com/apache/incubator-yunikorn-core/pkg/common/configs"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/resources"
+	"gotest.tools/assert"
 )
 
 // create the root queue, base for all testing
-func createRootQueue() (*QueueInfo, error) {
+func createRootQueue(t *testing.T) *QueueInfo {
```
yangwwei, line 31: Same as before, I don't think it is a good idea to check the nil error inside of such helper functions. Unless you can print the trace correctly. Such as: https://github.com/apache/incubator-yunikorn-core/blob/366b635676e86e250d759f234fe1bfa44e4322c5/pkg/scheduler/tests/mock_rm_callback.go#L116

pkg/scheduler/placement/fixed_rule_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/apache/incubator-yunikorn-core/pkg/cache"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/configs"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/security"
+	"gotest.tools/assert"
```
wilfred-s, line 27: Please run make lint over the change I think it will fail on ordering of the imports
An other tool you have is to just run goimports locally to see what it complains about: `goimports -d -local "github.com/apache/incubator-yunikorn" .`
This same comment applies to lots of import changes in this PR

pkg/cache/queue_info_test.go
```diff
@@ -56,10 +59,7 @@ func createUnManagedQueue(parentQI *QueueInfo, name string, parent bool) (*Queue
 // base test for creating a managed queue
 func TestQueueInfo(t *testing.T) {
 	// create the root
-	root, err := createRootQueue()
-	if err != nil {
-		t.Fatalf("failed to create basic root queue: %v", err)
-	}
+	root := createRootQueue(t)
```
wilfred-s, line 62: This will hide the error from the caller and make tracking more difficult

pkg/cache/queue_info_test.go
wilfred-s: Too complex and too much overhead, just leave it as is.

pkg/cache/partition_info_test.go
wilfred-s: Which takes us back to what we had already, just replace `if err` with the assert. There is no need to add this extra function call: it does not make it any more readable and it does not reduce the lines of code.

wilfred-s: make sure that when this change is reversed that the err variable is not shadowed like the change in `TestAddNewNode`

pkg/cache/queue_info_test.go
wilfred-s: This will also require reversing variable creation and checks for shadowing of `err`

pkg/scheduler/scheduling_application_test.go
```diff
@@ -89,14 +89,11 @@ func TestAppReservation(t *testing.T) {
 		t.Error("new app should not have reservations for unknown node")
 	}
 
-	queue, err := createRootQueue(nil)
-	if err != nil {
-		t.Fatalf("queue create failed: %v", err)
-	}
+	queue := createDefaultRootQueueProtected(t)
```
wilfred-s, line 92: Same as for the info objects, don't hide the error in a function.
This reversal will also require variable declaration to be reversed to prevent shadowing

pkg/scheduler/scheduling_partition_test.go
```diff
@@ -43,6 +43,12 @@ func newTestPartition() (*partitionSchedulingContext, error) {
 	return newPartitionSchedulingContext(info, rootSched), nil
 }
 
+func newTestPartitionProtected(t *testing.T) *partitionSchedulingContext {
+	partition, err := newTestPartition()
+	assert.NilError(t, err, "test partition create failed with error")
+	return partition
+}
+
```
wilfred-s, line 51: Same as for the info objects, don't hide the error in a function.
This reversal will also require variable declaration to be reversed to prevent shadowing

pkg/scheduler/scheduling_queue_test.go
```diff
@@ -50,6 +49,13 @@ func createRootQueue(maxRes map[string]string) (*SchedulingQueue, error) {
 	return newSchedulingQueueInfo(root, nil), err
 }
 
+// create the root queue with no maxRes and fail if it couldn't be created
+func createDefaultRootQueueProtected(t *testing.T) *SchedulingQueue {
+	queue, err := createRootQueue(nil)
+	assert.NilError(t, err, "queue create failed")
+	return queue
+}
+
```
wilfred-s, line 58: Same as for the info objects, don't hide the error in a function.
This reversal will also require variable declaration to be reversed to prevent shadowing

pkg/scheduler/placement/fixed_rule_test.go
adamantal: I still got `-bash: goimports: command not found` but fixed it manually.

adamantal:
I hope this is what you meant @wilfred-s . Pls review

yangwwei:
Hi @wilfred-s  could you pls confirm if your comments have been addressed in the latest patch? Once you are good, feel free to merge this.

PR #138, opened 2020-05-05
[YUNIKORN-134] Post 0.8.0 release - update roadmap.

yangwwei:
This PR updates the roadmap.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -4,25 +4,27 @@
 
 **yunikorn-core**
 
-* Gang Scheduling.
-* Priority based preemption.
-* Support application priority.
-* Rich scheduler metrics.
-* Workload simulator.
-* Prometheus integration (phase 2).
-* Grafana integration.
+* [YUNIKORN-1](https://issues.apache.org/jira/browse/YUNIKORN-1): Support app/task priority aware scheduling.
+* [YUNIKORN-2](https://issues.apache.org/jira/browse/YUNIKORN-2): Gang Scheduling.
+* [YUNIKORN-21](https://issues.apache.org/jira/browse/YUNIKORN-21): Optimize node sorting algorithms.
+* [YUNIKORN-42](https://issues.apache.org/jira/browse/YUNIKORN-42): High efficient scheduling events framework phase 1.
+* [YUNIKORN-33](https://issues.apache.org/jira/browse/YUNIKORN-33): Performance benchmark with Kubemark.
+* [YUNIKORN-131](https://issues.apache.org/jira/browse/YUNIKORN-131): Prometheus integration - phase 2.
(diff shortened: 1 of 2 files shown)


docs/roadmap.md
```diff
@@ -4,25 +4,27 @@
...
-* Workload simulator.
-* Prometheus integration (phase 2).
-* Grafana integration.
+* [YUNIKORN-1](https://issues.apache.org/jira/browse/YUNIKORN-1): Support app/task priority aware scheduling.
+* [YUNIKORN-2](https://issues.apache.org/jira/browse/YUNIKORN-2): Gang Scheduling.
+* [YUNIKORN-21](https://issues.apache.org/jira/browse/YUNIKORN-21): Optimize node sorting algorithms.
+* [YUNIKORN-42](https://issues.apache.org/jira/browse/YUNIKORN-42): High efficient scheduling events framework phase 1.
+* [YUNIKORN-33](https://issues.apache.org/jira/browse/YUNIKORN-33): Performance benchmark with Kubemark.
+* [YUNIKORN-131](https://issues.apache.org/jira/browse/YUNIKORN-131): Prometheus integration - phase 2.
+* [YUNIKORN-132](https://issues.apache.org/jira/browse/YUNIKORN-132): Publish helm charts for convenient access.
```
sunilgovind, line 13: Do we need this? 
Is there any public helm repo in apache where we can publish same?

docs/roadmap.md
```diff
@@ -4,25 +4,27 @@
...
 
 **yunikorn-k8shim**
 
-* Security support.
-* Performance improvement: merging, packing requests.
-* 3rd Operator/controller integration, pluggable app discovery.
```
sunilgovind, line 19: How about detailed APP state management including FINISHED & FINISHING and other closing states?

docs/roadmap.md
yangwwei: No, I don't think apache has that.
If you are not comfortable with this, I am fine to remove it.

wilfred-s: the helm chart could be deployed and accessible from github by using `chart-releaser` the index.yaml file we can even host on the standard yunikorn website but this is a lot of work and would need good amount of thought, planning and build changes too.

I would prefer to not do this for now

yangwwei: OK, let me remove it.

docs/roadmap.md
yangwwei: That is more like a bug fix

docs/roadmap.md
sunilgovind: Ok.

docs/roadmap.md
sunilgovind: Basic APP state is there anyway, I am fine with this.

PR #140, opened 2020-05-08
[YUNIKORN-141] Fix flaky tests

yangwwei:


yangwwei:
hi @wilfred-s 

This bothers us too long, I spent a good amount of time today and hopefully now we have a fix. I've run the tests for quite a few times, seems to be much more stable now. The flaky tests actually were caused by a bunch of reasons.

### First, how to reproduce this?
A common issue we run into is, we do a `make test`, sometimes it fails for the first time, but then it always successful subsequentially. This was because once a go test has run, it will cache some test results, and next time those succeed tests will not run. Our flaky tests are mostly because they are conflicting with other tests. If only part of the tests were run, it mostly can work.

Try to run
```
go clean -cache -testcache
make test
```
you should be able to reproduce the issue.

### What caused the intermittent failures

There are a few things

#### 1. Services (cache, scheduler, rmProxy) were started but never stopped properly. 
In the smoke tests, we do start/stop the mocked cluster again and again. When we run these tests together. But when we do stop, none of them were stopped. This leaks a lot of goroutines and probably causes some conflicts.

**Fix**: add stop method for all services and make sure they are stopped while shutting down the mocked cluster.

#### 2. timing issue with `MultiStepSchedule`
MultiStepScheduler is supposed to run several scheduling cycles, and lots of tests are depending on a determined order of execution. However, internally we are handing events in async mode. Which means the order is not always determinable. E.g when you do `tryAllocate()`, the app or ask maybe not added to the cache/schedulingContext properly.

**Fix**: add `drain()` function for scheduler and cache, make sure for each step all events have been processed before going to next step. This way the processing will be more stable.

#### 3. `ms.mockRM.waitForAllocations` timeout is too short
Some tests called `ms.mockRM.waitForAllocations` and they fail intermittently. These tests only wait for allocation with 1s timeout. This usually is not a problem with our laptop, I presume it is much faster than docker containers running on github. 

**Fix**:  increase the timeout to 3s. BTW, waiting for a bit longer of time is not a problem, unless something really bad happens and every one of these functions wait for 3s (need to fix anyway).

yangwwei:
Move to #147

yangwwei:
Close this and move changes to #147

PR #143, opened 2020-05-12
[YUNIKORN-135] Move the scheduler-interface related constants from the markdown to the common package

adamantal:


adamantal:
> I think we need to distinguish between two sets of constants:
> 
> * shared between shim(s) and core
> * shim or core specific constants
> 
> Any constant that is used as a key in messages that flow between the shim and the core should be in the SI. Constants that are used as values should not be in the SI. Each part can set its own defaults and we should not mix that into the specification for the messages.
> 
> As an example: rack name information
> 
> * the key that specifies the rack name attribute should be in the SI
> * the default value should _not_ be in the SI

I don't see any scheduler-local constants here, so I keep this PR as it is.

adamantal:
Compile pass and UT pass against https://github.com/apache/incubator-yunikorn-scheduler-interface/pull/15 latest patch

pkg/scheduler/tests/mock_scheduler.go
```diff
@@ -81,8 +81,8 @@ func (m *mockScheduler) addNode(nodeID string, resource *si.Resource) error {
 			{
 				NodeID: nodeID,
 				Attributes: map[string]string{
-					"si.io/hostname": nodeID,
-					"si.io/rackname": "rack-1",
+					"si/hostname": nodeID,
+					"si/rackname": "rack-1",
```
wilfred-s, line 85: None of the tests use these values and for clarity we should not set them.
For all tests the same comment: lines should be removed

pkg/cache/node_info_test.go
```diff
@@ -21,9 +21,9 @@ package cache
 import (
 	"testing"
 
+	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/common"
```
wilfred-s, line 24: Check the sort order
This belongs to  the local imports below.

pkg/scheduler/helper.go
```diff
@@ -18,15 +18,15 @@
 
 package scheduler
 
-import "github.com/apache/incubator-yunikorn-core/pkg/api"
+import "github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/common"
```
wilfred-s, line 21: This whole file should be removed as it is not used.

pkg/cache/node_info_test.go
adamantal: Done

pkg/scheduler/helper.go
adamantal: Indeed. Deleted the whole file.

pkg/scheduler/tests/mock_scheduler.go
adamantal: Fixed.

adamantal:
Thanks for the review @wilfred-s. Fixed the items.

PR #144, opened 2020-05-13
[YUNIKORN-151] fix TestUnManagedSubQueues

wilfred-s:
Simple fix for a unit test failure and fix left over `err != nil` checks


wilfred-s:
Yes the lint check is failing because of YUNIKORN-139.
We need the SHA for `origin/HEAD` to correctly run the lint checks but in the checked out code we do not seem to have a `origin/HEAD`

wilfred-s:
closing: current test case can stay need to extend with a new case and not fix the current one

PR #145, opened 2020-05-13
[YUNIKORN-99] Add state based app sorting

wilfred-s:
To allow FIFO scheduling to work when lots of applications are submitted
an extra state is added to the applications. This state shows a newly
starting application with just one allocation. If more allocations are
added the application goes into a running state. The starting state
times out after 5 minutes to prevent starvation of other apps.

A new sorting policy for applications in a queue is added which only
allows one starting application to be scheduled at a time. If there are
no starting applications in the list a new application withut
allocations, accepted state, is added to the list of apps to schedule.

The sorting policy can be set on a leaf queue by specifying "status" as
the sort policy.


wilfred-s:
closed by wrong commit message tag, re-opening

pkg/scheduler/policies/sorting_policy.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+)
+
+// Sort type for queues & apps.
+type SortPolicy int
+
+const (
+	FifoSortPolicy   SortPolicy = iota // fair based on usage
```
adamantal, line 29: I think the comments are wrong here.

pkg/scheduler/policies/sorting_policy.go
```diff
@@ -0,0 +1,51 @@
...
+	FairSortPolicy                     // first in first out, submit time
+	StateAwarePolicy                   // only 1 app in starting state
+	UndefinedApp                       // not initialised or parsing failed
+)
+
+func (s SortPolicy) String() string {
+	return [...]string{"fifo", "fair", "status", "undefined"}[s]
```
adamantal, line 36: Shouldn't it be something like "stateaware" instead of "status"?

pkg/scheduler/sorters.go
```diff
@@ -47,39 +39,101 @@ func sortQueue(queues []*SchedulingQueue, sortType SortType) {
...
+			r := sortedApps[j]
 			return l.ApplicationInfo.SubmissionTime < r.ApplicationInfo.SubmissionTime
 		})
+	case policies.StateAwarePolicy:
+		sortedApps = stateAwareFilter(apps)
+		// Sort by submission time oldest first
+		sort.SliceStable(sortedApps, func(i, j int) bool {
+			l := sortedApps[i]
+			r := sortedApps[j]
+			return l.ApplicationInfo.SubmissionTime < r.ApplicationInfo.SubmissionTime
```
adamantal, line 71: The state aware policy looks orthogonal to the choice between Fifo and Fair. In other words: a user could only use state aware policy with the Fifo ordering (but not with fair). I think we should enable this feature for the user.

pkg/scheduler/sorters_test.go
```diff
@@ -344,6 +369,64 @@ func TestSortAppsFair(t *testing.T) {
 	assertAppList(t, list, []int{1, 3, 2, 0})
 }
 
+func TestSortAppsStateAware(t *testing.T) {
+	// stable sort is used so equal values stay were they were
```
adamantal, line 373:
```suggestion
	// stable sort is used so equal values stay where they were
```

pkg/cache/application_info.go
```diff
@@ -73,45 +81,117 @@ func (ai *ApplicationInfo) GetAllAllocations() []*AllocationInfo {
...
 	ai.leafQueue = leaf
 	ai.QueueName = leaf.GetQueuePath()
 }
 
 // Add a new allocation to the application
 func (ai *ApplicationInfo) addAllocation(info *AllocationInfo) {
-	ai.lock.Lock()
-	defer ai.lock.Unlock()
+	// progress the state based on where we are, we cannot fail in this case
+	// do NOT change the order as that will break the state
```
adamantal, line 182: I'd prefer a switch statement here. One could easily miss this point.

pkg/cache/application_info.go
```diff
@@ -73,45 +81,117 @@ func (ai *ApplicationInfo) GetAllAllocations() []*AllocationInfo {
...
+		//nolint: errcheck
+		_ = ai.HandleApplicationEvent(RunApplication)
+	}
+	if ai.IsAccepted() {
+		//nolint: errcheck
+		_ = ai.HandleApplicationEvent(StartApplication)
+	}
+	// add the allocation
+	ai.Lock()
+	defer ai.Unlock()
```
yangwwei, line 194: Do we need these state checks?
I think we should leverage the state machine. I think we can define an `AddAllocationEvent`. 
Then in the state machine, we can have something like the following:

```
Event: AddAllocationEvent
src: Accepted
dst: Starting

Event: AddAllocationEvent
src: Starting
dst: Runing
```

pkg/scheduler/scheduling_application.go
```diff
@@ -207,10 +224,28 @@ func (sa *SchedulingApplication) addAllocationAsk(ask *schedulingAllocationAsk)
...
+				zap.Error(err))
+		}
+	}
+	if sa.isWaiting() {
+		if err := sa.ApplicationInfo.HandleApplicationEvent(cache.RunApplication); err != nil {
+			log.Logger().Debug("Application state change failed while adding new ask",
+				zap.String("currentState", sa.ApplicationInfo.GetApplicationState()),
+				zap.Error(err))
+		}
+	}
```
yangwwei, line 244: Similar to the comments I had for the `addAllocation`, I think we can achieve this by leveraging the state machine.

pkg/scheduler/scheduler.go
```diff
@@ -407,12 +417,6 @@ func (s *Scheduler) processApplicationUpdateEvent(ev *schedulerevent.SchedulerAp
 				acceptedApps = append(acceptedApps, &si.AcceptedApplication{
 					ApplicationID: app.ApplicationID,
 				})
-				// app is accepted by scheduler
-				err = app.HandleApplicationEvent(cache.AcceptApplication)
-				if err != nil {
-					log.Logger().Debug("cache event handling error returned",
-						zap.Error(err))
-				}
```
yangwwei, line 415: The Accepted state means the app is accepted by the state (as the opposite to Rejected). This is a proper place for transiting the app to `Accepted`. Why do we want to move this to another place?

pkg/scheduler/scheduling_application.go
```diff
@@ -183,6 +183,23 @@ func (sa *SchedulingApplication) removeAllocationAsk(allocKey string) int {
...
+		oldState := sa.ApplicationInfo.GetApplicationState()
+		if err := sa.ApplicationInfo.HandleApplicationEvent(cache.WaitApplication); err != nil {
+			log.Logger().Warn("Application state not changed to Waiting while removing ask(s)",
+				zap.String("oldState", oldState),
+				zap.Int("releaseAllocs", toRelease),
+				zap.Error(err))
+		} else {
+			log.Logger().Debug("Application state change to Waiting while removing ask(s)",
+				zap.String("oldState", oldState),
+				zap.Int("releaseAllocs", toRelease))
```
yangwwei, line 200: dup logging? we already did the logging for each time when app state changes

pkg/scheduler/scheduling_application.go
```diff
@@ -633,3 +668,44 @@ func (sa *SchedulingApplication) recoverOnNode(node *SchedulingNode, ask *schedu
...
+		//nolint: errcheck
+		_ = sa.ApplicationInfo.HandleApplicationEvent(cache.StartApplication)
+		_ = sa.ApplicationInfo.HandleApplicationEvent(cache.RunApplication)
+	}
+	// log unexpected state
+	if !sa.ApplicationInfo.IsRunning() {
+		log.Logger().Warn("State of recovered app is not the expected RUNNING state",
+			zap.String("state", sa.ApplicationInfo.GetApplicationState()))
+	}
+}
```
yangwwei, line 711: Similarly, can we handle a `FinishRecoveryEvent` in the state machine?

pkg/scheduler/sorters.go
```diff
@@ -47,39 +39,101 @@ func sortQueue(queues []*SchedulingQueue, sortType SortType) {
...
+	}
+	return filteredApps
+}
+
+// This filter only allows one (1) application with a state that is not running in the list of candidates.
+// The preference is a state of Starting. If we can not find an app with a starting state we will use an app
+// with an Accepted state. However if there is an app with a Starting state even with no pending resource
+// requests, no Accepted apps can be scheduled. An app in New state does not have any asks and can never be
+// scheduled.
+func stateAwareFilter(apps map[string]*SchedulingApplication) []*SchedulingApplication {
```
yangwwei, line 94: we need UT for this function

pkg/scheduler/sorters.go
```diff
@@ -47,39 +39,101 @@ func sortQueue(queues []*SchedulingQueue, sortType SortType) {
...
+				if !foundStarting && (acceptedApp == nil || acceptedApp.ApplicationInfo.SubmissionTime > app.ApplicationInfo.SubmissionTime) {
+					acceptedApp = app
+				}
+				continue
+			}
+			// this is a running or starting app add it to the list
+			filteredApps = append(filteredApps, app)
+		}
+	}
+	// just add the accepted app if we need to: apps are not sorted yet
```
yangwwei, line 119: if the apps are not sorted. when all the apps are in `Accepted` state, how can we ensure the 1st application can be filtered out and scheduled first?

pkg/scheduler/sorters.go
yangwwei, line 94: one issue for this algorithm is
If I have a bunch of apps, one is `Starting` and the rest is `Running`, then only the `Starting` app can be filtered out. If my `Starting` app happens to be a single pod app. Even the pod is running,  it will need to wait for at least 5 minutes before other apps can be scheduled.

pkg/scheduler/policies/sorting_policy.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// Sort type for queues & apps.
+type SortPolicy int
+
+const (
+	FifoSortPolicy   SortPolicy = iota // fair based on usage
+	FairSortPolicy                     // first in first out, submit time
+	StateAwarePolicy                   // only 1 app in starting state
+	UndefinedApp                       // not initialised or parsing failed
```
yangwwei, line 32: UndefinedApp -> Undefined

pkg/cache/application_state.go
```diff
@@ -31,76 +31,102 @@ import (
...
-	New ApplicationState = iota
+	New applicationState = iota
 	Accepted
-	Rejected
+	Starting
 	Running
+	Waiting
+	Rejected
 	Completed
 	Killed
```
yangwwei, line 63: I found the meaning of these states is really confusing.
Can you please add the comment here, and also in a doc. I think we need a design doc for this as well.

pkg/scheduler/sorters.go
yangwwei: Basically, I think the `Running` state should not be depending on the fact that there is a 2nd allocation that happened for the app. Instead, we should read the info from the app, each app can have a field, e.g `org.apache.yunikorn/minTaskNum` (default=1). When this is 1, then once 1 allocation is made, we can say the app is `Running`. Later on, we can leverage this field for FIFO based Gang scheduling.

pkg/scheduler/policies/sorting_policy.go
wilfred-s: fixed

pkg/scheduler/policies/sorting_policy.go
wilfred-s: changed the name

pkg/scheduler/sorters_test.go
wilfred-s: fixed it 7 times in the file

pkg/cache/application_info.go
wilfred-s: simplified this in combination with the remark from @yangwwei

pkg/cache/application_info.go
wilfred-s: Using the existing RunApplication event for it, updated the state transitions to allow this.

pkg/scheduler/scheduler.go
wilfred-s: The Accepted state means that the app is ready for scheduling. When we add the app it is not ready for scheduling it it only ready for scheduling when we add an ask to the app.
The application is in a new state until we get that ask added. The application is rejected by the scheduler if it cannot place the app or it already exists etc. So the app exists in a state that is un-schedulable.

pkg/scheduler/scheduling_application.go
wilfred-s: Same as earlier using the run event

pkg/scheduler/policies/sorting_policy.go
wilfred-s: fixed

pkg/cache/application_state.go
wilfred-s: First doc info added to the jira, need to write a full doc for it, including graphs.
We need to fully document all states that we have in the system.

pkg/scheduler/scheduling_application.go
wilfred-s: yes I overlooked that one

pkg/scheduler/scheduling_application.go
wilfred-s: Also moved this to the same event as the other transitions.

pkg/scheduler/sorters.go
wilfred-s: We can not use this sorting for gang scheduling, there is no dependency between the asks. That means that if the app sets the minimum to 5 the scheduler will just wait with progressing the state of the application. Allocations will still be 1 at a time and can be interrupted by other apps.. This means that there is no guarantee that the scheduler will allocate one or more of them at the same time or even within a limited amount of time.

Setting a default of 1 is not the correct setting. The default would mean that the Starting state is always skipped unless the submitter says otherwise. In essence you turn off the Starting state and thus state aware sorting, except for those apps that want to use it. You need to set a default that will force all apps to go through the starting state for 1 allocation _unless_ the app specifies differently.

yangwwei: do we have any other way to resolve this problem? this is a blocker to me

wilfred-s:
The lint remarks are not simple to fix: if one file is changed the other files in the package are showing up. Fixing the issues would bloat the patch unnecessarily. Leaving it as is for now.

Unit tests pass.

pkg/scheduler/sorters.go
wilfred-s: We can add two policies to implement the sorting and still use the state filtering. That would be like adding  a second state aware policy and renaming what we have now.
i.e. StateAwareFIFO and StateAwareFair

Is that what you mean, as that is really simple to implement.

pkg/scheduler/sorters.go
wilfred-s: Which part is a blocker: the gang side is not possible to solve as this sorting was never designed to solve gang scheduling. Gang scheduling is also not related to the state change that is or is not triggered. A gang would be a set of asks to return only if that set is complete. The app could ask once or more times for a gang, it might also not be the first request coming in.

Two things:
I chose the timeout of 5 minutes, no idea if that is too long or not as the default.

For the single container/pod we can leverage a tag or label on the application. It will not change anything on the filtering side. The only impact will be in the app when moving into and out of the `Starting` state. If the tag or label is set the app itself quickly auto progresses using a `RunApplication` event. We do not skip the state we move the app out of the `Starting` state almost immediately. That solves the single container/pod issue.

Requiring or even allowing multiple allocations to move out of `Starting` is not really a use case that will show up as far as I can assess. It does not cover a gang type request, it does also not fit in with any application type I know.

wilfred-s: We can use a simple label like this: `yunikorn.apache.org/singleAllocation`
Value set I don't think matters, it is passed as a string. The application will progress out of the `Starting` state as follows:
- Any non empty value --> there is just one allocation, _immediate_ auto progress.
- Not present or empty --> progress on the a second allocation or after the default timeout.

yangwwei: based on the discussion. Let's group un-named pods into one single app per namespace. this can resolve the issue.

yangwwei:
hi @wilfred-s this LGTM. once the UT is fixed, feel free to merge this PR.

wilfred-s:
The failing unit test `TestSchedulerRecoveryWhenPlacementRulesApplied` is covered in YUNIKORN-155 it is a known flaky test.
Based on the approval ready for merging.

PR #147, opened 2020-05-13
[YUNIKORN-141] Fix flaky tests

yangwwei:


yangwwei:
hi @wilfred-s

This bothers us too long, I spent a good amount of time today and hopefully now we have a fix. I've run the tests for quite a few times, seems to be much more stable now. The flaky tests actually were caused by a bunch of reasons.

## First, how to reproduce this?
A common issue we run into is, we do a make test, sometimes it fails for the first time, but then it always successful subsequentially. This was because once a go test has run, it will cache some test results, and next time those succeed tests will not run. Our flaky tests are mostly because they are conflicting with other tests. If only part of the tests were run, it mostly can work.

### Try to run

```
go clean -cache -testcache
make test
```
you should be able to reproduce the issue.

## What caused the intermittent failures

There are a few things

### 1. Services (cache, scheduler, rmProxy) were started but never stopped properly.

In the smoke tests, we do start/stop the mocked cluster again and again. When we run these tests together. But when we do stop, none of them were stopped. This leaks a lot of goroutines and probably causes some conflicts.

**Fix**: add stop method for all services and make sure they are stopped while shutting down the mocked cluster.

### 2. timing issue with MultiStepSchedule

MultiStepScheduler is supposed to run several scheduling cycles, and lots of tests are depending on a determined order of execution. However, internally we are handing events in async mode. Which means the order is not always determinable. E.g when you do `tryAllocate()`, the app or ask maybe not added to the `cache/schedulingContext` properly.

**Fix**: add a small sleep (100ms) per scheduling cycle, this gives the rest of the goroutines always have the chance to run.

### 3. scheduling_partition_test is not stable

Such as `TestTryAllocate`, this fails intermittently. Because it relies on the order of queue sorting. See

```
// create a set of queues and apps: app-1 2 asks; app-2 1 ask (same size)
// leaf1 will have an app with 2 requests and thus more unconfirmed resources compared to leaf2
// this should filter up in the parent and the 1st allocate should show an app-1 allocation
// the ask with the higher priority is the second one added alloc-2 for app-1
```
this makes sense but our code doesn't handle that. When sorting queues, we only check the fairness. But actually, when queue fairness is the same, we can try to allocate resources for the queue that has more pending resources.

**Fix**: If that is equal, we can compare the pending resource, and allocate resources for the queue that has more pending resources.

### 4. TestUnManagedSubQueues fails

This is a regression from #135 . cc @adamantal
Just fixed in the latest PR

wilfred-s:
#148 has been merged, should we close this?

PR #151, opened 2020-05-19
[YUNIKORN-169] Panic when removing allocation ask with inflight allocation

yangwwei:
Fix the race condition while allocating an ask and at the same time the ask is removed by app.

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -145,7 +145,7 @@ func (s *Scheduler) processAllocationProposal(allocProposal *commonevents.Alloca
 		return fmt.Errorf("cannot find scheduling partition %s, for allocation ID %s", allocProposal.PartitionName, allocProposal.AllocationKey)
 	}
 
-	return partition.confirmAllocation(allocProposal.ApplicationID, allocProposal.NodeID, allocProposal.AllocationKey, confirm)
+	return partition.confirmAllocation(allocProposal, confirm)
 }
 
 // When a new app added, invoked by external
@@ -295,6 +295,9 @@ func (s *Scheduler) recoverExistingAllocations(existingAllocations []*si.Allocat
 	}
 }
 
+// cache sends the SchedulerAllocationUpdatesEvent to the scheduler, this is to notify the scheduler
+// to update its internal states and keep in sync with the cache. at this point, allocations are either
+// already allocated or released in the cache.
 func (s *Scheduler) processAllocationUpdateEvent(ev *schedulerevent.SchedulerAllocationUpdatesEvent) {
 	if len(ev.ExistingAllocations) > 0 {
 		// in recovery mode, we only expect existing allocations being reported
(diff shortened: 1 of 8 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -315,9 +318,24 @@ func (s *Scheduler) processAllocationUpdateEvent(ev *schedulerevent.SchedulerAll
...
 		alloc := ev.AcceptedAllocations[0]
 		// Update pending resource
-		if err := s.confirmAllocationProposal(alloc); err != nil {
+		if err := s.confirmAllocationProposal(alloc.AllocationProposal); err != nil {
 			log.Logger().Error("failed to confirm allocation proposal",
 				zap.Error(err))
+			// if the scheduler could not confirm the proposal,
+			// we need to notify the cache to revert the allocations
+			for _, allocInfo := range alloc.Allocations {
+				s.eventHandlers.CacheEventHandler.HandleEvent(&cacheevent.ReleaseAllocationsEvent{
```
wangdatan, line 327: Since Once Cache received ReleaseAllocationsEvent, Cache will notify RMProxy: 

```
			m.notifyRMAllocationReleased(rmID, releasedAllocations, toReleaseAllocation.ReleaseType, toReleaseAllocation.Message)
```

For this case (scheduler cannot confirm allocation), I think we should skip notifying RM, because this is not an allocation RM aware of.

pkg/scheduler/scheduling_partition.go
```diff
@@ -439,36 +440,45 @@ func (psc *partitionSchedulingContext) allocate(alloc *schedulingAllocation) boo
 // Called for both allocations from reserved as well as for direct allocations.
 // The unreserve is already handled before we get here so there is no difference in handling.
 // Lock free call this must be called holding the context lock
-func (psc *partitionSchedulingContext) confirmAllocation(appID, nodeID, allocKey string, confirm bool) error {
+func (psc *partitionSchedulingContext) confirmAllocation(proposal *commonevents.AllocationProposal, confirm bool) error {
```
wangdatan, line 443: We also need to notify RMProxy about the confirmed allocation, previously Cache notifies RMProxy about confirmed allocation, now since scheduler will do confirmation and have chances to fail, we need to move the notification scheduler (once allocation is confirmed).

pkg/scheduler/scheduler.go
yangwwei: Currently, shim has no action to take for released allocations.
But this makes sense, it's better to add another `ReleaseType` in the scheduler-interface for this case, maybe`AllocationReleaseResponse_REJECTED_BY_SCHEDULER `? We can track this enhancement in a separate JIRA.

pkg/scheduler/scheduling_partition.go
yangwwei: It makes sense. I've moved the notification code to the scheduler, right after the proposal is confirmed.

pkg/scheduler/scheduler.go
wangdatan: I don't suggest to change scheduler interface for the purpose. scheduler-interface is supposed to be used for communication between core and shim, not within the shim. 

Or we can add a field to indicate the allocation is a "transient allocation" so cache can find there's no need to send the information back to RMProxy.

I think it is fine to have a separate JIRA because notifying RM about the extra released resource should be fine. But let's fix it by 0.9 release (mark priority to be blocker).

wangdatan:
@wilfred, I think the patch should solve the out of sync problem, can you add more details about the problem you saw?

wilfred-s:
I opened #152 which handles the sync issue between the scheduler and the cache. It also fixes two race conditions found while testing.

wangdatan:
@wilfred-s is the #152 is a superset of #151 or it is a separate one?

wilfred-s:
> @wilfred-s is the #152 is a superset of #151 or it is a separate one?

It is a superset of this with a number of extra changes that are needed to fix the inconsistencies and another race condition I saw during testing

PR #152, opened 2020-05-20
[YUNIKORN-169] panic in ask release

wilfred-s:
When an ask is released while an allocation for that ask is processed
the scheduler panics. The panic is caused by a race between two event
handling routines updating the same app. The changes remove the need for
the ask to be accessed on confirmation.
The inflight allocation is now also correctly removed from the cache
when the confirmation fails.

A race condition in reservation removal was fixed.
A sync issue between cache and scheduling node for available resources
was fixed.

### DIFF
--- a/pkg/cache/application_info.go
+++ b/pkg/cache/application_info.go
@@ -73,6 +73,23 @@ func (ai *ApplicationInfo) GetAllAllocations() []*AllocationInfo {
 	return allocations
 }
 
+// Return the allocation for the allocation key. This will return the first allocation for the
+// allocation key. It will only work for the case that the repeat is 1.
+func (ai *ApplicationInfo) GetAllocationUUID(allocKey string) string {
+	ai.lock.RLock()
+	defer ai.lock.RUnlock()
+
+	// just get the first one, ignore multiple repeats for now.
+	// we cannot detect multiple repeats easily here.
+	for uuid, alloc := range ai.allocations {
+		if alloc.AllocationProto.AllocationKey == allocKey {
+			return uuid
+		}
+	}
+	// not found return an empty string
+	return ""
+}
+
 // Return the current state for the application.
 // The state machine handles the locking.
 func (ai *ApplicationInfo) GetApplicationState() string {
--- a/pkg/cache/application_info.go
+++ b/pkg/cache/application_info.go
@@ -24,9 +24,15 @@ import (
 	"time"
 
 	"github.com/looplab/fsm"
+	"go.uber.org/zap"
 
 	"github.com/apache/incubator-yunikorn-core/pkg/common/resources"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/security"
+	"github.com/apache/incubator-yunikorn-core/pkg/log"
+)
+
+var (
+	startingTimeout = time.Minute * 5
 )
 
 /* Related to applications */
@@ -43,7 +49,9 @@ type ApplicationInfo struct {
 	allocatedResource *resources.Resource        // total allocated resources
 	allocations       map[string]*AllocationInfo // list of all allocations
 	stateMachine      *fsm.FSM                   // application state machine
-	lock              sync.RWMutex
+	stateTimer        *time.Timer                // timer for state time
+
+	sync.RWMutex
 }
 
 // Create a new application
@@ -63,8 +71,8 @@ func NewApplicationInfo(appID, partition, queueName string, ugi security.UserGro
 
 // Return the current allocations for the application.
 func (ai *ApplicationInfo) GetAllAllocations() []*AllocationInfo {
-	ai.lock.RLock()
-	defer ai.lock.RUnlock()
+	ai.RLock()
+	defer ai.RUnlock()
 
 	var allocations []*AllocationInfo
 	for _, alloc := range ai.allocations {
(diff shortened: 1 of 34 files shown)


pkg/cache/application_info.go
```diff
@@ -73,6 +73,23 @@ func (ai *ApplicationInfo) GetAllAllocations() []*AllocationInfo {
...
+	// we cannot detect multiple repeats easily here.
+	for uuid, alloc := range ai.allocations {
+		if alloc.AllocationProto.AllocationKey == allocKey {
+			return uuid
+		}
+	}
+	// not found return an empty string
+	return ""
+}
+
```
yangwwei, line 92: This is less efficient, and it assumes a proposal only has 1 allocation. 
To fix the uuid issue, can we use the approach we used in #151 ? That said, to define https://github.com/yangwwei/incubator-yunikorn-core/blob/8f2ffe4cbd54ad2e49bb193a938de0a0bd048009/pkg/scheduler/schedulerevent/events.go#L39. And use this like: https://github.com/yangwwei/incubator-yunikorn-core/blob/8f2ffe4cbd54ad2e49bb193a938de0a0bd048009/pkg/scheduler/scheduler.go#L326-L337.

pkg/scheduler/tests/scheduler_reservation_test.go
```diff
@@ -382,6 +383,86 @@ func TestAddNewNode(t *testing.T) {
 	assert.Equal(t, 0, ms.getPartitionReservations()[appID], "reservations not removed from partition")
 }
 
+// simple reservation one app one queue
+func TestUnReservationAndDeletion(t *testing.T) {
```
yangwwei, line 387: This needs to run multiple times, one time is not enough to reproduce the issue.

wilfred-s: It crashed for me and the logs show that the ask is removed before the allocation is processed. If that does not happen the allocation will be confirmed. The checks at the end of the test case should not return no allocations on the nodes as they are not in the correct state etc.
That would cause the test case to fail.

yangwwei: I usually only get this reproduced after a few runs.  I don't think it crashes every time.

> If that does not happen the allocation will be confirmed. The checks at the end of the test case should not return no allocations on the nodes as they are not in the correct state etc.

If that does not happen, then the allocation proposal was not made because it is already removed. That doesn't cover the case we are expecting it to cover.

pkg/cache/application_info.go
wilfred-s: We have no proper support multiple repeats at the moment.

Let me explain what I think is the problem with multiple repeats and with what you are proposing in code:
If the asks has multiple repeats we might have allocated one or more of those repeats. We're currently just link the ask to the allocation. One ask multiple allocations. Allocations are added and removed based on UUID. So far all is good.

In this case we are removing an ask. One of the repeats of that ask is in the process of being allocated. That ask *can* have a number of confirmed allocations and has at *least* one allocation in flight otherwise we would not have to clean up where we are. There might be unallocated repeats also but they are not important.
We need to clean up the in flight allocation that is one allocation. We cannot just remove all allocations that have been added and confirmed. They should not be impacted as they are not related to the current change. If those allocations need to be removed then that is an allocation removal request. 

Those earlier allocated ask repeats might have been running for a while and because an outstanding ask gets deleted I cannot and may not remove confirmed allocations.
For the case that there are multiple allocation in flight there is a whole other complexity that needs to be looked at.

In the proposed code of PR #151 you are not only cancelling the inflight allocation but also all confirmed allocations, that would be an unexpected behaviour of cancelling an ask.

yangwwei: > We have no proper support multiple repeats at the moment.
I agree. But we should not break the semantics as much as possible.

>In the proposed code of PR #151 you are not only cancelling the inflight allocation but also all confirmed allocations, that would be an unexpected behaviour of cancelling an ask.
To get more consistency, we can do the following

From the cache, it first generates the allocation, and then send exact same message to the scheduler for confirmation. along with the proposal itself. Each time it only handles one allocation for one proposal.

```
m.EventHandlers.SchedulerEventHandler.HandleEvent(&schedulerevent.SchedulerAllocationUpdatesEvent{
		AcceptedAllocations: []*schedulerevent.AcceptedAllocationProposal{
			{
				AllocationProposal: proposal,
				Allocation: allocInfo.AllocationProto,
			},
		},
	})
```

In the scheduler, it processes the proposal, if it fails, notify cache to release the allocation:

```
alloc := ev.AcceptedAllocations[0]
		// Update pending resource
		if err := s.confirmAllocationProposal(alloc.AllocationProposal); err != nil {
			log.Logger().Error("failed to confirm allocation proposal",
				zap.Error(err))
			// if the scheduler could not confirm the proposal,
			// we need to notify the cache to revert the allocations				s.eventHandlers.CacheEventHandler.HandleEvent(&cacheevent.ReleaseAllocationsEvent{
					AllocationsToRelease: []*commonevents.ReleaseAllocation{
						{
							UUID:          alloc.Allocation,
							ApplicationID: alloc.AllocationProposal.ApplicationID,
							PartitionName: alloc.AllocationProposal.PartitionName,
							Message:       "scheduler is unable to confirm the allocation",
							ReleaseType:   si.AllocationReleaseResponse_STOPPED_BY_RM,
						},
					},
				})
		}
```

pkg/scheduler/tests/scheduler_reservation_test.go
wilfred-s: Running multiple times only lowers the chance of that happening but it still does not guarantee that the events come in correctly. The only way that could be done is to stop part of the handling or manipulating the objects directly.

pkg/cache/application_info.go
wilfred-s: New commit has a simple fix by passing back the UUID of the allocation in the proposal instead of the whole allocation with all the things that hang of it.

yangwwei:
The latest commit only adds the UUID to the `AllocationProposal`, this should be OK for now.
+1, let's merge this.
Thanks @wilfred-s

PR #154, opened 2020-05-21
[YUNIKORN-168] make test has become chatty due to verbose clean

HuangTing-Yao:
ignore the output which from make clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,7 @@ test: clean
 
 # Simple clean of generated files only (no local cleanup).
 .PHONY: clean
-@clean:
-	go clean -cache -testcache -r -x ./...
+clean:
+	@echo "cleaning up caches and output"
+	go clean -cache -testcache -r -x ./... 2>&1 >/dev/null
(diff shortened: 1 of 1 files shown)


HuangTing-Yao:
@wilfred-s If we add @ to `rm -rf` which in `clean`, the large amount of remove command log still happen, so I add the @ to the front of clean target. Here is the output of `make test` .
test
But when I run `make clean`, the output is 
clean
I am not sure that this modify is correct or not.

wilfred-s:
I was thinking more of something like this:
```
.PHONY: clean
clean:
	@echo "cleaning up caches and output"
	go clean -cache -testcache -r -x ./... 2>&1 >/dev/null
	-rm -rf _output
```
That gives me an output of:
```
cleaning up caches and output
go clean -cache -testcache -r -x ./... 2>&1 >/dev/null
rm -rf _output
```

HuangTing-Yao:
@wilfred-s I think your way is much better, I will modify this commit.

PR #157, opened 2020-05-22
[YUNIKORN-155] data race in unit test: TestSchedulerRecoveryWhenPlace…

kingamarton:
…mentRulesApplied

### DIFF
--- a/pkg/cache/application_info.go
+++ b/pkg/cache/application_info.go
@@ -254,8 +254,8 @@ func (ai *ApplicationInfo) GetTag(tag string) string {
 }
 
 func (ai *ApplicationInfo) String() string {
-	ai.lock.Lock()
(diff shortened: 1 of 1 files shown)


pkg/cache/application_info.go
```diff
@@ -170,3 +171,10 @@ func (ai *ApplicationInfo) GetTag(tag string) string {
 	}
 	return tagVal
 }
+
+func (ai *ApplicationInfo) String() string {
+	ai.lock.Lock()
+	defer ai.lock.Unlock()
```
yangwwei, line 177: Hi @kingamarton 
I guess this only needs a Read lock, correct?

kingamarton: Yes, @yangwwei you are right. I will change it.

PR #163, opened 2020-06-01
[YUNIKORN-195] Set unmanaged queue sort policy

wilfred-s:
Set the queue application sort policy on a new unmanaged queue based on
the policy set on the parent. This a partial solution for a child queue
template in YUNIKORN-193.

### DIFF
--- a/pkg/cache/queue_info.go
+++ b/pkg/cache/queue_info.go
@@ -42,13 +42,14 @@ const (
 
 // The queue structure as used throughout the scheduler
 type QueueInfo struct {
-	Name       string
-	Parent     *QueueInfo        // link to the parent queue
-	Properties map[string]string // this should be treated as immutable the value is a merge of parent(s)
-	// properties with the config for this queue only manipulated during creation
-	// of the queue or via a queue configuration update
+	Name   string
+	Parent *QueueInfo // link to the parent queue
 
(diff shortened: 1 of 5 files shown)


pkg/cache/queue_info.go
```diff
@@ -350,6 +353,40 @@ func (qi *QueueInfo) MarkQueueForRemoval() {
...
+// Set the properties that the unmanaged child queue inherits from the parent
+// only called on create of a new unmanaged queue
+// This currently only sets the sort policy as it is set on the parent
+// Further implementation is part of YUNIKORN-193
+func (qi *QueueInfo) setTemplateProperties() {
+	qi.Lock()
+	defer qi.Unlock()
+	// for a leaf queue pull out all values from the template and set each of them
+	// See YUNIKORN-193: for now just copy one attr from parent
+	if qi.isLeaf {
```
yangwwei, line 376: It looks like this is limiting to a 2 level setup.
I wonder why it is not just inheriting the properties from its parent, no matter it is a leaf queue or not.

yangwwei: The rest of the patch looks good. I think we can get this merged and unblock the new version. For more improvements, we can address in https://issues.apache.org/jira/browse/YUNIKORN-193

PR #166, opened 2020-06-10
[YUNIKORN-99] State and sort policy documentation

wilfred-s:
Documentation of the sorting policies defined in the core for applications,
nodes and requests. This includes the defaults used by the scheduler and
the way to change the configuration.
Documentation for the yunikorn core state machines with events. This is
a required part for the stateaware policy.


docs/sorting_policy.md
```diff
@@ -0,0 +1,129 @@
...
+    * one application in the _starting_ state
+    * one application in the _accepted_ state
+* applications must have outstanding requests
+
+The application in the _accepted_ state is only added if there is no application in the _starting_ state.
+The application in the _starting_ state does not have to have outstanding requests.
+Any application in the _starting_ state will prevent _accepted_ applications from being added to the filtered list.
+
+After the list is filtered all applications are sorted on create time.
+
```
yangwwei, line 86: It would be nice to give an example to illustrate how this works, e.g with Spark jobs.

docs/sorting_policy.md
```diff
@@ -0,0 +1,129 @@
...
+      queues:
+      - name: sandbox
+        properties:
+          application.sort.policy: fifo
+```
+
+The only applications that are considered during scheduling must have outstanding requests.
+A filter is applied _while_ sorting the applications to remove all that do not have outstanding requests.
+
+### FifoSortPolicy
```
yangwwei, line 48: actually a lot of users are confused about the FIFO policy.
one common misunderstanding is when FIFO is applied, does it mean job2 won't be scheduled until job1 is done? I think we need some more doc to explain this.

yangwwei: Let me create a lower priority jira to track this. We can do this later.

yangwwei:
Created https://issues.apache.org/jira/browse/YUNIKORN-219 for the follow up tasks.

PR #167, opened 2020-06-11
[YUNIKORN-216] Fixed helm chart link in deployment doc

kingamarton:


yangwwei:
The UT failure was because a race: https://issues.apache.org/jira/browse/YUNIKORN-221. The doc fix looks good.

PR #168, opened 2020-06-12
[YUNIKORN-222] Race in state change

wilfred-s:
If the application allocation is not confirmed before the ask is updated
the state is changed from accepted to running (via waiting) instead of
accepted to starting. This breaks the stateaware scheduling policy of a
queue.

### DIFF
--- a/pkg/cache/cachetestutils.go
+++ b/pkg/cache/cachetestutils.go
@@ -38,6 +38,11 @@ func AddAllocationToApp(app *ApplicationInfo, alloc *AllocationInfo) {
 	app.addAllocation(alloc)
 }
 
+// Remove allocation to cache app for tests
+func RemoveAllocationFromApp(app *ApplicationInfo, uuid string) {
+	app.removeAllocation(uuid)
+}
+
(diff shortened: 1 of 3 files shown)


yangwwei:
hi @wilfred-s 

The change itself looks good. But I do not fully understand why this causes the problem.
From `Accepted` to `Waiting`, this is triggered when `removeAllocationAsk ` is called, which is called when there is a container released. But when we run a spark job, when the driver is allocated, the job is at `Accepted` state, then there will be other executors adding to this job. there is no release action could happen here, why it would go to `Waiting` and `Running` then?

wilfred-s:
The race is linked to when we update the repeat on the ask. I removed the transition check from the repeat update. When we create a proposal we lower the repeat, if the cache pushes back we cancel the proposal and increase the ask again. During proposal processing we also change the pending, inflight and allocated values. These changes race with each other because the cache is leading for an allocation while the scheduler is leading for all other.
The scheduler does not get told that the allocation is confirmed until after the fact.

wilfred-s:
Spoken off-line with Weiwei, merging the change as he approved

PR #172, opened 2020-06-18
[YUNIKORN-107] Display additional queue information

kingamarton:


pkg/cache/queue_info_test.go
```diff
@@ -425,6 +425,8 @@ func TestGetQueueInfos(t *testing.T) {
 	assert.NilError(t, err, "failed to create queue: %v", err)
 	err = parent.IncAllocatedResource(parentUsed, false)
 	assert.NilError(t, err, "failed to increment allocated resource: %v", err)
+	parent.properties = make(map[string]string)
+	parent.properties[configs.ApplicationSortPolicy] = "fifo"
```
yangwwei, line 429: The UT is verifying the properties set in the parent level.
Can we add some more UT for this, including:
- add multiple properties in leaf queue
- test coverage with unmanaged queues' properties (see more in https://github.com/apache/incubator-yunikorn-core/pull/163)
- test no properties set for any of the queues

kingamarton: I deleted this changes from this test case, so this will be the case when no properties are set
Also added a new test case when I set some properties in the leaf queues as well as in the root queue, and also added some un managed leaf queues.

PR #173, opened 2020-06-19
[YUNIKORN-231] Add ability to update pod condition based on accurate scheduling state

yangwwei:
On cloud env, when partition resource (physical nodes) resource is not enough, but YK still has enough capacity, we do want to trigger auto-scaling; but if queue capacity is exceeded, we do not want to trigger auto-scaling.

Introduce a scheduler plugin to send some supplementary info to the shim to handle this accordingly. This one depends on https://github.com/apache/incubator-yunikorn-scheduler-interface/pull/21.


wilfred-s:
I kicked of the build again to get the tests to run.
Travis failed to run test and lint targets

Is this really still in a draft or is this good to go?

pkg/scheduler/scheduling_queue.go
```diff
@@ -499,6 +499,37 @@ func (sq *SchedulingQueue) getHeadRoom() *resources.Resource {
 	return resources.ComponentWiseMin(headRoom, parentHeadRoom)
 }
 
+// this function returns the max headRoom of a queue
+// this doesn't get the partition resources into the consideration
+func (sq *SchedulingQueue) getMaxHeadRoom() *resources.Resource {
```
wilfred-s, line 504: The getMaxHeadRoom and getHeadRoom share most of the code.
Can we simplify this by taking the recursive code out for the two different calculations:
```
func (sq *SchedulingQueue) getHeadRoom() *resources.Resource {
	var parentHeadRoom *resources.Resource
	if sq.parent != nil {
		parentHeadRoom = sq.parent.getHeadRoom()
	}
	return sq.internalHeadRoom(parentHeadRoom)
}

func (sq *SchedulingQueue) getMaxHeadRoom() *resources.Resource {
	var parentHeadRoom *resources.Resource
	if sq.parent != nil {
		parentHeadRoom = sq.parent.getMaxHeadRoom()
	} else {
		return nil
	}
	return sq.internalHeadRoom(parentHeadRoom)
}
```

And then in the `internalHeadRoom` the real calculation:
```
func (sq *SchedulingQueue) internalHeadRoom(parentHeadRoom *resources.Resource) *resources.Resource {
	sq.RLock()
	defer sq.RUnlock()
	headRoom := sq.QueueInfo.GetMaxResource()

	// if we have no max set headroom is always the same as the parent
	if headRoom == nil {
		return parentHeadRoom
	}
	// calculate unused
	headRoom.SubFrom(sq.allocating)
	headRoom.SubFrom(sq.QueueInfo.GetAllocatedResource())
	// check the minimum of the two: parentHeadRoom is nil for root
	if parentHeadRoom == nil {
		return headRoom
	}
	return resources.ComponentWiseMin(headRoom, parentHeadRoom)
}
```

pkg/scheduler/scheduling_queue_test.go
```diff
@@ -688,11 +692,295 @@ func TestHeadroom(t *testing.T) {
...
 	}
 	headRoom = leaf2.getHeadRoom()
+	maxHeadRoom = leaf2.getMaxHeadRoom()
+	if !resources.Equals(res, headRoom) || !resources.Equals(res, maxHeadRoom)  {
+		t.Errorf("leaf1 queue head room not as expected %v, got: %v (err %v)", res, headRoom, err)
+	}
+}
+
+func TestMaxHeadroom(t *testing.T) {
+	// create the root: nil test
```
wilfred-s, line 707: I am missing tests with a maxResource set on the root queue.
You can set it on the root queue via the `qi.setMaxResource()`

pkg/plugins/plugins.go
```diff
@@ -50,6 +50,12 @@ func RegisterSchedulerPlugin(plugin interface{}) {
 		plugins.reconcilePlugin = t
 		registered = true
 	}
+	if t, ok := plugin.(ContainerSchedulingStateUpdater); ok {
+		log.Logger().Debug("register scheduler plugin",
+			zap.String("type", "ContainerSchedulingStateUpdater"))
```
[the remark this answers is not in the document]
yangwwei: Yes. simplified and updated.

go.mod
```diff
@@ -51,3 +51,5 @@ require (
 	gotest.tools v2.2.0+incompatible
 	honnef.co/go/tools v0.0.1-2020.1.3 // indirect
 )
+
+replace github.com/apache/incubator-yunikorn-scheduler-interface => /Users/wyang/workspace/github/wyang/incubator-yunikorn-scheduler-interface
```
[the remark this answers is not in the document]
yangwwei: This is because apache/incubator-yunikorn-scheduler-interface#21 has not been merged. I will remove this once that one is merged.

pkg/plugins/plugins.go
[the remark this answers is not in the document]
yangwwei: But we could not move to `switch` because the logic here allows an object that implements multiple plugin interfaces. I added a UT for this too in the latest patch.

pkg/scheduler/scheduling_application.go
```diff
@@ -418,12 +431,34 @@ func (sa *SchedulingApplication) tryAllocate(headRoom *resources.Resource, ctx *
...
 		if nodeIterator := ctx.getNodeIterator(); nodeIterator != nil {
 			alloc := sa.tryNodes(request, nodeIterator)
-			// have a candidate return it
-			if alloc != nil {
+			if alloc == nil {
+				// we have enough headroom, but we could not find a node for this request,
+				// this can happen when non of the nodes is qualified for this request,
+				// by satisfying both conditions:
+				//   1) node has enough resources;
+				//   2) node satisfies all placement constraints of the request (e.g predicates)
```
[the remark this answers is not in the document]
yangwwei: I think it is working like this:
1. if we visit all nodes, no allocation or reservation is found (alloc==nil), then we do trigger the auto-scaling for this pod.
2. if we have a reservation, we hold on triggering the auto-scaling for it; but we will need to revisit the request again, that will hit  -> 1 again. this way we still need to trigger the auto-scaling for it.

pkg/scheduler/scheduling_queue.go
yangwwei: Yeah. this is brilliant, I just updated the code like this.

pkg/scheduler/scheduling_queue_test.go
yangwwei: That is part of the coverage: https://github.com/yangwwei/incubator-yunikorn-core/blob/d0eeadf50a51cafaca84ae7b2e7b78bf69ee0954/pkg/scheduler/scheduling_queue_test.go#L817-L832

pkg/scheduler/scheduling_application.go
```diff
@@ -408,6 +408,19 @@ func (sa *SchedulingApplication) sortRequests(ascending bool) {
 	}
 }
 
+// update container scheduling state to shim if the plugin is registered
+func (sa *SchedulingApplication) updateContainerSchedulingStateIfNeeded(ask *schedulingAllocationAsk,
```
wilfred-s, line 412: The `IfNeeded` is not correct. There is no state or other check here we call always if the plugin is configured.
Should be `updateContainerSchedulingState`

pkg/scheduler/scheduling_queue_test.go
wilfred-s: These tests can be greatly simplified. We also should not mix the two headroom calculation tests.
filed follow up jira to fix this

pkg/scheduler/scheduling_application.go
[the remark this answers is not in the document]
wilfred-s: In this case we might not have been able to allocate or reserve due to affinity. Scaling up a node will not help. This is the case that in earlier discussions we tried to prevent from causing the cluster to scale up.

In a smaller cluster we might have already reserved all nodes and thus return no allocation. These reservations would have already triggered scale up.
I think we need to be conservative in this case and scale up on reservation:
```
	if alloc.result == reserved {
		sa.updateContainerSchedulingState(...)
	}
	return alloc
```

pkg/scheduler/scheduling_queue_test.go
yangwwei: OK, thanks!

pkg/scheduler/scheduling_application.go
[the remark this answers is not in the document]
yangwwei: There are a few issues if we only trigger auto-scale based on reservation:
- Compute nodes might be 0, that means no reservation at all at the initial phase
- On a small cluster, this approach will dramatically slow down the pace of auto-scaling. If we have lots of pods pending, it may only trigger the auto-scaling 1 node at a time.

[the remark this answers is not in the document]
wilfred-s: The scale up with 0 nodes will not be triggered here at all. It will have been triggered above in the `FitIn()` check. The `headRoom` check fails and the `maxHeadRoom` passes. We thus scale immediately. We cannot reserve in that case or even try to allocate if there are no nodes.

In small clusters we also do not want to go overboard and keep scaling up. If there is just one node, or maybe a couple of nodes, in the cluster it is also better to not go and trigger large numbers of scale up. The scaled up node will more often than not fit a number of requests. That should thus account for the backlog. 
If we still run out we keep scaling up. Keep in mind that we also scale up if the container does not fit in the headroom. So for large container we might have already triggered scale up due to headroom restrictions.

yangwwei:
> I have one last comment: if we allocate the request the pod will be assigned etc. Does that automatically trigger an update of the state or do we need to handle that too?

If we allocate the pod, the pod state will be changed to PodScheduled=true (from false), this will be done by kubelet, not the scheduler. Something like:

```
Conditions:                                                                                                          
Type              Status                                                                                                    
Initialized       True                                                                                                           
Ready             True
ContainersReady   True 
PodScheduled      True
```

pkg/scheduler/scheduling_application.go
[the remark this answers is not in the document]
yangwwei: this is no longer valid with the latest changes.

pkg/scheduler/outstanding_requests.go
```diff
@@ -0,0 +1,43 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package scheduler
+
+// an outstanding request is a request that fits into the queue capacity but in
+// pending state because there is no enough partition resources to allocate it
+type outstandingRequests struct {
```
wangdatan, line 23: Regarding style: Why use typedef struct instead of a simple array[]?

pkg/scheduler/scheduling_queue.go
```diff
@@ -482,9 +482,26 @@ func (sq *SchedulingQueue) getHeadRoom() *resources.Resource {
 	if sq.parent != nil {
 		parentHeadRoom = sq.parent.getHeadRoom()
 	}
+	return sq.internalHeadRoom(parentHeadRoom)
+}
+
+// this function returns the max headRoom of a queue
+// this doesn't get the partition resources into the consideration
+func (sq *SchedulingQueue) getMaxHeadRoom() *resources.Resource {
```
wangdatan, line 490: Maybe I missed something here, I think we don't need getMaxHeadRoom (and it should be Headroom instead of HeadRoom).

What we really need is to be able to pass the parent's Headroom (instead of getting from parent.Headroom, which is be used by scheduling).

Also, this patch doesn't take hiearchical resource quota into account. We can do this better by doing the following logic. 

```
// Return total assigned resource of this queue
queue.getQueueOutstandingRequests(parentHeadroom, outstandingRequests) *Resource{
	myHeadroom = queue.getHeadRoom(parentHeadroom)

	totalOutstandingResource := 0

	if queue.isLeafQueue() {
       for _, app := range sq.sortApplications() {
       	  app.getOutstandingRequests(myHeadRoom, total)
       }
	} else {

	   for child : queue.children() {
	       assigned := child.getQueueOutstandingRequests(myHeadroom, outstandingRequests)

	       myHeadroom -= assigned

	       totalOutstandingResource += assigned 

	       if myHeadroom <= 0 {
	           return totalOutstandingResource
	       }
	   }
	}
}

// And for root, it passes nil for parentHeadroom.
```

pkg/scheduler/scheduler.go
```diff
@@ -110,6 +111,39 @@ func (s *Scheduler) internalPreemption() {
...
 }
 
+func (s *Scheduler) internalInspectOutstandingRequests() {
+	for {
+		time.Sleep(1000 * time.Millisecond)
+		s.inspectOutstandingRequests()
+	}
+}
+
+func (s *Scheduler) inspectOutstandingRequests() {
```
wangdatan, line 121: Suggest to rename it to `UpdateUnschedulableStateForOutstandingRequests`

pkg/scheduler/scheduling_application.go
```diff
@@ -409,6 +409,21 @@ func (sa *SchedulingApplication) sortRequests(ascending bool) {
 	}
 }
 
+func (sa *SchedulingApplication) getOutstandingRequests(headRoom *resources.Resource, total *outstandingRequests) {
```
wangdatan, line 412: Suggest to rename to `getAllocatableOutstandingRequests`

pkg/scheduler/scheduling_partition.go
```diff
@@ -356,6 +356,17 @@ func (psc *partitionSchedulingContext) removeSchedulingNode(nodeID string) {
 	}
 }
 
+func (psc *partitionSchedulingContext) calculateOutstandingRequests() *outstandingRequests {
```
wangdatan, line 359: Suggest to rename to `getAllocatableOutstandingRequests`

pkg/scheduler/scheduling_queue.go
```diff
@@ -558,6 +575,19 @@ func (sq *SchedulingQueue) tryAllocate(ctx *partitionSchedulingContext) *schedul
 	return nil
 }
 
+func (sq *SchedulingQueue) getQueueOutstandingRequests(total *outstandingRequests) {
```
wangdatan, line 578: Suggest to rename to `getAllocatableOutstandingRequests`

wilfred-s:
Looking at the change using the re-implemented logic I think we have a far more fundamental issue. Specially when we take into account a queue hierarchy.
It does not seem to be stable over time and work consistently over all the sorting algorithms we can use. This marking only seems to work consistently for the FIFO sorting.

The outstanding requests are not tracked and a new list is build each time we run the checks. If we have a change in the pending requests for the queues the order of the sorted queues can change. Queues are always sorted based fair. If there is a change a queue which was at the end of the could end up at the front. That could lead to us looking at a different set of requests.
example: root -> parent -> leaf1 & leaf2. Parent has a limit set.
Leaf1 has app1 and leaf2 has app2. First time the check runs the queue order is leaf1, leaf2. App1 requests are marked for to trigger scale up. Second time the order is changed to leaf2, leaf1. We now mark the requests for app2 to trigger scale up. We thus overshoot the parent limit.

Then we have similar issues inside a queue. If there is a change within the requests for an application we can cause issues at two levels: inside an app and over multiple apps.
Using a priority sorting a newly added request with higher priority will be placed at the start of the list. Even if we already had marked enough request to trigger upscaling we now mark the new one wit the higher priority for scale up  as they are the first in the list.
Same for a fair sorting over applications. Newly added requests can change the order of the applications inside a queue. An application that was first in the list might shift to the back. This will cause a different app with new requests to be looked at first and again we will overshoot.
This mechanism can also not support application priorities at the application level for the same reason as mentioned above.

We need a mechanism that works, and is reproducible, if you use a queue hierarchy with any sorting mechanism at all levels: queues and apps.

wangdatan:
@wilfred-s I think what you suggested make sense, and (if you remember) that is why initially in YuniKorn core implementation, we have logic to get "allocatable resource request" first (which consider fairness for the "potential allocated" resource).  

If we want to achieve a close-to-truth outstanding resource request sorting which covers point you suggested, it will be a major effort which will take months and ideally it needs to exchange information with autoscaler. 

My suggestion is to get this done in phases, for phase #1, I think it is relatively safe to assume the following thing: 

**Resource requests are relatively homogeneous**: which means if new node added for app1's total 200 GB pending mem, even if app1 somehow cannot consume the new node resource immediately (maybe caused by FIFO order). there will be some other pending pods can pick it up and use it.  

If we agree with the above assumption, Weiwei's patch should be a good starting point for phase#1. If we want to do it better, here're follow-ups we can do:  

- P1: Prioritize daemonset/service PODs for queues which are under utilized. Daemonset/service pods are likely to be heterogeneous and requires anti-affinity, etc.
- P1: Do the algorithm in two passes: first pass will select queue which has usage under guaranteed capacity (usage ratio < 1.0), and fill them to >=1. second pass will fill queues to maximum capacity.  
- P2: Sort queues/apps based on the possible resource usage. 

Thoughts?

wilfred-s:
I have opened #181 which handles all the points talked about in the previous update:
It provides a an implementation that works: in a hierarchy, time stable and for all sorting policies we have.
Plugin and interface changes are the same as in this PR.

pkg/scheduler/scheduling_queue.go
wilfred-s: I do not see how this would solve the hierarchy issue.
It might even cause more issues as the headroom for a queue might be different for different locations in the queue hierarchy. The normal headroom includes the max from the root (i.e cluster size) so we cannot use that and need to use the max as defined below to start with.

pkg/scheduler/scheduling_application.go
wilfred-s: The requests are not allocatable, if they were we would and should not see them in the list.

pkg/scheduler/scheduling_queue_test.go
```diff
@@ -688,11 +694,296 @@ func TestHeadroom(t *testing.T) {
...
+
+	headRoom = leaf2.getMaxHeadRoom()
+	assert.Assert(t, headRoom == nil, "headRoom of leaf2 should be nil because no max set for all queues")
+
+	// recreate the structure, set max capacity only in parent level
+	// structure is:
+	// root (max: nil)
+	//   - parent (max: 20,8)
+	//     - leaf1 (max: nil)  (alloc: 5,3)
+	//     - leaf2 (max: nil)  (alloc: 6,4)
```
wilfred-s, line 773: This is testing the special case of the below setup with one leaf with a max. Both do the same for the leaf queue that has no limit set.
Please remove this case.

pkg/scheduler/scheduling_queue_test.go
```diff
@@ -688,11 +694,296 @@ func TestHeadroom(t *testing.T) {
...
+	if !resources.Equals(res, headRoom) {
+		t.Errorf("leaf2 queue head room not as expected %v, got: %v (err %v)", res, headRoom, err)
+	}
+
+	// recreate the structure, set max capacity in root level, verify that will be ignored
+	// structure is:
+	// root (max: 1, 1)
+	//   - parent (max: 20,8)
+	//     - leaf1 (max: nil)  (alloc: 5,3)
+	//     - leaf2 (max: nil)  (alloc: 6,4)
```
wilfred-s, line 825: This does not add anything that is/cannot tested by calling getMaxHeadroom on the root queue directly.
Should be rolled into the `nil` tests at the start.
Please remove this case.

pkg/scheduler/scheduling_queue_test.go
```diff
@@ -688,11 +694,296 @@ func TestHeadroom(t *testing.T) {
...
+		t.Errorf("leaf2 queue head room not as expected %v, got: %v (err %v)", res, headRoom, err)
+	}
+
+	// recreate the structure, set max capacity in parent and a leaf queue
+	// structure is:
+	// root (max: nil)
+	//   - parent (max: 20,8)
+	//     - parent1 (max: nil)
+	//       - parent2 (max: 10,4)
+	//         - leaf1 (max: nil)   (alloc: 3,3)
```
wilfred-s, line 937: This tests the same as a parent with a leaf below it and does not include any new test case.
please remove this case

pkg/scheduler/scheduling_queue_test.go
```diff
@@ -827,3 +1118,157 @@ func TestIsEmpty(t *testing.T) {
...
+	assert.Equal(t, len(queue2Total.getRequests()), 0)
+
+	rootTotal = newOutstandingRequests()
+	root.getQueueOutstandingRequests(rootTotal)
+	assert.Equal(t, len(rootTotal.getRequests()), 5)
+}
+
+func TestGetOutstandingRequestWithStateAwareSortingPolicy(t *testing.T) {
+	const app1ID = "app1"
+	const app2ID = "app2"
```
wilfred-s, line 1209: A policy must not impact the tests. The outcome of the policy is a filtered list of sorted applications. The internal working of the policy is tested in the policy testing.
The policies internal working will never directly impact the tests.
Please remove the test.

pkg/scheduler/scheduling_queue.go
wilfred-s: See above, not sure that this is allocatable

pkg/scheduler/scheduling_partition.go
wilfred-s: see above, not sure that this is allocatable

pkg/scheduler/scheduling_queue_test.go
```diff
@@ -827,3 +1118,157 @@ func TestIsEmpty(t *testing.T) {
 	leaf.addSchedulingApplication(app)
 	assert.Equal(t, leaf.isEmpty(), false, "queue with registered app should not be empty")
 }
+
+func TestGetOutstandingRequest(t *testing.T) {
+	const app1ID = "app1"
+	const app2ID = "app1"
```
wilfred-s, line 1124: The logic for calculating this is in the app and not in the queue.
If the app returns the correct thing the queue just add them at this point. We should test the app not the queue.

pkg/scheduler/outstanding_requests.go
yangwwei: Removed.

pkg/scheduler/scheduling_application.go
yangwwei: Since there is no consensus on the name of this function, I'd prefer to keep it as it is for now.
We can revisit this if we can come up with a better name, this is a non-functional change, not critical.

pkg/scheduler/scheduler.go
yangwwei: I found `UpdateUnschedulableStateForOutstandingRequests` is also a bit confusing. Instead of updating this function name, I added more code comment to explain what it does:

```
// inspect on the outstanding requests for each of the queues,
// update request state accordingly to shim if needed.
// this function filters out all outstanding requests that being
// skipped due to insufficient cluster resources and update the
// state through the ContainerSchedulingStateUpdaterPlugin in order
// to trigger the auto-scaling.
```

pkg/scheduler/scheduling_queue.go
yangwwei: based on the discussion, we don't have to solve this for now. correct?

pkg/scheduler/scheduling_queue_test.go
yangwwei: Removed

pkg/scheduler/scheduling_queue_test.go
yangwwei: removed

pkg/scheduler/scheduling_queue_test.go
yangwwei: Removed

pkg/scheduler/scheduling_queue_test.go
yangwwei: Removed.

pkg/scheduler/scheduling_queue_test.go
yangwwei: we can add more tests for `app.getOutstandingRequests()`, but this is covering more code paths by including the queue, I suggest to keep it.

pkg/scheduler/scheduling_queue.go
wilfred-s: You cannot solve this without the changes as proposed in #173.
Without that making any changes here does not help. So no, it just will not work correctly in a queue hierarchy.

pkg/scheduler/scheduling_queue_test.go
wilfred-s: we'll take this up via the follow up jira

PR #174, opened 2020-06-21
[YUNIKORN-226] Remove panic calls from scheduler

HuangTing-Yao:


pkg/cache/cluster_info.go
```diff
@@ -85,7 +85,7 @@ func (m *ClusterInfo) handleSchedulerEvents() {
 		case *commonevents.RemoveRMPartitionsEvent:
 			m.processRemoveRMPartitionsEvent(v)
 		default:
-			panic(fmt.Sprintf("%s is not an acceptable type for scheduler event.", reflect.TypeOf(v).String()))
+			log.Logger().Warn(fmt.Sprintf("%s is not an acceptable type for scheduler event.", reflect.TypeOf(v).String()))
```
yangwwei, line 88: hi @HuangTing-Yao 

This should be at the ERROR level.
Another thing is, instead of using `fmt`, can you pls follow the logging format such as https://github.com/apache/incubator-yunikorn-core/blob/772e3cbdbdbda397468f19e09f137433caf5f144/pkg/scheduler/scheduler.go#L149-L153.

PR #177, opened 2020-06-25
[YUNIKORN-241] priority sorting quick fix

wilfred-s:
Quick fix to change priority sorting to FIFO sorting for requests based on
the fact that the shim does not set priority.


wilfred-s:
The unit test failure is due to the known data race logged as YUNIKORN-202

PR #178, opened 2020-06-25
[YUNIKORN-256] remove error as a return from SendEvent

adamantal:


wilfred-s:
I have moved this to draft as it will need a go.mod change after we have changed the interface and committed it

adamantal:
Merged master and removed the fail case in event_publihser_test.go. Since we have no dependency on the SI I'll convert this back from draft.

PR #179, opened 2020-06-26
[YUNIKORN-251] yunikorn-core recovers allocation UUID from the reported existing allocations

yangwwei:
This is needed by https://github.com/apache/incubator-yunikorn-k8shim/pull/140.
The shim side changes make sure the UUID is set per existing allocation, the core side needs to use the same UUID to track these recovered allocations.


wilfred-s:
I just thought of a huge issue: if the shim has not passed in a UUID for the allocations we will lose track of the allocation.
If the `UUID == ""` it should fail. I added the changes on top of what you did and will commit when we pass the build

PR #181, opened 2020-07-07
[YUNIKORN-231] update pod condition for scale up

wilfred-s:
To support a cluster that can grow or shrink (auto scale) we need to
provide information back to an auto scaler. Auto scaling can not just be
triggered by an outstanding request as we have queue quotas to consider.

We should only mark a request to trigger an auto scale if the queue, and
the parent queues, can still accommodate the request within the max queue's
limits.
If enough requests have been marked to trigger an upscale changes to the
application or request ordering do not cause new requests to be marked for
upscaling as that can lead to over provisioning of nodes.


wangdatan:
@wilfred-s , thanks for the patch, If I understand this patch correct, I think this patch doesn't consider the sorting order based on upscaling resource.

I can see the upside of this patch is, it is part of scheduler allocation logic. But it is also the downside of the patch: it is too intrusive and the following logic is quite different from of existing allocation logic. 

```
   if resources.FitIn(maxHeadRoom, request.AllocatedResource) {
   }
```

I hope I didn't miss anything: even though this patch took a different approach, but what this patch can achieve is pretty similar to #173 plus my comment https://github.com/apache/incubator-yunikorn-core/pull/173#discussion_r449813724

wilfred-s:
> If I understand this patch correct, I think this patch doesn't consider the sorting order based on upscaling resource.

The effect achieved is not comparable to #173 the base change using a separate routine to trigger upscaling will never be stable over time as it does not track the old state. Every single time it runs it assumes nothing has happened up until that point in time. We need to track the old state to get the stability over time. You cannot get it without tracking.

>I can see the upside of this patch is, it is part of scheduler allocation logic. But it is also the downside of the patch: it is too intrusive and the following logic is quite different from of existing allocation logic.

The point that this logic is called we're in the track that skips allocating. Instead of skipping it directly we now check if we need to trigger an upscale. It is the most logical location to make this decision as we have done all the sorting already.

The sorting order is considered. As an example: if we order based on request priority, we will honour the request with the highest priority first even if that was not the one that originally triggered the scale up for the application. Similar for the sorting order over applications. If the queue has triggered upscale for an app as that was the first app in the sorting order and circumstances change pushing that app to the back then that will be honoured too as part of the normal scheduling.

wilfred-s:
closing for now, the code has changed to much to apply this to the HEAD now.

PR #184, opened 2020-07-13
[YUNIKORN-278] Outstanding requests in queues with no max capacity set should always trigger auto-scaling.

yangwwei:
Currently, `getMaxHeadRoom()` returns nil when capacity is not set, we expect this queue has unlimited capacity; however, `resources.FitIn(headRoom, request.AllocatedResource)` will always return false in such case
this causes the problem if the queue has no capacity set, pods will not be set to unschedulable to trigger auto-scaling (we expect this is always setting to unschedulable)


wilfred-s:
I kicked of the build again. I was not expecting the failures seen. As soon as that is done I'll have another look but the code change itself good for now.

PR #185, opened 2020-07-16
[YUNIKORN-287] reserved ask release double release

wilfred-s:
When an ask gets released that is reserved and being allocated the
counters tracking reservations can be updated twice for one ask release.
The counters keep track of the number of reservations on a queue or
partition.

This can lead to other reservations being ignored. Nodes that have been
reserved by other asks for the same app will be skipped during
scheduling. In a small cluster this could lead to resource starvation


wilfred-s:
I have added a test case which fails without the changes and passes with the changes.

pkg/scheduler/scheduling_application.go
```diff
@@ -325,39 +328,50 @@ func (sa *SchedulingApplication) reserve(node *SchedulingNode, ask *schedulingAl
...
+	if num, err = node.unReserve(sa, ask); err != nil {
+		return 0, err
+	}
+	// if the unreserve worked on the node check the app
 	if _, found := sa.reservations[resKey]; found {
-		if err := node.unReserve(sa, ask); err != nil {
-			return err
+		// worked on the node means either found or not but no error, log difference here
+		if num == 0 {
+			log.Logger().Debug("reservation not found while removing from node, app has reservation",
```
wangdatan, line 360: I think we should use Info here.

pkg/scheduler/scheduling_application.go
```diff
@@ -325,39 +328,50 @@ func (sa *SchedulingApplication) reserve(node *SchedulingNode, ask *schedulingAl
...
+				zap.String("appID", sa.ApplicationInfo.ApplicationID),
+				zap.String("nodeID", node.NodeID),
+				zap.String("ask", ask.AskProto.AllocationKey))
 		}
 		delete(sa.reservations, resKey)
-		return nil
+		return 1, nil
 	}
 	// reservation was not found
 	log.Logger().Debug("reservation not found while removing from app",
```
wangdatan, line 369: same comment, suggest to use info here.

wilfred-s: will fix on commit

pkg/scheduler/scheduling_application.go
wilfred-s: will fix on commit

PR #190, opened 2020-07-29
[YUNIKORN-332] Add events for reserved pods

adamantal:


adamantal:
Started testing it, but the results were different from the expectation (not all events that I expected from the pods showed up in K8s event system). Will take a look at it, in the meantime I uploaded this PR as draft to be reviewed.

adamantal:
Plenty of questions here, let me try to summarize first where is the latest status on this.

1. The current implementation works as follows:
- If the pod's resource request can be accomodated, it will be started without any event created neither for the pod nor the node. This currently also means the "PodBindSuccessful" event emitted by the shim itself - I did not touch this part.
```
$ kubectl describe pod pod-with-enough-resources
...
Events:
  Type     Reason             Age                 From                     Message
  ----     ------             ----                ----                     -------
  Normal   Scheduling         26m                 yunikorn                 default/task0 is queued and waiting for allocation
  Normal   Scheduled          26m                 yunikorn                 Successfully assigned default/task0 to node docker-desktop
  Normal   PodBindSuccessful  26m                 yunikorn                 Pod default/task0 is successfully bound to node docker-desktop
  Normal   Started            23m (x4 over 25m)   kubelet, docker-desktop  Started container sleep-30s
  Normal   Pulling            22m (x5 over 25m)   kubelet, docker-desktop  Pulling image "alpine:latest"
  Normal   Pulled             22m (x5 over 25m)   kubelet, docker-desktop  Successfully pulled image "alpine:latest"
  Normal   Created            22m (x5 over 25m)   kubelet, docker-desktop  Created container sleep-30s
```
- If the pod's resource usage is too high to be spawned in some node, it will be "reserved" by the scheduler:
From yunikorn-scheduler log:
```
2020-08-03T15:50:13.614+0200	INFO	scheduler/scheduling_partition.go:529	allocation ask is reserved	{"appID": "application-sleep-0003", "queue": "root.default", "allocationKey": "a9446376-981c-4499-8ce3-855eb8571ba9", "node": "docker-desktop"}
```
From k8s commands:
```
AAntal-MBP15:yunikorn-k8shim adamantal$ kubectl get events --field-selector=involvedObject.kind=Node
LAST SEEN   TYPE     REASON                        OBJECT                MESSAGE
6m41s       Normal   NodeHasSufficientMemory       node/docker-desktop   Node docker-desktop status is now: NodeHasSufficientMemory
6m41s       Normal   NodeHasNoDiskPressure         node/docker-desktop   Node docker-desktop status is now: NodeHasNoDiskPressure
6m41s       Normal   NodeHasSufficientPID          node/docker-desktop   Node docker-desktop status is now: NodeHasSufficientPID
6m21s       Normal   RegisteredNode                node/docker-desktop   Node docker-desktop event: Registered Node docker-desktop in Controller
6m20s       Normal   Starting                      node/docker-desktop   Starting kube-proxy.
2m33s       Normal   NodeAccepted                  node/docker-desktop   node docker-desktop is accepted by the scheduler
107s        Normal   AllocationAskReservedOnNode   node/docker-desktop   Ask a9446376-981c-4499-8ce3-855eb8571ba9 from application application-sleep-0003 is reserved on this node
```
but I also have to add that there's no trace of the `AllocationAskReservedOnNode` in the output of the `kubectl describe node docker-desktop` command:
```
Events:
  Type    Reason                   Age                From                        Message
  ----    ------                   ----               ----                        -------
  Normal  Starting                 41m                kubelet, docker-desktop     Starting kubelet.
  Normal  NodeHasSufficientMemory  41m (x8 over 41m)  kubelet, docker-desktop     Node docker-desktop status is now: NodeHasSufficientMemory
  Normal  NodeHasNoDiskPressure    41m (x8 over 41m)  kubelet, docker-desktop     Node docker-desktop status is now: NodeHasNoDiskPressure
  Normal  NodeHasSufficientPID     41m (x7 over 41m)  kubelet, docker-desktop     Node docker-desktop status is now: NodeHasSufficientPID
  Normal  NodeAllocatableEnforced  41m                kubelet, docker-desktop     Updated Node Allocatable limit across pods
  Normal  Starting                 41m                kube-proxy, docker-desktop  Starting kube-proxy.
```
This might be a bug.

So my questions are:
1. What is exactly the "reservation" event you described [in the description of the jira](https://issues.apache.org/jira/browse/YUNIKORN-332)? Keep in mind that we already have a `PodBindSuccessful` event emitted by shim - put a bookmark with a `TODO` to the code to show where I believe such events should be emitted from the core.
2. Should such reservation event be exposed for pods of any kind (those that can be started and those that can not)? If yes, I suggest to rename these events because reservation in the YuniKorn scheduler seems to be a very different concept.
3. If a pod gets deleted, the app will not be unreserved due to lack of application lifecycle. Therefore the implementation of emitting `AllocationAskUnreservedFromNode` event is not working. The other problem is that this reservation is seemingly bound to the application and not the pod itself. I feel some discrepancy here: the pod could not reserved to a node anyways if it does not have the required resources.

I'm eager to hear all your thoughts on this @yangwwei @wilfred-s @kingamarton how to move forward.

P.s.: found a data race which is seemingly unrelated to this patch, filed [YUNIKORN-342](https://issues.apache.org/jira/browse/YUNIKORN-342).

yangwwei:
> What is exactly the "reservation" event you described in the description of the jira?

I have commented on the JIRA: https://issues.apache.org/jira/browse/YUNIKORN-332?focusedCommentId=17170587&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17170587. Please take a look.

> Keep in mind that we already have a PodBindSuccessful event emitted by shim - put a bookmark with a TODO to the code to show where I believe such events should be emitted from the core.

The `PodBindSuccessful` is from the shim side, the core side has no sense about when a pod gets bound to a node. So I don't think there is any problem here.

> Should such reservation event be exposed for pods of any kind (those that can be started and those that can not)?

Not sure I get this. Could you please elaborate?

> If a pod gets deleted, the app will not be unreserved due to lack of application lifecycle.

Could you please double-check. When a pod is deleted, we should have trigger the preservation logic for the corresponding ask, otherwise we will have some problem here.

> I feel some discrepancy here: the pod could not reserved to a node anyways if it does not have the required resources.

We do necessary checks before reserving a pod onto a node. Such as: https://github.com/apache/incubator-yunikorn-core/blob/357d71fa290619e583b5da0484215c5d927807b7/pkg/scheduler/scheduling_node.go#L318-L325. We also do check the predicates to make sure the pod can fit into the target node.

adamantal:
Created a new implementation which is based on the allocation result flag of `schedulingAllocation`. If the result is changed (like `none->reserved` transition) then an event is triggered.

I tested this on my local, and found the following:
- if we have lots of free space I see no reserved events (it has been moved to `allocated` directly - no reserve is taken place)
- if we have limited free resources in the queue I see reserved events for pending asks - I am not sure if this is expected or not
- from the node's perspective I can see the reserved events, but I see no unreserve events for pods that got deleted. I think the reason behind this is that the YK deletes the `schedulingAllocation` instead of updating its allocation result to `unreserve`.

@yangwwei @wilfred-s please let me know if you have any questions regarding this patch. I'd appreciate your time, if you can review it.

wilfred-s:
This is a lot of change for something that can be simply handled as emitting an event in the `partitionSchedulingContext` (psc).
All allocation results are processed in the `allocate()` inside the psc. We exactly know what is happening and process the resulting allocation to make sure the internals of the core are up to date. This handles all changes from reserved to unreserved, reserved to allocated on the same node or not etc. The logic is already there to do all of this.

The `schedulingAllocation` is a transient container object, we create a new one every time. It is just there to communicate the result we should not add logic to it. The object gets created and dropped later.

This change should be as simple as adding an emit for an event in the `allocate()` based on the updates we do there.

adamantal:
Created the new implementation based on your comment.

Tested locally on docker-desktop. The pod without sufficient resource got reserved event but I still don't see any unreserved event if I delete the pod.

Would appreciate the reviews.

yangwwei:
> Tested locally on docker-desktop. The pod without sufficient resource got reserved event but I still don't see any unreserved event if I delete the pod.

Good catch. The logic in this PR only handles the reserve/unreserve in the allocating phase. This should be enough for reservation, but for unreservation, there are some other places triggered this, such as removing a pod, removing a node etc. For example https://github.com/apache/incubator-yunikorn-core/blob/c52493985cce6fe598ac9c6e569c5cea9ef15b63/pkg/scheduler/scheduling_application.go#L150-L167

I believe @wilfred-s has a better thought of how to get these covered.

adamantal:
I jumped back to this issue, but I don't know how can we cover all the cases systematically. Any idea @wilfred-s to drop here?

adamantal:
Uploaded initial implementation based on the discussion with Wilfred.

Did so to check lint and UTs, but it's not done as I haven't written any unit tests yet. Will updated the PR with the tests soon.

adamantal:
Added unit tests to the patch.

Also slightly modified the patch: since the current implementation of the event cache allows only one event per object in every time period (2 sec by default), it is pointless to add more than one event per reservation or allocation.
Adjusted the code and the tests accordingly.

pkg/scheduler/scheduling_application.go
```diff
@@ -569,21 +596,33 @@ func (sa *SchedulingApplication) tryNodes(ask *schedulingAllocationAsk, nodeIter
...
 			if len(reservedAsks) > 0 {
 				nodeID := strings.TrimSuffix(reservedAsks[0], "|"+allocKey)
+				appID := sa.ApplicationInfo.ApplicationID
 				log.Logger().Debug("allocate picking reserved ask during non reserved allocate",
-					zap.String("appID", sa.ApplicationInfo.ApplicationID),
+					zap.String("appID", appID),
 					zap.String("nodeID", nodeID),
 					zap.String("allocationKey", allocKey))
+				EmitUnReserveEventForNode(ask.AskProto.AllocationKey, ask.ApplicationID, nodeID)
+				EmitAllocatedReservedEvents(ask.AskProto.AllocationKey, ask.ApplicationID, node.NodeID)
```
wilfred-s, line 625: You already have `appID` stored locally why not reuse it?

pkg/scheduler/scheduling_application.go
```diff
@@ -624,13 +663,43 @@ func (sa *SchedulingApplication) tryNodes(ask *schedulingAllocationAsk, nodeIter
...
 		return alloc
 	}
 	// ask does not fit, skip to next ask
 	return nil
 }
 
+func EmitAllocatedReservedEvents(allocKey, appID, nodeID string) {
+	err := events.EmitAllocatedReservedEvent(allocKey, appID, nodeID)
+	if err != nil {
+		log.Logger().Warn("could not emit allocatedReserved events to shim",
```
wilfred-s, line 684: be consistent when using logging level: seems to change from WARN to DEBUG for similar errors.

pkg/events/events_test.go
```diff
@@ -68,3 +71,119 @@ func TestEmptyFields(t *testing.T) {
...
+	cache := GetEventCache()
+	cache.StartService()
+
+	allocKey := "test-alloc-2"
+	appID := "app-2"
+	nodeID := "node-2"
+	err := EmitReserveEvent(allocKey, appID, nodeID)
+	assert.NilError(t, err, "expected EmitReserveEvent to run without errors")
+
+	assertEmitRequestAppAndNodeEvents(t, allocKey, appID, nodeID, "AppReservedNode")
```
wilfred-s, line 98: Test coverage for the error cases could be easily extended by passing in empty strings.
That would cover the codecov remarks

pkg/events/events.go
```diff
@@ -58,3 +58,73 @@ func CreateNodeEventRecord(objectID, reason, message string) (*si.EventRecord, e
...
+	if eventCache == nil {
+		// The event cache is disabled.
+		// It's not an error, let's return.
+		return nil
+	}
+	msg := fmt.Sprintf(msgFmt, allocKey, appID, nodeID)
+
+	event, err := CreateNodeEventRecord(nodeID, reason, msg)
+	if err != nil {
+		return err
```
wilfred-s, line 111: Generic question: if this event create fails does that also mean all next events should or must fail. Could we have a case that just the nodeID is an issue and should we send out the rest of the events in that case?
I think we need to be as smart as possible and sent as much as we can (i.e. fail only what really must fail)

wilfred-s (changes requested):

pkg/scheduler/scheduling_application.go
adamantal: You're right. Changed the WARNs to DEBUGs.

pkg/events/events_test.go
adamantal: Added tests for those in the `events` package, but I could not get triggered the errors in `scheduling_application.go`.

pkg/events/events.go
adamantal: You're right. Modified that part, also adjusted the unit tests.

adamantal:
@wilfred-s do you have any objections against merging this?

PR #191, opened 2020-07-29
[YUNIKORN-331] Reduce the verbosity of the logs when predicate fails

adamantal:


yangwwei:
hi @adamantal 

Thanks for working on this. Some more places might also need to be checked, such as:

https://github.com/apache/incubator-yunikorn-core/blob/d2a76c0d7000bb50e377f170538938236c1144c5/pkg/scheduler/scheduling_node.go#L216-L219

https://github.com/apache/incubator-yunikorn-core/blob/d2a76c0d7000bb50e377f170538938236c1144c5/pkg/scheduler/scheduling_node.go#L225-L228

adamantal:
Thanks for the review @yangwwei. Removed those log entries as well.

PR #192, opened 2020-07-29
[YUNIKORN-204] Post app state changes to shim side via UpdatedApplication

kingamarton:
@wilfred-s , @yangwwei , @TaoYang526 Can you please check this one when you have some time?


pkg/cache/application_info.go
```diff
@@ -52,9 +55,21 @@ type ApplicationInfo struct {
 	stateMachine      *fsm.FSM                   // application state machine
 	stateTimer        *time.Timer                // timer for state time
 
+	//TODO create a struct for it
+	statusChangeEventHandler commonevents.EventHandler
```
yangwwei, line 59: change this to

```
eventHandlers handler.EventHandlers
```

just to keep consistency with `cluster_info`, `scheduler` etc

pkg/cache/application_info.go
```diff
@@ -112,6 +127,23 @@ func (ai *ApplicationInfo) IsWaiting() bool {
...
+
+	if err ==nil && ai.statusChangeEventHandler != nil {
+	ai.statusChangeEventHandler.HandleEvent(
+		&rmevent.RMApplicationUpdateEvent{
+			RmID:                 ai.rmId,
+			AcceptedApplications: make([]*si.AcceptedApplication, 0),
+			RejectedApplications: make([]*si.RejectedApplication, 0),
+			UpdatedApplications: updatedApps,
+		})
+	}
```
yangwwei, line 146: do not handle this here, we can add this to the callback while state changes. https://github.com/apache/incubator-yunikorn-core/blob/a4a44d7e05d16aa712e1777c85f09e4d55466830/pkg/cache/application_state.go#L102-L108.

This is called whenever there is a state change
> // 6. enter_state - called after entering all states

Like the rest of callbacks in `application_state.go`, you can create a function in `application_info.go`, such as

```
func (ai *ApplicationInfo) onStateChange() {
  log.Logger().Debug("Application state transition",
    zap.String("appID", event.Args[0].(*ApplicationInfo).ApplicationID),
    zap.String("source", event.Src),
    zap.String("destination", event.Dst),
    zap.String("event", event.Event))

  // TODO : move the event handling here
} 
```

pkg/cache/application_info.go
kingamarton: done

pkg/cache/application_info.go
kingamarton: > Hi @kingamarton
> please see my comments inline. Also could you please add some UT for this as well?

@yangwwei do you have any good idea on how can we test in from the unit test (without writing integration test)? I was thinking to write some test for the onStatusChange method, but I don't have any good idea what can we test there. THis is something what we can easily test in out integration tests, but for unit tests if you have any good ideas, please let me know.

yangwwei: @kingamarton I think you can do something similar to https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/cache/application_state_test.go. You can create some app with `NewApplicationInfoWithEventHandler `, then a mocked implementation of the eventHandler, then verify the state info is actually handled.

yangwwei:
Looks good. I think we need to relax the `codecov/patch`, otherwise it is too easy to fail. 
I will merge this PR shortly. Thanks @kingamarton

PR #193, opened 2020-07-30
[YUNIKORN-311] Code coverage is not updating for the core builds

adamantal:


adamantal:
I checked the logs of the codecov report, and the hash matches to me.
Tried another approach: added `codecov.yaml` and set the `wait_for_ci` flag to false. This should theoretically force the codecov bot to comment on the PR. Let's wait.

wilfred-s:
Other builds are pushing codecov updates into the PR. I think we currently have this fixed.
Should we close this?

adamantal:
Yes, I think so. Closed.

PR #194, opened 2020-08-03
[YUNIKORN-273] expose config via rest

wilfred-s:
Expose the running config via a web call in both yaml (default) and json
(query parameter "json" required). The end point is "/ws/v1/config".

The yaml response is a multi line marshalled yaml object.
The json response is a one line marshalled json object.
The sha512 checksum is provided as part of the response. The checksum is
not part of the marshalled data. The checksum is printed as a 32 byte
hexadecimal sha256 string.

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -296,21 +296,15 @@ func getContainerHistory(w http.ResponseWriter, r *http.Request) {
 
 func getClusterConfig(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
-	jsonOutput := false
-	query := r.URL.Query()
-	// check if we have a request for json output
-	if len(query) != 0 {
-		if _, ok := query["json"]; ok {
-			jsonOutput = true
-		}
-	}
 
+	conf := configs.ConfigContext.Get(gClusterInfo.GetPolicyGroup())
 	var marshalledConf []byte
 	var err error
-	conf := configs.ConfigContext.Get(gClusterInfo.GetPolicyGroup())
-	if jsonOutput {
+	// check if we have a request for json output
+	if r.Header.Get("Accept") == "application/json" {
 		marshalledConf, err = json.Marshal(&conf)
 	} else {
+		w.Header().Set("Content-Type", "application/x-yaml; charset=UTF-8")
 		marshalledConf, err = yaml.Marshal(&conf)
 	}
 	if err != nil {
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -19,6 +19,7 @@ package webservice
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"runtime"
@@ -26,6 +27,7 @@ import (
 	"strings"
 
 	"go.uber.org/zap"
+	"gopkg.in/yaml.v2"
 
 	"github.com/apache/incubator-yunikorn-core/pkg/cache"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/configs"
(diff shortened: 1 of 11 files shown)


pkg/webservice/handlers.go
```diff
@@ -291,3 +293,33 @@ func getContainerHistory(w http.ResponseWriter, r *http.Request) {
 	}
 
 }
+
+func getClusterConfig(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	jsonOutput := false
```
yangwwei, line 299: Normally if an HTTP client expects a certain format of the response from the server, an HTTP header such as "Accept: application/json" needs to be set. Instead of adding a query parameter to do so. This is referring to doc like https://restfulapi.net/content-negotiation/.

pkg/webservice/handlers.go
```diff
@@ -291,3 +293,33 @@ func getContainerHistory(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+	var marshalledConf []byte
+	var err error
+	conf := configs.ConfigContext.Get(gClusterInfo.GetPolicyGroup())
+	if jsonOutput {
+		marshalledConf, err = json.Marshal(&conf)
+	} else {
+		marshalledConf, err = yaml.Marshal(&conf)
```
yangwwei, line 314: I am not sure if yaml is a valid response format for a RESTful API. 
I think json is good enough because the rest of our API all return json. Shall we keep consistency?

pkg/webservice/handlers.go
wilfred-s: changed, default is yaml

pkg/webservice/handlers.go
wilfred-s: REST does not specify, require or limit anything we return it can be any text form as well as binary.

We should provide yaml as that is the format the original file is in. We should not expect that the client **must** implement a json to yaml conversion tool before it can use the file. It might also lead to data issues.

wilfred-s:
The failed check passes locally: the `scheduler_smoke_test` file was not committed, adding the file with the review changes.

yangwwei:
hi @wilfred-s thanks for the update. could you please check: `TestConfigScheduler` is still failing.

wilfred-s:
> hi @wilfred-s thanks for the update. could you please check: `TestConfigScheduler` is still failing.

This is a different failure not caused by the checksum changes. The failure is caused by the event handling. The logs show that the new queue is created in the cache but event processing has not been triggered in the scheduler yet. This should be fixed as part of the event updates and cache rewrite. Test passes on re-run.

yangwwei:
hi @wilfred-s  could you pls rebase it? Once all CI jobs pass, we can get it merged. THX

pkg/webservice/handlers.go
sunilgovind: I think its good to see the default the response as yaml.
As I see, its not good to compare headers like r.Header.Get. If possible, could we some how have a cleaner interfaces to find/get some header if present. Some time, comparison can have issues in unicode, caps etc etc. It's better we have a util package/class to handle some checks in a standard and use some bool/enum to handle the further processing.

pkg/webservice/routes.go
```diff
@@ -24,74 +24,82 @@ import (
...
+	route{
+		"Scheduler",
+		"GET",
+		"/ws/v1/config",
+		getClusterConfig,
+	},
+
 	// endpoint to validate conf
-	Route{
+	route{
```
sunilgovind, line 88: I think this is not a correct way to handle a POST for validate. Recheck with REST mode.

Also, its better we clearly move all PUT, POST, DELETE etc to another cleaner interface or under some group so that we can protect all WRITE ops in a more secure way later from simple GET calls.

pkg/webservice/handlers.go
wilfred-s: The header retrieval is hidden inside the go packages. It already handles the unicode and caps etc. Predefined MIME headers should however only be ascii, specially for the ones we use. This is from the comments in the go source code:
```
// Get gets the first value associated with the given key. If
// there are no values associated with the key, Get returns "".
// It is case insensitive; textproto.CanonicalMIMEHeaderKey is
// used to canonicalize the provided key. To access multiple
// values of a key, or to use non-canonical keys, access the
// map directly.
```
So we're covered unless we come up with some special non standard headers.

pkg/webservice/routes.go
wilfred-s: The validate-conf URL will be updated as part of YUNIKORN-366 and its sub tasks

PR #195, opened 2020-08-04
[YUNIKORN-343] Provide a placement rule that can support attaching NSs to parent queue

adamantal:
I uploaded this preliminary implementation - not yet ready for review.
I will also add a documentation patch parallel to this PR which can be merged as well.

Will add UT to cover the new behaviour.


adamantal:
Added documentation in https://github.com/apache/incubator-yunikorn-site/pull/13

wilfred-s:
Closing as the approach is now to use the existing tag rules per YUNIKORN-368

PR #197, opened 2020-08-12
[YUNIKORN-369] Set default go version to 1.15.

yangwwei:


yangwwei:
@wilfred-s pls help to review

.travis.yml
```diff
@@ -21,7 +21,7 @@ os:
   - linux
 
 go:
-  - "1.12"
+  - "1.15"
```
sunilgovind, line 24: Most of dev env is still in 1.12.
Any particular reason to move to 15?

yangwwei: Hi @sunilgovind  According to https://endoflife.date/go, v1.12 was released 1.5 years ago and it is marked as EOL.

wilfred-s (changes requested):
Based on the go language specs the mod change unlocks go 1.15 language features. It has nothing to do with the version we compile or support. This is the definition: https://golang.org/doc/go1.12#modules
`The go directive in a go.mod file now indicates the version of the language used by the files within that module.`
Unless you really use the features you can still compile the project with an older version even if this is set to a later version. I am thus not sure we should or must update the directive unless we want to unlock new language features and are going to use them. The version in the file is not the minimum version of the tool chain needed to compile the code, it just causes extra messages to be printed if the compilation fails.
case 1:
code uses 1.15 features -> go mod shows 1.12 
 compile with 1.12 -> syntax error
 compile with 1.15 -> all OK
case 2:
code uses 1.15 features -> go mod shows 1.15
 compile with 1.12 -> syntax error, but says code requires version 1.15
 compile with 1.15 -> all OK
case 3:
code uses 1.12 features -> go mod shows 1.15
 compile with 1.12 -> all OK
 compile with 1.15 -> all OK

yangwwei:
hi @wilfred-s 

Since go 1.12 is EOL, I think we should update the version to 1.15.
We do not rely on any of 1.15 features directly today, but that doesn't mean we cannot (or not need) to include some new features moving forward. So today, we are most likely in your `case 3`.
We need to be careful to define what version of go can be supported, today we are saying 1.11+. But if someday we added some necessary changes that only supported by later go versions, we will need to visit our dependency check and doc, we can make decision later if we need to include that change or not.

wilfred-s:
> Since go 1.12 is EOL, I think we should update the version to 1.15.

The `go.mod` version has nothing to do with being EOL or not. It is purely about what we allow at a language level and the way the compile fails.

> We do not rely on any of 1.15 features directly today, but that doesn't mean we cannot (or not need) to include some new features moving forward. So today, we are most likely in your `case 3`.

Yes but it does not enforce anything. That is why I have a change for the Makefile that will enforce the `go.mod` version as a minimum go version for the build. I think that is far more important and we can change the mod file when we really leverage the new language functionality.

> We need to be careful to define what version of go can be supported, today we are saying 1.11+. But if someday we added some necessary changes that only supported by later go versions, we will need to visit our dependency check and doc, we can make decision later if we need to include that change or not.

I have a change ready to go that will make sure the `go.mod` version is the same as we check in the Makefile. It will also make sure that the installed version that is used is at _least_ the version we define in `go.mod`. If it is not lining up we fail the build. (YUNIKORN-387)

wilfred-s (approved):
approved, without the go.mod update
merging

PR #198, opened 2020-08-18
[YUNIKORN-366] Add rest API to update queue configuration

kingamarton:
This is still a wip PR. I have to add the following things:

- [x] unit tests
- [x] revert the configmap change if saving the config fails
- [x] return if there is already one update in progress
- [x] find out how to prevent dupe refresh by configwatcher
- [x] update interface dependency


pkg/webservice/handlers.go
```diff
@@ -339,3 +342,55 @@ func getClusterConfig(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
 }
+
+var mutex sync.Mutex
```
yangwwei, line 346: can we use the RWMutex defined here: https://github.com/apache/incubator-yunikorn-core/blob/742e21611d15da8e3c8d5441f8d3b42d8532634c/pkg/webservice/webservice.go#L41. But looks like we need to move it out of the constructor.

pkg/webservice/handlers.go
```diff
@@ -339,3 +342,55 @@ func getClusterConfig(w http.ResponseWriter, r *http.Request) {
...
+func updateConfig(w http.ResponseWriter, r *http.Request) {
+	mutex.Lock()
+	defer mutex.Unlock()
+	writeHeaders(w)
+	requestBytes, err := ioutil.ReadAll(r.Body)
+
+	if err == nil {
```
yangwwei, line 353: when `err != nil` we should log an error in the server-side and response the client with an error too.,

pkg/webservice/handlers.go
```diff
@@ -339,3 +342,55 @@ func getClusterConfig(w http.ResponseWriter, r *http.Request) {
...
+		// validation is already called when loading the config
+		schedulerConf, err2 := configs.LoadSchedulerConfigFromByteArray(requestBytes)
```
yangwwei, line 355: Can we re-use err instead of having multiple ones err2, err3, and err4?

pkg/webservice/handlers.go
```diff
@@ -339,3 +342,55 @@ func getClusterConfig(w http.ResponseWriter, r *http.Request) {
...
+
+func buildUpdateResponse(success bool, reason string, w http.ResponseWriter) {
+	var result dao.UpdateConfResponse
+	result.Success = success
+	result.Reason = reason
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
+}
+func saveConfigmap(conf string) (string, error) {
```
yangwwei, line 384: Rename `saveConfigmap` to `updateConfiguration`

pkg/cache/cluster_info.go
```diff
@@ -711,3 +720,29 @@ func (m *ClusterInfo) processRemovedApplication(event *cacheevent.RemovedApplica
...
+	}
+	if len(deletedPartitions) > 0 {
+		result2 := m.sendDeletedPartitionsToScheduler(deletedPartitions)
+		if !result2.Succeeded {
+			return fmt.Errorf(result2.Reason)
+		}
+	}
+	return nil
+}
+
```
yangwwei, line 748: It is still possible that failure might happen while trying to delete some partitions.
The `updatedPartition` is already processed, but `deletedPartition` is not, the config becomes inconsistent.
I am not sure if we can have a more robust way to handle this based on the current structure, maybe we can leave this as the future work. As the auto-refresh code is exactly the same.

kingamarton:
> hi @kingamarton
> 
> Thanks for the PR. Except for the changes in this PR, could you please check the logic here:
> 
> https://github.com/apache/incubator-yunikorn-core/blob/742e21611d15da8e3c8d5441f8d3b42d8532634c/pkg/common/configs/configwatcher.go#L93-L97
> 
> , see if this is enough to prevent a dup refresh?

@yangwwei, I already checked that part and I am afraid that a dup refresh is not possible(or it not straightforward how to do it) to prevent in every case because the config reloader is triggered at the moment when the configmap is changed. However I will have to find out a solution for this issue, because I debugged the actual workflow and I can see that after the configuration is refreshed in the scheduler, the configwatcher will revert the changes, because it is triggered, but since there is a delay in observing the changes in the pod, it detects the difference between the pod(still having the old config) and the scheduler (having already the new config), and will change the configuration in the scheduler.

pkg/plugins/types.go
```diff
@@ -62,3 +63,7 @@ type ContainerSchedulingStateUpdater interface {
 	// the shim side implementation must be thread safe
 	Update(request *si.UpdateContainerSchedulingStateRequest)
 }
+
+type ConfigMapPlugin interface {
+	UpdateConfigMap(args *si.ConfigMapArgs) (string, error)
```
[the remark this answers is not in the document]
kingamarton: Renamed

pkg/webservice/handlers.go
kingamarton: Done

pkg/webservice/handlers.go
kingamarton: I added a log message when creating the response. Also handled this case as well.

pkg/webservice/handlers.go
kingamarton: Done

pkg/webservice/handlers.go
kingamarton: Done

pkg/cache/cluster_info.go
kingamarton: Yes, this can be an issue. I suggest to think about it in a follow up task, because this is how it works the auto-refresh as well.

yangwwei:
> @yangwwei, I already checked that part and I am afraid that a dup refresh is not possible(or it not straightforward how to do it) to prevent in every case because the config reloader is triggered at the moment when the configmap is changed. However I will have to find out a solution for this issue, because I debugged the actual workflow and I can see that after the configuration is refreshed in the scheduler, the configwatcher will revert the changes, because it is triggered, but since there is a delay in observing the changes in the pod, it detects the difference between the pod(still having the old config) and the scheduler (having already the new config), and will change the configuration in the scheduler.

Got it, this is exactly what I was concerned about.
I think there is an easier solution to fix this. On the shim side, we can add a CLI option, e.g `enableConfigHotRefresh`. If this value is `false`, we can bypass https://github.com/apache/incubator-yunikorn-k8shim/blob/ac345b38432b416aede47e3d51d5e120011c56fd/pkg/cache/context.go#L283-L293. This code triggers the config refresh.

kingamarton:
> > @yangwwei, I already checked that part and I am afraid that a dup refresh is not possible(or it not straightforward how to do it) to prevent in every case because the config reloader is triggered at the moment when the configmap is changed. However I will have to find out a solution for this issue, because I debugged the actual workflow and I can see that after the configuration is refreshed in the scheduler, the configwatcher will revert the changes, because it is triggered, but since there is a delay in observing the changes in the pod, it detects the difference between the pod(still having the old config) and the scheduler (having already the new config), and will change the configuration in the scheduler.
> 
> Got it, this is exactly what I was concerned about.
> I think there is an easier solution to fix this. On the shim side, we can add a CLI option, e.g `enableConfigHotRefresh`. If this value is `false`, we can bypass https://github.com/apache/incubator-yunikorn-k8shim/blob/ac345b38432b416aede47e3d51d5e120011c56fd/pkg/cache/context.go#L283-L293. This code triggers the config refresh.
> 
> We should only allow ONE updateConf mechanism at runtime, either
> 
> 1. Via configmap hot refresh
> 2. Via UpdateConf REST API
> 
> not both. When the hot refresh is enabled. We should deny the UpdateConf calls from the REST API.

@yangwwei, by limiting the mechanisms might solve some issues, but for me is strange to limit this, once both of them are available. What would be the default one? I can imagine use cases when editing directly the configmap is easier, and some other cases when the REST API can be used. How will the user change what he/she want's to use?

sunilgovind:
I think this makes sense to have either one of the option to be enabled in YK.
1. Via configmap hot refresh
2. Via UpdateConf REST API
At the time when we are expecting to have the config update from rest api, its always gonna be a mess and complex to watch for configmap (some one could edit via kube apis). And as a system, we will be patching holes all around.

I am not seeing a major reason to support both together. So +1 to this approach.

yangwwei:
Agree with @sunilgovind . We should be more restrictive about this. I don't think we need to support both options in a deployment.

pkg/cache/cluster_info.go
```diff
@@ -47,8 +48,8 @@ type ClusterInfo struct {
 	pendingSchedulerEvents chan interface{}
 
 	// RM Event Handler
-	EventHandlers handler.EventHandlers
-
+	EventHandlers    handler.EventHandlers
+	skipConfigUpdate bool
```
yangwwei, line 52: based on the discussion, I don't think we need this flag anymore.
Let's keep the core side change as simple as possible

pkg/cache/cluster_info.go
```diff
@@ -483,12 +484,33 @@ func (m *ClusterInfo) processRMRegistrationEvent(event *commonevents.RegisterRME
 // Updated and deleted partitions can not fail on the scheduler side.
 // Locking occurs by the methods that are called, this must be lock free.
 func (m *ClusterInfo) processRMConfigUpdateEvent(event *commonevents.ConfigUpdateRMEvent) {
+	if m.skipConfigUpdate {
```
yangwwei, line 487: remove this flag

pkg/webservice/handlers.go
```diff
@@ -339,3 +341,63 @@ func getClusterConfig(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			log.Logger().Error("Configuration rollback failed")
+		}
+		buildUpdateResponse(false, err.Error(), w)
+		return
+	}
+	buildUpdateResponse(true, "", w)
+}
+
+func buildUpdateResponse(success bool, reason string, w http.ResponseWriter) {
```
yangwwei, line 378: for the REST API response. If update was successful, we should do

```
json.NewEncoder(w).Encode(result)
```

however, if the update was not successful, I think we should return an error code to the client

```
http.Error(w, err.Error(), http.StatusConflict))
```

where the StatusConflict means:  The request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request.

pkg/cache/cluster_info.go
kingamarton: removed it

pkg/webservice/handlers.go
kingamarton: Changed it

pkg/cache/cluster_info.go
kingamarton: You are right, I removed it

pkg/webservice/handlers.go
kingamarton: @yangwwei , If we return StatusConflict in case of failure, I think there is no need to have an UpdateConfResponse, because we can set the status code to 200-OK and return a message that the config update was successful, so I removed the UpdateConfResponse and set 200-OK in case of success.

pkg/webservice/handlers.go
```diff
@@ -339,3 +341,67 @@ func getClusterConfig(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	err = gClusterInfo.UpdateSchedulerConfig(schedulerConf)
+	if err != nil {
+		// revert configmap changes
+		_, err := updateConfiguration(oldConf)
+		if err != nil {
+			log.Logger().Error("Configuration rollback failed")
+		}
+		buildUpdateResponse(false, err.Error(), w)
```
yangwwei, line 372: these lines seem to be problematic.
line 368 has overwritten the value of `err`, and line 372 is out of the `err != nil` check.
that means if `err == nil`, line 372 will give an NPE error

yangwwei (changes requested):

pkg/webservice/handlers.go
kingamarton: Good catch! I was just focusing to fix the lint issue :). Fixed it

yangwwei:
All look good to me now.  +1

PR #199, opened 2020-08-26
[YUNIKORN-374] Use same GRPC version in the core and the scheduler-interface

yangwwei:


wilfred-s:
This change is looking at the GPRC version why are we making changes to the prometheus version also? The prometheus changes are handled under YUNIKORN-375. Can we split these two as per the jiras?

yangwwei:
hi @wilfred-s sure. Let me split the changes. Now the change only include the gRPC version update. Pls take another look.

wilfred-s (changes requested):
I think we need to also update the scheduler interface version in the mod file, we should first commit  apache/incubator-yunikorn-scheduler-interface#25
Beside that it looks good.

wilfred-s (approved):
lgtm, approved

PR #200, opened 2020-08-27
[YUNIKORN-375] Use compatible Prometheus client library

yangwwei:
Better to use a more stable Prometheus version, this PR is based on

```
go get github.com/prometheus/client_golang@v0.9.2 
```


go.mod
```diff
@@ -23,15 +23,17 @@ go 1.12
...
 	github.com/apache/incubator-yunikorn-scheduler-interface v0.9.1-0.20200827013520-ec2f681ecb5b
+	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/golang/protobuf v1.4.0 // indirect
 	github.com/gorilla/mux v1.7.3
 	github.com/kr/text v0.2.0 // indirect
 	github.com/looplab/fsm v0.1.0
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
 	github.com/pkg/errors v0.9.1 // indirect
-	github.com/prometheus/client_golang v1.4.0
+	github.com/prometheus/client_golang v0.9.2
```
sunilgovind, line 33: @yangwwei why are we bumping down the version from a stable 1.x to a 0.x version? if there are compatibility issues, it may get missed behind the scenes. 
How could we verify that we dont have a breakage?

wilfred-s: That is exactly what I was looking at. 0.9.2 is not the latest on the 0.9.x branch either, 0.9.4 is.
We are downgrading to an almost 2 year old release which has had two updates on top.
I can understand that we might need to stay at a pre 1.0 release due to breaking changes and k8s or etcd lagging behind but we should be at least at the latest 0.9.x release.

yangwwei: Hi @sunilgovind @sunilgovind the reason to use 0.9.2 is that this is the version used by k8s 1.16. Just want to keep the consistency because we will need to push metrics to the compatible Prometheus that works with 1.16. We are not using new APIs so I guess we are fine with downgrade. What do you think?

wilfred-s (approved):
lgtm, approved 
0.9.4 is the latest we can do for now
The fact that the core and shim are build as one executable pushes the dependencies that are used to build the shim into the core. We need to work on a way to de-couple this as we do not wantto be held to k8s dependencies for the core and other pieces we have.

PR #201, opened 2020-09-01
[YUNIKORN-390] Panic in placement rule evaluation

wilfred-s:
Placement rules can specify an arbitrary parent rule and a create flag for
the parent rule. If the parent rule is configured with the create flag set
to "true" and returns a non-existing queue the scheduler panics.

The fact that the parent queue does not exist means it cannot be a leaf
queue and the isLeafQueue() check should be skipped.
Adding tests to cover the parent rule returns to all rules.

### DIFF
--- a/pkg/scheduler/placement/tag_rule_test.go
+++ b/pkg/scheduler/placement/tag_rule_test.go
@@ -144,3 +144,110 @@ partitions:
 		t.Errorf("tag rule with parent queue incorrect queue '%s', error %v", queue, err)
 	}
 }
+
+func TestTagRuleParent(t *testing.T) {
+	// Create the structure for the test
+	data := `
+partitions:
+  - name: default
+    queues:
+      - name: testchild
+      - name: testparent
+        parent: true
+`
+	partInfo, err := CreatePartitionInfo([]byte(data))
+	assert.NilError(t, err, "Partition create failed with error")
+	tags := make(map[string]string)
+	user := security.UserGroup{
+		User:   "testuser",
+		Groups: []string{},
+	}
+
+	// trying to place in a child using a parent, fail to create child
+	conf := configs.PlacementRule{
+		Name:   "tag",
+		Value:  "label1",
+		Create: false,
+		Parent: &configs.PlacementRule{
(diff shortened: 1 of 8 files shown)


pkg/scheduler/placement/tag_rule_test.go
```diff
@@ -144,3 +144,112 @@ partitions:
 		t.Errorf("tag rule with parent queue incorrect queue '%s', error %v", queue, err)
 	}
 }
+
+// Create the structure for the parent rule tests
+// shared by a number of rule tests
+const confParentChild = `
```
adamantal, line 150: I don't see why it has been put to the tag rule. IDEs provide quick jump to the definition, but it's a bit counterintuitive. I think this can be moved to `rule_test.go` or `test_rule.go`, or even a helper go file can be created for this.

adamantal (changes requested):
Thanks for the patch @wilfred-s.
It's nice that you've made test cases to support the fix.

One minor nit: 
- I would also add a test case with creatable parent but with filter defined which prevent the user from submitting the app to the queue.

pkg/scheduler/placement/tag_rule_test.go
wilfred-s: We cannot move it to `test_rule.go` as that is a part of the production code (i.e. does not end in `_test.go`) and it really does not belong there.
The `rule_test.go` is testing the interface for the rule object, also not really the place to set it as it is not even referenced there. Probably placing it in the top of one of the test files would have been better, just forgot to do that. Adding a new file for just two constants seems a bit much.
There is no real solution that solves it nicely without the need to jump. It might be better for readability to add it to the `rule_test.go` file. It still requires the jumping to find the value, We can do that a a bit of an in between solution, are you OK with that then?

adamantal: Yes, let's keep it here then.

wilfred-s:
If there are no more questions outstanding can someone please approve the change so we can merge this?

yangwwei (approved):
LGTM

PR #202, opened 2020-09-02
[YUNIKORN-395] Clean up the old docs in core/shim/web/scheduler-interface repos

adamantal:
See my comment on the jira. We still need to figure out what to do with the remaining ones.


yangwwei:
Hi @adamantal  Second thought, I don't think we need to have the build steps all documented here. Since it is so simple, we can cover it in the dev guide. We will suggest the user read the dev guide anyway before making changes in this repo. I just made a minor change to do so.

adamantal:
Thanks for the modification @yangwwei . I agree with that. The Travis failure is unrelated to this doc modification, so let me retrigger with a sync from master.

yangwwei (approved):
LGTM. Let's track the remaining things in separate JIRAs.
Thanks for getting this done @adamantal !

PR #203, opened 2020-09-07
[YUNIKORN-397] Makefile use go version from go.mod

wilfred-s:
Instead of hard coding a go version in the Makefile use the version
from the go.mod file. The build will fail if the installed tools version
is lower than the version from the go.mod file.


wilfred-s:
Pushing changes to build in all 3 repos

PR #204, opened 2020-09-07
[YUNIKORN-324]Add rest API to retrieve cluster resource utilization

HuangTing-Yao:
just return the utilization of memory and vcore.
@yangwwei pls review when you have some time. Thanks.

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -86,11 +86,16 @@ func getClusterInfo(w http.ResponseWriter, r *http.Request) {
 func getClusterUtilization(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	var clusterUtil []*dao.ClustersUtilDAOInfo
+	var utilizations []*dao.ClusterUtilDAOInfo
 	lists := gClusterInfo.ListPartitions()
 	for _, k := range lists {
 		partition := gClusterInfo.GetPartition(k)
-		utilizations := getClusterUtilJSON(partition)
-
+		for name := range partition.GetTotalPartitionResource().Resources {
+			if name == "memory" || name == "vcore" {
+				util := getClusterUtilJSON(partition, name)
+				utilizations = append(utilizations, util)
+			}
+		}
 		clusterUtil = append(clusterUtil, &dao.ClustersUtilDAOInfo{
 			PartitionName: partition.Name,
 			ClustersUtil:  utilizations,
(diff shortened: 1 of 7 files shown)


yangwwei:
Commented a while back, but I forgot to submit ...

kingamarton:
@HuangTing-Yao  there is an open Jira for refactoring the REST API: https://issues.apache.org/jira/browse/YUNIKORN-413
I am wondering if we can solve this issue according to the new design: https://docs.google.com/document/d/1KgoGqmNGR7TK3yBeqmefiFso2_lQujA_TzlsU0SsG1o/edit#heading=h.9aay2d6eltya
Can you please check if it is possible?
Also can you please add some more unit tests to have the patch coverage check passing?

HuangTing-Yao:
@kingamarton There aren't progress update on yunikorn-413. Perhaps we can done this one first?

kingamarton:
> @kingamarton There aren't progress update on yunikorn-413. Perhaps we can done this one first?

@TaoYang526 I think yes. At least when doing the refactoring this part will already be OK. Let me ping Mani on the Jira to see his opinion.

kingamarton:
> > @kingamarton There aren't progress update on yunikorn-413. Perhaps we can done this one first?
> 
> @TaoYang526 I think yes. At least when doing the refactoring this part will already be OK. Let me ping Mani on the Jira to see his opinion.

@HuangTing-Yao since there is no progress with the refactoring, I think it will OK to follow the old design.

pkg/webservice/handlers.go
```diff
@@ -188,6 +211,31 @@ func getClusterJSON(name string) *dao.ClusterDAOInfo {
...
+		total = 0
+	}
+	used, ok = partition.Root.GetAllocatedResource().Resources[name]
+	if !ok {
+		used = 0
+	}
+	if total < used || total == 0 {
+		percent = int64(0)
+	} else {
+		percent = ((int64(used) * int64(100)) / int64(total))
```
kingamarton, line 228: This needs some protection against integer overflow. Both positive and negative. For example if If used = maxInt64, available is 10 then we will have an overflow here. You can find an example in the resources.go fileÉ https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/common/resources/resources.go#L264-L279

kingamarton: Also I think we need to cover the edge cases as well with unit test: nil values, zero values, inf values

kingamarton (changes requested):

pkg/webservice/handlers.go
kingamarton: In resources.go file there is already a method for calculating absolute usage: func CalculateAbsUsedCapacity(capacity, used *Resource) *Resource please use that one if possible.

kingamarton (changes requested):

pkg/webservice/handlers.go
HuangTing-Yao: done

kingamarton (approved):
Thank you @HuangTing-Yao for addressing the comments. LGTM

PR #206, opened 2020-09-11
[YUNIKORN-404] Config update breaks when there is upper case queue names

yangwwei:
This is actually due to the queue was named as upper case "X", internally we are converting all upper case letters in the queue to lower cases, see code here. When the scheduler tries to update the queues, it got confused because it thinks it visited "root.x", but not "root.X". So it marks the queue "root.X" for removal.


wilfred-s (approved):
Change looks good

PR #207, opened 2020-09-17
[YUNIKORN-405] Added checksum validation

kingamarton:


yangwwei (commented):
I just added a minor comment. Overall the PR looks good, it has good UT coverage, which is good.
The concern I have is not for the changes itself, but the workflow when a client starts to consume the API.
If the checksum failed, updateConfig fails, the client-side could be very confusing... what do we expect them to do? Is it supposed to ask them to call a GET to reload the latest config from YK, and overwrite their existing configs in whatever storage? What if users already made some changes in their local, how they can restore that, and apply to the latest content? This is very complex.

The point to have a source-of-truth discussion is to make this model simpler. If we allow user to update configs via configUpdate API, that means the client side will have storage used as the source-of-truth for configs, not something in YK's memory. Think YARN as an example, the storage will be either a file, or ZK, or DB. This way, client just need to maintain the configs in their end, overwrite the YK side changes when needed.

This is the reason why I did not want to add this change at the begnning.

pkg/common/configs/config.go
```diff
@@ -142,6 +142,20 @@ type LoadSchedulerConfigFunc func(policyGroup string) (*SchedulerConfig, error)
...
+	conf, err := ParseAndValidateConfig(content)
+	if err != nil {
+		return nil, err
+	}
+	// Create a sha256 checksum for this validated config
+	PopulateChecksum(content, conf)
+	return conf, err
+}
+
+func PopulateChecksum(content []byte, conf *SchedulerConfig) {
```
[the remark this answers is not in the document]
kingamarton: Since we call this function from handler.go as well, we need to keep it public.

kingamarton:
> I just added a minor comment. Overall the PR looks good, it has good UT coverage, which is good.
> The concern I have is not for the changes itself, but the workflow when a client starts to consume the API.
> If the checksum failed, updateConfig fails, the client-side could be very confusing... what do we expect them to do? Is it supposed to ask them to call a GET to reload the latest config from YK, and overwrite their existing configs in whatever storage? What if users already made some changes in their local, how they can restore that, and apply to the latest content? This is very complex.
@yangwwei  yes, in case of stale config the client will bee notified that they should update their configuration before applying their changes. If we will not add the checksum validation, we might be in the case that the user is performing some unwanted changes, like deleting some changes made by an another user, without even noticing it. @wilfred-s  any thoughts on this?

yangwwei:
> If we will not add the checksum validation, we might be in the case that the user is performing some unwanted changes, like deleting some changes made by an another user, without even noticing it

Correct. But I think such logic should be built outside of YuniKorn. The client-side stores the configuration and use that as the source of truth, the client-side can avoid concurrent updates, stale configs, etc, which seems to be an easier way for integration.

wilfred-s:
> If the checksum failed, updateConfig fails, the client-side could be very confusing... what do we expect them to do? Is it supposed to ask them to call a GET to reload the latest config from YK, and overwrite their existing configs in whatever storage? What if users already made some changes in their local, how they can restore that, and apply to the latest content? This is very complex.

Yes, the clients needs to do the GET and update its configuration. It is not complex. It is for the client to handle that issue.
If they really do not care they can do a GET, replace the checksum and push the change again. However, it does put the onus on the client to do the right thing.
The simple thing that the checksum prevents is clients from updating the configuration without thinking about whom else might have updated the running configuration. It could be that the client is out of sync, been down for a while or the user has left an old update sitting in an uncommitted state, changes could have been made via a different client too.
I would see a client workflow as follows:
1. GET config 
2. make change (keep delta) 
3. PUT changed config
4. PUT:  OK update checksum, persist config
  PUT: fail, show message and either go directly back to 1 and let user redo it from scratch or
5. GET config, apply delta 
6. Show changed config (delta applied) go back to 2.

>The point to have a source-of-truth discussion is to make this model simpler. If we allow user to update configs via configUpdate API, that means the client side will have storage used as the source-of-truth for configs, not something in YK's memory. Think YARN as an example, the storage will be either a file, or ZK, or DB. This way, client just need to maintain the configs in their end, overwrite the YK side changes when needed.

The YARN API is a single change at a time: add, remove or update a single queue or update a set of values in the global config. YARN is the source of truth and *implicitly* expects that the client knows the current state. In YARN you pass a delta to be applied on top of the current config, it does not accept a complete new configuration. It still can fail every single change. A completely different point to start from.

For us we *cannot* set the source of truth outside of YuniKorn. We do not control the client, which would mean we give up control completely. YuniKorn has to be the source of truth, with the checksum check we at least push the client to think about what they are doing.
YuniKorn runs with a configuration. It has queues and applications running in that queue etc. A configuration change that would just blindly overwrite what is running within the system is dangerous. YuniKorn needs to be able to push back and in some way say _Are you sure that you want to do this?_ If it is then accept it and make the change, with all the side effects it has.

yangwwei:
hi @wilfred-s  the workflow you suggested

> GET config
make change (keep delta)
PUT changed config
PUT: OK update checksum, persist config
PUT: fail, show message and either go directly back to 1 and let user redo it from scratch or
GET config, apply delta
Show changed config (delta applied) go back to 2.

This is exactly why I thought it is too complex... imagine we are building a service to manage the queue structure where user can modify, track versioning, rollback to a previous version, etc.  This service has to maintain the configs, e.g in its DB. Which one will be the source of truth, the configs in DB, or the configs in YK? Looks like you are suggesting using the YK side configs as the source of truth. Then how can the service deal with its own data, or it will be a stateless service? How can they implement version control, and rollback features, etc?

All things we want to do is to avoid a harmful update (I agree this is important). But I think such logic could be complex and it should be built outside of YK.

yangwwei:
hi @kingamarton I had a long discussion with @wilfred-s today about this. The decision we come up with is to get this one merged. And we need to add the doc in http://yunikorn.apache.org/docs/next/api/scheduler to doc the API, and also the notes for the client who wants to use the API. Could you help to do that?

pkg/common/configs/config.go
```diff
@@ -35,7 +35,7 @@ import (
 // set of scheduler resources.
 type SchedulerConfig struct {
 	Partitions []PartitionConfig
-	Checksum   [32]byte `yaml:"-" json:"-"`
+	Checksum   [32]byte `yaml:",omitempty" json:",omitempty"`
```
wilfred-s, line 38: If we marshal the checksum in the config there should be no need to handle them separately in the code.
This seems to be a double up with handling in the code

pkg/webservice/handlers.go
```diff
@@ -378,6 +385,20 @@ func updateConfig(w http.ResponseWriter, r *http.Request) {
 	buildUpdateResponse(true, "", w)
 }
 
+func getConfigurationString(requestBytes []byte, checksum [32]byte) string {
+	newConf := string(requestBytes)
+	checksumString := fmt.Sprintf("%v", checksum)
+	checksumString = strings.ReplaceAll(checksumString, " ", ",")
+	checksumString = "checksum: " + checksumString
```
wilfred-s, line 392: In the `getClusterConfig()` we use the following string to write the checksum:
`sha256 := fmt.Sprintf("\nsha256 checksum: %X", conf.Checksum)`
The `%X` converts the checksum into a hex string which is the same for both json and yaml. 

I cannot assess if this works for both or how this relates to the current layout we have in the get call.

wilfred-s (commented):
Two points around the checksum need clarification

pkg/common/configs/config.go
kingamarton: I handled them separately  because in case of config update the input will have the new configuration with the base configuration checksum.

pkg/webservice/handlers.go
kingamarton: That will not work for this case, because in this case we need to have an array with the checksum bytes delimited by ",". In getClusterConfig() it will result in a human readable checksum string

wilfred-s:
Now that the yaml and json marshaller write the checksum we generate the checksum over a config that already has a checksum set. We need to make sure that we "clear out" the checksum from the object before we generate the checksum.
This also needs to be clearly documented: the checksum is only calculated for a scheduler object with an unset or empty  checksum

kingamarton:
I solved the conflicts and excluded the checksum from the checksum calculation. I will need to do some manual testing when the shim will use the changes from YK-317. (There will be some changes what needs to be done, because there are some compilation failures if I just change the core version)

pkg/common/configs/config.go
```diff
@@ -35,7 +36,7 @@ import (
 // set of scheduler resources.
 type SchedulerConfig struct {
 	Partitions []PartitionConfig
-	Checksum   [32]byte `yaml:"-" json:"-"`
+	Checksum   string `yaml:",omitempty" json:",omitempty"`
```
wilfred-s, line 39: I am wondering if we should split this in two. Checksum is always the current config checksum and "UpdateChecksum" is the checksum of the config we are updating. The reuse of the same field for two purposes is a bit strange and looks really weird in the code. What do you think?

pkg/common/configs/config.go
```diff
@@ -191,6 +204,13 @@ func resolveConfigurationFileFunc(policyGroup string) string {
 	}
 	return filePath
 }
+func GetConfigurationString(requestBytes []byte) string {
+	conf := string(requestBytes)
+	if strings.Contains(conf, "checksum") {
+		conf = strings.Split(conf, "checksum: ")[0]
```
wilfred-s, line 210: Can we rely on the fact that the checksum will always be the last part of the marshalled request?
The go yaml or json code might do that but there is no guarantee that it will happen for all requests that come in. We need to be really careful here.

wilfred-s (changes requested):
codecov complains about new lines in the diff not being hit by the test code, we should really hit all

pkg/common/configs/config.go
kingamarton: I agree that it is a bit strange, but if we would introduce a new `UpdateChecksum` that will be weird as well, because we will have to hack around the marshalling part. Also I can image that it may confuse the users as well, because while creating the input for the request it seems natural to extend it with `Checksum: .....`, how the getConfig is returning it, instead of changing it to `UpdateChecksum: ....`

pkg/webservice/handlers.go
```diff
@@ -495,6 +496,11 @@ func updateConfig(w http.ResponseWriter, r *http.Request) {
 	buildUpdateResponse(nil, w)
 }
 
+func isChecksumValid(checksum string) bool {
+	actualConf := configs.ConfigContext.Get(schedulerContext.GetPolicyGroup())
+	return actualConf.Checksum == checksum
```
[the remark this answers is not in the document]
kingamarton: It shouldn't be nil. Changed to one statement

pkg/common/configs/config.go
kingamarton: You are right. I modified it to parse the checksum wherever it is placed. If it is placed in the middle of the request, it will consider the first 64 characters after the `"Checksum: "` string

wilfred-s (approved):
LGTM
I have two minor nits in the test code where the `assert.NilError()` checks are separated from the function call they check and will fix them on commit.

wilfred-s:
Merged, although github says that there are no merge conflicts the code would not compile after merging. The changes in YUNIKORN-418 #231 renamed a function call which caused a compilation error.
Fixed as part of the commit, marking

PR #211, opened 2020-10-01
[YUNIKORN-429] Set default logging level to INFO

adamantal:
Rationalized a few log items in the core:
- increased log level from DEBUG to INFO when:
  - adding a new managed queue 
  - validating configs (once per partition). More granular logs are kept in DEBUG level
  - registering scheduler plugins
- added a few extra lines in `entrypoint.go` to display what services have been started with YuniKorn

When the log levels are increased from DEBUG to INFO we will get rid of the following entries:
1. All the DEBUG level FTM log entries, like this:
```
2020-10-01T14:40:12.912+0200	DEBUG	scheduler/scheduler.go:231	enqueued event	{"eventType": "*schedulerevent.SchedulerUpdatePartitionsConfigEvent", "eventError": "json: unsupported type: chan *commonevents.Result", "currentQueueSize": 0}
```
I think we're fine to remove all these.

2. Outstanding request inspection:
```
2020-10-01T14:40:13.888+0200	DEBUG	scheduler/scheduler.go:129	inspect outstanding requests
2020-10-01T14:42:24.923+0200	DEBUG	scheduler/scheduler.go:135	outstanding request	{"appID": "batch-sleep-job-1", "allocationKey": "3574688c-009b-452b-bdf8-2b66d499e02f"}
```
I think both of them can be kept in debug level, because in case of a busy cluster than can be lots of these logs entries, and by default this log appears in every second. I added an aggregated variation about the outstanding request in `inspectOutstandingRequests()` in INFO level. This gives the information about the size of the pending requests, but does not flood the logs.

3. Queue cleaner
```
2020-10-01T14:41:02.925+0200	DEBUG	scheduler/partition_manager.go:62	time consumed for queue cleaner	{"duration": "24.489µs"}
```
I think this can be kept in debug level as well, it is not an essential information.

4. Metrics collection
```
2020-10-01T14:41:12.887+0200	DEBUG	metrics/metrics_collector.go:58	Adding current status to historical partition data
```
This log appears in every minute. I think if we display in `entrypoint.go` that YuniKorn have started the metrics collection there's no need to display this in every minute.

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -377,8 +377,6 @@ func Validate(newConfig *SchedulerConfig) error {
 			partition.Name = DefaultPartition
 		}
 		// check the queue structure
-		log.Logger().Info("checking partition",
-			zap.String("partitionName", partition.Name))
 		err := checkQueuesStructure(&partition)
 		if err != nil {
 			return err
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -377,7 +377,7 @@ func Validate(newConfig *SchedulerConfig) error {
 			partition.Name = DefaultPartition
 		}
 		// check the queue structure
-		log.Logger().Debug("checking partition",
+		log.Logger().Info("checking partition",
 			zap.String("partitionName", partition.Name))
 		err := checkQueuesStructure(&partition)
 		if err != nil {
--- a/pkg/cache/queue_info.go
+++ b/pkg/cache/queue_info.go
@@ -88,7 +88,7 @@ func NewManagedQueue(conf configs.QueueConfig, parent *QueueInfo) (*QueueInfo, e
 		}
 	}
 
-	log.Logger().Debug("queue added",
+	log.Logger().Info("queue added",
 		zap.String("queueName", qi.Name),
 		zap.String("queuePath", qi.GetQueuePath()))
 	return qi, nil
--- a/pkg/entrypoint/entrypoint.go
+++ b/pkg/entrypoint/entrypoint.go
@@ -65,6 +65,7 @@ func startAllServicesWithParameters(opts StartupOptions) *ServiceContext {
 	var eventCache *events.EventCache
 	var eventPublisher events.EventPublisher
 	if opts.eventCacheEnabled {
+		log.Logger().Info("creating event cache")
 		events.CreateAndSetEventCache()
 		eventCache = events.GetEventCache()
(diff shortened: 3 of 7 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -377,7 +377,7 @@ func Validate(newConfig *SchedulerConfig) error {
 			partition.Name = DefaultPartition
 		}
 		// check the queue structure
-		log.Logger().Debug("checking partition",
+		log.Logger().Info("checking partition",
```
yangwwei, line 380: Do we need to include this in the INFO level, this line of log is not very informative won't be very much useful.

yangwwei (commented):
hi @adamantal 

Thanks for working on this. I am good with most of the changes in this PR, just 1 minor comment.
Except for these, another thing is I think we have some missing INFO level logs, have you checked if we have had the logs for such places, e.g

1. scheduler creates an allocationProposal
2. scheduler confirms/rejects a proposal
3. scheduler releases an ask/allocation (success/fail)
4. scheduler creates/deletes a reservation
5. scheduler add/remove nodes (success/fail)
6. placement rule logs, like places an app to a queue, creating a queue etc

pkg/common/configs/configvalidator.go
adamantal: Remove the unnecessary log.

adamantal:
Hi @yangwwei,

Used docker-desktop to validate the messages locally. I uploaded the INFO level logs of running some sleep jobs (a few of them were pending due to insufficient resources).
You can check too, that 1-3, 5-6 has traces in the logs.

One missing piece that I found is the reservation. There were pending pods in my scenario, but the reservation did not get triggered, so I could not see the related logs. Also I added an extra log in the reservation and the unreservation phase that may be helpful in the future.

I hope that this make sense. If you have any concerns please let me know.

yangwwei:
Hi @adamantal  Thanks for validating this, looks good. For the reservation/unreservation, we do a bunch of checks to make sure the reservation is only created on some valid nodes. You can try the following steps:
- submit a pod and get it running
- submit another pod that ask for more resources than what is available on the node, but less than the node total
- wait for a few seconds (we have a reservation-delay)
- check if reservation happens

for the changes in this PR, I think they look good enough. Pls double check the reservation-related logs, if there is any missing things, we can track in a different JIRA.

yangwwei:
Since this is only changing logs, code coverage failure won't be a concern. Merging this shortly.

yangwwei (approved):

PR #214, opened 2020-10-05
[YUNIKORN-428] Reduce the warning logs in resources.go

adamantal:
Reduced the log level from WARN to INFO

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -792,7 +792,7 @@ func CalculateAbsUsedCapacity(capacity, used *Resource) *Resource {
 				absResValue = math.MinInt64
 			}
 		} else {
-			log.Logger().Warn("Cannot calculate absolute usage for resource because of missing usage information",
+			log.Logger().Info("Cannot calculate absolute usage for resource because of missing usage information",
 				zap.String("resource name", resourceName))
 			continue
 		}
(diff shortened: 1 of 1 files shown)


pkg/common/resources/resources.go
```diff
@@ -792,7 +792,7 @@ func CalculateAbsUsedCapacity(capacity, used *Resource) *Resource {
 				absResValue = math.MinInt64
 			}
 		} else {
-			log.Logger().Warn("Cannot calculate absolute usage for resource because of missing usage information",
+			log.Logger().Info("Cannot calculate absolute usage for resource because of missing usage information",
```
yangwwei, line 795: Today, we are reporting back all resource types, so we constantly see these lines because resource types "pods", "vcore", "ephemeral-storage", "hugepages-1Gi" etc are not tracked in the scheduler. So setting at INFO level won't seem to be enough for this. I'd suggest removing this line.
Or we will need some mechanism to do white-listed resource names, that will introduce bigger changes for which I don't think it is necessary for now.

yangwwei (changes requested):

adamantal:
Thanks for the review. 

I think it still worth to display the information about the missing resource, but I also agree that this log is too much.

I made it DEBUG level, and only display one aggregated log about the every resource that we're missing information about. I hope you will agree with this resolution.

yangwwei:
> I made it DEBUG level, and only display one aggregated log about the every resource that we're missing information about. I hope you will agree with this resolution.

OK, I think this is fine. Thanks!

yangwwei (approved):
LGTM

PR #215, opened 2020-10-13
[YUNIKORN-352] Validate queue capacity setup

kingamarton:


pkg/common/configs/configvalidator.go
```diff
@@ -72,37 +72,68 @@ func checkACL(acl string) error {
...
-	if resource.Guaranteed != nil && len(resource.Guaranteed) != 0 {
-		_, err := checkResource(resource.Guaranteed)
-		if err != nil {
-			return err
+		if cur.Resources.Guaranteed != nil {
+			if !resources.FitIn(curGuaranteedRes, sum) {
+				return fmt.Errorf("queue %s has guaranteed-resources (%v) smaller than sum of children guaranteed resources (%v)", cur.Name, curGuaranteedRes, sum)
+			}
+		} else {
+			cur.Resources.Guaranteed = sum.ToConf()
```
yangwwei, line 116: this seems to be modifying the value: `cur *QueueConfig`.
Could you pls check if this is the case? I don't think we should modify the configs.

pkg/common/configs/configvalidator.go
```diff
@@ -72,37 +72,68 @@ func checkACL(acl string) error {
...
+// - child or children cannot have higher maximum or guaranteed limits than parents
+// - children (added together) cannot have a higher guaranteed setting than a parent
+func checkResourceConfigurationsForQueue(cur *QueueConfig, parent *QueueConfig) error {
+	// If cur has children, make sure sum of children's guaranteed <= cur.guaranteed
+	curGuaranteedRes, err := resources.NewResourceFromConf(cur.Resources.Guaranteed)
+	if err != nil {
+		return err
+	}
+	if curGuaranteedRes.HasNegativeValue() {
+		return fmt.Errorf("invalid guaranteed resource for queue %s, cannot be negative", cur.Name)
```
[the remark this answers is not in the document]
kingamarton: done

pkg/common/configs/configvalidator.go
```diff
@@ -72,37 +72,68 @@ func checkACL(acl string) error {
...
+	}
+	curMaxRes, err := resources.NewResourceFromConf(cur.Resources.Max)
+	if err != nil {
+		return err
+	}
+	if curMaxRes.HasNegativeValue() {
+		return fmt.Errorf("invalid max resource for queue %s, cannot be negative", cur.Name)
```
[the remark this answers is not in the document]
kingamarton: done

pkg/common/configs/configvalidator.go
kingamarton: Yes, it is the case. This is exactly the same check we had before in the `initializer.go` file. I just moved the check here, to have it in the validation phase. The guaranteed resource was modified there as well if it wasn't set. I haven't removed it because I didn't wanted to change the check's behaviour and I think it makes sense to set the guaranteed resource for a parent queue = to the sum of children's guaranteed resource it it is not set. @yangwwei please let me know if it makes sense for you.

yangwwei: If that's the case, I am OK with the change.
The concern I had earlier was because this is a validation function, a validation function is not supposed to modify the original value of the object.
By looking how we use the validation, it seems to be safe though.

yangwwei (approved):

PR #216, opened 2020-10-14
YUNIKORN-414: Add partitions REST API to fetch list of partitions

manirajv06:
First cut implementation. Need to add few more attributes in JSON response.

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -54,13 +54,13 @@ type PartitionContext struct {
 	allocations            map[string]*objects.Allocation  // allocations
 	placementManager       *placement.AppPlacementManager  // placement manager for this partition
 	partitionManager       *partitionManager               // manager for this partition
-	StateMachine           *fsm.FSM                        // the state of the partition for scheduling
-	StateTime              time.Time                       // last time the state was updated (needed for cleanup)
+	stateMachine           *fsm.FSM                        // the state of the partition for scheduling
+	stateTime              time.Time                       // last time the state was updated (needed for cleanup)
 	isPreemptable          bool                            // can allocations be preempted
 	rules                  *[]configs.PlacementRule        // placement rules to be loaded by the scheduler
 	userGroupCache         *security.UserGroupCache        // user cache per partition
 	totalPartitionResource *resources.Resource             // Total node resources
-	NodeSortingPolicy      *policies.NodeSortingPolicy     // Global Node Sorting Policies
+	nodeSortingPolicy      *policies.NodeSortingPolicy     // Global Node Sorting Policies
 
 	sync.RWMutex
 }
@@ -76,8 +76,8 @@ func newPartitionContext(conf configs.PartitionConfig, rmID string, cc *ClusterC
 	pc := &PartitionContext{
 		Name:         conf.Name,
 		RmID:         rmID,
-		StateMachine: objects.NewObjectState(),
-		StateTime:    time.Now(),
+		stateMachine: objects.NewObjectState(),
+		stateTime:    time.Now(),
 		applications: make(map[string]*objects.Application),
 		reservedApps: make(map[string]int),
 		nodes:        make(map[string]*objects.Node),
@@ -136,10 +136,10 @@ func (pc *PartitionContext) initialPartitionFromConfig(conf configs.PartitionCon
 	case policies.BinPackingPolicy, policies.FairnessPolicy:
 		log.Logger().Info("NodeSorting policy set from config",
 			zap.String("policyName", configuredPolicy.String()))
-		pc.NodeSortingPolicy = policies.NewNodeSortingPolicy(conf.NodeSortPolicy.Type)
+		pc.nodeSortingPolicy = policies.NewNodeSortingPolicy(conf.NodeSortPolicy.Type)
 	case policies.Unknown:
 		log.Logger().Info("NodeSorting policy not set using 'fair' as default")
-		pc.NodeSortingPolicy = policies.NewNodeSortingPolicy("fair")
+		pc.nodeSortingPolicy = policies.NewNodeSortingPolicy("fair")
 	}
 	return nil
 }
@@ -248,23 +248,25 @@ func (pc *PartitionContext) markPartitionForRemoval() {
 // Get the state of the partition.
 // No new nodes and applications will be accepted if stopped or being removed.
 func (pc *PartitionContext) isDraining() bool {
-	return pc.StateMachine.Current() == objects.Draining.String()
+	return pc.stateMachine.Current() == objects.Draining.String()
 }
 
 func (pc *PartitionContext) isRunning() bool {
-	return pc.StateMachine.Current() == objects.Active.String()
+	return pc.stateMachine.Current() == objects.Active.String()
 }
 
 func (pc *PartitionContext) isStopped() bool {
-	return pc.StateMachine.Current() == objects.Stopped.String()
+	return pc.stateMachine.Current() == objects.Stopped.String()
 }
 
 // Handle the state event for the partition.
 // The state machine handles the locking.
 func (pc *PartitionContext) handlePartitionEvent(event objects.ObjectEvent) error {
-	err := pc.StateMachine.Event(event.String(), pc.Name)
+	err := pc.stateMachine.Event(event.String(), pc.Name)
 	if err == nil {
-		pc.StateTime = time.Now()
+		pc.Lock()
+		defer pc.Unlock()
+		pc.stateTime = time.Now()
 		return nil
 	}
 	// handle the same state transition not nil error (limit of fsm).
@@ -897,7 +899,7 @@ func (pc *PartitionContext) unReserve(app *objects.Application, node *objects.No
 // Sorting should use a copy of the node list not the main list.
 func (pc *PartitionContext) getNodeIteratorForPolicy(nodes []*objects.Node) interfaces.NodeIterator {
 	pc.RLock()
-	configuredPolicy := pc.NodeSortingPolicy.PolicyType
+	configuredPolicy := pc.nodeSortingPolicy.PolicyType
 	pc.RUnlock()
 	if configuredPolicy == policies.Unknown {
 		return nil
@@ -1226,4 +1228,18 @@ func (pc *PartitionContext) cleanupExpiredApps() {
 		delete(pc.applications, app.ApplicationID)
 		pc.Unlock()
 	}
+}
+
+func (pc *PartitionContext) GetCurrentState() string {
+	return pc.stateMachine.Current()
+}
+
+func (pc *PartitionContext) GetStateTime() time.Time {
+	pc.RLock()
+	defer pc.RUnlock()
+	return pc.stateTime
+}
+
+func (pc *PartitionContext) GetNodeSortingPolicy() *policies.NodeSortingPolicy {
+	return pc.nodeSortingPolicy
 }
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -264,8 +264,6 @@ func (pc *PartitionContext) isStopped() bool {
 func (pc *PartitionContext) handlePartitionEvent(event objects.ObjectEvent) error {
 	err := pc.stateMachine.Event(event.String(), pc.Name)
 	if err == nil {
-		pc.Lock()
-		defer pc.Unlock()
 		pc.stateTime = time.Now()
 		return nil
 	}
@@ -898,9 +896,7 @@ func (pc *PartitionContext) unReserve(app *objects.Application, node *objects.No
 // Get the iterator for the sorted nodes list from the partition.
 // Sorting should use a copy of the node list not the main list.
 func (pc *PartitionContext) getNodeIteratorForPolicy(nodes []*objects.Node) interfaces.NodeIterator {
-	pc.RLock()
-	configuredPolicy := pc.nodeSortingPolicy.PolicyType
-	pc.RUnlock()
+	configuredPolicy := pc.GetNodeSortingPolicy()
 	if configuredPolicy == policies.Unknown {
 		return nil
 	}
@@ -1240,6 +1236,8 @@ func (pc *PartitionContext) GetStateTime() time.Time {
 	return pc.stateTime
 }
 
-func (pc *PartitionContext) GetNodeSortingPolicy() *policies.NodeSortingPolicy {
-	return pc.nodeSortingPolicy
+func (pc *PartitionContext) GetNodeSortingPolicy() policies.SortingPolicy {
+	pc.RLock()
+	defer pc.RUnlock()
+	return pc.nodeSortingPolicy.PolicyType
 }
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -262,6 +262,8 @@ func (pc *PartitionContext) isStopped() bool {
 // Handle the state event for the partition.
 // The state machine handles the locking.
 func (pc *PartitionContext) handlePartitionEvent(event objects.ObjectEvent) error {
+	pc.Lock()
+	defer pc.Unlock()
 	err := pc.stateMachine.Event(event.String(), pc.Name)
 	if err == nil {
 		pc.Lock()
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -54,13 +54,13 @@ type PartitionContext struct {
 	allocations            map[string]*objects.Allocation  // allocations
 	placementManager       *placement.AppPlacementManager  // placement manager for this partition
 	partitionManager       *partitionManager               // manager for this partition
-	stateMachine           *fsm.FSM                        // the state of the partition for scheduling
-	stateTime              time.Time                       // last time the state was updated (needed for cleanup)
+	StateMachine           *fsm.FSM                        // the state of the partition for scheduling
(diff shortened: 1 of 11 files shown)


pkg/webservice/routes.go
```diff
@@ -126,6 +126,14 @@ var webRoutes = routes{
 		getContainerHistory,
 	},
 
+	//
+	route{
+		"Partitions",
+		"GET",
+		"/ws/v1/partitions",
+		getPartitions,
+	},
```
yangwwei, line 135: Are we going to support the sub-path `ws/v1/partitions/{partitionName}`?

pkg/webservice/dao/partition_info.go
```diff
@@ -24,6 +24,14 @@ type PartitionDAOInfo struct {
 	Queues        QueueDAOInfo      `json:"queues"`
 }
 
+type PartitionInfo struct {
+	Name              string            `json:"name"`
+	Capacity          PartitionCapacity `json:"capacity"`
+	IsPremptable      bool              `json:"isPremptable"`
```
yangwwei, line 30: remove this

pkg/webservice/dao/partition_info.go
```diff
@@ -24,6 +24,14 @@ type PartitionDAOInfo struct {
 	Queues        QueueDAOInfo      `json:"queues"`
 }
 
+type PartitionInfo struct {
```
yangwwei, line 27: Can we add some fields about the partition state?
- state
- lastStateTransitionTime

yangwwei (changes requested):

pkg/webservice/dao/partition_info.go
```diff
@@ -33,3 +41,11 @@ type NodeInfo struct {
 	NodeID     string `json:"nodeId"`
```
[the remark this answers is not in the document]
manirajv06: We are using PartitionInfo (not PartitionDAOInfo) struct for this API. May be cleaning up PartitionDAOInfo struct can be taken up later in separate jira?

pkg/webservice/routes.go
manirajv06: As of now, We don't have such plan.

pkg/webservice/dao/partition_info.go
```diff
@@ -24,12 +24,29 @@ type PartitionDAOInfo struct {
 	Queues        QueueDAOInfo      `json:"queues"`
 }
 
+type PartitionInfo struct {
+	Name                    string            `json:"name"`
+	Capacity                PartitionCapacity `json:"capacity"`
```
sunilgovind, line 29: This capacity will be the configured capacity one, correct?

pkg/webservice/dao/partition_info.go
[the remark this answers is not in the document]
sunilgovind: Lets take this up separate, thats more cleaner and better.

pkg/webservice/handlers.go
```diff
@@ -527,3 +527,45 @@ func updateConfiguration(conf string) (string, error) {
...
+	partitionInfo.NodeSortingPolicy = partitionContext.GetNodeSortingPolicy().String()
+
+	applicationsInfo := dao.Applications{}
+	appList := partitionContext.GetApplications()
+	applicationsState := make(map[string]int)
+
+	for _, app := range appList {
+		applicationsState[app.GetApplicationState()]++
+	}
+	applicationsInfo.Running = applicationsState["Running"]
```
sunilgovind, line 564: could we optimize this code a bit. too much of impl related to comparison etc.

pkg/webservice/dao/partition_info.go
manirajv06: As and when any new node gets added, Node's total resource would get added to this capacity. Similarly, for node removal resource would be subtracted from this capacity.

pkg/webservice/handlers.go
manirajv06: Can you please check the recent changes? It has been taken care.

yangwwei: hi @manirajv06  for this part. Can we avoid using hardcoded states for Applications struct.
Maybe we can change it to

```
type PartitionInfo struct {
   ...
   appCounts map[string]int
}
```
then we can simply aggregate the count of apps by their state. What do you think?

manirajv06: hi @yangwwei  Sorry for the delay. 

applicationsState in handlers#getPartition() is a map contains the state wise count. 

applicationsInfo ( dao.Applications{} ) in handlers#getPartition() has been used to display the state wise count (after aggregating some app states. For ex, Killed and Rejected as Failed) in JSON output.

I am not sure how defining appCounts map[string]int in PartitionInfo struct helps in this case as we are manipulating some app states. Please correct my understanding if needed.

pkg/webservice/dao/partition_info.go
sunilgovind: ok. thanks

pkg/webservice/handlers.go
```diff
@@ -521,3 +521,39 @@ func updateConfiguration(conf string) (string, error) {
...
+
+func getPartitions(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+
+	var partitionsInfo []*dao.PartitionInfo
+	lists := schedulerContext.GetPartitionMapClone()
+	for _, partitionContext := range lists {
+		partitionInfo := &dao.PartitionInfo{}
+		partitionInfo.Name = partitionContext.Name
+		partitionInfo.State = partitionContext.StateMachine.Current()
```
yangwwei, line 533: It is not a good idea to expose `StateMachine` and `StateTime`, they are pretty internal objects.
Instead, we can create a `GetCurrentState()` function that returns a string. Same for the stateTime.

pkg/scheduler/partition.go
```diff
@@ -54,13 +54,13 @@ type PartitionContext struct {
 	allocations            map[string]*objects.Allocation  // allocations
 	placementManager       *placement.AppPlacementManager  // placement manager for this partition
 	partitionManager       *partitionManager               // manager for this partition
-	stateMachine           *fsm.FSM                        // the state of the partition for scheduling
-	stateTime              time.Time                       // last time the state was updated (needed for cleanup)
```
yangwwei, line 58: See my comments later for this, we should not expose the internal objects out of its package.

pkg/scheduler/partition.go
```diff
@@ -54,13 +54,13 @@ type PartitionContext struct {
...
+	StateMachine           *fsm.FSM                        // the state of the partition for scheduling
+	StateTime              time.Time                       // last time the state was updated (needed for cleanup)
 	isPreemptable          bool                            // can allocations be preempted
 	rules                  *[]configs.PlacementRule        // placement rules to be loaded by the scheduler
 	userGroupCache         *security.UserGroupCache        // user cache per partition
 	totalPartitionResource *resources.Resource             // Total node resources
-	nodeSortingPolicy      *policies.NodeSortingPolicy     // Global Node Sorting Policies
+	NodeSortingPolicy      *policies.NodeSortingPolicy     // Global Node Sorting Policies
```
yangwwei, line 63: Do not expose this field as well. We can introduce a function to return the sorting policy instead.

pkg/webservice/dao/partition_info.go
```diff
@@ -24,12 +24,29 @@ type PartitionDAOInfo struct {
...
 	NodeID     string `json:"nodeId"`
 	Capability string `json:"capability"`
 }
+
+type Applications struct {
+	Total     int `json:"total"`
+	Running   int `json:"running"`
+	Pending   int `json:"pending"`
+	Completed int `json:"completed"`
+	Failed    int `json:"failed"`
```
yangwwei, line 51: My earlier comment is to remove hardcoded state names.
Because if we modified app state in the scheduler package, and most likely we are going to do so. It is going to break this. And I see we are doing some formatting like:

```
applicationsInfo.Running = applicationsState["Running"]
applicationsInfo.Pending = applicationsState["Waiting"] + applicationsState["Accepted"] + applicationsState["Starting"] + applicationsState["New"]
applicationsInfo.Completed = applicationsState["Completed"]
applicationsInfo.Failed = applicationsState["Killed"] + applicationsState["Rejected"]
applicationsInfo.Total = applicationsInfo.Running + applicationsInfo.Pending + applicationsInfo.Completed + applicationsInfo.Failed
```
I'd rather stay with the initial values. Because we are documenting all the app states carefully, see http://yunikorn.apache.org/docs/next/design/scheduler_object_states#application-state. We need to keep the REST API response align with the doc. So I think returning its original map plus a total number is enough. Does that make sense to you?

yangwwei (commented):
hi @manirajv06 catching up with this, thanks for the updates. Please see my comments. Thanks!

pkg/webservice/handlers.go
manirajv06: Taken care.

pkg/scheduler/partition.go
manirajv06: Taken care.

pkg/webservice/dao/partition_info.go
manirajv06: I was trying to sum up certain states under common state. However, showing up each state count is also detailed enough.

pkg/scheduler/partition.go
```diff
@@ -1099,3 +1099,15 @@ func (pc *PartitionContext) removeAllocationAsk(appID string, allocationKey stri
...
+	return pc.stateMachine.Current()
+}
+
+func (pc *PartitionContext) GetStateTime() time.Time {
+	return pc.stateTime
+}
+
+func (pc *PartitionContext) GetNodeSortingPolicy() *policies.NodeSortingPolicy {
+	return pc.nodeSortingPolicy
+}
```
yangwwei, line 1113: I think we need to add a read lock for `GetStateTime()`, and the write lock in `handlePartitionEvent()`. because `handlePartitionEvent()` and the REST API could be called at the same time.

```
func (pc *PartitionContext) handlePartitionEvent(event objects.ObjectEvent) error {
	err := pc.stateMachine.Event(event.String(), pc.Name)
	if err == nil {
                pc.Lock()
		pc.stateTime = time.Now()
                pc.Unlock()
		return nil
	}
	// handle the same state transition not nil error (limit of fsm).
	if err.Error() == "no transition" {
		return nil
	}
	return err
}
```

`GetCurrentState()` and `GetNodeSortingPolicy()` should be safe because state machine Current() holds the lock already, and the nodeSortingPolicy only gets modified when updating the config, that is a swapping of the old object.

yangwwei (commented):
Thanks for the update @manirajv06 , I have one more comment. And could you please check the UT failure and the lint failure logged here: https://travis-ci.com/github/apache/incubator-yunikorn-core/builds/214290145? Thanks

pkg/scheduler/partition.go
manirajv06: Taken care.

pkg/scheduler/partition.go
```diff
@@ -262,6 +262,8 @@ func (pc *PartitionContext) isStopped() bool {
 // Handle the state event for the partition.
 // The state machine handles the locking.
 func (pc *PartitionContext) handlePartitionEvent(event objects.ObjectEvent) error {
+	pc.Lock()
+	defer pc.Unlock()
```
yangwwei, line 266: Can we move this to line 269?
I think we do not need to lock the partition when calling `pc.stateMachine.Event(event.String(), pc.Name)`

yangwwei (changes requested):
Looks almost good, just added one more comment in handlePartitionEvent. And pls fix the remaining lint issue, thanks!

yangwwei:
hi @manirajv06  Looks almost good, but still need to fix the lint issue

```
$ make lint
running golangci-lint
checking against commit sha 2bf276764227
pkg/scheduler/objects/application.go:141:1: receiver name ai should be consistent with previous receiver name sa for Application (golint)
func (ai *Application) IsCompleted() bool {
^
pkg/scheduler/objects/application.go:145:1: receiver name ai should be consistent with previous receiver name sa for Application (golint)
func (ai *Application) IsKilled() bool {
^
pkg/scheduler/objects/application.go:149:1: receiver name ai should be consistent with previous receiver name sa for Application (golint)
func (ai *Application) IsRejected() bool {
^
Makefile:58: recipe for target 'lint' failed
make: *** [lint] Error 1
```

you can run `make lint` locally to ensure there is no more lint issues before updating the PR. Thanks

manirajv06:
Fixed.

yangwwei (approved):
+1. I will commit this shortly. Thanks @manirajv06 !

yangwwei:
Oops, I can't get this merged as there are some conflicts.
Could you pls fix the conflict? Thanks!

pkg/webservice/dao/partition_info.go
wilfred-s: I do not see this structure being used anywhere we should remove it.

pkg/webservice/handlers_test.go
```diff
@@ -607,3 +627,46 @@ func TestGetNodesUtilJSON(t *testing.T) {
...
+	resp := &MockResponseWriter{}
+	var partitionInfo []dao.PartitionInfo
+	getPartitions(resp, req)
+	err = json.Unmarshal(resp.outputBytes, &partitionInfo)
+	assert.NilError(t, err, "failed to unmarshal applications dao response from response body: %s", string(resp.outputBytes))
+	for _, part := range partitionInfo {
+		if part.Name == partitionName {
+			assert.Equal(t, part.Name, "[rm-123]default")
+			assert.Equal(t, part.NodeSortingPolicy, "fair")
+			assert.Equal(t, part.Applications["total"], 1)
```
wilfred-s, line 664: We should also test the count for the state the application is in.
```
assert.Equal(t, part.Applications["new"], 1)
```
can we also extend this to multiple apps, in different states and make sure they all count correctly and sum up to the total.

wilfred-s (changes requested):

pkg/scheduler/partition.go
```diff
@@ -264,6 +264,8 @@ func (pc *PartitionContext) isStopped() bool {
 func (pc *PartitionContext) handlePartitionEvent(event objects.ObjectEvent) error {
 	err := pc.stateMachine.Event(event.String(), pc.Name)
 	if err == nil {
+		pc.Lock()
+		defer pc.Unlock()
```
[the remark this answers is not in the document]
yangwwei: hi @wilfred-s  does this [my earlier comment ](https://github.com/apache/incubator-yunikorn-core/pull/216#discussion_r563348013)make sense to you?

pkg/webservice/handlers_test.go
manirajv06: Improved UT coverage.

pkg/webservice/dao/partition_info.go
```diff
@@ -35,18 +35,10 @@ type PartitionInfo struct {
 
 type PartitionCapacity struct {
 	Capacity     string `json:"capacity"`
-	UsedCapacity string `json:"usedCapacity"`
+	UsedCapacity string `json:"usedcapacity"`
```
manirajv06, line 38: Renaming as before because this change should not break the existing use cases. Once new API has been adopted completely, we can do the naming with similar convention.

pkg/scheduler/partition.go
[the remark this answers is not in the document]
wilfred-s: For Time the [documentation](https://golang.org/pkg/time/#Time) shows:
> A Time value can be used by multiple goroutines simultaneously except that the methods GobDecode, UnmarshalBinary, UnmarshalJSON and UnmarshalText are not concurrency-safe.

Which would mean that we do not need a lock.

pkg/scheduler/partition.go
```diff
@@ -1208,3 +1210,17 @@ func (pc *PartitionContext) removeAllocationAsk(appID string, allocationKey stri
...
+}
+
+func (pc *PartitionContext) GetStateTime() time.Time {
+	pc.RLock()
+	defer pc.RUnlock()
+	return pc.stateTime
+}
+
+func (pc *PartitionContext) GetNodeSortingPolicy() *policies.NodeSortingPolicy {
+	return pc.nodeSortingPolicy
```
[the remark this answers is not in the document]
wilfred-s: Can we change this to:
```
func (pc *PartitionContext) GetNodeSortingPolicy() String {
	pc.RLock()
	defer pc.RUnlock()
	return pc.nodeSortingPolicy.PolicyType.String()
}
```
That is inline with the `getNodeIteratorForPolicy()` call.

pkg/scheduler/partition.go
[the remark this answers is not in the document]
yangwwei: Good point, agree we can remove this. Thanks!

pkg/scheduler/partition.go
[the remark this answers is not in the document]
manirajv06: Taken care, But returning "SortingPolicy instead of string" approach is re-usable. Hope this is fine.

pkg/scheduler/partition.go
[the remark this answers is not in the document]
manirajv06: Taken care.

pkg/scheduler/partition.go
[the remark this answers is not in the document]
wilfred-s: with the reuse I am ok with this

wilfred-s (approved):
LGTM, thanks for sticking with me on this long review.

PR #218, opened 2020-10-20
[YUNIKORN-445] Panic in Resource.String()

wilfred-s:
The String method of Resource is not nil safe which causes a panic to be
logged instead of a proper message.


yangwwei (approved):
+1

yangwwei:
Hi @wilfred-s  seems like there is a conflict, could you please fix it?

kingamarton (approved):

yangwwei:
hi @wilfred-s  could u pls fix the conflict and merge this?

PR #220, opened 2020-10-30
[YUNIKORN-456] Added ENV var to disable reservation

kingamarton:


pkg/scheduler/scheduling_application.go
```diff
@@ -598,7 +606,7 @@ func (sa *SchedulingApplication) tryNodes(ask *schedulingAllocationAsk, nodeIter
 		// nothing allocated should we look at a reservation?
 		// TODO make this smarter a hardcoded delay is not the right thing
 		askAge := time.Since(ask.getCreateTime())
-		if allowReserve && askAge > reservationDelay {
+		if !reservationDisabled && allowReserve && askAge > reservationDelay {
```
yangwwei, line 609: Minor thing: could we move the `reservationDisabled ` check to the former line in this function:
```
allowReserve := len(reservedAsks) < int(ask.pendingRepeatAsk)
```

pkg/scheduler/scheduling_application.go
```diff
@@ -28,14 +28,22 @@ import (
...
 )
 
+const DisableReservationVarName = "DISABLE_RESERVATION"
+
 var reservationDelay = 2 * time.Second
+var reservationDisabled = isReservationDisabled()
+
+func isReservationDisabled() bool {
+	return common.GetBoolEnvVar(DisableReservationVarName, false)
+}
```
yangwwei, line 46: This will lookup for the env var all the time in each scheduling cycle, this might be too expensive.
To update the env vars, we require a restart (i.e update the ENV VARs defined in the deployment yaml file and restart the scheduler). In this case, I suggest we parse the env vars into a common struct `schedulerOptions`, and pass that to `ClusterSchedulingContext`. So the scheduling code can directly get the options from the context.

yangwwei (changes requested):

yangwwei (approved):
LGTM

PR #222, opened 2020-11-24
[YUNIKORN-465] scheduler health check REST API

kingamarton:
I still have to add the test cases for the change. @wilfred-s , @yangwwei please check and add your ideas.
Shortly what I added in the PR:

- checked for scheduling errors: even if right now are used for restored nodes, now can be an alarm if we have some, if we will extend its usage in the future it will be even better
- checked for failed nodes: I think it means that something is/was wrong with our internal state, if for example we have a node for a non existing partition.
- added the consistency check
- added the reservationNr/nodeNr ratio check. Here I don't know yet what should be the limit when we can say that everything is OK. Any ideas?

I skipped the latency check at the moment. I like the idea of measuring the time until an ask is confirmed. I think we can add this measurement to the metrics as well. And than we can include in the health check as well.

### DIFF
--- a/pkg/webservice/dao/scheduler-health.go
+++ b/pkg/webservice/dao/scheduler-health.go
@@ -1,0 +1,37 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package dao
+
+type SchedulerHealthDAOInfo struct {
+	Healthy              bool                         `json:"healthy"`
+	SchedulingErrors     int                          `json:"schedulingErrors"`
+	FailedNodes          int                          `json:"failedNodes"`
+	UnhealthyPartition   map[string]*PartitionDAOInfo `json:"unhealthyPartitions"`
+	UnhealthyNodes       map[string]*NodeDAOInfo      `json:"unhealthyNodes"`
+	ReservationNodeRatio float32                      `json:"reservationRatio"`
+}
+
+func (s *SchedulerHealthDAOInfo) SetHealthStatus() {
+	s.Healthy =
+		s.SchedulingErrors == 0 &&
+			s.FailedNodes == 0 &&
+			len(s.UnhealthyPartition) == 0 &&
+			len(s.UnhealthyNodes) == 0
+	//TODO: what is a healthy reservation/node ratio?
+}
--- a/pkg/webservice/dao/scheduler-health.go
+++ b/pkg/webservice/dao/scheduler-health.go
@@ -19,19 +19,27 @@
 package dao
 
 type SchedulerHealthDAOInfo struct {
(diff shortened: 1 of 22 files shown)


pkg/webservice/dao/scheduler-health.go
```diff
@@ -0,0 +1,37 @@
...
+
+package dao
+
+type SchedulerHealthDAOInfo struct {
+	Healthy              bool                         `json:"healthy"`
+	SchedulingErrors     int                          `json:"schedulingErrors"`
+	FailedNodes          int                          `json:"failedNodes"`
+	UnhealthyPartition   map[string]*PartitionDAOInfo `json:"unhealthyPartitions"`
+	UnhealthyNodes       map[string]*NodeDAOInfo      `json:"unhealthyNodes"`
+	ReservationNodeRatio float32                      `json:"reservationRatio"`
```
yangwwei, line 27:
```
type SchedulerHealthDAOInfo struct {
   Healthy bool
   HealthChecks []HealthCheckInfo
}

type HealthCheckInfo struct {
   Name string
   Description string // describe what we are checking here
   Succeed bool
   DiagnosisMessage string // non-empty if succeed=false, some human-readable message to explain why this fails
}
```

pkg/webservice/handlers.go
```diff
@@ -494,6 +495,95 @@ func updateConfig(w http.ResponseWriter, r *http.Request) {
 	}
 	buildUpdateResponse(nil, w)
 }
+func checkScheduingErrors(metrics metrics2.CoreSchedulerMetrics, w http.ResponseWriter, result *dao.SchedulerHealthDAOInfo) {
+	schedulingErrors, err := metrics.GetSchedulingErrors()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
```
yangwwei, line 502: what the error message looks like when multiple of these checks are failed?

pkg/webservice/dao/scheduler-health.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func (s *SchedulerHealthDAOInfo) SetHealthStatus() {
+	s.Healthy =
+		s.SchedulingErrors == 0 &&
+			s.FailedNodes == 0 &&
+			len(s.UnhealthyPartition) == 0 &&
+			len(s.UnhealthyNodes) == 0
+	//TODO: what is a healthy reservation/node ratio?
```
yangwwei, line 36: if we do not know what's the reasonable ratio, can we add this value to the returned message and mark this check always passed? So we can observe what kind of values we can get from the real cluster. maybe later we have a better idea of how to set this up.

pkg/webservice/routes.go
```diff
@@ -197,4 +197,11 @@ var webRoutes = routes{
 		Pattern:     "/debug/pprof/trace",
 		HandlerFunc: pprof.Trace,
 	},
+	// endpoint to check health status
+	route{
+		"Scheduler",
+		"GET",
+		"/ws/v1/scheduler/health",
```
yangwwei, line 204: better to rename this "/ws/v1/scheduler/healthcheck", which is more self-explanatory

pkg/webservice/dao/scheduler-health.go
kingamarton: I think it makes no sense to collect the successful checks as well. I would remove the `Succeed` flag from the `HealthCheckInfo` and collect only the failed infos.

yangwwei: It makes some sense to display what kind of checks have been included. When that gets rendered in UI, it could give us a list of checks, and the individual states. If this is not adding too much work, I would rather include this info.

pkg/webservice/routes.go
kingamarton: Done

pkg/webservice/dao/scheduler-health.go
kingamarton: Ok. I have included it

yangwwei (approved):
Looks good. Thanks @kingamarton

PR #223, opened 2020-11-26
YUNIKORN-415: Add "queues" REST API to fetch list of queues

manirajv06:
First cut implementation.

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -303,8 +303,8 @@ func (sq *Queue) UpdateSortType() {
 }
 
 func (sq *Queue) GetQueuePath() string {
-	sq.Lock()
-	defer sq.Unlock()
+	sq.RLock()
+	defer sq.RUnlock()
 	return sq.QueuePath
 }
 
@@ -416,6 +416,8 @@ func (sq *Queue) GetPartitionQueues() dao.PartitionQueueDAOInfo {
 			queueInfo.Children = append(queueInfo.Children, child.GetPartitionQueues())
 		}
 	}
+	sq.RLock()
+	defer sq.RUnlock()
 	queueInfo.QueueName = sq.GetQueuePath()
 	queueInfo.Status = sq.stateMachine.Current()
 	queueInfo.MaxResource = sq.maxResource.DAOString()
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -409,6 +409,28 @@ func (sq *Queue) GetQueueInfos() dao.QueueDAOInfo {
 	return queueInfo
 }
 
+func (sq *Queue) GetPartitionQueues() dao.PartitionQueueDAOInfo {
+	queueInfo := dao.PartitionQueueDAOInfo{}
+	if len(sq.children) > 0 {
+		for _, child := range sq.GetCopyOfChildren() {
+			queueInfo.Children = append(queueInfo.Children, child.GetPartitionQueues())
+		}
+	}
+	queueInfo.QueueName = sq.GetQueuePath()
+	queueInfo.Status = sq.stateMachine.Current()
+	queueInfo.MaxResource = sq.maxResource.DAOString()
+	queueInfo.GuaranteedResource = sq.guaranteedResource.DAOString()
+	queueInfo.AllocatedResource = sq.allocatedResource.DAOString()
+	queueInfo.IsLeaf = sq.IsLeafQueue()
+	queueInfo.IsManaged = sq.IsManaged()
+	if sq.parent == nil {
+		queueInfo.Parent = ""
+	} else {
+		queueInfo.Parent = sq.parent.GetQueuePath()
+	}
+	return queueInfo
+}
+
 // Return the pending resources for this queue
 func (sq *Queue) GetPendingResource() *resources.Resource {
 	sq.RLock()
--- a/pkg/webservice/dao/queue_info.go
+++ b/pkg/webservice/dao/queue_info.go
@@ -31,3 +31,16 @@ type QueueCapacity struct {
 	UsedCapacity    string `json:"usedcapacity"`
 	AbsUsedCapacity string `json:"absusedcapacity"`
 }
+
+type PartitionQueueDAOInfo struct {
+	QueueName          string                  `json:"queuename"`
+	Status             string                  `json:"status"`
+	Partition          string                  `json:"partition"`
+	MaxResource        string                  `json:"maxResource"`
+	GuaranteedResource string                  `json:"guaranteedResource"`
+	AllocatedResource  string                  `json:"allocatedResource"`
+	IsLeaf             bool                    `json:"isLeaf"`
+	IsManaged          bool                    `json:"isManaged"`
+	Parent             string                  `json:"parent"`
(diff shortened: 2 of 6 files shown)


yangwwei:
hi @manirajv06  thanks for working on this. Is this PR ready for review? I see some travis job failures so I am not sure if this is ready for review. Thanks

sunilgovind (commented):
It seems your working branch is not in sync with current master. @manirajv06

pkg/scheduler/objects/queue.go
```diff
@@ -401,6 +401,28 @@ func (sq *Queue) GetQueueInfos() dao.QueueDAOInfo {
...
 
+func (sq *Queue) GetPartitionQueues() dao.PartitionQueueDAOInfo {
+	queueInfo := dao.PartitionQueueDAOInfo{}
+	if len(sq.children) > 0 {
+		for _, child := range sq.GetCopyOfChildren() {
+			queueInfo.Children = append(queueInfo.Children, child.GetPartitionQueues())
+		}
+	}
+	queueInfo.QueueName = sq.GetQueuePath()
+	queueInfo.Status = sq.stateMachine.Current()
```
sunilgovind, line 412: Is it safe to get this variable w/o readlock ?

pkg/webservice/handlers.go
```diff
@@ -521,3 +522,27 @@ func updateConfiguration(conf string) (string, error) {
 	}
 	return "", fmt.Errorf("config plugin not found")
 }
+
+func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+	if len(vars) == 0 {
+		http.Error(w, "Mandatory parameters are missing in URL path. Please check the usage documentation", http.StatusBadRequest)
```
sunilgovind, line 530: Do we need to return from here?

pkg/scheduler/objects/queue.go
manirajv06: Added readlock to ensure its safety. Also replaced write lock with read lock in GetQueuePath() which got introduced as part of cache refactoring. Read lock is good enough in this case. Please verify this as well.

sunilgovind: Looks good

sunilgovind: looks good

sunilgovind:
@manirajv06 could you please help to check with the codecov report as well?

pkg/webservice/routes.go
```diff
@@ -127,6 +127,13 @@ var webRoutes = routes{
 		getContainerHistory,
 	},
 
+	route{
+		"Scheduler",
+		"GET",
+		"/ws/v1/partition/{partition}/queues",
+		getPartitionQueues,
```
kingamarton, line 134: Now we have two endpoint for retrieving the queue infos. Shouldn't we delete the old one? Or what is the plan, how will the transition to the new API will be done?

pkg/webservice/handlers.go
```diff
@@ -521,3 +523,24 @@ func updateConfiguration(conf string) (string, error) {
 	}
 	return "", fmt.Errorf("config plugin not found")
 }
+
+func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+	_, partitionExists := vars["partition"]
+	if len(vars) == 0 || !partitionExists {
+		http.Error(w, "Mandatory parameters are missing in URL path. Please check the usage documentation", http.StatusBadRequest)
```
kingamarton, line 532: Please extend the error message with the missing parameter names.

pkg/webservice/handlers.go
```diff
@@ -521,3 +523,24 @@ func updateConfiguration(conf string) (string, error) {
 	}
 	return "", fmt.Errorf("config plugin not found")
 }
+
+func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+	_, partitionExists := vars["partition"]
```
kingamarton, line 530: Please save here the partition name instead of ignoring it, so in the line 536 you can directly use it

pkg/webservice/dao/queue_info.go
```diff
@@ -31,3 +31,16 @@ type QueueCapacity struct {
...
+	Status             string                  `json:"status"`
+	Partition          string                  `json:"partition"`
+	MaxResource        string                  `json:"maxResource"`
+	GuaranteedResource string                  `json:"guaranteedResource"`
+	AllocatedResource  string                  `json:"allocatedResource"`
+	IsLeaf             bool                    `json:"isLeaf"`
+	IsManaged          bool                    `json:"isManaged"`
+	Parent             string                  `json:"parent"`
+	Children           []PartitionQueueDAOInfo `json:"children"`
+}
```
kingamarton, line 46: We already have defined a `QueueDAOInfo`, I think we should reuse/modify that one instead of creating a new DAO structure for the Queues.

pkg/webservice/handlers.go
```diff
@@ -521,3 +523,24 @@ func updateConfiguration(conf string) (string, error) {
...
+	if len(vars) == 0 || !partitionExists {
+		http.Error(w, "Mandatory parameters are missing in URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	var partitionQueuesDAOInfo dao.PartitionQueueDAOInfo
+	var partition = schedulerContext.GetPartition(vars["partition"])
+	if partition != nil {
+		partitionQueuesDAOInfo = partition.GetPartitionQueues()
+	} else {
+		http.Error(w, "partition not found", http.StatusInternalServerError)
```
kingamarton, line 540: Wouldn't be better to return with BasRequest error in this case as well? If the partition is not found that means that the user entered a wrong partition name.

kingamarton (changes requested):
Can you please add some more test cases? The patch coverage is very low.

pkg/webservice/routes.go
manirajv06: Idea is to migrate the users to newly designed API's and slowly retire old one's. Hence, didn't touched the old places, in fact created a completely new dao's, struct etc and planning to take care of all clean up activities after the adoption through separate jira's. This way, we need not worry to about backward compatibility etc. Hope it makes sense.

pkg/webservice/handlers.go
manirajv06: Taken care.

pkg/webservice/handlers.go
manirajv06: Taken care.

pkg/webservice/dao/queue_info.go
manirajv06: Comment https://github.com/apache/incubator-yunikorn-core/pull/223#discussion_r563832692 covers this too.

pkg/webservice/handlers.go
manirajv06: Taken care.

manirajv06:
> Can you please add some more test cases? The patch coverage is very low.

As a general thing, Developed test cases to ensure API return results as expected. Let me see..

manirajv06:

kingamarton (approved):

PR #224, opened 2020-11-27
YUNIKORN-416: Add "applications" REST API to fetch list of applications

manirajv06:
First cut implementation.

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -40,6 +40,7 @@ import (
 	"github.com/apache/incubator-yunikorn-core/pkg/scheduler/objects"
 	"github.com/apache/incubator-yunikorn-core/pkg/webservice/dao"
 	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
+
 	"github.com/gorilla/mux"
 )
 
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -40,7 +40,6 @@ import (
 	"github.com/apache/incubator-yunikorn-core/pkg/scheduler/objects"
 	"github.com/apache/incubator-yunikorn-core/pkg/webservice/dao"
 	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
-
 	"github.com/gorilla/mux"
 )
 
@@ -662,5 +661,57 @@ func getPartitionNodes(w http.ResponseWriter, r *http.Request) {
 		}
 	} else {
 		buildJSONErrorResponse(w, "Partition not found", http.StatusBadRequest)
+	}
+}
+
+func getQueueApplications(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+
+	partition, partitionExists := vars["partition"]
+	if !partitionExists {
+		buildJSONErrorResponse(w, "Partition is missing in URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	queueName, queueNameExists := vars["queue"]
+	if !queueNameExists {
+		buildJSONErrorResponse(w, "Queue is missing in URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	queueErr := validateQueue(queueName)
+	if queueErr != nil {
+		buildJSONErrorResponse(w, queueErr.Error(), http.StatusBadRequest)
+		return
+	}
+	if len(vars) != 2 {
+		buildJSONErrorResponse(w, "Incorrect URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	var appExistsInPartition = false
+	var appExistsInQueue = false
+	var appsDao []*dao.ApplicationDAOInfo
+	lists := schedulerContext.GetPartitionMapClone()
+	for _, partitionContext := range lists {
+		if partitionContext.Name == partition {
+			appExistsInPartition = true
+			appList := partitionContext.GetApplications()
+			for _, app := range appList {
+				if strings.EqualFold(queueName, app.GetQueueName()) {
+					appExistsInQueue = true
+					appsDao = append(appsDao, getApplicationJSON(app))
+				}
+			}
+		}
+	}
+	if !appExistsInPartition {
+		buildJSONErrorResponse(w, "Partition not found", http.StatusBadRequest)
+		return
+	}
+	if !appExistsInQueue {
+		buildJSONErrorResponse(w, "Queue not found", http.StatusBadRequest)
+		return
+	}
+	if err := json.NewEncoder(w).Encode(appsDao); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/webservice/handlers.go
```diff
@@ -521,3 +522,49 @@ func updateConfiguration(conf string) (string, error) {
...
+			appList := partition.GetApplications()
+			for _, app := range appList {
+				if strings.EqualFold(queueName, app.GetQueueName()) {
+					queueExists = true
+					appsDao = append(appsDao, getApplicationJSON(app))
+				}
+			}
+		}
+	}
+	if !partitionExists {
```
sunilgovind, line 559: Do we need to loop to find the partition exists or not?

pkg/webservice/handlers.go
```diff
@@ -521,3 +522,49 @@ func updateConfiguration(conf string) (string, error) {
...
+		return
+	}
+	var partitionExists = false
+	var queueExists = false
+	var appsDao []*dao.ApplicationDAOInfo
+	lists := schedulerContext.GetPartitionMapClone()
+	for _, partition := range lists {
+		if partition.Name == vars["partition"] {
+			partitionExists = true
+			appList := partition.GetApplications()
```
sunilgovind, line 550: Is it safe to loop application here?

manirajv06: I think so as it is in handler side unlike variables in core objects/classes.

sunilgovind: @wilfred-s cud u please share your thought on this comment?

yangwwei: This function is supposed to get all applications from a queue, why we need to loop the partition?
Can we leverage `queue#getCopyOfApps()`?

pkg/webservice/handlers.go
```diff
@@ -521,3 +523,43 @@ func updateConfiguration(conf string) (string, error) {
...
+
+func getQueueApplications(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+	_, partitionExists := vars["partition"]
+	_, queueExists := vars["queue"]
+	var queueName = vars["queue"]
+
+	if len(vars) != 2 || !partitionExists || !queueExists {
+		http.Error(w, "Mandatory parameters are missing in URL path. Please check the usage documentation", http.StatusBadRequest)
```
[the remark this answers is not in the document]
manirajv06: Makes sense. Taken care.

manirajv06:
> his function is supposed to get all applications from a queue, why we need to loop the partition?
> Can we leverage queue#getCopyOfApps()?

This has been already taken care in earlier commits. Can you check the recent changes?

yangwwei:
> > his function is supposed to get all applications from a queue, why we need to loop the partition?
> > Can we leverage queue#getCopyOfApps()?
> 
> This has been already taken care in earlier commits. Can you check the recent changes?

Hi @manirajv06  I did read the latest change. Currently, it is like:

```
var queue = schedulerContext.GetQueue(queueName, vars["partition"])
if queue != nil {
  appList := partition.GetApplications()
  for _, app := range appList {
    if strings.EqualFold(queueName, app.GetQueueName()) {
      appsDao = append(appsDao, getApplicationJSON(app))
  }
}
```

this loops all apps in the partition. I think we can change this to:

```
var queue = schedulerContext.GetQueue(queueName, vars["partition"])
if queue != nil {
  appList := queue.GetCopyOfApps() // This needs to expose getCopyOfApps()
  ...
}
```
does. that make sense to you?

Second, looks like the code coverage has been dropped. Could you please add some negative tests, e.g wrong HTTP request to improve the UT coverage?

manirajv06:
Taken care. Added more test cases.

manirajv06:

kingamarton (approved):
The change looks good to me now. @yangwwei since you ded a previous review, can you please briefly check it again?
@manirajv06 can you please resolve the conflicts?

pkg/webservice/handlers_test.go
```diff
@@ -836,3 +838,89 @@ func TestMetricsNotEmpty(t *testing.T) {
...
+	assert.Equal(t, 1, len(part.GetApplications()))
+
+	NewWebApp(schedulerContext, nil)
+
+	var req *http.Request
+	req, err = http.NewRequest("GET", "/ws/v1/partition/default/queue/root.default/applications", strings.NewReader(""))
+	vars := map[string]string{
+		"partition": "[rm-123]default",
+		"queue":     "root.default",
+	}
```
yangwwei, line 869: Why the URL has partition: "default", but here in the vars we add "[rm-123]default" ?
There is some inconsistency here. Could you please clarify?
IMO, we should just use the partition name (without the clusterID) here, because clusterID is something internal.

pkg/webservice/handlers.go
```diff
@@ -610,3 +612,50 @@ func getPartitions(w http.ResponseWriter, r *http.Request) {
...
+		buildJSONErrorResponse(w, "Incorrect URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	queueErr := validateQueue(queueName)
+	if queueErr != nil {
+		http.Error(w, queueErr.Error(), http.StatusBadRequest)
+		return
+	}
+	var appsDao []*dao.ApplicationDAOInfo
+	var partition = schedulerContext.GetPartition(vars["partition"])
```
yangwwei, line 640: What is the value of `vars["partition"]` here?
I think we should expect the partition name (without clusterID), e.g "default"
To get this to work, I think there needs some more work here. I think we need to add a function to retrieve the partition with the short form name. E.g `GetPartitionWithoutClusterID()`. Right now, this function `schedulerContext.GetPartition()` seems to require the full name form: `[rmID]default`.
could you please check this?

yangwwei (changes requested):
hi @manirajv06 I have added some comments. please check. Pls also test this locally and upload some REST API requests/responses to make sure this works. Thanks!

pkg/webservice/handlers.go
```diff
@@ -610,3 +612,50 @@ func getPartitions(w http.ResponseWriter, r *http.Request) {
...
+	if !queueExists {
+		buildJSONErrorResponse(w, "Queue is missing in URL path. Please check the usage documentation", http.StatusBadRequest)
+	}
+	if len(vars) != 2 {
+		buildJSONErrorResponse(w, "Incorrect URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	queueErr := validateQueue(queueName)
+	if queueErr != nil {
+		http.Error(w, queueErr.Error(), http.StatusBadRequest)
```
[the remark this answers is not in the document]
manirajv06: Taken care

pkg/webservice/handlers_test.go
manirajv06: Will address in https://issues.apache.org/jira/browse/YUNIKORN-678

yangwwei (approved):
Thanks @manirajv06 . As discussed offline, I will get this merged.
Pls work on https://issues.apache.org/jira/browse/YUNIKORN-678 to solve the remaining review comments. Thanks!

PR #225, opened 2020-12-02
[YUNIKORN-285] change lint sha detection

wilfred-s:
In a branch or PR the CI build has a limited history and often a
detached head. The lint sha detection does not work in all cases.
Adding detached head detection and changing the default reference for
the sha to HEAD^
Bring travis checkout depth inline with the shim.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,11 +54,6 @@ all:
 .PHONY: lint
 # Run lint against the previous commit for PR and branch build
 # In dev setup look at all changes on top of master
-REV := "HEAD^"
-DETACHED := $(findstring "fatal:", $(shell git symbolic-ref HEAD))
-ifndef DETACHED
-REV := "origin/HEAD"
-endif
 lint:
 	@echo "running golangci-lint"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -52,6 +52,13 @@ all:
 	$(MAKE) -C $(dir $(BASE_DIR)) build
 
 .PHONY: lint
+# Run lint against the previous commit for PR and branch build
+# In dev setup look at all changes on top of master
+REV := "HEAD^"
+DETACHED := $(findstring "fatal:", $(shell git symbolic-ref HEAD))
```
kingamarton, line 58: can `git symbolic-ref` have some fatal messages in case of some other errors than a detached head? I think it would be better to check the return status of `git symbolic-ref`, since it will return 1 if is not a symbolic ref, and 128 for some other errors.

wilfred-s: Simplified the check, it did not work in all cases either...

kingamarton (approved):

PR #226, opened 2020-12-03
YUNIKORN-417: Add "nodes" REST API to fetch list of nodes

manirajv06:
First cut implementation

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -661,6 +661,6 @@ func getPartitionNodes(w http.ResponseWriter, r *http.Request) {
 			buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
 		}
 	} else {
-		buildJSONErrorResponse(w, "Partition not found", http.StatusInternalServerError)
+		buildJSONErrorResponse(w, "Partition not found", http.StatusBadRequest)
 	}
 }
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -637,3 +637,26 @@ func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
 		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
 	}
 }
+
+func getPartitionNodes(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+	_, partitionExists := vars["partition"]
+	if len(vars) == 0 || !partitionExists {
+		buildJSONErrorResponse(w, "Mandatory parameters are missing in URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	partition := schedulerContext.GetPartition(vars["partition"])
+	if partition != nil {
+		var nodesDao []*dao.NodeDAOInfo
+		for _, node := range partition.GetNodes() {
+			nodeDao := getNodeJSON(node)
+			nodesDao = append(nodesDao, nodeDao)
+		}
+		if err := json.NewEncoder(w).Encode(nodesDao); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/webservice/handlers.go
```diff
@@ -39,6 +39,8 @@ import (
 	"github.com/apache/incubator-yunikorn-core/pkg/scheduler/objects"
 	"github.com/apache/incubator-yunikorn-core/pkg/webservice/dao"
 	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
+
+	"github.com/gorilla/mux"
```
kingamarton, line 43: This is licensed under BSD 3-Clause "New" or "Revised" License. I am not very familiar with the licensing aspects. Can you check if it is OK to use this? I think we need to mention this at least in some Notice file

kingamarton (changes requested):

pkg/webservice/handlers.go
yangwwei: hi @kingamarton  I think this license is fine, and actually, this dependency is already been used by YK. See more info from the dependency doc in https://cwiki.apache.org/confluence/display/INCUBATOR/YuniKornProposal. Close this as resolved.

pkg/webservice/handlers.go
```diff
@@ -521,3 +523,26 @@ func updateConfiguration(conf string) (string, error) {
 	}
 	return "", fmt.Errorf("config plugin not found")
 }
+
+func getPartitionNodes(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+	_, partitionExists := vars["partition"]
+	if len(vars) == 0 || !partitionExists {
+		http.Error(w, "Mandatory parameters are missing in URL path. Please check the usage documentation", http.StatusBadRequest)
```
yangwwei, line 532: Please explicitly include what parameter is missing in the error message.

pkg/webservice/handlers.go
```diff
@@ -521,3 +523,26 @@ func updateConfiguration(conf string) (string, error) {
...
+		return
+	}
+	partition := schedulerContext.GetPartition(vars["partition"])
```
yangwwei, line 535: Can we change line 530 to

```
partitionName, partitionExists := vars["partition"]
```

and here we can reuse the object:
```
partition := schedulerContext.GetPartition(partitionName)
```

manirajv06:

pkg/webservice/handlers.go
```diff
@@ -610,3 +612,30 @@ func getPartitions(w http.ResponseWriter, r *http.Request) {
...
+	partition, partitionExists := vars["partition"]
+	if !partitionExists {
+		buildJSONErrorResponse(w, "Partition is missing in URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	if len(vars) != 1 {
+		buildJSONErrorResponse(w, "Incorrect URL path. Please check the usage documentation", http.StatusBadRequest)
+		return
+	}
+	partitionContext := schedulerContext.GetPartition(partition)
```
kingamarton, line 628: This part is is the same for all the PR's. Can you please extract it in a method and reuse it?

kingamarton (changes requested):
The changes overall looks good to me. I left just on minor comment.

Can you please also rebase it and fix thee conflicts?

manirajv06:
> The changes overall looks good to me. I left just on minor comment.
> 
> Can you please also rebase it and fix thee conflicts?

done

pkg/webservice/handlers.go
manirajv06: > This part is is the same for all the PR's. Can you please extract it in a method and reuse it?

Sure, will take this up in a follow up jira once all API changes goes into trunk.

kingamarton (approved):

PR #227, opened 2020-12-09
[YUNIKORN-480] root queue max resource update

wilfred-s:
When updating the configuration the root queue max resources should not
be set to the value specified in the config. The root queue represents
the total resources available on the cluster and is updated when nodes
are added removed.
Existing tests in the queue already cover this as part of the test queue
creation. A new test is added to the partition tests to cover the config
update.

### DIFF
--- a/pkg/scheduler/partition_test.go
+++ b/pkg/scheduler/partition_test.go
@@ -1071,6 +1071,7 @@ func TestUpdateRootQueue(t *testing.T) {
 	}
 
 	err = partition.updatePartitionDetails(conf)
+	assert.NilError(t, err, "partition update failed")
 	// resources should not have changed
 	assert.Assert(t, resources.Equals(res, partition.totalPartitionResource), "partition resource not set as expected")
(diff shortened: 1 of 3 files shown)


pkg/scheduler/partition_test.go
```diff
@@ -1042,3 +1042,39 @@ func TestScheduleRemoveReservedAsk(t *testing.T) {
...
+				Queues:    nil,
+			},
+		},
+		PlacementRules: nil,
+		Limits:         nil,
+		Preemption:     configs.PartitionPreemptionConfig{},
+		NodeSortPolicy: configs.NodeSortingPolicy{},
+	}
+
+	err = partition.updatePartitionDetails(conf)
```
kingamarton, line 1073: Please check that the error is nil. Also lint will report an issue for this line I think.

kingamarton (changes requested):

pkg/scheduler/partition_test.go
wilfred-s: added the missing assert

kingamarton (approved):
LGTM, +1. I will merge the changes as soon as as we will have the check result as well.

PR #228, opened 2020-12-09
[YUNIKORN-481] deadlock when removing allocation ask

wilfred-s:
When removing an allocationAsk the partition was locked before the
application scheduling itself does it the other way around which can
cause a deadlock if the ask is removed while the application is
scheduled.
Possible deadlock in the node usage calculation removed.

Removal of an allocation contains a possible deadlock which has not been
handled as YUNIKORN-461 is open for changing that behaviour already.

E2E tests run as part of the k8shim cover the testing.


kingamarton (approved):
The changes LGTM. Let's wait for the pre commit checks and then from my side it can be merged.

PR #230, opened 2020-12-16
[YUNIKORN-484] Added waiting state timeout

kingamarton:
Added waiting state timeout. After it times out the app is progressing into the Completed state and is removed from the queue and partition. 
Also a completedApp map is created for storing the completed apps for a while. Right now I set it to 30 days. After it times out, the application will be deleted from this queue as well, and it won't be tracked anymore.
@yangwwei , @wilfred-s  please check this PR when you will have some time, and let me know your thoughts.

There is one use case what is not working properly: during the recovery, the application will be created in the New state, even if it was in Completed one. For fixing this I think we will need some change in the shim side.

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -43,7 +43,7 @@ var (
 	reservationDelay = 2 * time.Second
 	startingTimeout  = 5 * time.Minute
 	waitingTimeout   = 30 * time.Second
-	completedTimeout = 15 * 24 * time.Hour
+	completedTimeout = 3 * 24 * time.Hour
 )
 
 type Application struct {
@@ -149,7 +149,7 @@ func (sa *Application) IsCompleted() bool {
 }
 
 func (sa *Application) IsMarkedForRemoval() bool {
-	return sa.stateMachine.Is(Deleting.String())
+	return sa.stateMachine.Is(Expired.String())
 }
 
 // Handle the state event for the application.
@@ -186,7 +186,7 @@ func (sa *Application) OnStateChange(event *fsm.Event) {
 // Set the starting timer to make sure the application will not get stuck in a starting state too long.
 // This prevents an app from not progressing to Running when it only has 1 allocation.
 // Called when entering the Starting state by the state machine.
-func (sa *Application) SetStateTimer(timeout time.Duration, currentState string, event applicationEvent) {
+func (sa *Application) setStateTimer(timeout time.Duration, currentState string, event applicationEvent) {
 	log.Logger().Debug("Application state timer initiated",
 		zap.String("appID", sa.ApplicationID),
 		zap.String("state", sa.stateMachine.Current()),
@@ -213,7 +213,7 @@ func (sa *Application) timeoutTimer(expectedState string, event applicationEvent
 // Clear the starting timer. If the application has progressed out of the starting state we need to stop the
 // timer and clean up.
 // Called when leaving the Starting state by the state machine.
-func (sa *Application) ClearStateTimer() {
+func (sa *Application) clearStateTimer() {
 	if sa == nil || sa.stateTimer == nil {
 		return
 	}
@@ -977,6 +977,16 @@ func (sa *Application) SetQueue(queue *Queue) {
 	defer sa.Unlock()
 	sa.QueueName = queue.QueuePath
 	sa.queue = queue
+}
+
+// remove the leaf queue the application runs in, used when completing the app
+func (sa *Application) unSetQueue() {
+	if sa.queue != nil {
+		sa.queue.RemoveApplication(sa)
+	}
+	sa.Lock()
+	defer sa.Unlock()
+	sa.queue = nil
 }
 
 // get a copy of all allocations of the application
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -43,6 +43,7 @@ var (
 	reservationDelay = 2 * time.Second
 	startingTimeout  = 5 * time.Minute
 	waitingTimeout   = 30 * time.Second
+	completedTimeout = 30 * 24 * time.Hour
 )
 
 type Application struct {
@@ -140,6 +141,10 @@ func (sa *Application) IsCompleted() bool {
 	return sa.stateMachine.Is(Completed.String())
 }
 
+func (sa *Application) IsMarkedForRemoval() bool {
+	return sa.stateMachine.Is(Deleting.String())
+}
+
 // Handle the state event for the application.
 // The state machine handles the locking.
 func (sa *Application) HandleApplicationEvent(event applicationEvent) error {
@@ -177,12 +182,14 @@ func (sa *Application) OnStateChange(event *fsm.Event) {
 func (sa *Application) SetStateTimer() {
 	log.Logger().Debug("Application state timer initiated",
 		zap.String("appID", sa.ApplicationID),
-		zap.String("state", sa.stateMachine.Current()),
-		zap.Duration("timeout", startingTimeout))
-	if sa.IsWaiting() {
-		sa.stateTimer = time.AfterFunc(waitingTimeout, sa.timeOutWaiting)
-	} else {
+		zap.String("state", sa.stateMachine.Current()))
+	switch {
+	case sa.IsStarting():
 		sa.stateTimer = time.AfterFunc(startingTimeout, sa.timeOutStarting)
+	case sa.IsWaiting():
+		sa.stateTimer = time.AfterFunc(waitingTimeout, sa.timeOutWaiting)
+	case sa.IsCompleted():
+		sa.stateTimer = time.AfterFunc(completedTimeout, sa.timeOutCompleted)
 	}
 }
 
@@ -218,6 +225,17 @@ func (sa *Application) timeOutWaiting() {
 
 		//nolint: errcheck
 		_ = sa.HandleApplicationEvent(CompleteApplication)
+	}
+}
+
+func (sa *Application) timeOutCompleted() {
+	if sa.IsCompleted() {
+		log.Logger().Debug("Application in completed state timed out: marking it for removal ",
+			zap.String("applicationID", sa.ApplicationID),
+			zap.String("state", sa.stateMachine.Current()))
+
+		//nolint: errcheck
+		_ = sa.HandleApplicationEvent(DeleteApplication)
 	}
 }
 
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -53,17 +53,21 @@ type Application struct {
 	SubmissionTime time.Time
 
 	// Private fields need protection
-	queue             *Queue                    // queue the application is running in
-	pending           *resources.Resource       // pending resources from asks for the app
-	reservations      map[string]*reservation   // a map of reservations
-	requests          map[string]*AllocationAsk // a map of asks
-	sortedRequests    []*AllocationAsk
-	user              security.UserGroup     // owner of the application
-	tags              map[string]string      // application tags used in scheduling
-	allocatedResource *resources.Resource    // total allocated resources
-	allocations       map[string]*Allocation // list of all allocations
-	stateMachine      *fsm.FSM               // application state machine
-	stateTimer        *time.Timer            // timer for state time
+	queue                *Queue                    // queue the application is running in
+	pending              *resources.Resource       // pending resources from asks for the app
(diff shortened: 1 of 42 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -42,6 +42,7 @@ import (
 var (
 	reservationDelay = 2 * time.Second
 	startingTimeout  = 5 * time.Minute
+	waitingTimeout   = 5 * time.Minute
```
yangwwei, line 45: 5 minutes is too long, which means a job finishes and the user needs to wait for 5 minutes until they see the job is completed.  I think we should make this more aggressive, e.g 30s.  Otherwise, users will continue to complain about the issue.

pkg/scheduler/objects/application.go
```diff
@@ -200,6 +204,17 @@ func (sa *Application) timeOutStarting() {
 	}
 }
 
+func (sa *Application) timeOutWaiting() {
+	if sa.IsWaiting() {
+		log.Logger().Warn("Application in waiting state timed out: marking it as completed",
+			zap.String("applicationID", sa.ApplicationID),
+			zap.String("state", sa.stateMachine.Current()))
```
yangwwei, line 211: I don't think this is a WARN. We can change this to DEBUG safely

pkg/scheduler/objects/application.go
```diff
@@ -61,7 +62,8 @@ type Application struct {
 	allocatedResource *resources.Resource    // total allocated resources
 	allocations       map[string]*Allocation // list of all allocations
 	stateMachine      *fsm.FSM               // application state machine
-	stateTimer        *time.Timer            // timer for state time
+	startingTimer     *time.Timer            // timer for state time
+	waitingTimer      *time.Timer            // timer for state time
```
yangwwei, line 66: Can we use 1 timer? I like the previous stateTimer better.
since this is a state machine, each time the app must be in one and only one state
we just need to make sure the timer is set properly when entering the state and reset when it exits from the state. we can set different duration for the different state.

yangwwei:
hi @kingamarton  is this still a draft? could you pls update this. you can run some tests with your changes and see if these works end to end. The shim changes should have done in https://issues.apache.org/jira/browse/YUNIKORN-478.

kingamarton:
> hi @kingamarton is this still a draft? could you pls update this. you can run some tests with your changes and see if these works end to end. The shim changes should have done in https://issues.apache.org/jira/browse/YUNIKORN-478.

Hi @yangwwei , yes it is still in progress, because I have to address your review comments, to add test cases, and also implement the cleanup and find a way to keep the completed apps after we remove them from the queue.

yangwwei:
hi @kingamarton  thanks for the update. I think this can work,  but we maybe able to improve this further.
The idea is to abstract the common things into a struct, something like:

We can define a `StateTimers` and `StateTimer` type like the following:
```
type StateTimers map[applicationState]StateTimer
type StateTimer struct {
  timeout time.Duration
  callback func() // executes when it times out
}

func (s *StateTimer) Start() {}
func (s *StateTimer) Stop() {}

```

then we can init the stateTimers in each app like

```
Application{
  ...
  stateTimers : StateTimers{
		Waiting : {
			timeout: 3*time.Second,
			callback: func() {
				// do something
			},
		},
	}
}
```

and then in the state machine, we can simplify the code to something like:

```
// this gets executed in every state transition
fsm.Callbacks{
  "enter_state": func(event *fsm.Event) {
      // blablabla...
      if st, ok := stateTimers[currentState]; ok {
          // a state timer was set
          st.start()
      }
  },
}
```

similarly to the "leave_state" callbacks.
If we can do this, this can be further utilized by other state machines, which gives us a general way to handle state timeouts. Do you think this makes sense to you?

wilfred-s:
We can only ever have one state timer in an application there is no need for multiple state timers as an application can only be in one state at a time. The timer is an app instance variable so we always need to call using the app instance.

The start call is always the same: `app.SetTimer(app.timeoutX, currentState, event)`. The start of the timer will always call the exact same method on the app. The `timeoutX` is directly linked to the app and the state it is in. The other 2 values are fixed per state.
The generic callback that is set with the `time.AfterFunc` will also be the same for all states: `app.TimeoutTimer(currentState, event)`

Leave state will always be the same call. Nothing special happens there: `app.ClearTimer()`
That could be defined as a fsm callback on `leave_state` always just need to make sure it is nil safe etc.

That is a simpler solution than having a map as all state related details are in the ApplicationState and the functions are totally generic defined in the app where they must be defined even for the map based setup.

yangwwei:
> We can only ever have one state timer in an application there is no need for multiple state timers as an application can only be in one state at a time. The timer is an app instance variable so we always need to call using the app instance.

This is a good point. I was saying the same too. We can store a common timer instance in the `StateTimers` struct. And then all the impl details can be hidden in this struct. I was suggesting this way because I think this is 1) easier to read; 2) common impl and can be further re-used by other state machines. However, it is totally up to you and @kingamarton which way to implement this. I should be fine with either approach. Thanks.

wilfred-s (changes requested):
If we just need to export the states etc for test there might be a simpler way to create a call that export the fsm `SetState(string)` on the app that will allow us to do testing easier.

wilfred-s:
> This is a good point. I was saying the same too. We can store a common timer instance in the StateTimers struct. And then all the impl details can be hidden in this struct. I was suggesting this way because I think this is 1) easier to read; 2) common impl and can be further re-used by other state machines. However, it is totally up to you and @kingamarton which way to implement this. I should be fine with either approach. Thanks

That is possible, the struct is a bit more overhead. It needs to be instantiated for each app and stores duplicate info. With large numbers of apps that might be sub optimal.

pkg/scheduler/objects/application.go
kingamarton: changed to 30s

pkg/scheduler/objects/application.go
kingamarton: Done

pkg/scheduler/partition.go
```diff
@@ -919,6 +921,9 @@ func (pc *PartitionContext) GetApplications() []*objects.Application {
 	for _, app := range pc.applications {
 		appList = append(appList, app)
 	}
+	for _, app := range pc.completedApplications {
+		appList = append(appList, app)
+	}
```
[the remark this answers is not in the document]
kingamarton: I spitted it up.

pkg/scheduler/objects/application_state.go
```diff
@@ -61,10 +63,11 @@ const (
 	Rejected
 	Completed
 	killed
+	Deleting
 )
 
 func (as applicationState) String() string {
-	return [...]string{"New", "Accepted", "Starting", "Running", "Waiting", "Rejected", "Completed", "killed"}[as]
+	return [...]string{"New", "Accepted", "Starting", "Running", "Waiting", "Rejected", "Completed", "killed", "Deleting"}[as]
```
yangwwei, line 70: Suggest to rename `Deleting` to `Expired`.
And corresponding event rename to `expireApplication`

pkg/scheduler/partition.go
```diff
@@ -49,6 +49,7 @@ type PartitionContext struct {
 	// Private fields need protection
 	root                   *objects.Queue                  // start of the queue hierarchy
 	applications           map[string]*objects.Application // applications assigned to this partition
+	completedApplications  map[string]*objects.Application // completed applications from this partition
```
yangwwei, line 52: let's remove this map, such overhead seems not to be necessary
because the app cleanup happens in the background, we can iterate the application's map and delete apps that has expired, it won't cause any perf issue

pkg/scheduler/objects/application_state.go
```diff
@@ -134,7 +144,15 @@ func NewAppState() *fsm.FSM {
 			},
 			fmt.Sprintf("enter_%s", Completed.String()): func(event *fsm.Event) {
 				metrics.GetSchedulerMetrics().IncTotalApplicationsCompleted()
+				setTimer(completedTimeout, event, deleteApplication)
```
wilfred-s, line 147: We also need to disconnect the app from the queue and vice versa:
```
app.unSetQueue()
```
And in the app have unSetQueue() as:
```
// remove the leaf queue the application runs in, used when completing the app
func (sa *Application) unSetQueue() {
	sa.queue.RemoveApplication()
	sa.Lock()
	defer sa.Unlock()
	sa.queue = nil
}
```
This should be done in combination with the removal of completedApplications in the PartitionContext

pkg/scheduler/partition.go
```diff
@@ -1112,3 +1123,25 @@ func (pc *PartitionContext) removeAllocationAsk(appID string, allocationKey stri
 		}
 	}
 }
+
+// Move all the completed apps into the completedApp list
+// Delete all the applications marked for removal
+func (pc *PartitionContext) cleanupApps() {
```
wilfred-s, line 1129: This we can now drive from the state change since we remove the second map.

pkg/scheduler/partition.go
```diff
@@ -922,6 +923,16 @@ func (pc *PartitionContext) GetApplications() []*objects.Application {
 	return appList
 }
 
+func (pc *PartitionContext) GetCompletedApplications() []*objects.Application {
+	pc.RLock()
+	defer pc.RUnlock()
+	var appList []*objects.Application
+	for _, app := range pc.completedApplications {
+		appList = append(appList, app)
```
wilfred-s, line 931: this needs to become a filtering loop as we only have one map: we can directly filter on the `Expired` state if I am correct

wilfred-s (changes requested):
after long discussions we have come to the conclusion that we can better support the use case with a single map

pkg/scheduler/objects/application_state.go
kingamarton: Renamed it

yangwwei (commented):
Looks all good except for one minor thing.
@kingamarton  pls check, thanks!

pkg/scheduler/partition.go
```diff
@@ -1208,3 +1219,11 @@ func (pc *PartitionContext) removeAllocationAsk(appID string, allocationKey stri
 		}
 	}
 }
+
+func (pc *PartitionContext) cleanupExpiredApps() {
+	for _, app := range pc.GetAppsByState(objects.Expired.String()) {
+		if app.IsMarkedForRemoval() {
```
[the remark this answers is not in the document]
kingamarton: You are right! I will remove the check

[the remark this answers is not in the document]
yangwwei: hi @kingamarton is this comment addressed? I am still seeing the check here.

yangwwei (changes requested):

pkg/scheduler/objects/application.go
```diff
@@ -172,37 +186,41 @@ func (sa *Application) OnStateChange(event *fsm.Event) {
...
 // Called when entering the Starting state by the state machine.
-func (sa *Application) SetStartingTimer() {
-	log.Logger().Debug("Application Starting state timer initiated",
+func (sa *Application) setStateTimer(timeout time.Duration, currentState string, event applicationEvent) {
+	log.Logger().Debug("Application state timer initiated",
 		zap.String("appID", sa.ApplicationID),
-		zap.Duration("timeout", startingTimeout))
-	sa.stateTimer = time.AfterFunc(startingTimeout, sa.timeOutStarting)
+		zap.String("state", sa.stateMachine.Current()),
+		zap.Any("timeout", timeout))
```
wilfred-s, line 193: Can we use `zap.Duration()` for this log entry?

pkg/scheduler/partition.go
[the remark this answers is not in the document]
wilfred-s: I think we still need to remove that line.

pkg/scheduler/partition.go
```diff
@@ -1208,3 +1219,11 @@ func (pc *PartitionContext) removeAllocationAsk(appID string, allocationKey stri
 		}
 	}
 }
+
+func (pc *PartitionContext) cleanupExpiredApps() {
```
wilfred-s, line 1223: The `PartitionManager` is a different go routine. That means the delete below can cause a race condition.
We either need to lock around the delete or make the function locked. I think speed wise we should just lock around the delete.

wilfred-s (changes requested):
minor remarks beside that the change looks good.

wilfred-s (approved):
LGTM committing

PR #234, opened 2021-01-06
[YUNIKORN-499] race condition in sorting apps

wilfred-s:
While sorting applications the pending resources of an application are
accessed. This could happen while a different routine is updating the
pending resources when a Ask is modified.
Replace the pending resource object instead of cloning on retrieval as it
is faster.


yangwwei (approved):
+1, thanks for the quick fix @wilfred-s

PR #236, opened 2021-01-06
[YUNIKORN-476] placeholder allocation and swap

wilfred-s:
Logic for the placeholder allocation and swap of the placeholder with a
real allocation.
Includes partial unit tests, no smoke tests.
Requires the changes from YUNIKORN-497 from the SI

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -111,13 +111,23 @@ func (cc *ClusterContext) schedule() {
 		if psc.isStopped() {
 			continue
 		}
-		// try reservations first
-		alloc := psc.tryReservedAllocate()
-		// nothing reserved that can be allocated try normal allocate
+		// placeholder replacement first
+		alloc := psc.tryPlaceholderAllocate()
+		// try reservations second
 		if alloc == nil {
-			alloc = psc.tryAllocate()
+			alloc = psc.tryReservedAllocate()
+			// nothing reserved that can be allocated try normal allocate
+			if alloc == nil {
+				alloc = psc.tryAllocate()
+			}
 		}
 		if alloc != nil {
+			if alloc.Result == objects.Replaced {
+				// communicate the removal to the RM
+				cc.notifyRMAllocationReleased(psc.RmID, alloc.Releases, si.AllocationRelease_PLACEHOLDER_REPLACED, "replacing UUID: "+alloc.UUID)
+				// only communicate the release rest happens on shim callback
+				continue
+			}
 			// TODO: The alloc is passed to the RM twice why do we need event + callback?
 			// See YUNIKORN-462, there are two separate communications for the same allocation
 			// between the core and the shim they should be merged into one communication.
@@ -448,6 +458,7 @@ func (cc *ClusterContext) processApplications(request *si.UpdateRequest) {
 			continue
 		}
 		// convert and resolve the user: cache can be set per partition
+		// need to do this before we create the application
 		ugi, err := partition.convertUGI(app.Ugi)
 		if err != nil {
 			rejectedApps = append(rejectedApps, &si.RejectedApplication{
@@ -457,7 +468,7 @@ func (cc *ClusterContext) processApplications(request *si.UpdateRequest) {
 			continue
 		}
 		// create a new app object and add it to the partition (partition logs details)
-		schedApp := objects.NewApplication(app.ApplicationID, app.PartitionName, app.QueueName, ugi, app.Tags, cc.rmEventHandler, request.RmID)
+		schedApp := objects.NewApplication(app, ugi, cc.rmEventHandler, request.RmID)
 		if err = partition.AddApplication(schedApp); err != nil {
 			rejectedApps = append(rejectedApps, &si.RejectedApplication{
 				ApplicationID: app.ApplicationID,
@@ -694,7 +705,8 @@ func (cc *ClusterContext) processAllocationReleases(releases []*si.AllocationRel
 	for _, toRelease := range releases {
 		partition := cc.GetPartition(toRelease.PartitionName)
 		if partition != nil {
-			allocs := partition.removeAllocation(toRelease.ApplicationID, toRelease.UUID)
+
+			allocs := partition.removeAllocation(toRelease)
 			// notify the RM of the exact released allocations
 			if len(allocs) > 0 {
 				cc.notifyRMAllocationReleased(rmID, allocs, si.AllocationRelease_STOPPED_BY_RM, "allocation remove as per RM request")
@@ -745,6 +757,8 @@ func (cc *ClusterContext) notifyRMAllocationReleased(rmID string, released []*ob
 	}
 	for _, alloc := range released {
 		releaseEvent.ReleasedAllocations = append(releaseEvent.ReleasedAllocations, &si.AllocationRelease{
+			ApplicationID:   alloc.ApplicationID,
+			PartitionName:   alloc.PartitionName,
 			UUID:            alloc.UUID,
 			TerminationType: terminationType,
 			Message:         message,
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -125,34 +125,8 @@ func (cc *ClusterContext) schedule() {
 			if alloc.Result == objects.Replaced {
 				// communicate the removal to the RM
 				cc.notifyRMAllocationReleased(psc.RmID, alloc.Releases, si.AllocationRelease_PLACEHOLDER_REPLACED, "replacing UUID: "+alloc.UUID)
-				// only communicate the release rest happens on shim callback
-				continue
-			}
-			// TODO: The alloc is passed to the RM twice why do we need event + callback?
-			// See YUNIKORN-462, there are two separate communications for the same allocation
-			// between the core and the shim they should be merged into one communication.
-
-			// communicate the allocation to the RM
-			cc.rmEventHandler.HandleEvent(&rmevent.RMNewAllocationsEvent{
-				Allocations: []*si.Allocation{alloc.NewSIFromAllocation()},
-				RmID:        psc.RmID,
-			})
-			// if reconcile plugin is enabled, re-sync the cache now.
-			// before deciding on an allocation, call the reconcile plugin to sync scheduler cache
-			// between core and shim if necessary. This is useful when running multiple allocations
-			// in parallel and need to handle inter container affinity and anti-affinity.
-			if rp := plugins.GetReconcilePlugin(); rp != nil {
-				if err := rp.ReSyncSchedulerCache(&si.ReSyncSchedulerCacheArgs{
-					AssumedAllocations: []*si.AssumedAllocation{
-						{
-							AllocationKey: alloc.AllocationKey,
-							NodeID:        alloc.NodeID,
-						},
-					},
-				}); err != nil {
-					log.Logger().Error("failed to sync shim",
-						zap.Error(err))
-				}
+			} else {
+				cc.notifyRMNewAllocation(psc.RmID, alloc)
 			}
 		}
 	}
@@ -706,7 +680,7 @@ func (cc *ClusterContext) processAllocationReleases(releases []*si.AllocationRel
 		partition := cc.GetPartition(toRelease.PartitionName)
 		if partition != nil {
 
-			allocs := partition.removeAllocation(toRelease)
+			allocs, confirmed := partition.removeAllocation(toRelease)
 			// notify the RM of the exact released allocations
 			if len(allocs) > 0 {
 				cc.notifyRMAllocationReleased(rmID, allocs, si.AllocationRelease_STOPPED_BY_RM, "allocation remove as per RM request")
@@ -715,6 +689,10 @@ func (cc *ClusterContext) processAllocationReleases(releases []*si.AllocationRel
 				toReleaseAllocations = append(toReleaseAllocations, &si.ForgotAllocation{
 					AllocationKey: alloc.AllocationKey,
 				})
+			}
+			// notify the RM of the confirmed allocations (placeholder swap & preemption)
+			if confirmed != nil {
+				cc.notifyRMNewAllocation(rmID, confirmed)
 			}
 		}
 	}
@@ -731,7 +709,7 @@ func (cc *ClusterContext) processAllocationReleases(releases []*si.AllocationRel
 				ForgetAllocations: toReleaseAllocations,
 			})
 			if err != nil {
-				log.Logger().Error("failed to sync shim",
+				log.Logger().Error("failed to sync shim on allocation release",
 					zap.Error(err))
 			}
 		}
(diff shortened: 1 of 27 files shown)


wilfred-s:
I want to add further tests after the weekend. The code coverage should not drop specially as this is a new often executed  allocation code.
With the latest events change we get the full gang working based on the offline tests wih @yangwwei

yangwwei:
hi @TaoYang526  thanks for the draft PR. At the first glance, I wasn't able to fully understand some of the places. After reading more info from the previous design doc in https://issues.apache.org/jira/browse/YUNIKORN-1. They start to make sense. I pretty like interfaces defined in this PR, the confusion part is `queue_request_manager_plugin.go`.  I feel we can somehow optimize this to make it simpler. I will look into more details and get back to you. Thanks.

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,121 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+			// mark placeholder as released
+			phFit.released = true
+			// update just the node to make sure we keep its spot
+			// no queue update as we're releasing the placeholder and are just temp over the size
+			if !node.AddAllocation(alloc) {
+				log.Logger().Debug("Node update failed unexpectedly",
+					zap.String("applicationID", sa.ApplicationID),
+					zap.String("ask", reqFit.String()),
+					zap.String("placeholder", phFit.String()))
+				return nil
```
yangwwei, line 734: We do not need to wait for a placeholder to release before updating the node resources?
This seems diverse from the normal processing logic. 
If we can merge this part to the normal processing code, can we do something like the following:

1. Placeholder allocate fails (doesn't fit)
2. Mark a placeholder to release, "park" the request
3. Wait until the placeholder is released in the shim, do normal allocation for the request

This means: the placeholder wasn't good, so we try to schedule the pod as before.

pkg/scheduler/partition.go
```diff
@@ -320,6 +320,22 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
+		// retrieve the max set
+		if maxQueue := queue.GetMaxQueueSet(); maxQueue != nil {
+			if !resources.FitIn(maxQueue, placeHolder) {
+				return fmt.Errorf("queue %s cannot fit application %s: task group request %s larger than max queue allocation", queueName, appID, placeHolder.String())
+			}
+		}
+		// check the queue sorting
+		if !queue.SupportTaskGroup() {
+			return fmt.Errorf("queue %s cannot run application %s with task group request: unsupported sort type", queueName, appID)
+		}
```
yangwwei, line 336: Thanks for adding this check!

pkg/scheduler/objects/application.go
yangwwei: > We do not need to wait for a placeholder to release before updating the node resources?
> This seems diverse from the normal processing logic.
> If we can merge this part to the normal processing code, can we do something like the following:
> 
> 1. Placeholder allocate fails (doesn't fit)
> 2. Mark a placeholder to release, "park" the request
> 3. Wait until the placeholder is released in the shim, do normal allocation for the request
> 
> This means: the placeholder wasn't good, so we try to schedule the pod as before.

Just add to this. One concern I have is we may go over queue resource quota with this.
Imagine a case we have 10 placeholders and 10 real pods for an app, at the beginning we have 10 placeholders allocated, where we have 

10 placeholders 1GB * 10 = 10GB
  - app: 10GB in placeholders
  - nodes: 10GB in allocated resources
  - queue: 10GB allocated resources

imagine the worst-case scenario, the placeholders don't fit, and we start to fallback allocating the real pods on other nodes. Then we get

  - app: 10GB in placeholders, 10GB in allocated resources
  - nodes: 20GB
  - queue: 10GB (* dangerous)

*dangerous: both app and nodes are tracking 20GB used resources for this app, but for the queue, it is 10GB. If the queue has a resource quota set to 20GB. It might be possible that before shim releases all 10 placeholders, the core will schedule other pods from other apps, that will goes over the quota defined for this queue.

wilfred-s: That would break the gang. If the placeholder is allocated the queue usage is reserved for the gang. When we release the placeholder there is no guarantee that another application in the queue is not going to take that space. This could be an older application or even a newer application.
It would in (almost) all cases directly break gang scheduling

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,121 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+						zap.Error(err))
+				}
+				return alloc
+			}
+			iterator.Reset()
+		}
+	}
+	// we checked all placeholders and asks nothing worked as yet
+	// pick the first fit and try all nodes if that fails give up
+	if phFit != nil && reqFit != nil {
```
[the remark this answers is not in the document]
wilfred-s: We cannot reuse that logic as it does not allow us to link back to the replacement. We also need to be careful with the release if no allocation is made. The queue usage would drop if we release without allocating and the resource could be used by another application in the queue. That would break the gang as it might not be possible o get that back.

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,121 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+	// get all the requests from the app sorted in order
+	for _, request := range sa.sortedRequests {
+		// skip placeholders they follow standard allocation
+		// this should also be part of a task group just make sure it is
+		if request.placeholder || request.taskGroupName == "" {
+			continue
+		}
+		// walk over the placeholders, allow for processing all as we can have multiple task groups
+		phAllocs := sa.getPlaceholderAllocations()
+		for _, ph := range phAllocs {
```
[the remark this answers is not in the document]
wilfred-s: Yes we do, we do not know if the placeholder that we get is for the specific TaskGroup. Since we can have multiple task groups on each application we need to find one that matches.
There is no reason to assume that if one placeholder fails they all fail. The expected cause for failures to allocate the real pod would be that the node predicates fail. This could be a temporary issue or just for that one specific node. Assuming the same state for all nodes is wrong. In the case that all placeholders have issues there is something strange going on.
We can also not run the normal scheduling cycle (see next review comment).

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,121 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+			var ok bool
+			var node *Node
+			for iterator.HasNext() {
+				node, ok = iterator.Next().(*Node)
+				if !ok {
+					log.Logger().Warn("Node iterator failed to return a node")
+					return nil
+				}
+				if node.NodeID == ph.NodeID {
+					break
```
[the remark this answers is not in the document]
wilfred-s: This issue resolves itself with the signature change.

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,121 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+			// we could have already released this placeholder and are waiting for the shim to confirm
+			// and check that we have the correct task group before trying to swap
+			if ph.released || request.taskGroupName != ph.taskGroupName {
+				continue
+			}
+			if phFit == nil && reqFit == nil {
+				phFit = ph
+				reqFit = request
+			}
+			// this is inefficient but needs to do for now as we only have the ID not the handle on the node itself
```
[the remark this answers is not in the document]
wilfred-s: A new TODO will cause the lint check to fail and thus the build to be marked as failed. That is why I did not add it.
The solution is rather simple for this issue and would require a change of the function signature to:
```
TryPlaceholderAllocate(nodeIterator func() interfaces.NodeIterator, node func() *objects.Node)
```
that will allow us to just retrieve the node and remove a lot of the iteration. I will push a patch for that.

pkg/scheduler/partition.go
```diff
@@ -704,6 +720,57 @@ func (pc *PartitionContext) tryReservedAllocate() *objects.Allocation {
...
+		return pc.replace(alloc)
+	}
+	return nil
+}
+
+// Process the allocation and make the left over changes in the partition.
+func (pc *PartitionContext) replace(alloc *objects.Allocation) *objects.Allocation {
+	pc.Lock()
+	defer pc.Unlock()
+	// partition is locked nothing can change from now on
```
kingamarton, line 742: I think this comment line is redundant, since from lines 740 and 741, we already know that the partition is locked and that it cannot change

pkg/scheduler/partition.go
```diff
@@ -704,6 +720,57 @@ func (pc *PartitionContext) tryReservedAllocate() *objects.Allocation {
...
+			allocationUUID := common.GetNewUUID()
+			log.Logger().Warn("UUID clash, random generator might be lacking entropy",
+				zap.String("uuid", alloc.UUID),
+				zap.String("new UUID", allocationUUID))
+			if pc.allocations[allocationUUID] == nil {
+				alloc.UUID = allocationUUID
+				break
+			}
+		}
+	}
```
kingamarton, line 764: This is exactly the same that we have in the lines 822-835. Please extract this into a new function and reuse it.

kingamarton (changes requested):

pkg/scheduler/objects/application.go
yangwwei: hi @wilfred-s even with the approach in this PR, gang scheduling is not 100% guaranteed. If we can't utilize the placeholders for the real pods and schedule them on some other nodes. We do not know how many of them can be allocated successfully and how many will keep pending. It still breaks the gang scheduling semantics.

The case I described above is pretty bad. The scheduler data should always be consistent, few dimensions are: used resources by all apps in a queue should equal to used resources of this queue. This is a more serious problem than breaking gang scheduling in the worst-case scenario.

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,121 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+			// skip the node if conditions can not be satisfied
+			if !node.preAllocateConditions(reqFit.AllocationKey) {
+				continue
+			}
+			// allocation worked: on a non placeholder node update result and return
+			alloc := NewAllocation(common.GetNewUUID(), node.NodeID, reqFit)
+			// double link to make it easier to find
+			alloc.Releases = []*Allocation{phFit}
+			phFit.Releases = []*Allocation{alloc}
+			alloc.Result = Replaced
```
yangwwei, line 724: This doesn't seem to be correct.
if placeholder is on node `A`, and the alloc is made on node `B`
here we send the shim to release the placeholder, which is on `A`
once shim confirms that the placeholder has been released, in `partition#removeAllocation()`, it replaces the allocation with placeholder allocation, which is on `B`.

yangwwei: Already got the answer from @wilfred-s , closing this one.

pkg/scheduler/objects/application.go
wilfred-s: Unless you can in some way guarantee that a node cannot change over time there is no way that you can get that 100% guarantee for gang scheduling. This is just not possible, we're close to it but not there. What this approach tries to do is making sure that the exception case is a small as possible and handled as well as possible. Preventing breakage where possible.

An early release of a placeholder from the queue's total use will break it at the point that we have told the application that the gang is reserved. In the case that there is a limit on the queue it could even mean that we cannot get that quota back as an other application in the queue has started using it. It might thus cause such a problem that the gang can never be fully allocated.
The only other option is to reserve the node so we do not lose the node. Then on the release confirmation from the placeholder finish the real allocation. The release would be processed on the node and in the app. We would need to push the real allocation through a special allocation cycle to update the "new" node. Then update the app with the final result.
We would also somehow make sure that the reservation is removed from the normal `TryReserveAllocate` cycle adding another complexity. The node that got reserved will be unavailable for scheduling all the time the release is being processed. That also does not feel correct and really complex.

For the temporary discrepancy: it *only* happens if all the placeholder run on nodes that can not for some reason support the real allocation (it is the failure case). It lasts only for the time it takes K8s to release the placeholder and it is all back in sync again.

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,121 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+					log.Logger().Warn("Node iterator failed to return a node")
+					return nil
+				}
+				if node.NodeID == ph.NodeID {
+					break
+				}
+			}
+			// got the node run same checks as for reservation (all but fits)
+			// resource usage should not change anyway between placeholder and real one
+			if node != nil && node.preReserveConditions(request.AllocationKey) {
```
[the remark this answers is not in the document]
wilfred-s: Filed a follow up jira for that, not sure where it should happen as we can do that in either the core or the shim

pkg/scheduler/objects/application.go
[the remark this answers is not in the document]
wilfred-s: We iterate over all placeholders as not all placeholders might belong to the same TaskGroup and thus cannot be used. We must find the first placeholder that fits the taskgroup.

We then retain the first fit on purpose. If we cannot replace any of the placeholder with any of the requests outstanding we try to allocate that one. It is not an arbitrary choice but it is the first request, based on the request sorting, that still has a placeholder that can be replaced. It is the request that we should allocate first before anything else.

pkg/scheduler/partition.go
wilfred-s: refactored into a new `checkUUID()` method

pkg/scheduler/objects/application.go
wilfred-s: comments updated in two locations to clear this up: it is doing the right thing release on node A and adds on node B

pkg/scheduler/context.go
```diff
@@ -111,38 +111,22 @@ func (cc *ClusterContext) schedule() {
 		if psc.isStopped() {
 			continue
 		}
-		// try reservations first
-		alloc := psc.tryReservedAllocate()
-		// nothing reserved that can be allocated try normal allocate
+		// placeholder replacement first
+		alloc := psc.tryPlaceholderAllocate()
```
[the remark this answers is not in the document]
wilfred-s: swapped the logic: first reserve, then placeholder and normal last

pkg/scheduler/partition.go
wilfred-s: removed

pkg/scheduler/objects/application.go
```diff
@@ -604,6 +630,112 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+	// we checked all placeholders and asks nothing worked as yet
+	// pick the first fit and try all nodes if that fails give up
+	if phFit != nil && reqFit != nil {
+		for iterator.HasNext() {
+			node, ok := iterator.Next().(*Node)
+			if !ok {
+				log.Logger().Warn("Node iterator failed to return a node")
+				return nil
+			}
+			if err := node.preAllocateCheck(reqFit.AllocatedResource, reservationKey(nil, sa, reqFit), false); err != nil {
```
kingamarton, line 701: Most of the steps we are performing in the lines 701 - 734 is the same as we are doing in the `tryNodes` part. Can you please export and reuse the common things if possible?

yangwwei:
I had a long discussion with @wilfred-s about the main concern I have about this PR.
Like we discussed [in this comment](https://github.com/apache/incubator-yunikorn-core/pull/236#discussion_r554867952) , it is not likely something that can be easily fixed without causing the code to be more complex. I've already done some basic tests, the current code works. Our conclusion is to get this merged, run more tests (including the negative scenarios), revisit this if we find any issue. For the review stage, we don't need to track ourselves here.
So I am +1 on the change and let's get this merged.
The tests are passing, the code coverage can be improved in a follow-up JIRA. Also @kingamarton please file a JIRA to address the latest comment you added, they make sense.

PR #238, opened 2021-01-22
YUNIKORN-468: Node resource updates should trigger cleanup

manirajv06:
@wilfred-s WIP patch with lot of "todo" comments to confirm the overall idea/direction. Can you please take a quick look?


wilfred-s:
I think we need to first sort out how and what we are going to do here.
Lets start the discussion on the jira so we can figure out:
- what we want the behaviour to be
- how does this relate to occupied resources of the node (non yunikorn scheduled ones)
- what we want to make configurable
- how that interacts with apps and queues and the rest of preemption

At the moment this is not a high priority as K8s can not change a node's size as far as I know (i.e. nodes are not fixed in mem etc after creation). Before we thus dive in lets figure out what we want/need.

craigcondit:
Closing as abandoned.

PR #242, opened 2021-01-28
[YUNIKORN-461] remove allocations list from partition

wilfred-s:
The allocations map in the partition is not needed anymore after the
cache removal. Updating and keeping the map in sync with the other
objects in the scheduler can cause a deadlock.

Remove the map and replace it with a simple counter for the web UI.
Simplify locking where possible to remove partition locks.

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -28,7 +28,6 @@ import (
 	"github.com/looplab/fsm"
 	"go.uber.org/zap"
 
-	"github.com/apache/incubator-yunikorn-core/pkg/common"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/configs"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/resources"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/security"
@@ -51,7 +50,6 @@ type PartitionContext struct {
 	applications           map[string]*objects.Application // applications assigned to this partition
 	reservedApps           map[string]int                  // applications reserved within this partition, with reservation count
 	nodes                  map[string]*objects.Node        // nodes assigned to this partition
-	allocations            map[string]*objects.Allocation  // allocations
 	placementManager       *placement.AppPlacementManager  // placement manager for this partition
 	partitionManager       *partitionManager               // manager for this partition
 	stateMachine           *fsm.FSM                        // the state of the partition for scheduling
@@ -61,6 +59,7 @@ type PartitionContext struct {
 	userGroupCache         *security.UserGroupCache        // user cache per partition
 	totalPartitionResource *resources.Resource             // Total node resources
 	nodeSortingPolicy      *policies.NodeSortingPolicy     // Global Node Sorting Policies
+	allocations            int                             // Number of allocations on the partition
 
 	sync.RWMutex
 }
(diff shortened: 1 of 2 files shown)


pkg/scheduler/partition.go
```diff
@@ -1051,7 +1022,9 @@ func (pc *PartitionContext) addAllocation(alloc *objects.Allocation) error {
 	node.AddAllocation(alloc)
 	app.RecoverAllocationAsk(alloc.Ask)
 	app.AddAllocation(alloc)
-	pc.allocations[alloc.UUID] = alloc
+
+	// track the number of allocations
+	pc.allocations += 1
```
manirajv06, line 1027: Any reasons for not using updateAllocationCount()? Similarly, in other places too.

manirajv06, line 1027: Any reasons for not using updateAllocationCount()? Similarly, in other places too.

pkg/scheduler/partition.go
```diff
@@ -61,6 +59,7 @@ type PartitionContext struct {
 	userGroupCache         *security.UserGroupCache        // user cache per partition
 	totalPartitionResource *resources.Resource             // Total node resources
 	nodeSortingPolicy      *policies.NodeSortingPolicy     // Global Node Sorting Policies
+	allocations            int                             // Number of allocations on the partition
```
yangwwei, line 62: why we still need to track the number of allocations?
if this is just for UT, I think we should remove them.

yangwwei (commented):
+1 to remove the allocations map

pkg/scheduler/partition.go
wilfred-s: Yes, `updateAllocationCount()` is a locked call. The point we are at already has a lock on the partition. We cannot lock and introducing a function to update that does not take a lock is extra overhead we do not need.
That is the case for all points we do a direct update on the counter.

pkg/scheduler/partition.go
wilfred-s: No the UI uses this and shows this as number when it gets the partition overview.

yangwwei: Is it better to re-implement `GetTotalAllocationCount()`, and simply loop all the apps and sum allocations of all apps?
So we don't need to maintain a separate `allocations` count here.

wilfred-s: That would require us to grab a copy of the apps (i.e. lock the partition). Then we need to lock each app when we get the allocation count from the app. Really heavy and not scalable as soon as we go above a small number of apps and allocations per app.

yangwwei: I don't think that's gonna be a problem for perf. It only needs the app's read lock to get a count. And this is only called via REST API. But if we do that, the code will be simpler, and fewer objects to maintain (worry about keeping them inconsistent all the time). And note, update the count here is not a lock-free call, it holds an even bigger partition lock.

yangwwei (approved):
The rest of the changes look good to me. The only thing is do we really need to maintain allocations number in the partition. I will open a follow up JIRA to discuss. That is less important. +1 to get this merged.

pkg/scheduler/partition.go
yangwwei: Created https://issues.apache.org/jira/browse/YUNIKORN-526 to follow up on this.

PR #243, opened 2021-01-31
[YUNIKORN-528] Nil pointer exception while receiving releasing same allocation request more than once

yangwwei:
the root cause is when the shim deletes a placeholder, it can trigger 2 events sometimes,
- Pod Update
- Pod Delete
When a pod updated to TERMINATED state and when a pod gets DELETED, the shim will send a release request to the core. But when there is a second release request, as the previous one already removed the allocation, then we are hitting the Nil pointer.

To fix this, we need to fix the both core and shim side to avoid NPE. This is the PR for the core side, we need to add a nil pointer check while replacing an allocation.


wilfred-s (approved):
LGTM, the real fix is on the shim side in YUNIKORN-229.

PR #248, opened 2021-02-07
[YUNIKORN-541] Unexpected WARN log in node.go#refreshAvailableResource

TaoYang526:


wilfred-s (approved):
LGTM
I thought we had already fixed this one earlier but I must be wrong

PR #249, opened 2021-02-08
[YUNIKORN-542] App submission time in nano seconds

wilfred-s:
Times in the rest interface should be in nano seconds. The application
submission time is converted into seconds causing display issues on
the web UI.


yangwwei (approved):
LGTM

PR #251, opened 2021-02-18
[YUNIKORN-505] placeholder allocate tests

wilfred-s:
Additional tests for placeholder allocation. Introducing a new plugin
for mocking the predicate plugin.
Miscellaneous Logging fixes to help troubleshoot failures and progress of
allocations.

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -424,11 +424,13 @@ func (cc *ClusterContext) processApplications(request *si.UpdateRequest) {
 		partition := cc.GetPartition(app.PartitionName)
 		if partition == nil {
 			msg := fmt.Sprintf("Failed to add application %s to partition %s, partition doesn't exist", app.ApplicationID, app.PartitionName)
-			log.Logger().Info(msg)
 			rejectedApps = append(rejectedApps, &si.RejectedApplication{
 				ApplicationID: app.ApplicationID,
 				Reason:        msg,
 			})
+			log.Logger().Info("Failed to add application to non existing partition",
+				zap.String("applicationID", app.ApplicationID),
+				zap.String("partitionName", app.PartitionName))
 			continue
 		}
 		// convert and resolve the user: cache can be set per partition
@@ -439,6 +441,10 @@ func (cc *ClusterContext) processApplications(request *si.UpdateRequest) {
 				ApplicationID: app.ApplicationID,
 				Reason:        err.Error(),
 			})
+			log.Logger().Info("Failed to add application to partition (user rejected)",
+				zap.String("applicationID", app.ApplicationID),
+				zap.String("partitionName", app.PartitionName),
+				zap.Error(err))
 			continue
 		}
 		// create a new app object and add it to the partition (partition logs details)
@@ -448,11 +454,20 @@ func (cc *ClusterContext) processApplications(request *si.UpdateRequest) {
 				ApplicationID: app.ApplicationID,
 				Reason:        err.Error(),
 			})
+			log.Logger().Info("Failed to add application to partition (placement rejected)",
+				zap.String("applicationID", app.ApplicationID),
+				zap.String("partitionName", app.PartitionName),
+				zap.Error(err))
 			continue
 		}
 		acceptedApps = append(acceptedApps, &si.AcceptedApplication{
 			ApplicationID: schedApp.ApplicationID,
 		})
+		log.Logger().Info("Added application to partition",
+			zap.String("applicationID", app.ApplicationID),
+			zap.String("partitionName", app.PartitionName),
+			zap.String("requested queue", app.QueueName),
+			zap.String("placed queue", schedApp.GetQueueName()))
 	}
 
 	// Respond to RMProxy with accepted and rejected apps if needed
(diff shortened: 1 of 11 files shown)


pkg/scheduler/partition.go
```diff
@@ -380,9 +379,6 @@ func (pc *PartitionContext) removeApplication(appID string) []*objects.Allocatio
 			}
 		}
 	}
-	log.Logger().Debug("application removed from the scheduler",
-		zap.String("queue", queueName),
-		zap.String("applicationID", appID))
```
kingamarton, line 385: I think this log message can be useful. Right now we have no logs when we remove an application

pkg/scheduler/partition.go
```diff
@@ -585,10 +580,6 @@ func (pc *PartitionContext) AddNode(node *objects.Node, existingAllocations []*o
 
 	// Node is added update the metrics
 	metrics.GetSchedulerMetrics().IncActiveNodes()
-	log.Logger().Info("added node to partition",
-		zap.String("partitionName", pc.Name),
-		zap.String("nodeID", node.NodeID),
-		zap.String("partitionResource", pc.totalPartitionResource.String()))
```
kingamarton, line 591: This one can be helpful as well, I think. Maybe at debug level is enough, but I don't think we should remove it.

pkg/scheduler/partition_test.go
```diff
@@ -1221,3 +1222,237 @@ func TestUpdateNode(t *testing.T) {
...
+	err = partition.AddApplication(app)
+	if err == nil {
+		t.Error("app-1 should be rejected due to TG request")
+	}
+
+	limit = map[string]string{"first": "100"}
+	partition, err = newLimitedPartition(limit)
+	assert.NilError(t, err, "partition create failed")
+	err = partition.AddApplication(app)
+	assert.NilError(t, err, "app-1 should have been added to the partition")
```
kingamarton, line 1244: Besides checking for error please check if the application was added to the application list as well.

kingamarton (changes requested):

pkg/scheduler/partition.go
wilfred-s: this has been moved to an INFO level message in the partition to standardise the way we log.

pkg/scheduler/partition.go
wilfred-s: We have more logs now, as part of the change we log this in the context at an INFO level to make sure we log all possible failure cases [moved entry](https://github.com/apache/incubator-yunikorn-core/pull/251/files#diff-a6bde2f6900fb05707fa68db3365f08a79642f39ea629ad5b489d05d3cacc7b3R497-R499)

pkg/scheduler/context.go
```diff
@@ -626,27 +656,18 @@ func (cc *ClusterContext) processAsks(request *si.UpdateRequest) {
 			continue
 		}
 
-		// if app info doesn't exist, reject the request
-		app := partition.getApplication(siAsk.ApplicationID)
-		if app == nil {
-			msg := fmt.Sprintf("Failed to find application %s, for allocation %s", siAsk.ApplicationID, siAsk.AllocationKey)
-			log.Logger().Info(msg)
-			rejectedAsks = append(rejectedAsks,
```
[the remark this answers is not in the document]
wilfred-s: This has all been folded into one call. Instead of first checking the app on the partition and then calling the app directly we call the add on the partition which calls the add on the app. This brings it inline with the other checks we do and the siAsk is the loop variable and has not changed. The error still exists and still gets [returned](https://github.com/apache/incubator-yunikorn-core/pull/251/files)

pkg/scheduler/partition_test.go
wilfred-s: done

pkg/scheduler/partition.go
```diff
@@ -1185,18 +1180,29 @@ func (pc *PartitionContext) removeAllocation(release *si.AllocationRelease) ([]*
...
+}
+
+// Remove the allocation Ask from the specified application
+// NOTE: this is a lock free call. It must NOT be called holding the PartitionContext lock.
+func (pc *PartitionContext) addAllocationAsk(siAsk *si.AllocationAsk) error {
+	app := pc.getApplication(siAsk.ApplicationID)
+	if app == nil {
+		return fmt.Errorf("failed to find application %s, for allocation ask %s", siAsk.ApplicationID, siAsk.AllocationKey)
 	}
+	// remove the allocation asks from the app
```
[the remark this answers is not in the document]
wilfred-s: done

pkg/scheduler/partition.go
```diff
@@ -1185,18 +1180,29 @@ func (pc *PartitionContext) removeAllocation(release *si.AllocationRelease) ([]*
...
+	}
+	// remove the allocation asks from the app
+	reservedAsks := app.RemoveAllocationAsk(allocationKey)
+	// update the partition if the asks were reserved (clean up)
+	if reservedAsks != 0 {
+		pc.unReserveCount(appID, reservedAsks)
+	}
+}
+
+// Remove the allocation Ask from the specified application
```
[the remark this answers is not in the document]
wilfred-s: done

yangwwei (approved):
Looks good. @wilfred-s could u pls get this one merged? Thanks!

PR #252, opened 2021-02-24
[YUNIKORN-548] Allow AppID reuse after the app is completed

kingamarton:
The completed apps are moved into a separate map

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -69,9 +69,9 @@ type Application struct {
 	execTimeout          time.Duration          // execTimeout for the application run
 	// appTimer             *time.Timer            // application run timer
 
-	rmEventHandler handler.EventHandler
-	rmID           string
-	callback       func(appID string)
+	rmEventHandler    handler.EventHandler
+	rmID              string
+	completedCallback func(appID string)
 
 	sync.RWMutex
 }
@@ -1160,11 +1160,11 @@ func (sa *Application) GetTag(tag string) string {
 func (sa *Application) SetCompletedCallback(callback func(appID string)) {
 	sa.Lock()
 	defer sa.Unlock()
-	sa.callback = callback
+	sa.completedCallback = callback
 }
 
-func (sa *Application) completedCallback() {
-	if sa.callback != nil {
-		go sa.callback(sa.ApplicationID)
+func (sa *Application) executeCompletedCallback() {
+	if sa.completedCallback != nil {
+		go sa.completedCallback(sa.ApplicationID)
 	}
 }
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -71,6 +71,7 @@ type Application struct {
 
 	rmEventHandler handler.EventHandler
 	rmID           string
+	callback       func(appID string)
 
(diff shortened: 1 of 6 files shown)


kingamarton:
First I wanted to append the date and time to the app ID when it was completed, but it is causing a lot of troubles around locking, so I removed it. We have the submission time, so we can differentiate the apps with the same ID according to their submission time.

pkg/scheduler/objects/application.go
```diff
@@ -71,6 +71,7 @@ type Application struct {
 
 	rmEventHandler handler.EventHandler
 	rmID           string
+	callback       func(appID string)
```
yangwwei, line 74: Need a more explicit callback name for this function, easier for people to understand what this is for

yangwwei (commented):
Looks good. Just add one small comment, and also please add some UT coverage for this callback mechanism. Thanks

pkg/scheduler/partition.go
```diff
@@ -78,6 +79,7 @@ func newPartitionContext(conf configs.PartitionConfig, rmID string, cc *ClusterC
 		stateMachine: objects.NewObjectState(),
 		stateTime:    time.Now(),
 		applications: make(map[string]*objects.Application),
+		completedApplications: make(map[string]*objects.Application),
```
wilfred-s, line 82: please run `gofmt` over the file

pkg/scheduler/partition.go
```diff
@@ -967,6 +980,11 @@ func (pc *PartitionContext) GetAppsByState(state string) []*objects.Application
 			appList = append(appList, app)
 		}
 	}
+	for _, app := range pc.completedApplications {
+		if app.CurrentState() == state {
+			appList = append(appList, app)
```
wilfred-s, line 985: We should not range over both lists. If the state is _completed_ we range over the `completedApplications` map in all other cases we range over the `applications` map.

pkg/scheduler/partition.go
```diff
@@ -1222,3 +1240,12 @@ func (pc *PartitionContext) GetNodeSortingPolicy() policies.SortingPolicy {
 	defer pc.RUnlock()
 	return pc.nodeSortingPolicy.PolicyType
 }
+
+func (pc *PartitionContext) moveCompletedApp(appID string) {
+	pc.Lock()
+	defer pc.Unlock()
+	if app, ok := pc.applications[appID]; ok {
+		delete(pc.applications, appID)
+		pc.completedApplications[appID] = app
```
wilfred-s, line 1249: We need to be careful: there is no reason to assume that the same application ID is not re-used multiple times in the time we keep the old application around. A simple time stamp addition will allow us to re-use the appID multiple times:
```
// new ID as completedApplications map key, use negative value to get a divider
newID := appID + strconv.FormatInt(-(time.Now()).Unix(), 10)
```
The `newID` is then only used as the key in the _completedApplications_ map. No updates in the application object.

wilfred-s (changes requested):
we should rename the `callback` variable to something linked to the completed state like `completedCallback` to make it really clear what it does like Weiwei proposed

wilfred-s (approved):
Changes are looking good, will commit when the build returns a pass

PR #253, opened 2021-03-02
[YUNIKORN-551] Deadlock when removing node

wilfred-s:
When a node gets removed the node alocations get removed as part of the
cleanup phase. The node removal acquires the partition lock and does not
release it until all actions are done. This could cause a deadlock with
scheduling if the node being removed contains allocations for the same
application that is scheduled.
The node removal should only lock while updating the partition and
release before updating the apps and queues. Since node removal is
called as part of the cleanup of a failed add of a node the add flow is
changed inline with the removal to minimise lock time.

The partition lock was also held while placing applications. This is not
required as placing an application does not modify the partition until
everything is done. The lock should only be acquired at that point.

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -530,37 +530,44 @@ func (cc *ClusterContext) updateNodes(request *si.UpdateRequest) {
 		}
 
 		if partition == nil {
-			log.Logger().Info("Failed to add node to non existing partition",
+			log.Logger().Info("Failed to update node on non existing partition",
 				zap.String("nodeID", update.NodeID),
-				zap.String("partitionName", update.Attributes[siCommon.NodePartition]))
+				zap.String("partitionName", update.Attributes[siCommon.NodePartition]),
+				zap.String("nodeAction", update.Action.String()))
 			continue
 		}
 
-		if node, ok := partition.nodes[update.NodeID]; ok {
-			switch update.Action {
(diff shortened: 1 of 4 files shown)


wilfred-s:
I checked this change also on top of the changes for YUNIKORN-460 (PR #250) and the overlap is not a problem the code touched by the 2 changes is limited to a couple of lines with simple to fix merge clashes

yangwwei (commented):
Thank you @wilfred-s  for the quick fix. I have gone through the changes, I think it's gonna fix the problem. Have you tried to add some unit tests to cover this? I think we can possibly do something like run the scheduling at the same time as node removal in the smoke test.

yangwwei (approved):
Checked the changes again, I did not find any issue.
Approving this now.

yangwwei:
hi @wilfred-s regarding your comment

> I checked this change also on top of the changes for YUNIKORN-460 (PR #250) and the overlap is not a problem the code touched by the 2 changes is limited to a couple of lines with simple to fix merge clashes

do you mean we are good to get this merged, this change won't cause any conflicts to PR #250?

wilfred-s:
> do you mean we are good to get this merged, this change won't cause any conflicts to PR #250?

Yes that is what I meant. There is one set of lines that overlaps between the two changes which is easily fixed even on commit. The renamed termination type is the reason for the overlap.

yangwwei:
Thanks for the clarification, merged the PR.

PR #254, opened 2021-03-09
[YUNIKORN-557] Move sync shim cache error messages to DEBUG level

yangwwei:
Remove the false alarm error messages.
To completely get rid of this, we need to see how to fix https://issues.apache.org/jira/browse/YUNIKORN-462


wilfred-s (approved):
I am OK with the direction of the change and will commit the change.

I am not OK with the description, it clearly needs to reference YUNIKORN-462 in the commit message. We should leave a comment in the code pointing to the real fix.

Will change on commit

PR #255, opened 2021-03-12
[YUNIKORN-564] Improve the upscaling performance for StateAware policy.

yangwwei:
For more description about this improvement, pls see https://issues.apache.org/jira/browse/YUNIKORN-564.
The fix here is to add a `filterApp` in the sortApplication function, when it is false, skip the filtering.


wilfred-s (changes requested):
I would not push the `filterApps` flag into the sorting code. StateAware is FIFO with a filter, that is how we have designed and documented it. 
Instead I would handle the flag in the `sortApplications()` of the queue. If the flag is false and sort type is StateAware we fall back to FIFO. That keeps all code changes within the queue code.
It also keeps the sorter code cleaner by hiding the optimisation in the queue (looking towards the plugability of the sorter code later on)

yangwwei:
> I would not push the `filterApps` flag into the sorting code. StateAware is FIFO with a filter, that is how we have designed and documented it.
> Instead I would handle the flag in the `sortApplications()` of the queue. If the flag is false and sort type is StateAware we fall back to FIFO. That keeps all code changes within the queue code.
> It also keeps the sorter code cleaner by hiding the optimisation in the queue (looking towards the plugability of the sorter code later on)

Updated, could u pls help to review this again? thx

wilfred-s (approved):
LGTM, will commit

PR #256, opened 2021-03-12
[YUNIKORN-567] Queue resources are not cleaned up after placeholder cleanup

kingamarton:
After this fix, the queue resources are decreased with the allocated placeholder resources as well during an application removal


wilfred-s (approved):
LGTM

PR #259, opened 2021-03-17
[YUNIKORN-578] dynamic queue quota for gang size

wilfred-s:
A dynamic queue which is linked to a namespace and the quota linked to
that fails to properly enforce gang request size against that quota.
This only happens for the first application submitted to the namespace
and queue. The order for queue create, app addition and quota checks is
incorrect.


yangwwei (approved):
LGTM

PR #260, opened 2021-03-17
[YUNIKORN-574] Wait for placeholder cleanup

kingamarton:
With this fix a new state is introduced: Failing state. Also, Waiting state is renamed to Completing.
When we cleanup the placeholders, before moving the application into the terminating state (Completed or Failed)
we wait for the shim to release the allocations first.

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -42,8 +42,8 @@ import (
 var (
 	reservationDelay          = 2 * time.Second
 	startingTimeout           = 5 * time.Minute
-	waitingTimeout            = 30 * time.Second
-	completedTimeout          = 3 * 24 * time.Hour
+	completingTimeout         = 30 * time.Second
+	terminatedTimeout         = 3 * 24 * time.Hour
 	defaultPlaceholderTimeout = 15 * time.Minute
 )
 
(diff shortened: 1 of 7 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -1303,3 +1281,52 @@ func (sa *Application) notifyRMAllocationAskReleased(rmID string, released []*Al
...
+
+func (sa *Application) isApplicationEmpty(skipPlaceholders bool) bool {
+	if skipPlaceholders {
+		return resources.IsZero(sa.pending) && resources.IsZero(sa.allocatedResource)
+	}
+	return resources.IsZero(sa.pending) && resources.IsZero(sa.allocatedResource) && resources.IsZero(sa.allocatedPlaceholder)
+}
+
+// lock free call, this should be called when holding a lock on the application object
+func (sa *Application) handleStateChange() {
```
wilfred-s, line 1301: This looks really complex. We have discussed this off-line and  come to a nicer solution.
Please update with the discussed changes.

wilfred-s (changes requested):
Please add unit test for the pieces that are missing, if we missed any bits.

pkg/scheduler/objects/application.go
kingamarton: I made the changes we discussed about.
I also kept the `func (sa *Application) failAppIfPossible()` function, because I checked the code and when there is a partition removed, we want to fail the application as well, and in this case we might have no allocations, to trigger the transition to Failed state.

yangwwei:
Thanks for addressing the review comments. LGTM, I'll merge this.

yangwwei (approved):

wilfred-s (commented):
LGTM, just to clean up the state

PR #261, opened 2021-03-19
[YUNIKORN-586] enhance placeholder cleanup on timeout

wilfred-s:
When the placeholder processing times out we could have placeholders in
two states: released or not released.
In the clean up we need to account for that. Placeholders that are marked
for release have a real allocation linked and are being processed by the
shim. The timeout should not release those placeholders.
This also needs to happen when the application is in Completing state.
On confirmation by the shim these placeholder replacements will trigger
a state change back to Running for the application. The placeholders
should thus no longer be available.

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -266,12 +266,29 @@ func (sa *Application) timeoutPlaceholderProcessing() {
 	defer sa.Unlock()
 	switch {
 	// Case 1: if all app's placeholders are allocated, only part of them gets replaced, just delete the remaining placeholders
-	case (sa.IsRunning() || sa.IsStarting()) && !resources.IsZero(sa.allocatedPlaceholder):
+	case (sa.IsRunning() || sa.IsStarting() || sa.IsCompleting()) && !resources.IsZero(sa.allocatedPlaceholder):
+		var toRelease []*Allocation
+		replacing := 0
 		for _, alloc := range sa.getPlaceholderAllocations() {
+			// skip over the allocations that are already marked for release, they will be replaced soon
+			if alloc.released {
+				replacing++
+				continue
+			}
 			alloc.released = true
+			toRelease = append(toRelease, alloc)
 		}
+		log.Logger().Info("Placeholder timeout, releasing placeholders",
+			zap.String("AppID", sa.ApplicationID),
+			zap.Int("placeholders being replaced", replacing),
+			zap.Int("releasing placeholders", len(toRelease)))
+		sa.notifyRMAllocationReleased(sa.rmID, toRelease, si.TerminationType_TIMEOUT, "releasing allocated placeholders on placeholder timeout")
 	// Case 2: in every other case fail the application, and notify the context about the expired placeholder asks
 	default:
+		log.Logger().Info("Placeholder timeout, releasing asks and placeholders",
+			zap.String("AppID", sa.ApplicationID),
+			zap.Int("releasing placeholders", len(sa.allocations)),
+			zap.Int("releasing asks", len(sa.requests)))
 		// change the status of the app to Failing. Once all the placeholders are cleaned up, if will be changed to Failed
 		if err := sa.HandleApplicationEvent(FailApplication); err != nil {
 			log.Logger().Debug("Application state change failed when placeholder timed out",
@@ -281,8 +298,9 @@ func (sa *Application) timeoutPlaceholderProcessing() {
 		}
 		sa.notifyRMAllocationAskReleased(sa.rmID, sa.getAllRequests(), si.TerminationType_TIMEOUT, "releasing placeholders asks on placeholder timeout")
 		sa.removeAsksInternal("")
+		// all allocations are placeholders but GetAllAllocations is locked and cannot be used
+		sa.notifyRMAllocationReleased(sa.rmID, sa.getPlaceholderAllocations(), si.TerminationType_TIMEOUT, "releasing allocated placeholders on placeholder timeout")
 	}
-	sa.notifyRMAllocationReleased(sa.rmID, sa.getPlaceholderAllocations(), si.TerminationType_TIMEOUT, "releasing allocated placeholders on placeholder timeout")
 	sa.clearPlaceholderTimer()
 }
 
@@ -1276,6 +1294,10 @@ func (sa *Application) executeTerminatedCallback() {
 }
 
 func (sa *Application) notifyRMAllocationReleased(rmID string, released []*Allocation, terminationType si.TerminationType, message string) {
+	// only generate event if needed
+	if len(released) == 0 {
+		return
+	}
 	releaseEvent := &rmevent.RMReleaseAllocationEvent{
 		ReleasedAllocations: make([]*si.AllocationRelease, 0),
 		RmID:                rmID,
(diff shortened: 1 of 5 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -266,12 +266,29 @@ func (sa *Application) timeoutPlaceholderProcessing() {
 	defer sa.Unlock()
 	switch {
 	// Case 1: if all app's placeholders are allocated, only part of them gets replaced, just delete the remaining placeholders
-	case (sa.IsRunning() || sa.IsStarting()) && !resources.IsZero(sa.allocatedPlaceholder):
+	case (sa.IsRunning() || sa.IsStarting() || sa.IsCompleting()) && !resources.IsZero(sa.allocatedPlaceholder):
```
kingamarton, line 269: Why do we need here the `sa.IsCompleting()` check?
If the application is in completing state that means that the completing timer is started, so if it will timeout right after the placeholder timer, we might send the same allocation release request to the shim twice.
If we want to keep that check here, the state timer must be cleared.

pkg/scheduler/objects/application.go
```diff
@@ -1308,14 +1334,3 @@ func (sa *Application) notifyRMAllocationAskReleased(rmID string, released []*Al
...
-// Auto progress the application when it enters the Failing state if there is nothing to clean up.
-// Since this is called by the _locked_ state machine while processing an event we cannot call back
-// into the statemachine directly and we need a go routine to avoid a deadlock.
-func (sa *Application) failAppIfPossible() {
-	if resources.IsZero(sa.pending) && resources.IsZero(sa.allocatedResource) && resources.IsZero(sa.allocatedPlaceholder) {
-		// The event handling cannot fail
-		//nolint: errcheck
-		go sa.HandleApplicationEvent(FailApplication)
-	}
-}
```
kingamarton, line 1321: I think we will need this change, because we want to fail the apps if there is a partition removed. In this case we might have an application with no allocations and asks, so if we are not moving it from Failing to Failed, it will be stuck into Failing state.

kingamarton (changes requested):

pkg/scheduler/objects/application.go
wilfred-s: > If the application is in completing state that means that the completing timer is started, so if it will timeout right after the placeholder timer, we might send the same allocation release request to the shim twice.

There is no reason to assume that the timeouts will . The placeholder timer could time out 1 sec after entering the completing state.

What this check does is that it only removes placeholders that have not been replaced yet. If we move into the _Completing_ state we can do that with placeholder still allocated on the app. Any of those placeholders that have been marked as `released` have been communicated to the shim already for a replacement. They have a real allocation linked to it which is not confirmed yet as the core is waiting for the shim to confirm the release. If the replacement comes back from the shim we add the real allocation to the application. At that point we move back out of _Completing_ and we move back into the _Running_ state. As soon as that happens the timing out of the state _Completing_ will become a noop as the state is not the expected state.

That is the specific case we're trying to solve. The case 1 comment should be updated to cover all the use cases properly.

pkg/scheduler/objects/application.go
wilfred-s: When the partition is removed there is no way to get to the applications any more. Everything is gone nothing is kept. It does not matter what happens in the core as long as everything that needs to be released is communicated to the shim.
If we want to keep history around for a removed shim than this might become important but at the moment _everything_ is deleted from memory when the partition is removed.

pkg/scheduler/objects/application_state.go
```diff
@@ -99,7 +99,7 @@ func NewAppState() *fsm.FSM {
 				Dst:  Completed.String(),
 			}, {
 				Name: FailApplication.String(),
-				Src:  []string{Accepted.String(), New.String(), Running.String(), Starting.String(), Completing.String()},
+				Src:  []string{New.String(), Accepted.String(), Starting.String(), Running.String()},
```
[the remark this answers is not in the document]
wilfred-s: See my comments above.

pkg/scheduler/objects/application_state.go
```diff
@@ -152,9 +152,6 @@ func NewAppState() *fsm.FSM {
 				app.executeTerminatedCallback()
 				app.clearPlaceholderTimer()
 			},
-			fmt.Sprintf("enter_%s", Failing.String()): func(event *fsm.Event) {
-				event.Args[0].(*Application).failAppIfPossible()
-			},
```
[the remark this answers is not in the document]
wilfred-s: partition removal leaves no trace of anything

pkg/scheduler/objects/application_test.go
```diff
@@ -19,16 +19,49 @@
...
+	"github.com/apache/incubator-yunikorn-core/pkg/rmproxy/rmevent"
+	"github.com/apache/incubator-yunikorn-scheduler-interface/lib/go/si"
 	"gotest.tools/assert"
 
 	"github.com/apache/incubator-yunikorn-core/pkg/common"
 	"github.com/apache/incubator-yunikorn-core/pkg/common/resources"
 )
 
+// basic app creating with timeout checks
+func TestNewApplication(t *testing.T) {
```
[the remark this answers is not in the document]
wilfred-s: It really tests the call to NewApplication. I'll add further checks to show that it is not just for the placeholder timeout.

pkg/scheduler/objects/application.go
kingamarton: Even if there is no reason to assume that the timeouts will follow quickly, it can easily happen. 

- Let's say we timeout the placeholders, and send the placeholder release request to the shim. As we know it can take a longer period for the shim to remove them, so in the core we will still have the placeholders marked as removed, until we have an answer from the shim.
- Right after that, the completing state times out(before we can get the replacement fro the shim). Since there we don't have this check for the already released placeholders, the core will send the remaining placeholders to the shim again.

So I think we need to mark the placeholders as released when we timeout the Completing state as well, and also we need to clear the completing state timer if we timeout the placeholders. There is no reason to keep it running until it will timeout.

pkg/scheduler/objects/application.go
kingamarton: That mens that Failing the application here: https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/scheduler/partition_manager.go#L131 makes no sense?

wilfred-s: Not for the core side. However that event might trigger a state change which might trigger updates to be send out to the shim.
About 15 line later we call back into the context and remove the partition from the context. You cannot access it anymore. I think there needs to be a broader discussion about historic information tracking in the objects. The core does not do any of that at the moment which we really need to revisit. At that point it will become important **but** that is a larger change that needs to be implemented.

pkg/scheduler/objects/application.go
wilfred-s: starting point of both cases:
- enter _Completing_ state, placeholder allocated
- placeholder timeout triggered: release all non replaced placeholder

case 1:
- in _Completing_, timeout placeholder triggered
- earlier replaced placeholder allocation is confirmed released by the shim, allocation is swapped
- application returns to _Running_
- --> placeholders that are timed out get removed
- --> placeholders that were replaced get swapped and app runs to a normal finish. All based on the shim responses

case 2:
- in _Completing_, timeout placeholder triggered
- no placeholder releases are confirmed
- _Completing_ timeout triggered, all placeholder will be released (again)
- wait for all placeholders to be released, all timers are cleaned up (existing behaviour)
- --> placeholders that are timed out get removed
- --> placeholders that were replaced get swapped and app runs to a normal finish. All based on the shim responses
- --> if all placeholders are removed and the app has not transitioned back to _Running_ it will transition to _Completed_

And yes in case it might happen that an original replacement request comes back before the second release requested. This is not a problem as that just simply triggers a change back into _Running_ for the app. All placeholders will be released either as a replacement or just a simple release. Changing the _Completing_ to _Completed_ transition itself is not what

PR #264, opened 2021-04-08
[YUNIKORN-631] undefined max resource handling

wilfred-s:
### What is this PR for?
The documentation is out of sync with the implementation for the max
resource setting on the queue. Undefined resources are not limited
according to the documentation but the behaviour is a limit of 0.

The documentation describes the correct behaviour.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-631

Unit tests have been added and updated to check the correct behaviour

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -395,7 +395,8 @@ func subNonNegative(left, right *Resource) (*Resource, string) {
 	}
 	return out, message
 }
-// Check if smaller fits in larger, negative values will be treated as 0
+
+// Check if smaller fits in larger
 // Types not defined in the larger resource are considered 0 values for Quantity
 // A nil resource is treated as an empty resource (all types are 0)
 func FitIn(larger, smaller *Resource) bool {
@@ -403,9 +404,9 @@ func FitIn(larger, smaller *Resource) bool {
 }
 
 // Check if smaller fits in the defined resource
-// Types not defined in resource this is called against are skipped and considered maximum value for Quantity
+// Types not defined in resource this is called against are considered the maximum value for Quantity
 // A nil resource is treated as an empty resource (no types defined)
-func (r *Resource) FitIn(smaller *Resource) bool {
+func (r *Resource) FitInMaxUndef(smaller *Resource) bool {
 	return r.fitIn(smaller, true)
 }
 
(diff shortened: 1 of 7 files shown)


pkg/common/resources/resources.go
```diff
@@ -395,12 +395,26 @@ func subNonNegative(left, right *Resource) (*Resource, string) {
 	}
 	return out, message
 }
-
-// Check if smaller fitin larger, negative values will be treated as 0
-// A nil resource is treated as an empty resource (zero)
+// Check if smaller fits in larger, negative values will be treated as 0
+// Types not defined in the larger resource are considered 0 values for Quantity
+// A nil resource is treated as an empty resource (all types are 0)
 func FitIn(larger, smaller *Resource) bool {
```
kingamarton, line 401: I think we should rename the two FitIn functions: `func FitIn(larger, smaller *Resource) bool` and` func (r *Resource) FitIn(smaller *Resource) bool` to a name what reflects the difference between them. Having the same name is easy to mix them, also while reading the code is not obvious what will happen with the undefined resources.

kingamarton (changes requested):

pkg/common/resources/resources.go
wilfred-s: I thought about that but could not really think of a good name, that is why I left it.

The difference is in the description of the function: type is not defined becomes 0, type is not defined becomes max Quantity (i.e. max int64) 
Since Quantity can never be larger and thus will always fit, the value check is skipped:
```
 if skipUndef && !ok {
  continue
}
```
implements the difference between the two.

kingamarton: What about `FitInSkipUndef` for the one where we handle it as max, and left the other one as it is

wilfred-s: The skip undefined points to how it was implemented. It is faster than setting the value to max int64 and doing the comparison. I was trying to capture the max in the name and had `FitInMaxUndef` as my early choice.

kingamarton: I am fine with `FitInMaxUndef` as well, just we need to make the difference obvious between the two methods.

wilfred-s: updated the name and fixed some typos in the comments

kingamarton (approved):

PR #266, opened 2021-04-15
YUNIKORN-637: Core side changes of YUNIKORN-490

manirajv06:
### What is this PR for?
Core side changes of YUNIKORN-490

### What type of PR is it?
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-637

### Questions:


wilfred-s (changes requested):
I don't think we need to import api with an alias (apiCommon). There is no other api package.
That will also limit the changes needed.

I also think that the go.mod file change is missing from the commit. It should point to:
github.com/apache/incubator-yunikorn-scheduler-interface v0.0.0-20210226143918-19a5cca5e428
Please notice the special semver v0.0.0 we should have used that for a long time...

wilfred-s:
Before we commit with the new SI reference we need to run a `go mod tidy`.
There is a lot of dangling references.
@manirajv06: can you please run that after you update and add go.mod and go.sum updates to the change.

manirajv06:
> I don't think we need to import api with an alias (apiCommon). There is no other api package.
> That will also limit the changes needed.
> 
> I also think that the go.mod file change is missing from the commit. It should point to:
> `github.com/apache/incubator-yunikorn-scheduler-interface v0.0.0-20210420020902-c2149f9635f6`
> Please notice the special semver v0.0.0 we should have used that for a long time...

Taken care.

manirajv06:
> Before we commit with the new SI reference we need to run a `go mod tidy`.
> There is a lot of dangling references.
> @manirajv06: can you please run that after you update and add go.mod and go.sum updates to the change.

Taken care.

wilfred-s:
Not sure why it was closed, re-opening for finalising

wilfred-s (approved):
LGTM, nothing left

PR #267, opened 2021-04-21
[YUNIKORN-645] Metrics endpoint doesn't export any metrics

kingamarton:
### What is this PR for?

After YUNIKORN-418 we are using a custom http handler, what while reading the response checks if the response status is changed. If yes, the further reading was skipped. When accessing the metrics endpoint, the response is sent is multiple sub parts. Since after the first one the status will be changed, the processing of the response is skipped. 

With this PR I extended the check to skip the further processing if the status is changed and it is not 200-OK.

### What type of PR is it?
* [X] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-645

### Questions:

### DIFF
--- a/pkg/webservice/webservice.go
+++ b/pkg/webservice/webservice.go
@@ -81,7 +81,7 @@ type YResponseWriter struct {
 
 func (rw *YResponseWriter) Write(bytes []byte) (int, error) {
 	rw.body = bytes
-	if rw.statusCode == -1 {
+	if rw.statusCode == -1  || rw.statusCode == http.StatusOK {
 		_, err := rw.ResponseWriter.Write(bytes)
 		if err != nil {
 			log.Logger().Error(fmt.Sprintf("Unable to serve response. Problem in writing the response \"%s\". Reason: %s", string(bytes), err.Error()))
--- a/pkg/webservice/webservice.go
+++ b/pkg/webservice/webservice.go
@@ -81,7 +81,7 @@ type YResponseWriter struct {
 
 func (rw *YResponseWriter) Write(bytes []byte) (int, error) {
 	rw.body = bytes
-	if rw.statusCode == -1  || rw.statusCode == http.StatusOK {
+	if rw.statusCode == -1 || rw.statusCode == http.StatusOK {
 		_, err := rw.ResponseWriter.Write(bytes)
 		if err != nil {
 			log.Logger().Error(fmt.Sprintf("Unable to serve response. Problem in writing the response \"%s\". Reason: %s", string(bytes), err.Error()))
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -22,8 +22,6 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
-	"gopkg.in/yaml.v2"
-	"gotest.tools/assert"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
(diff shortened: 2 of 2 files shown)


kingamarton:
@manirajv06 can you please check this fix?

pkg/webservice/webservice.go
```diff
@@ -81,7 +81,7 @@ type YResponseWriter struct {
 
 func (rw *YResponseWriter) Write(bytes []byte) (int, error) {
 	rw.body = bytes
-	if rw.statusCode == -1 {
+	if rw.statusCode == -1 || rw.statusCode == http.StatusOK {
```
yangwwei, line 84: This seems to fix the problem we saw here, but this line could still cause issues.
For example, what if we get some other HTTP code defined in https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html. Such as `Accepted`, `Created`, etc.  This will still skip writing the body to the client-side, which is dangerous.

yangwwei (changes requested):

wilfred-s:
I think we need to wind back the introduction of the `YResponseWriter` If there are calls that need specific handling of the errors then that should be part of the defined `HandlerFunc` which we specify in the web routes. We can still use a predefined error mechanism to get the details.
Beside the issue that was mentioned by @yangwwei for the other _successful_ status codes I can also see an issue around error handling and non success codes. If we set `http.StatusInternalServerError` because we have encountered a problem when writing the body via the json encoder, we try to redo that same thing again. Similar for any error in the 400 range.

pkg/webservice/webservice.go
manirajv06: Yes, correct. We should have taken a generic approach. Taken care now.

manirajv06:
> Beside the issue that was mentioned by @yangwwei for the other _successful_ status codes I can also see an issue around error handling and non success codes. If we set `http.StatusInternalServerError` because we have encountered a problem when writing the body via the json encoder, we try to redo that same thing again. Similar for any error in the 400 range.

Intent was to convey the message to the client in case of any issues in sending json formatted error response. To avoid the cyclic dependency, was trying to send the message in plain text format with `http.StatusInternalServerError` because client depends on status code for first level checks before even trying to interpret the actual message (response body).

Now, logging only as ERROR and onus comes on the Infra admin to understand the reasons as part of general debugging. Thoughts?

wilfred-s:
I understand what the intent was but there are a number of issues that show up.
The code must be able to handle _any_ HTTP response code. We also need to be able to handle 3rd party pieces like prometheus or other endpoints that will go through in the future. We have full control over our code and instead of an indirect update we can directly push the json formatted error we need into the message. There would not be any wrapping needed.

That is in line with the current way we return a custom json object in the `validateConf()`, be it with a `dao.ValidateConfResponse` object. The same can be used for errors too (using `w.WriteHeader()` with the appropriate http status code).

manirajv06:
Idea behind using wrapper is to ensure all REST API's starts sending even JSON formatted error responses in single go instead of doing this in piece meal manner (i.e API wise).  In later approach, We have to make changes for each API individually and requires co-ordination with UI side as well for each API. Where as, earlier approach of using wrapper to send JSON formatted error responses requires minimal work and co-ordination on both sides.

wilfred-s:
This was fixed via #268. It removes the wrapping response handler.

PR #268, opened 2021-04-25
YUNIKORN-645. Fixed status code issues, error handling issues etc

manirajv06:
### What is this PR for?

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -211,6 +211,14 @@ func writeHeaders(w http.ResponseWriter) {
 	w.Header().Set("Access-Control-Allow-Headers", "X-Requested-With,Content-Type,Accept,Origin")
 }
 
+func buildJSONErrorResponse(w http.ResponseWriter, detail string, code int) {
+	w.WriteHeader(code)
+	errorInfo := dao.NewYAPIError(nil, code, detail)
+	if jsonErr := json.NewEncoder(w).Encode(errorInfo); jsonErr != nil {
+		log.Logger().Error(fmt.Sprintf("Problem in sending error response in JSON format. Error response: %s", detail))
+	}
+}
+
 func getClusterJSON(partition *scheduler.PartitionContext) *dao.ClusterDAOInfo {
 	clusterInfo := &dao.ClusterDAOInfo{}
 	clusterInfo.TotalApplications = strconv.Itoa(partition.GetTotalApplicationCount())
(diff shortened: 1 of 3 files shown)


pkg/webservice/webservice.go
```diff
@@ -81,7 +82,7 @@ type YResponseWriter struct {
 
 func (rw *YResponseWriter) Write(bytes []byte) (int, error) {
 	rw.body = bytes
-	if rw.statusCode == -1 {
+	if rw.statusCode == -1 || (rw.statusCode >= http.StatusOK && rw.statusCode <= http.StatusIMUsed) {
```
yangwwei, line 85: I still have concerns about checking the status code. This doesn't seem to be a reliable solution. I suggest reverting the wrapper change and continue to use the default writer, which is safer for me. If you look at https://golang.org/pkg/net/http/#Client.Get, the go client library will return an error with some internal logic and check over the HTTP status code. If the wrapper on the server-side wrongly operates on responses, we are in trouble.

yangwwei (changes requested):
hi @manirajv06 

Pls see my comments. I have concerns that the wrapper may cause more problems. To solve the original problem, can we use this approach:

```
// ToJson() wraps the error messages to JSON format
http.Error(w, utils.ToJson(err), http.StatusInternalServerError)
```

I think this can solve the issue and also a safer approach. Let me know if this makes sense.

wilfred-s (approved):
LGTM

kingamarton (changes requested):
The fix itself LGTM, I added one test case in the following PR: https://github.com/apache/incubator-yunikorn-core/pull/267/files
Please add that UT to your change, so this bug will be covered as well.

manirajv06:
> The fix itself LGTM, I added one test case in the following PR: https://github.com/apache/incubator-yunikorn-core/pull/267/files
> Please add that UT to your change, so this bug will be covered as well.

Taken care

yangwwei:
Cool, thank you  @manirajv06.
Can you please fix the lint issue: 
```
pkg/webservice/handlers_test.go:43: File is not `goimports`-ed with -local github.com/apache/incubator-yunikorn (goimports)
	"github.com/prometheus/client_golang/prometheus/promhttp"
make: *** [Makefile:59: lint] Error 1
Error: Process completed with exit code 2.
```

pkg/webservice/handlers.go
```diff
@@ -211,6 +211,14 @@ func writeHeaders(w http.ResponseWriter) {
 	w.Header().Set("Access-Control-Allow-Headers", "X-Requested-With,Content-Type,Accept,Origin")
 }
 
+func buildJSONErrorResponse(w http.ResponseWriter, detail string, code int) {
+	w.WriteHeader(code)
+	errorInfo := dao.NewYAPIError(nil, code, detail)
+	if jsonErr := json.NewEncoder(w).Encode(errorInfo); jsonErr != nil {
```
yangwwei, line 217: One doubt here: I think this line will write the response to the client with a successful code, correct?
how can we make sure the client-side can get the error code? Having the error code encoded in the response body is not enough.

yangwwei (changes requested):

pkg/webservice/handlers.go
manirajv06: No, @yangwwei. We are writing the HTTP status code as part of response headers as well. Actual response body has been generated based on the following struct:

	StatusCode  int    `json:"status_code"`
	Message     string `json:"message"`
	Description string `json:"description"`

In general, client reads the status code from response headers and takes decision accordingly. I assuming this should be the case even with our UI component.

yangwwei: Oh, misread that part. Thanks for the explaination.

yangwwei (approved):
LGTM

PR #270, opened 2021-05-04
YUNIKORN-545: JSON based API error responses

manirajv06:
### What is this PR for?

Return JSON based error responses for all REST API's.

### What type of PR is it?
Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-545

### Questions:


manirajv06:
@wilfred-s @yangwwei can you please take a look?

wilfred-s (approved):
LGTM
This should not "break" the web UI as we still set the same error codes in the response.
Formatting might be thrown off a bit by the different error. 
Follow up in a new jira for the web UI component

PR #271, opened 2021-05-05
Core side changes of YUNIKORN-657

yuchaoran-apple:
### What is this PR for?
Expose the reason of scheduling failures to the shim side. Shim side changes are at https://github.com/apache/incubator-yunikorn-k8shim/pull/258

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-657

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -166,7 +166,7 @@ func (sa *Application) IsFailed() bool {
 	return sa.stateMachine.Is(Failed.String())
 }
 
-// Handle the state event for the application.
+// HandleApplicationEvent handles the state event for the application.
 // The state machine handles the locking.
 func (sa *Application) HandleApplicationEvent(event applicationEvent) error {
 	err := sa.stateMachine.Event(event.String(), sa)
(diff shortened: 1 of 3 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -177,13 +177,28 @@ func (sa *Application) HandleApplicationEvent(event applicationEvent) error {
...
+}
+
+func (sa *Application) OnStateChange(event *fsm.Event, eventInfo string) {
 	updatedApps := make([]*si.UpdatedApplication, 0)
+	var message string
+	if len(eventInfo) == 0 {
+		message = fmt.Sprintf("Status change triggered by the event: %s", event.Event)
+	} else {
+		message = eventInfo
+	}
```
yangwwei, line 196: feel free to remove the message "Status change triggered by the event", this is not informative, I think we can ignore it.

pkg/scheduler/objects/application.go
```diff
@@ -299,7 +314,7 @@ func (sa *Application) timeoutPlaceholderProcessing() {
 			zap.Int("releasing placeholders", len(sa.allocations)),
 			zap.Int("releasing asks", len(sa.requests)))
 		// change the status of the app to Failing. Once all the placeholders are cleaned up, if will be changed to Failed
-		if err := sa.HandleApplicationEvent(FailApplication); err != nil {
+		if err := sa.HandleApplicationEventWithInfo(FailApplication, "InsufficientResources"); err != nil {
```
yangwwei, line 317: How about changing the message to "ResourceReservationTimeout"?
after all we are doing the resource reservation for the app, this can indicate we were trying but unable to fulfill the requests for a given time.
InsufficientResources on the other hand might be too general
let me know your thought

yangwwei (changes requested):

pkg/scheduler/objects/application.go
yuchaoran-apple: Cool I feel the same. Removed it

pkg/scheduler/objects/application.go
yuchaoran-apple: Makes sense to me. Changed to `ResourceReservationTimeout`

yangwwei (approved):
LGTM

PR #272, opened 2021-05-07
YUNIKORN-573: Make sure all timer go routines are stopped after remov…

manirajv06:
### What is this PR for?

Ensure PlaceholderTimer has been stopped as part of app removal

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-573

Remove all allocations of an app and ensure placeholdertimer is nil

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -268,6 +268,9 @@ func (sa *Application) clearPlaceholderTimer() {
 	}
 	sa.placeholderTimer.Stop()
 	sa.placeholderTimer = nil
+	log.Logger().Debug("Application placeholder timer cleared",
+		zap.String("AppID", sa.ApplicationID),
+		zap.Duration("Timeout", sa.execTimeout))
 }
 
 func (sa *Application) timeoutPlaceholderProcessing() {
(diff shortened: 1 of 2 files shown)


kingamarton (changes requested):
I think you mixed up a couple of Jiras here:
- in the title you have YUNIKORN-573: Please fix that one, because the PR is linked to the wrong Jira
- in the commit description you mentioned YUNIKORN-641, but the fix has only one comment added and a unit test, the Jira is about publishing pod events to the real pod when a placeholder is timed out. Please set the Jira as a Draft if it is still in progress.

manirajv06:
> I think you mixed up a couple of Jiras here:
> 
> * in the title you have YUNIKORN-573: Please fix that one, because the PR is linked to the wrong Jira
> * in the commit description you mentioned YUNIKORN-641, but the fix has only one comment added and a unit test, the Jira is about publishing pod events to the real pod when a placeholder is timed out. Please set the Jira as a Draft if it is still in progress.

Modified the description with correct jira link. This PR is for YUNIKORN-573.

pkg/scheduler/objects/application.go
```diff
@@ -1267,6 +1270,7 @@ func (sa *Application) RemoveAllAllocations() []*Allocation {
 	// cleanup allocated resource for app (placeholders and normal)
 	sa.allocatedResource = resources.NewResource()
 	sa.allocatedPlaceholder = resources.NewResource()
+	sa.clearPlaceholderTimer()
```
kingamarton, line 1273: Please clear the state timer as well. This will be probably OK, but just to make sure it is cleared I would call a clear on that as well.

kingamarton (changes requested):

kingamarton (approved):

PR #275, opened 2021-05-27
[YUNIKORN-582] Introduced soft/hard gang scheduling styles

kingamarton:
### What is this PR for?
We ned to introduce gang scheduling styles to able able to chose the behaviour if a gang scheduling application times out.
There can be two kind of styles defined:

"hard" - fail the application after timeout
"soft" - schedule the application as a normal, non gang scheduling application

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-582

For now the build will fail, because it depends on the scheduler-interface.
After the shim side tracked under https://issues.apache.org/jira/browse/YUNIKORN-679  changes are ready, we will need to test it end to end.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -47,6 +47,12 @@ var (
 	defaultPlaceholderTimeout = 15 * time.Minute
 )
 
+
+const(
+	Soft string = "Soft"
+	Hard = "Hard"
+	)
+
 type Application struct {
 	ApplicationID  string
 	Partition      string
@@ -69,6 +75,7 @@ type Application struct {
 	stateTimer           *time.Timer            // timer for state time
 	execTimeout          time.Duration          // execTimeout for the application run
 	placeholderTimer     *time.Timer            // placeholder replace timer
+	gangSchedulingStyle  string                 // gang scheduling style can be hard (after timeout we fail the application), or soft (after timeeout we schedule it as a normal application)
 
 	rmEventHandler     handler.EventHandler
 	rmID               string
@@ -97,6 +104,13 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
 	if time.Duration(0) == placeholderTimeout {
 		placeholderTimeout = defaultPlaceholderTimeout
 	}
+	gangSchedStyle := siApp.GetGangSchedulingStyle()
+	if gangSchedStyle != Soft && gangSchedStyle != Hard {
+		log.Logger().Info("Unknown gang scheduling style, using hard style as default",
+			zap.String("gang scheduling style", gangSchedStyle))
+		gangSchedStyle = Hard
+	}
+	app.gangSchedulingStyle = gangSchedStyle
 	app.execTimeout = placeholderTimeout
 	app.user = ugi
 	app.rmEventHandler = eventHandler
@@ -164,6 +178,10 @@ func (sa *Application) IsFailing() bool {
 
 func (sa *Application) IsFailed() bool {
 	return sa.stateMachine.Is(Failed.String())
+}
+
+func (sa *Application) IsResuming() bool {
+	return sa.stateMachine.Is(Resuming.String())
 }
 
 // HandleApplicationEvent handles the state event for the application.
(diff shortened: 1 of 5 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -97,6 +104,13 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
 	if time.Duration(0) == placeholderTimeout {
 		placeholderTimeout = defaultPlaceholderTimeout
 	}
+	gangSchedStyle := siApp.GetGangSchedulingStyle()
+	if gangSchedStyle != Soft && gangSchedStyle != Hard {
+		log.Logger().Info("Unknown gang scheduling style, using hard style as default",
+			zap.String("gang scheduling style", gangSchedStyle))
+		gangSchedStyle = Hard
+	}
```
yangwwei, line 112: we should use `soft` as default

pkg/scheduler/objects/application.go
```diff
@@ -312,9 +330,14 @@ func (sa *Application) timeoutPlaceholderProcessing() {
 		log.Logger().Info("Placeholder timeout, releasing asks and placeholders",
 			zap.String("AppID", sa.ApplicationID),
 			zap.Int("releasing placeholders", len(sa.allocations)),
-			zap.Int("releasing asks", len(sa.requests)))
+			zap.Int("releasing asks", len(sa.requests)),
+			zap.String("used Scheduling type", sa.gangSchedulingStyle))
```
yangwwei, line 334: suggest to change "used scheduling type" to "gang scheduling style", just to keep the words consistent

pkg/scheduler/objects/application_state.go
```diff
@@ -106,6 +108,10 @@ func NewAppState() *fsm.FSM {
 				Name: FailApplication.String(),
 				Src:  []string{Failing.String()},
 				Dst:  Failed.String(),
+			}, {
+				Name: ResumeApplication.String(),
+				Src:  []string{New.String(), Accepted.String()},
+				Dst:  Resuming.String(),
```
yangwwei, line 114: YuniKorn__Online_Whiteboard_for_Visual_Collaboration

in this chart, I think when placeholder times out, the app could be in any state among [Accepted, Starting, Running, Waiting, Completing]. But we only handle the states `New` and `Accepted`, will this be able to cover all other cases?

yangwwei (changes requested):
hi @kingamarton pls see my comments. And also I think we need to add some more UT for this to cover different scenarios.

pkg/scheduler/objects/application.go
kingamarton: Since we are already using Hard by default in our 0.10 release, if we change it to Soft, can break some workflows, not? I used Hard as default, to keep the same behaviour as before.

pkg/scheduler/objects/application_state.go
kingamarton: In all other cases (if the app is Running, Starting or Completing) we don't fail the application. In this case we just delete the remaining placeholders and let the application finish. You can see this part here: https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/scheduler/objects/application.go#L296-L312

kingamarton: We don't have anymore Waiting state. It was changed to Completing, so I think the Soft style will have impact only for the other cases when we would fail the application.

kingamarton:
> hi @kingamarton pls see my comments. And also I think we need to add some more UT for this to cover different scenarios.

@yangwwei what cases do you mean more exactly?
In application_test.go we have a couple of test cases for the timeout. (we are already testing the following scenarios:
- the app is Completing when it times out - the style has no effect on this case
- the app is in starting state - again the style has no effect on this case
- the app is in Accepted state - here I split the test case into 2 cases, where the style is Soft, and when the Style is Hard)

yangwwei (approved):
+1

PR #279, opened 2021-06-23
YUNIKORN-580: Check if the placeholders and real resources are the same

manirajv06:
### What is this PR for?

For now, In case of resource differences between real and placeholder, event would be published to the real allocation using events plugin and logging it as warn. Need to re-visit this in future releases for better implementation.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-580

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -842,7 +842,7 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() interfaces.Nod
 						zap.Error(err))
 				}
 				// Is real allocation matches with placeholder resource?
-				if ! resources.Equals(alloc.AllocatedResource, ph.AllocatedResource) {
+				if !resources.Equals(alloc.AllocatedResource, ph.AllocatedResource) {
 					log.Logger().Warn("Real allocation is not matching with placeholder allocation", zap.String("Real alloc: ", alloc.AllocationKey), zap.String("Real alloc resource: ", alloc.AllocatedResource.DAOString()), zap.String("Placeholder alloc: ", ph.AllocationKey), zap.String("Placeholder alloc resource: ", ph.AllocatedResource.DAOString()))
 					// post scheduling events via the event plugin
 					if eventCache := events.GetEventCache(); eventCache != nil {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -841,6 +841,21 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() interfaces.Nod
 					log.Logger().Warn("ask repeat update failed unexpectedly",
 						zap.Error(err))
 				}
+				// Is real allocation matches with placeholder resource?
+				if ! resources.Equals(alloc.AllocatedResource, ph.AllocatedResource) {
+					// post scheduling events via the event plugin
+					if eventCache := events.GetEventCache(); eventCache != nil {
+						message := fmt.Sprintf("Real Pod allocation %s is not matching with placeholder allocation %s in application %s", request.AllocationKey, ph.AllocationKey, request.ApplicationID)
```
yuchaoran2011, line 848: Could you please add in the pod event numerical values of the resources? Something like "real resource: %s, placeholder resource: %s"?

yuchaoran2011 (changes requested):
Thanks for working on this! Is this PR only going to include the check for equality? If that's the case, do we have another JIRA to work on enforcing equality and fall back to non-gang scheduling if they are not equal?

manirajv06:
> Thanks for working on this! Is this PR only going to include the check for equality? If that's the case, do we have another JIRA to work on enforcing equality and fall back to non-gang scheduling if they are not equal?

Yes, but falling back to non-gang scheduling approach needs some more thought. We may or may not end up in following the same. We applied the simple and quick changes as we are nearing release 0.11 code freeze.

yuchaoran2011:
@manirajv06 Sounds good. Just a formatting issue needs to be fixed

yangwwei (changes requested):
Looks good, pls fix the pre-commit.
@manirajv06 have you run some local tests to ensure this works end to end? Thanks

kingamarton (commented):
The changes looks good to me. Have you performed some e3e testing to check if it is working and the events are published to the pod as expected?

cheersyang:
@manirajv06 , @kingamarton have done enough tests for this? This is the last issue for 0.11, we need to get this done and start the release procedure, thanks!!

kingamarton (approved):
I did some manual testing and it works expected (the warning message is logged and the event is published if the event cache is enabled).

PR #285, opened 2021-07-29
[YUNIKORN-758]Ignore unschedulable nodes for asks with certain tags

HuangTing-Yao:
### What is this PR for?
leverage allocationTags to skip the schedulable and reserved node check, and retain the resource check.
Also remove the schedulable and reserved filter from `getnode` func

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-758

### Questions:

### DIFF
--- a/pkg/scheduler/objects/node.go
+++ b/pkg/scheduler/objects/node.go
@@ -312,8 +312,8 @@ func (sn *Node) ReplaceAllocation(uuid string, replace *Allocation) {
 // Taking into account resources marked for preemption if applicable.
 // If the proposed allocation does not fit false is returned.
 // TODO: remove when updating preemption
-func (sn *Node) CanAllocate(res *resources.Resource, preemptionPhase bool) bool {
-	err := sn.preAllocateCheck(res, "", preemptionPhase)
+func (sn *Node) CanAllocate(res *resources.Resource, preemptionPhase, ignore bool) bool {
+	err := sn.preAllocateCheck(res, "", preemptionPhase, ignore)
 	return err == nil
 }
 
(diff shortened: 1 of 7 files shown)


ycr-oss:
Hi @HuangTing-Yao, what's the status of this PR? Do you need help?

HuangTing-Yao:
Hi @ycr-oss , I am glad that you can help, After my change in this PR, those are some UT failed.
I am trying figure it out, but with this PR, DaemonSet can be scheduled to cordon node in k8s.

ycr-oss:
@HuangTing-Yao Looks like the test failures all come from this file: https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/scheduler/partition_test.go. There are three failures at lines 478, 1069 and 1356 respectively. I'll take a look over the weekend

HuangTing-Yao:
@ycr-oss  Thanks, I will also keep fixing it.

wilfred-s (changes requested):
The fact that an ask can be scheduled on a node marked as unschedulable should be part of the allocation ask we store. We should not need to re-calculate this fact every single time.
* extend the ask with a flag to store this outcome.
* set the flag when the ask is created
* keep the logic in the `NewAllocationAsk()`, a placeholder should also not have this flag.

I think we have the wrong approach to really fix the daemon set scheduling here. See my comments in the umbrella jira.

chenya-zhang:
Hi @TaoYang526 , are you still working on this ticket? :)

HuangTing-Yao:
Hi @chenya-zhang @ycr-oss , I have finished this PR, wait for others review.

pkg/common/utils.go
```diff
@@ -113,3 +114,19 @@ func ConvertSITimeout(millis int64) time.Duration {
...
+		ignore = false
+	} else {
+		ignore, err = strconv.ParseBool(ignoreUnschedulable)
+		if err != nil {
+			log.Logger().Warn("Failed to convert allocationTag ignoreUnschedulable from string to bool")
+			ignore = false
+		}
+	}
+	return ignore
+}
```
yangwwei, line 132: this can be simplified to: 

```
if ignoreUnschedulable, ok := tag[xxx]; ok {
  // tag found
 if  ignore, err = strconv.ParseBool(ignoreUnschedulable); err == nil {
    // parse successful
    return ignore
  }
}

// every other cases, return false (default)
return false
```

pkg/common/utils_test.go
```diff
@@ -116,3 +118,18 @@ func TestConvertSITimeout(t *testing.T) {
 		})
 	}
 }
+
+func TestGetIgnoreUnschedulable(t *testing.T) {
```
yangwwei, line 122: can we cover the case where the tag value is not true or false?

pkg/scheduler/objects/node.go
```diff
@@ -358,20 +358,16 @@ func (sn *Node) preConditions(allocID string, allocate bool) bool {
...
+func (sn *Node) preAllocateCheck(res *resources.Resource, resKey string, preemptionPhase, ignoreUnschedulable bool) error {
+	// skip schedulable check if ignoreUnschedulable is true
+	if !ignoreUnschedulable {
+		// shortcut if a node is not schedulable
+		if !sn.IsSchedulable() {
+			log.Logger().Debug("node is unschedulable",
+				zap.String("nodeID", sn.NodeID))
+			return fmt.Errorf("pre alloc check, node is unschedulable: %s", sn.NodeID)
+		}
 	}
```
yangwwei, line 370: move into one if ?
```
if !sn.IsSchedulable() && !ignoreUnschedulable {
   ...
}
```

pkg/scheduler/objects/node.go
```diff
@@ -380,6 +376,12 @@ func (sn *Node) preAllocateCheck(res *resources.Resource, resKey string, preempt
 			return fmt.Errorf("pre alloc check: node %s reserved for different app or ask: %s", sn.NodeID, resKey)
 		}
 	}
+	// cannot allocate zero or negative resource
+	if !resources.StrictlyGreaterThanZero(res) {
+		log.Logger().Debug("pre alloc check: requested resource is zero",
+			zap.String("nodeID", sn.NodeID))
+		return fmt.Errorf("pre alloc check: requested resource is zero: %s", sn.NodeID)
+	}
```
yangwwei, line 384: why do we need to move these few lines?

yangwwei (changes requested):

pkg/common/utils.go
HuangTing-Yao: Sure, done.

pkg/common/utils_test.go
HuangTing-Yao: Sure

pkg/scheduler/objects/node.go
HuangTing-Yao: Done.

pkg/scheduler/objects/node.go
HuangTing-Yao: Already move these back.

yangwwei (approved):
LGTM

PR #287, opened 2021-08-02
[YUNIKORN-771] make license-check fix for linux

wilfred-s:
### What is this PR for?
The license check run as part of the PR workflow always passes when run on linux.
The recursive grep redirect to a variable does not work.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-771

NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,13 +70,25 @@ lint:
 	$${lintBin} run --new-from-rev=$${headSHA}
 
 .PHONY: license-check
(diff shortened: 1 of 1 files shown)


yangwwei (approved):
+1. thanks for fixing this @wilfred-s

yangwwei (approved):

PR #292, opened 2021-08-05
YUNIKORN-783 The properties are not propagated to dynamic queue when …

chia7712:
…the queue is under a non-leaf dynamical queue

### What is this PR for?
The non-leaf dynamical queue can't get properties so it can' pass any properties to its leaf queue.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-783

TestPropertiesFromNewDynamicQueue

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -193,10 +193,8 @@ func (sq *Queue) setTemplateProperties(parent map[string]string) {
 	}
 	// for a leaf queue pull out all values from the template and set each of them
 	// See YUNIKORN-193: for now just copy one attr from parent
-	if sq.isLeaf {
-		if parent[configs.ApplicationSortPolicy] != "" {
-			sq.properties[configs.ApplicationSortPolicy] = parent[configs.ApplicationSortPolicy]
-		}
+	if parent[configs.ApplicationSortPolicy] != "" {
+		sq.properties[configs.ApplicationSortPolicy] = parent[configs.ApplicationSortPolicy]
 	}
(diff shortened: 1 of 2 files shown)


wilfred-s:
The design was never intended for a hierarchical copy of the application sort policy. It was a simple fix for just the direct parent to a leaf queue. Anything else was supposed to become part of the template.

There is a jira open to add a template that we can set for inheritance. [YUNIKORN-193](https://issues.apache.org/jira/browse/YUNIKORN-193). It has not been worked on but that is the solution we should really implement.

From that point of view I am a -1 on this change.
I left comments in YUNIKORN-193 and will be happy to work with you on getting that template properly defined and implemented.

chia7712:
@wilfred-s thanks for your feedback.

> There is a jira open to add a template that we can set for inheritance. YUNIKORN-193. It has not been worked on but that is the solution we should really implement.

agreed. that is a better solution.

> The design was never intended for a hierarchical copy of the application sort policy. It was a simple fix for just the direct parent to a leaf queue. Anything else was supposed to become part of the template.

The configured queue can have hierarchical copy. By contrast, dynamic queue can't get that. It seems to me that is weird behavior. For example: I add sort policy to root but only configured children queues can handle spark job (the dynamic queue can't do that since it does not extend stateaware policy from root)

chia7712:
close this PR. #296 offers a better solution.

PR #294, opened 2021-08-06
[YUNIKORN-786] refactor TestHistoricalPartitionInfoUpdater to make it not depend on timer

chia7712:
…ot depend on timer

### What is this PR for?
Depending on timer makes TestHistoricalPartitionInfoUpdater unstable. It fails if the thread can't complete at the right time. Also, some methods are not covered by test.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-786

### Questions:

### DIFF
--- a/pkg/metrics/metrics_collector_test.go
+++ b/pkg/metrics/metrics_collector_test.go
@@ -27,32 +27,66 @@ import (
 	"github.com/apache/incubator-yunikorn-core/pkg/metrics/history"
 )
 
-func TestHistoricalPartitionInfoUpdater(t *testing.T) {
+func TestStop(t *testing.T) {
 	metricsHistory := history.NewInternalMetricsHistory(3)
-	setInternalMetricsCollectorTicker(5 * time.Millisecond)
-	metricsCollector := NewInternalMetricsCollector(metricsHistory)
+	metricsCollector := newInternalMetricsCollector(metricsHistory, 1*time.Second)
 	metricsCollector.StartService()
 
-	go func() {
-		metrics := GetSchedulerMetrics()
-		for i := 0; i < 3; i++ {
-			metrics.IncTotalApplicationsRunning()
-			metrics.AddAllocatedContainers(2)
-			time.Sleep(4 * time.Millisecond)
-		}
-	}()
-
-	time.Sleep(11 * time.Millisecond)
 	metricsCollector.Stop()
+	// wait for the thread to store record. it should not happen
+	time.Sleep(1500 * time.Millisecond)
+
+	records := metricsHistory.GetRecords()
+	assert.Equal(t, 3, len(records), "Expected exactly 3 history records")
+	for _, record := range records {
+		assert.Assert(t, record == nil, "The 1st item should be nil!")
+	}
+}
+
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics_collector_test.go
```diff
@@ -27,32 +27,66 @@ import (
...
+	time.Sleep(1500 * time.Millisecond)
 	metricsCollector.Stop()
 
+	records := metricsHistory.GetRecords()
+	assert.Equal(t, 3, len(records), "Expected exactly 3 history records")
+	for i, record := range records {
+		if i == 2 {
+			assert.Assert(t, record != nil, "The 1st item should NOT be nil!")
+		} else {
+			assert.Assert(t, record == nil, "All items should be nil!")
```
yangwwei, line 61: we may want to slight change the message:

```
if i == 2 {
  assert.Assert(t, record != nil, fmt.Sprintf("record should not be nil, index: %d", i))
} else {
  assert.Assert(t, record == nil, fmt.Sprintf("record should be nil, index: %d", i))
}
```

pkg/metrics/metrics_collector_test.go
```diff
@@ -27,32 +27,66 @@ import (
...
+
 	records := metricsHistory.GetRecords()
 	assert.Equal(t, 3, len(records), "Expected exactly 3 history records")
 	for i, record := range records {
 		switch i {
 		case 0:
-			if record != nil {
-				t.Fatal("The 1st item should be nil!")
-			}
+			assert.Assert(t, record == nil, "The 1st item should be nil!")
```
yangwwei, line 89: similar to the earlier comment: 
```
fmt.Sprintf("record should be nil, index: %d", i)
```

pkg/metrics/metrics_collector_test.go
```diff
@@ -27,32 +27,66 @@ import (
...
 		case 1:
 			assert.Equal(t, 2, record.TotalApplications, "Expected exactly 2 applications at 10 msec")
```
yangwwei, line 91: you may want to remove the word: `at 10msec` in these messages since now it is not depending on that.
and for this one (not introduced by this PR but still would be good to get it fixed)

```
case 2:
  assert.Equal(t, 3, record.TotalApplications, "Expected exactly 3 applications at 20 msec")
  assert.Equal(t, 6, record.TotalContainers, "Expected exactly 4 allocations at 20 msec")
}
```

the message for the last assert should be "expected exactly 6 allocations"

yangwwei (changes requested):
hi @chia7712 thanks for working on this. the UT coverage improved a lot with this refactorying. overall looks good, I have added some minor comments and pls take a look. thanks!

chia7712:
@yangwwei thanks for all comments! I have addressed them in last commit.

yangwwei (approved):
hi @chia7712  thanks for the quick update. looks good. +1

PR #297, opened 2021-08-11
[YUNIKORN-791] Allow to register nodes with infinite resources

kingamarton:
### What is this PR for?
To bypass the “placement” phase, the simplest approach is to register a node with infinite resources, so the pod resource always fits into this node.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-791

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -403,6 +403,13 @@ func FitIn(larger, smaller *Resource) bool {
 	return larger.fitIn(smaller, false)
 }
 
+// Check if smaller fits in larger
+// Types not defined in the larger resource are considered 0 values for Quantity
+// A nil resource is treated as max resource
+func FitInUndef(larger, smaller *Resource) bool {
+	return larger.fitIn(smaller, true)
+}
+
 // Check if smaller fits in the defined resource
 // Types not defined in resource this is called against are considered the maximum value for Quantity
 // A nil resource is treated as an empty resource (no types defined)
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -403,13 +403,6 @@ func FitIn(larger, smaller *Resource) bool {
 	return larger.fitIn(smaller, false)
 }
 
-// Check if smaller fits in larger
-// Types not defined in the larger resource are considered 0 values for Quantity
-// A nil resource is treated as max resource
-func FitInUndef(larger, smaller *Resource) bool {
-	return larger.fitIn(smaller, true)
-}
-
 // Check if smaller fits in the defined resource
 // Types not defined in resource this is called against are considered the maximum value for Quantity
 // A nil resource is treated as an empty resource (no types defined)
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -592,7 +592,7 @@ func (cc *ClusterContext) addNodes(request *si.UpdateRequest) {
 	rejectedNodes := make([]*si.RejectedNode, 0)
 	// if we have an unlimited node, reject all new nodes
 	// if we have an unlimited node between the new nodes, register only that node
-	unlimitedNodeRegistered, visitedNodes := cc.checkForUnlimitedNodes()
+	unlimitedNodeRegistered := cc.checkForUnlimitedNodes()
 	if unlimitedNodeRegistered {
 		for _, node := range request.NewSchedulableNodes {
 			rejectedNodes = append(rejectedNodes, &si.RejectedNode{
@@ -612,10 +612,11 @@ func (cc *ClusterContext) addNodes(request *si.UpdateRequest) {
 		sn := objects.NewNode(node)
 		// if the node is unlimited, check the following things:
 		// 1. reservation is enabled or not. If yes, reject the node
-		// 2. this is the first node. If not reject if
+		// 2. wre there any other nodes in the list, if yes reject the node
+		// 3. this is the first node. If not reject it
 		if sn.IsUnlimited() {
 			// if we have other nodes as well or we already have some registered nodes, reject the node registration
-			if len(request.NewSchedulableNodes) > 1 || visitedNodes > 0 {
+			if len(request.NewSchedulableNodes) > 1 {
 				rejectedNodes = append(rejectedNodes, &si.RejectedNode{
 					NodeID: node.NodeID,
 					Reason: "There are other nodes as well, registering unlimited nodes is forbidden",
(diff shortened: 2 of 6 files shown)


yangwwei:
thank you @kingamarton for working on this, and thank you @wilfred-s  for the review.
regarding the comments

> Can we not set the availableResource to nil and leave the normal processing flow unchanged? 

@wilfred-s  do you mean to set node available resources to nil? and that will represent infinite resources?

yangwwei:
Nice, if that works then it will be perfect. Sounds good to me. @kingamarton  could you pls try to make the changes and see if there is any gap to do so? If that also makes sense to you.  Thanks

kingamarton:
@wilfred-s 
> In the solution chosen you special case each action. This bypasses updating allocated resources completely. That means lots of small changes everywhere. Resource calculations should all be nil safe (if not that is a bug on that calculation). Can we not set the `availableResource` to nil and leave the normal processing flow unchanged? That would remove a lot of the maintenance burden for later and the node will still show "normal" values when displayed in the web UI.

The nil value in the resource calculation is handled as zero resource, so the check if the allocation fits in the node will fail: https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/common/resources/resources.go#L417-L419
Since the resource calculation is used in so many places, I think it is safer to implement the checks in multiple places than check the nil value meaning.

wilfred-s:
> * what should happen with the already registered nodes? It might happen then when we register the infinite node, there are already some nodes registered.  In this case we should reject the unlimited node registration, or we should delete the other nodes?

I think we need to reject the new node. You can only have an unlimited node registration as the first node in the cluster. After that it could cause all kinds of issues.
I do take into account also that we cannot have reservations on either. We have thus some configs to check and fail an unlimited node registration in a number of cases

> The nil value in the resource calculation is handled as zero resource, so the check if the allocation fits in the node will fail: https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/common/resources/resources.go#L417-L419
Since the resource calculation is used in so many places, I think it is safer to implement the checks in multiple places than check the nil value meaning.

The problem with implementing a check in a number of places is that it is easy to change the logic and break or bypass the checks. If we can get to a simple construct like what we did for the queue quota check with stating undefined is interpreted as not limited i.e. maximum value then we cannot break things.
We only have two places in the node where we check the available resources that is in `preAllocateCheck` and `AddAllocation`. In the current change you have 8 points to check for an unlimited node. Even special casing the call to `FitIn` with a nil check before the call would bring it down from 8 to 2. Those two spots are easily documented and simple to maintain.

kingamarton:
I have addressed your comments @wilfred-s . Thank you for the hints. I am writing now some new tests for the neewly introduced parts.

pkg/scheduler/context.go
```diff
@@ -590,8 +590,52 @@ func (cc *ClusterContext) updateNodes(request *si.UpdateRequest) {
...
+				AcceptedNodes: acceptedNodes,
+				RejectedNodes: rejectedNodes,
+			})
+		return
+	}
 	for _, node := range request.NewSchedulableNodes {
 		sn := objects.NewNode(node)
+		// if the node is unlimited, check the following things:
+		// 1. reservation is enabled or not. If yes, reject the node
+		// 2. wre there any other nodes in the list, if yes reject the node
```
yangwwei, line 615: typo "wre" -> "were"

yangwwei (approved):
Looks all good, great work @kingamarton .
Just a tiny typo found in comments, we can fix that while committing the changes.
@wilfred-s do you want to give another look before I committing this?

pkg/common/resources/resources.go
```diff
@@ -403,6 +403,13 @@ func FitIn(larger, smaller *Resource) bool {
 	return larger.fitIn(smaller, false)
 }
 
+// Check if smaller fits in larger
+// Types not defined in the larger resource are considered 0 values for Quantity
+// A nil resource is treated as max resource
+func FitInUndef(larger, smaller *Resource) bool {
```
wilfred-s, line 409: This is the same as calling `larger.FitInMaxUndef(smaller)` I don't think we need a new function for that.

pkg/scheduler/context.go
```diff
@@ -590,8 +590,52 @@ func (cc *ClusterContext) updateNodes(request *si.UpdateRequest) {
 func (cc *ClusterContext) addNodes(request *si.UpdateRequest) {
 	acceptedNodes := make([]*si.AcceptedNode, 0)
 	rejectedNodes := make([]*si.RejectedNode, 0)
+	// if we have an unlimited node, reject all new nodes
+	// if we have an unlimited node between the new nodes, register only that node
```
wilfred-s, line 594: This needs to be moved into the loop: nodes are linked to partitions. I _could_ have an unlimited node on one partition but not on the other:
- create a new node object
- get the partition: fail if partition not found (existing behaviour)
- check for the unlimited and limited combination:
  - if new node is unlimited registered partition nodes must be 0
  - if new node is limited registered nodes = 0 or larger than 1 OK otherwise check the single node and make sure it is not unlimited
- reject / accept nodes are processed as per the normal flow from this point

pkg/scheduler/context.go
```diff
@@ -844,3 +888,16 @@ func (cc *ClusterContext) GetNode(nodeID, partitionName string) *objects.Node {
 	}
 	return partition.GetNode(nodeID)
 }
+
+func (cc *ClusterContext) checkForUnlimitedNodes() bool {
```
wilfred-s, line 892: This check should be per partition. A node is linked to one partition. The call for this check must be after the partition retrieval above, see previous comment

wilfred-s (changes requested):
Looking much better and simpler: we should leverage the existing resource method for undefined.
Node registration needs a clean up and simplification

pkg/scheduler/context.go
kingamarton: We need to check for unlimited nodes cluster wide, no? If we moove it per partition, we might have an unlimited node in some other partition, than the onee we are checking. Is this acceptable?

pkg/scheduler/context.go
kingamarton: It is in the loop. I can moove itt after we check the partition, however we need some checks what did not depend on the partition, such as check if reservation is enabled, and check if theerare other nodes in teh `NewSchedulableNodes` list. I think we need this check as well, because if we check only the already registered nodes, the processing of the nodes will depend on the processing order and we might have different behaviours with the same imput list, what is not good.

I will moove the already reeegistered node check after we have the partition, and I will make it a per partition check, and I will keep the other checks at the beginning, since we don't need the partition for those checks and we can fail it earlier.

wilfred-s (approved):
lgtm.
Thank you for working with me on this one it has been a journey.

PR #298, opened 2021-08-11
YUNIKORN-792 Refactor sorting_policy.go

chia7712:
### What is this PR for?
there are two major changes.
1. remove policies.Undefined
1. move the code of logging unknown properties into separate function

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-792

### Questions:

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -40,6 +40,8 @@ const (
 	ApplicationSortPolicy = "application.sort.policy"
 )
 
+var KnownProperties = map[string]string{ApplicationSortPolicy: policies.DefaultSortPolicy.String()}
+
 // A queue can be a username with the dot replaced. Most systems allow a 32 character user name.
 // The queue name must thus allow for at least that length with the replacement of dots.
 var QueueNameRegExp = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,64}$`)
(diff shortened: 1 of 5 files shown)


wilfred-s:
I don't think this is a good idea at the moment. We need to finish the template first at that point we need to look at this.
We probably also want to go one step further:
* make a split between queue and application sorting at this time
* rename SortPolicy to ApplicationSorter (policy is implied) and give queue its own
* use an interface like implementation as we do for the node sorting.
* make Undefined the first value in the iota (default remains the Undefined value)
* update the way we process the policy and remove the need to call it from the partition for the root queue
* remove logging of unknown properties completely

Keeping Undefined as it will allow a next step, after this refactor, to specify a partition wide default in the config which we do not have as it is hard coded to Fair. The move of Undefined to the first entry also opens up an easier extension of types (like priority which has been talked about for a long time).

chia7712:
@wilfred-s thanks for the response. Is the Undefined also suitable to node sorting? It seems they have different designs as the similar flag (it was called Unknown) was removed recently. 

I will close this PR later :)

wilfred-s:
Yes I think we need to add it back again there too. I know that there is work in progress as a follow up on YUNIKORN-780 already and will take that up as part of that work.

PR #303, opened 2021-08-15
YUNIKORN-798 the REST APIs of getting queue application returns "Queu…

chia7712:
…e not found" even though the queue is existent

### What is this PR for?
the queue having no apps is considered nonexistent.
 
```
 	if !appExistsInQueue {
		buildJSONErrorResponse(w, "Queue not found", http.StatusBadRequest)
		return
	}
```

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-798

new unit test

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -687,28 +687,25 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 		buildJSONErrorResponse(w, "Incorrect URL path. Please check the usage documentation", http.StatusBadRequest)
 		return
 	}
-	var appExistsInPartition = false
-	var appExistsInQueue = false
-	var appsDao []*dao.ApplicationDAOInfo
+
 	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
-	if partitionContext != nil {
-		appExistsInPartition = true
-		appList := partitionContext.GetApplications()
-		for _, app := range appList {
-			if strings.EqualFold(queueName, app.GetQueueName()) {
-				appExistsInQueue = true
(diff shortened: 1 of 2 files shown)


pkg/webservice/handlers.go
```diff
@@ -687,28 +687,25 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-	var appExistsInPartition = false
-	var appExistsInQueue = false
-	var appsDao []*dao.ApplicationDAOInfo
+
 	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
-	if partitionContext != nil {
-		appExistsInPartition = true
-		appList := partitionContext.GetApplications()
```
chia7712, line 696: this call copy all applications of whole partition. this PR optimizes it by copying apps from only specify queue.

manirajv06 (commented):
Yes, "Queue not found" error message is misleading when queue doesn't have any applications. It should go something like "No applications.." etc.

Yes, optimisation is good and main outcome of this JIRA. Some minor comments..

chia7712:
> It should go something like "No applications.." etc.

Did you mean it should return error response if the queue has no applications? If so, it seems to me returning empty array is better way. It is normal that a queue has no application. Hence, the empty array is more suitable to be the response of such REST apis.

manirajv06 (approved):
Yes, returning [] as successful response makes sense. LGTM.

PR #304, opened 2021-08-16
[YUNIKORN-799] Fix NPE in application#GetQueueName

chia7712:
### What is this PR for?
the queue of application is reset to nil after application is completed. Hence, application#GetQueueName should return `QueueName` rather than `queue.QueuePath` to avoid NPE. Querying a cluster which has a completed application by REST API "/ws/v1/apps" can produce the NPE.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-799

new unit test

### Questions:

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -481,7 +481,7 @@ func (cc *ClusterContext) processApplications(request *si.UpdateRequest) {
 			zap.String("applicationID", app.ApplicationID),
 			zap.String("partitionName", app.PartitionName),
 			zap.String("requested queue", app.QueueName),
-			zap.String("placed queue", schedApp.GetQueueName()))
+			zap.String("placed queue", schedApp.GetQueuePath()))
 	}
 
 	// Respond to RMProxy with accepted and rejected apps if needed
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1112,7 +1112,7 @@ func (sa *Application) tryNode(node *Node, ask *AllocationAsk) *Allocation {
 func (sa *Application) GetQueueName() string {
 	sa.RLock()
 	defer sa.RUnlock()
-	return sa.queue.QueuePath
+	return sa.QueueName
 }
 
 func (sa *Application) GetQueue() *Queue {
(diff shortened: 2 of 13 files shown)


wilfred-s (changes requested):
This breaks the logging that is in `processApplications()` as part of the placing of an app in the queue. We explicitly want the queue path for the logging to show where exactly the app is placed. Changing `QueueName` to contain the full path will break other things.

We might need to change what is returned (i.e. change `GetQueueName` to return a local copy of the value. This probably means a change to `GetQueuePath` for the method also.

chia7712:
@wilfred-s thanks for this quick feedback.

> This breaks the logging that is in processApplications() as part of the placing of an app in the queue. We explicitly want the queue path for the logging to show where exactly the app is placed. Changing QueueName to contain the full path will break other things.

the logging happens only if `addApplication` succeeds (https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/scheduler/context.go#L466). The `addApplication` (https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/scheduler/partition.go#L366) updates both `queue` and `QueueName` at the same time so `QueueName` always contain the full path when `processApplications` logs it.

chia7712:
> We might need to change what is returned (i.e. change GetQueueName to return a local copy of the value. This probably means a change to GetQueuePath for the method also.

The latest commit includes a part of this comment. `GetQueueName` is renamed to `GetQueuePath`

wilfred-s:
@chia7712 please do not use force pushes, just add new commits to the branch. Force pushes remove the history and when reviewing it is often simpler to look at the specific change instead of everything from scratch. The final commit into the repo will squash all the commits into one anyway.

Also to make commits and linking easier make sure your first commit message has the layout: [_JIRA-ID_] \<description>

wilfred-s (approved):
LGTM

chia7712:
@wilfred-s thanks for your kind reminder!

PR #306, opened 2021-08-18
[YUNIKORN-804] ws/v1/clusters returns invalid JSON when there are mul…

chia7712:
…tiples partitions

### What is this PR for?
it returns "two arrays" when there are two partitions. We should merge them into single array. Also, the partition name should be included by each element.
This PR changes the API format of multiple partitions. However, it should be fine as we should not rely on invalid JSON.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-804

### Questions:


pkg/webservice/dao/cluster_info.go
```diff
@@ -18,6 +18,7 @@
 package dao
 
 type ClusterDAOInfo struct {
+	PartitionName         string `json:"partition"`
```
chia7712, line 21: I will file another PR to update APIs example.

manirajv06 (changes requested):
Can we create a new test method to keep the asserts rather than modifying TestGetPartitionQueuesHandler?

pkg/webservice/dao/cluster_info.go
chia7712: https://issues.apache.org/jira/browse/YUNIKORN-824

chia7712:
> Can we create a new test method to keep the asserts rather than modifying TestGetPartitionQueuesHandler?

sure!

manirajv06 (approved):

PR #307, opened 2021-08-19
[YUNIKORN-807] Improve performance of node sorting.

craigcondit:
### What is this PR for?
Improves the performance of node sorting by maintaining an ordered B-Tree of nodes rather than sorting the entire node list on demand. This patch does not introduce any external behavior changes.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-807

All local unit tests pass. Scheduling under load should be verified.

N/A

### Questions:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/HdrHistogram/hdrhistogram-go v1.0.1 // indirect
 	github.com/apache/incubator-yunikorn-scheduler-interface v0.11.1-0.20210825113556-88f21e44dc01
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/google/btree v1.0.1 // indirect
 	github.com/google/uuid v1.2.0
 	github.com/gorilla/mux v1.7.3
 	github.com/looplab/fsm v0.1.0
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/HdrHistogram/hdrhistogram-go v1.0.1 // indirect
 	github.com/apache/incubator-yunikorn-scheduler-interface v0.11.1-0.20210825113556-88f21e44dc01
 	github.com/beorn7/perks v1.0.1 // indirect
-	github.com/google/btree v1.0.1 // indirect
+	github.com/google/btree v1.0.1
 	github.com/google/uuid v1.2.0
 	github.com/gorilla/mux v1.7.3
 	github.com/looplab/fsm v0.1.0
--- a/pkg/scheduler/objects/node.go
+++ b/pkg/scheduler/objects/node.go
@@ -52,6 +52,7 @@ type Node struct {
 
 	preempting   *resources.Resource     // resources considered for preemption
 	reservations map[string]*reservation // a map of reservations
+	listeners    []NodeListener          // a list of node listeners
 
 	sync.RWMutex
 }
@@ -70,6 +71,7 @@ func NewNode(proto *si.NewNodeInfo) *Node {
 		occupiedResource:  resources.NewResourceFromProto(proto.OccupiedResource),
 		allocations:       make(map[string]*Allocation),
 		schedulable:       true,
+		listeners:         make([]NodeListener, 0),
 	}
 	// initialise available resources
 	var err error
@@ -137,6 +139,7 @@ func (sn *Node) GetCapacity() *resources.Resource {
 }
 
 func (sn *Node) SetCapacity(newCapacity *resources.Resource) *resources.Resource {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 	if resources.Equals(sn.totalResource, newCapacity) {
@@ -156,6 +159,7 @@ func (sn *Node) GetOccupiedResource() *resources.Resource {
 }
 
 func (sn *Node) SetOccupiedResource(occupiedResource *resources.Resource) {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 	if resources.Equals(sn.occupiedResource, occupiedResource) {
@@ -208,6 +212,7 @@ func (sn *Node) GetAllAllocations() []*Allocation {
 // This will cause the node to be skipped during the scheduling cycle.
 // Visible for testing only
 func (sn *Node) SetSchedulable(schedulable bool) {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 	sn.schedulable = schedulable
@@ -250,6 +255,7 @@ func (sn *Node) getPreemptingResource() *resources.Resource {
 
 // Update the number of resource tagged for preemption on this node
 func (sn *Node) IncPreemptingResource(preempting *resources.Resource) {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 
@@ -257,6 +263,7 @@ func (sn *Node) IncPreemptingResource(preempting *resources.Resource) {
 }
 
 func (sn *Node) decPreemptingResource(delta *resources.Resource) {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 	var err error
@@ -273,6 +280,7 @@ func (sn *Node) decPreemptingResource(delta *resources.Resource) {
 // is found the Allocation removed is returned. Used resources will decrease available
 // will increase as per the allocation removed.
 func (sn *Node) RemoveAllocation(uuid string) *Allocation {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 
@@ -294,6 +302,7 @@ func (sn *Node) AddAllocation(alloc *Allocation) bool {
 	if alloc == nil {
 		return false
 	}
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 	// check if this still fits: it might have changed since pre check
@@ -310,6 +319,7 @@ func (sn *Node) AddAllocation(alloc *Allocation) bool {
 // Replace the paceholder allocation on the node. No usage changes as the placeholder must
 // be the same size as the real allocation.
 func (sn *Node) ReplaceAllocation(uuid string, replace *Allocation) {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 
@@ -459,6 +469,7 @@ func (sn *Node) isReservedForApp(key string) bool {
 // The reservation is checked against the node resources.
 // If the reservation fails the function returns false, if the reservation is made it returns true.
 func (sn *Node) Reserve(app *Application, ask *AllocationAsk) error {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 	if len(sn.reservations) > 0 {
@@ -492,6 +503,7 @@ func (sn *Node) Reserve(app *Application, ask *AllocationAsk) error {
 // If the reservation does not exist it returns 0 for reservations removed, if the reservation is removed it returns 1.
 // The error is set if the reservation key cannot be generated.
 func (sn *Node) unReserve(app *Application, ask *AllocationAsk) (int, error) {
+	defer sn.notifyListeners()
 	sn.Lock()
 	defer sn.Unlock()
 	resKey := reservationKey(nil, app, ask)
@@ -538,4 +550,39 @@ func (sn *Node) UnReserveApps() ([]string, []int) {
 		askRelease = append(askRelease, num)
 	}
 	return appReserve, askRelease
+}
+
+func (sn *Node) AddListener(listener NodeListener) {
+	sn.Lock()
+	defer sn.Unlock()
+	sn.listeners = append(sn.listeners, listener)
+}
+
+func (sn *Node) RemoveListener(listener NodeListener) {
+	sn.Lock()
+	defer sn.Unlock()
+
+	newListeners := make([]NodeListener, 0)
+	for _, entry := range sn.listeners {
+		if entry == listener {
+			continue
+		}
+		newListeners = append(newListeners, entry)
+	}
+	sn.listeners = newListeners
+}
+
+// Notifies listeners of changes to this node. This method must not be called while locks are held.
+func (sn *Node) notifyListeners() {
+	for _, listener := range sn.getListeners() {
+		listener.NodeUpdated(sn)
+	}
+}
+
+func (sn *Node) getListeners() []NodeListener {
+	sn.RLock()
+	defer sn.RUnlock()
+	list := make([]NodeListener, len(sn.listeners))
+	copy(list, sn.listeners)
+	return list
 }
--- a/pkg/scheduler/objects/node_collection.go
+++ b/pkg/scheduler/objects/node_collection.go
@@ -36,7 +36,7 @@ type NodeCollection interface {
 	GetNode(nodeID string) *Node
 	GetNodeCount() int
 	GetNodes() []*Node
-	GetSchedulableNodeIterator() NodeIterator
+	GetNodeIterator() NodeIterator
 	SetNodeSortingPolicy(policy NodeSortingPolicy)
 	GetNodeSortingPolicy() NodeSortingPolicy
 }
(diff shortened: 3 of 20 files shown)


craigcondit:
Baseline performance (current master):

```
[pkg/scheduler/tests] $ time (go test -bench BenchmarkScheduling -run=^$ | grep performance_test.go)
    performance_test.go:117: Total time to add 500 node in 31.182905ms, 16034.426555 per second
    performance_test.go:172: Total time to allocate 10000 containers in 27.813578526s, 359.536620 per second
    performance_test.go:117: Total time to add 1000 node in 50.780149ms, 19692.734655 per second
    performance_test.go:172: Total time to allocate 10000 containers in 56.294724715s, 177.636538 per second
    performance_test.go:117: Total time to add 2000 node in 90.955818ms, 21988.697853 per second
    performance_test.go:172: Total time to allocate 10000 containers in 1m49.266661833s, 91.519223 per second
    performance_test.go:117: Total time to add 5000 node in 254.403584ms, 19653.811166 per second
    performance_test.go:172: Total time to allocate 10000 containers in 4m57.678494932s, 33.593290 per second

real	8m14.938s
user	46m11.660s
sys	0m20.570s
```

B-Tree Performance (this patch):

```
[pkg/scheduler/tests] $ time (go test -bench BenchmarkScheduling -run=^$ | grep performance_test.go)
    performance_test.go:117: Total time to add 500 node in 30.549695ms, 16366.775511 per second
    performance_test.go:172: Total time to allocate 10000 containers in 854.16043ms, 11707.402554 per second
    performance_test.go:117: Total time to add 500 node in 30.830001ms, 16217.968984 per second
    performance_test.go:172: Total time to allocate 10000 containers in 765.890195ms, 13056.701947 per second
    performance_test.go:117: Total time to add 500 node in 31.055905ms, 16099.997730 per second
    performance_test.go:172: Total time to allocate 10000 containers in 906.42909ms, 11032.302593 per second
    performance_test.go:117: Total time to add 500 node in 30.945478ms, 16157.449563 per second
    performance_test.go:172: Total time to allocate 10000 containers in 800.436131ms, 12493.189166 per second
    performance_test.go:117: Total time to add 500 node in 30.565935ms, 16358.079673 per second
    performance_test.go:172: Total time to allocate 10000 containers in 797.422808ms, 12540.398769 per second
    performance_test.go:117: Total time to add 1000 node in 51.26442ms, 19506.706601 per second
    performance_test.go:172: Total time to allocate 10000 containers in 1.344201034s, 7439.363419 per second
    performance_test.go:117: Total time to add 2000 node in 115.78496ms, 17273.400621 per second
    performance_test.go:172: Total time to allocate 10000 containers in 2.137938513s, 4677.402993 per second
    performance_test.go:117: Total time to add 5000 node in 324.152488ms, 15424.839189 per second
    performance_test.go:172: Total time to allocate 10000 containers in 3.953890445s, 2529.154548 per second

real	0m40.619s
user	10m51.105s
sys	0m17.104s
```

craigcondit:
Linter failure is due to an existing unused method. It is retriggering here because the method signature changed.

pkg/scheduler/objects/node_collection.go
```diff
@@ -0,0 +1,245 @@
...
+	nc.activeNodes.ReplaceOrInsert(*nref)
+}
+
+// Create a new collection for the given partition.
+func NewNodeCollection(partition string) NodeCollection {
+	return &baseNodeCollection{
+		Partition:   partition,
+		nsp:         nil,
+		nodes:       make(map[string]*nodeRef),
+		activeNodes: btree.New(5),
```
yangwwei, line 243: why degree=5? any reason behind this? Really want to understand more : )

pkg/scheduler/objects/nodesorting.go
```diff
@@ -31,6 +31,7 @@ import (
 type NodeSortingPolicy interface {
 	PolicyType() policies.SortingPolicy
 	SortNodes(nodes []*Node)
```
yangwwei, line 33: is `SortNodes()` still needed?
My understanding is that now we just need to score the nodes

pkg/scheduler/partition.go
```diff
@@ -622,11 +612,10 @@ func (pc *PartitionContext) updatePartitionResource(delta *resources.Resource) {
 func (pc *PartitionContext) addNodeToList(node *objects.Node) error {
 	pc.Lock()
 	defer pc.Unlock()
-	if pc.nodes[node.NodeID] != nil {
+	// Node can be added to the system to allow processing of the allocations
+	if pc.nodes.AddNode(node) != nil {
```
yangwwei, line 616: it's better to read the error message from the returned err:

```
if err := nc.nodes[node.NodeID]; err != nil {
  return fmt.Errorf("failed to add node %s to partition %s, error: %v", nc.Partition, node.NodeID, err)
}
```

pkg/scheduler/objects/nodesorting.go
```diff
@@ -62,6 +63,16 @@ func (fairnessNodeSortingPolicy) SortNodes(nodes []*Node) {
...
+func (binPackingNodeSortingPolicy) ScoreNode(node *Node) float64 {
+	// choose most loaded node first
+	return resources.LargestUsageShare(node.GetAvailableResource())
+}
+
+func (fairnessNodeSortingPolicy) ScoreNode(node *Node) float64 {
+	// choose least loaded node first
+	return -resources.LargestUsageShare(node.GetAvailableResource())
+}
+
```
yangwwei, line 75: our current node sorting algorithm is problematic... it is not caused by this PR, but still want to raise it up for attention. Simply using largest resource share for calculating the score is wrong, we need to consider all the resources. when there is only cpu and memory, memory will be always leading the calculation; but in many cases, cpu is a more expensive resource, our result is suboptimal. 
As a reference, we can look at how this is done in: https://github.com/kubernetes/kubernetes/tree/master/pkg/scheduler/framework/plugins/noderesources. we need something similar for our FAIR & Bin-packing policies. We can have a follow up JIRA for this.

yangwwei (changes requested):
hi @craigcondit thanks for working on this. I've gone through the PR, it looks pretty good. I left some comments. There is one comment I think we need input from @wilfred-s  as well. I've cc'd him specifically for that one. Thanks, great work!

pkg/scheduler/objects/node.go
```diff
@@ -128,6 +130,7 @@ func (sn *Node) GetCapacity() *resources.Resource {
 }
 
 func (sn *Node) SetCapacity(newCapacity *resources.Resource) *resources.Resource {
+	defer sn.notifyListeners()
```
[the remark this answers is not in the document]
craigcondit: @yangwwei, you are correct that notification is not strictly necessary in the short-circuit return. However, it is vital to avoid deadlocks that notifyListeners() is only called when the lock is not held, which is why we defer the notifyListeners() call to the end. defer statements are executed last-in, first-out, so this ensures that we have releases our lock before calling it. Conversely, we need the lock to be able to check the condition in line 137, so we have a chicken-and-egg problem with clearly taking / releasing locks when needed. The current approach may not be 100% optimal, but it is easy to reason about and causes no locking or deadlock issues.

pkg/scheduler/objects/node.go
```diff
@@ -147,6 +150,7 @@ func (sn *Node) GetOccupiedResource() *resources.Resource {
 }
 
 func (sn *Node) SetOccupiedResource(occupiedResource *resources.Resource) {
+	defer sn.notifyListeners()
```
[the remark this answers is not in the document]
craigcondit: See above.

pkg/scheduler/objects/node.go
```diff
@@ -264,6 +271,7 @@ func (sn *Node) decPreemptingResource(delta *resources.Resource) {
 // is found the Allocation removed is returned. Used resources will decrease available
 // will increase as per the allocation removed.
 func (sn *Node) RemoveAllocation(uuid string) *Allocation {
+	defer sn.notifyListeners()
```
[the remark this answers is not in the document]
craigcondit: logically yes, however, that would result in calling notifyListeners() with a lock held, leading to deadlock.

pkg/scheduler/objects/node.go
```diff
@@ -285,6 +293,7 @@ func (sn *Node) AddAllocation(alloc *Allocation) bool {
 	if alloc == nil {
 		return false
 	}
+	defer sn.notifyListeners()
```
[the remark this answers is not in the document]
craigcondit: Same as above.

pkg/scheduler/objects/node.go
```diff
@@ -505,3 +517,38 @@ func (sn *Node) UnReserveApps() ([]string, []int) {
...
+	}
+}
+
+func (sn *Node) getListeners() []NodeListener {
+	sn.RLock()
+	defer sn.RUnlock()
+	list := make([]NodeListener, len(sn.listeners))
+	copy(list, sn.listeners)
+	return list
+}
```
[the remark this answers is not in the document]
craigcondit: Listeners may take arbitrary time to execute, so holding a lock is ill-advised. More importantly, the listener will probably call back into the node to query information, leading to double-acquiring of the lock, which is not allowed and leads to a deadlock condition.

pkg/scheduler/objects/node_collection.go
```diff
@@ -0,0 +1,245 @@
...
+	// Node can be added to the system to allow processing of the allocations
+	node.AddListener(nc)
+	nref := nodeRef{
+		node:      node,
+		nodeScore: nc.scoreNode(node),
+	}
+	nc.nodes[node.NodeID] = &nref
+	if node.IsSchedulable() && !node.IsReserved() {
+		nc.activeNodes.ReplaceOrInsert(nref)
+	}
```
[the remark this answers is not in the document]
craigcondit: This PR is focused on maintaining existing behavior but improving performance. I'd prefer to avoid behavior changes in the code until they are needed. Could we defer that to a future ticket?

pkg/scheduler/objects/node_collection.go
craigcondit: The degree determines the minimum number of items per tree node, and this seemed like a reasonable choice given upper bound of nodes would be maybe 10k at most. Consider a "full" tree of degree 5:

level 0 (root): 5 children
level 1: 5 * 5 children = 25
level 2: 25 * 5 children = 625
level 3: 625 * 5 children = 3125
level 4: 3125 * 5 children = 15625

Degree=5 means we can handle > 3k nodes with only depth=3, and > 15k with depth=4. At the same time, we only have max of 5 nodes to compare against at each level.

pkg/scheduler/objects/nodesorting.go
craigcondit: Agree the current algorithm needs work, and that we should address in a followup JIRA.

pkg/scheduler/partition.go
craigcondit: Agreed. Will update this.

pkg/scheduler/objects/nodesorting.go
craigcondit: I'm a little hesitant to remove this just yet... there may still be a need for this moving forward?

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/HdrHistogram/hdrhistogram-go v1.0.1 // indirect
 	github.com/apache/incubator-yunikorn-scheduler-interface v0.11.0
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/google/btree v1.0.1 // indirect
```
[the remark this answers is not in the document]
craigcondit: not sure; go get added it. possibly because the main module in core doesn't reference it?

craigcondit:
@yangwwei, I've rebased on master and made the err-handling change you suggested. Can you re-review?

pkg/scheduler/objects/node.go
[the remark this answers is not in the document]
yangwwei: I see, thanks for the clarification.
I guess another approach is to send updated nodes to a channel, and in the node collection object, we run a goroutine to select messages from the channel to process. That can solve the locking problem as well.  Have you considered that option?

pkg/scheduler/objects/node_collection.go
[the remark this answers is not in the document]
yangwwei: Make sense, let's take care of that in different JIRAs.

pkg/scheduler/objects/node_collection.go
yangwwei: Is this correct? 
```
level 2: 25 * 5 children = 625
```
shouldn't this be 125? does this mean we need depth=4 to handle 3k+ nodes, and depth=5 to handle 15k+ nodes?
what is the trade-off to have more items per tree node comparing to fewer items? I am asking because we may want to achieve the best results for less than 2000 nodes.

pkg/scheduler/objects/nodesorting.go
yangwwei: +1

yangwwei (changes requested):

pkg/scheduler/objects/node.go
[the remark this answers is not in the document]
craigcondit: We don't want to introduce an async process to this. The current goroutine is waiting on us to return the list of candidate nodes. So when we release the current node lock, and reacquire it in the node listener itself, most of the time there will be no other waiters on that lock so it's very cheap (any other updates would likely be due to resource availability changes from a container starting or stopping, and these are infrequent enough to be considered an edge case). So we do our work (listener basically just queries the node to get updated info) and return quickly.

If we spawn another goroutine in the background, now we are returning back to our caller without having done the node update. And because we don't wait for the goroutine to finish, we could have several stack up and make it very difficult to reason about the current state of things.

The code as written is admittedly not the most efficient possible, but it is simple, and easy to follow. Additionally, in the worst case, we have a few updates that remove a node from the btree and re-add it unnecessarily, but these occur in the edge cases and don't impact performance in a meaningful way.

pkg/scheduler/objects/node_collection.go
craigcondit: Good catch. I think we want to optimize for larger cases though, as with smaller cluster sizes the number of containers we need to schedule per second is also less, so any algorithmic inefficiencies are mitigated.

I decided to do a little benchmarking to see if there's some clarity around an ideal degree value, and the results are interesting.

I ran 4 runs in a row for each of degree=3,4,5,6,7,8. I threw out the first run as a 'warm up' and then averaged the remaining 3 and calculated variance as well.

There's a lot of variance in the data, but surprisingly, all results show a net loss of throughput on 500 nodes vs. degree=5. So we appear to be handling that case optimally now.

At 1000 nodes, there's a modest (7-9%) gain in throughput for degree=6 and degree=7, and little change with 3, 4 or 8.
At 2000 nodes, we start to see significant gains (10%-18%) for degree=6 and degree=7, and again minimal change with 3, 4 or 8.
At 5000 nodes, the gains are again significant (15-16%) for degree=6,7, 11% for 8, with minimal change on 3 or 4.

It seems that degree=7 is optimal on >= 1000 nodes, with the largest gains of all the variants in best-case, average-case, and worst-case vs. 5. 

I also ran a few tests with very large (up to degree=10000), and the conclusion I come to is that the cost doesn't increase too much except at the very end (where we basically convert our B-Tree into a single node with a child list).

I'm updated the code to use 7, since that seems to be the optimal case, but the benchmarking seems to indicate that we're pretty safe regardless.

yangwwei:
hi @craigcondit could you pls fix the conflict,  since there is no more review comments, once the conflict is solved, I think we can get this one merged, thanks!

craigcondit:
Rebased on master and fixed compile / test / lint issues that came up after [YUNIKORN-812](https://issues.apache.org/jira/browse/YUNIKORN-812). However, due to the changes to reservation in that PR, I'm a little concerned that we may be missing some logic. @yangwwei, do you have any concerns?

craigcondit:
Latest patch removes some dead code and simplifies NodeCollection to include all nodes in the sorted list regardless of active / schedulable / reserved status.

craigcondit:
Updated patch to return only non-reserved nodes in iterator to be consistent with YUNIKORN-812. Also renamed node iterator method since it now includes non-schedulable nodes. Updated comments to clearly reflect the new contract.

yangwwei (approved):
All look good, I will commit this soon, thank you @craigcondit !

PR #308, opened 2021-08-20
[YUNIKORN-814] The template of non-leaf queue should not be override …

chia7712:
…by template of parent

### What is this PR for?
This is a bug introduced by YUNIKORN-193. The non-leaf queue get template from conf but the template is override when it is added to parent queue.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-814

### Questions:


pkg/scheduler/objects/queue.go
```diff
@@ -656,7 +656,10 @@ func (sq *Queue) addChildQueue(child *Queue) error {
 		}
 		// managed (configured) leaf queue can't use template
 	} else {
-		child.template = sq.template
+		// don't override the template of non-leaf queue
+		if child.template == nil {
```
wilfred-s, line 660: Please add logging for this case, something like inheriting child template for queue...

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
chia7712: done!

chia7712:
@wilfred-s Could you please take a look? the docs (https://github.com/apache/incubator-yunikorn-site/pull/73) is updated so this PR is required now :)

kingamarton (approved):

PR #311, opened 2021-08-21
[YUNIKORN-817] Avoid NPE in /ws/v1/nodes/utilization REST API when partition has no node

chia7712:
… no node

### What is this PR for?
`PartitionContext#totalPartitionResource` is null if the partition has no node. We need to check null before converting it to dao object. 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-817

### Questions:


yangwwei (approved):
+1

PR #312, opened 2021-08-22
[YUNIKORN-818] the response of REST APIs should use empty array rathe…

chia7712:
…r than null

### What is this PR for?
It seems to me empty array is more suitable than null.  The benefit is that user don't need to check null before looping the value. Maybe some json tools can convert null array to empty automatically. However, using empty array to replace null can save the rest of json tools (and programmers) from null hell. 

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-818

### Questions:


manirajv06 (approved):
changes looks good. thanks.

PR #318, opened 2021-08-31
[YUNIKORN-831] node sorting should check other resources of nodes instead of comparing node id directly

chia7712:
…tead of comparing node id directly

### What is this PR for?
Node sorting scores node according to `resources.LargestUsageShare`. However, the largest usage in our cluster is always storage (and all nodes have same storage capacity). It results in the order is always based on `node id`. Before this patch, the order will compare remaining "usage" (vcore, memory, etc) when largest usage is equal.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-831

### Questions:

### DIFF
--- a/pkg/scheduler/objects/node_collection.go
+++ b/pkg/scheduler/objects/node_collection.go
@@ -26,6 +26,7 @@ import (
 	"github.com/google/btree"
 	"go.uber.org/zap"
 
+	"github.com/apache/incubator-yunikorn-core/pkg/common/resources"
 	"github.com/apache/incubator-yunikorn-core/pkg/log"
 	"github.com/apache/incubator-yunikorn-core/pkg/metrics"
 )
@@ -42,8 +43,8 @@ type NodeCollection interface {
 }
 
 type nodeRef struct {
-	node      *Node   // node reference
-	nodeScore float64 // node score
+	node      *Node     // node reference
+	nodeScore []float64 // node score
 }
 
 func (nr nodeRef) Less(than btree.Item) bool {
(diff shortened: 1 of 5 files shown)


pkg/scheduler/objects/node_collection.go
```diff
@@ -211,13 +213,7 @@ func (nc *baseNodeCollection) NodeUpdated(node *Node) {
...
 	defer nc.Unlock()
 
-	nref := nc.nodes[node.NodeID]
-	if nref == nil {
-		return
-	}
-
-	updatedScore := nc.scoreNode(node)
-	if nref.nodeScore != updatedScore {
+	if nref, ok := nc.nodes[node.NodeID]; ok {
```
craigcondit, line 216: I don't think this is correct. When NodeUpdated() is called, it might be due to usage values changing, and therefore we need to calculate a new score and update nref.nodeScore. If the score has changed, the node needs to be removed from the BTree (looked up using previous version of nref) and re-added using the updated nref.

craigcondit (commented):
Generally looks pretty good, but I'd like @yangwwei to look at it as well. I did leave a comment regarding NodeUpdated(), as I think that is no longer correct.

pkg/scheduler/objects/node_collection.go
craigcondit: Actually, I may be mistaken. It looks like you're unconditionally removing and re-adding the node? If so, that also works.

chia7712: >  It looks like you're unconditionally removing and re-adding the node?

yep. That may be a small performance issue since it removes and re-add the node even if the resources are not changed. However, the code of checking the changes gets more complicated after the type is changed from `float64` to `[]float64`. I prefer to update node unconditionally to keep code simpler.

craigcondit: Makes sense.

craigcondit:
Upon testing this locally, it seems that we really do need to consider the fact that resources have different units. I've opened a new issue https://issues.apache.org/jira/browse/YUNIKORN-847 with PR https://github.com/apache/incubator-yunikorn-core/pull/323 that provides an alternative approach. The new PR uses configurable weights for each resource and normalizes resources into percentage occupied rather than considering the absolute values. By default, weights are equal for vcore and memory, and 0 for everything else (effectively ignoring them).

chia7712:
close as duplicate

PR #320, opened 2021-09-07
[YUNIKORN-838] Improve coverage of allocation_ask.go

0yukali0:
### What is this PR for?
Add unit test to improve coverage

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-838

Use command "make test" to create coverage.txt and then use "go tool cover -func=coverage.txt" to check

### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation_ask_test.go
+++ b/pkg/scheduler/objects/allocation_ask_test.go
@@ -158,35 +158,55 @@ func TestGetRequiredNode(t *testing.T) {
 
 func TestSetPriority(t *testing.T) {
 	var tests = []struct {
-		testMessage string
-		in          int32
-		out         int32
+		testMessage           string
+		in                    int32
+		out                   int32
+		isSiArg, isReassigned bool
+		reassign              int32
 	}{
-		{"no priority setting", -1, 0},
-		{"set default priority value", 0, 0},
-		{"set priority", 2, 2},
+		// Set priorirty via NewAllocationAsk
+		{"No priority setting", -1, 0, false, false, 0},
+		{"Set priority via new", 3, 3, true, false, 0},
+		// Set priority via SetPriority function
+		{"Set default priority value", 0, 0, false, false, 0},
+		{"Set priority", 2, 2, false, false, 0},
+		// change priority
+		{"Change priority", 2, 5, true, true, 5},
 	}
 
 	for _, tt := range tests {
 		res := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 10})
-		siAsk := &si.AllocationAsk{
-			AllocationKey:  "ask-1",
-			ApplicationID:  "app-1",
-			MaxAllocations: 1,
-			ResourceAsk:    res.ToProto(),
-		}
 		t.Run(tt.testMessage, func(t *testing.T) {
-			ask := NewAllocationAsk(siAsk)
-			if tt.in >= 0 {
-				priority := si.Priority_PriorityValue{PriorityValue: tt.in}
+			siAsk := &si.AllocationAsk{
+				AllocationKey:  "ask-1",
+				ApplicationID:  "app-1",
+				MaxAllocations: 1,
+				ResourceAsk:    res.ToProto(),
+			}
+
+			priority := si.Priority_PriorityValue{PriorityValue: tt.in}
+			if tt.isSiArg {
 				siAsk.Priority = &si.Priority{
 					Priority: &priority,
 				}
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/allocation_ask_test.go
```diff
@@ -155,3 +155,39 @@ func TestGetRequiredNode(t *testing.T) {
...
+			ResourceAsk:    res.ToProto(),
+		}
+		t.Run(tt.testMessage, func(t *testing.T) {
+			ask := NewAllocationAsk(siAsk)
+			if tt.in >= 0 {
+				priority := si.Priority_PriorityValue{PriorityValue: tt.in}
+				siAsk.Priority = &si.Priority{
+					Priority: &priority,
+				}
+				ask.setPriority(siAsk.Priority.GetPriorityValue())
```
yangwwei, line 185: I guess it's better to do this in a different way because this is not how we usually assign the priority value.
can we assign the value directly in `AllocationAsk`?

```
siAsk := &si.AllocationAsk{
  AllocationKey:  "ask-1",
  ApplicationID:  "app-1",
  MaxAllocations: 1,
  Priority:   xxxx,
  ResourceAsk:    res.ToProto(),
}
```

yangwwei (changes requested):

pkg/scheduler/objects/allocation_ask_test.go
0yukali0: Hi @yangwwei , thank for review.
Priority is usually assigned to si.AllocationAsk and handled via NewAllocationAsk function. 
I will additionally add this example in UT.
Test two ways which are NewAllocationAsk and SetPriority function to set priority.

0yukali0: I add two situation in this commit.
1.Set priority via "NewAllocationAsk" function
2.Priority initialized by "NewAllocationAsk"  is  assigned new value by SetPriority.

yangwwei:
hi @0yukali0  sorry for the late response. I just looked more into this PR.
The priority actually has 2 forms, 1 is with the int value, the other is the priorityClass which is a string value, today the latter form is not fully implemented, but that's part of the proto, so it will be good to cover that as well. And regarding the UT itself, I think it can be simplified for easier reading. Below is what I thought in my mind, pls take a look:

```
func newPriorityValue(v int32) *si.Priority{
	return &si.Priority{
		Priority: &si.Priority_PriorityValue{
			PriorityValue: v},
	}
}

func newPriorityClass(priorityClassName string) *si.Priority {
	return &si.Priority{
		Priority: &si.Priority_PriorityClassName{
			PriorityClassName: priorityClassName},
	}
}

func TestSetPriority(t *testing.T) {
	var tests = []struct {
		testMessage           string
		priority              *si.Priority
		out                   int32
	}{
		// Set priorirty via NewAllocationAsk
		{"No priority setting",nil, 0},
		{"Set empty priority", &si.Priority{}, 0},
		{"Set positive priority value", newPriorityValue(1), 1},
		{"Set negative priority value", newPriorityValue(-100), -100},
		{"Set priority class", newPriorityClass("p0"), 0},
	}

	for _, tt := range tests {
		res := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 10})
		t.Run(tt.testMessage, func(t *testing.T) {
			siAsk := &si.AllocationAsk{
				AllocationKey:  "ask-1",
				ApplicationID:  "app-1",
				MaxAllocations: 1,
				Priority:       tt.priority,
				ResourceAsk:    res.ToProto(),
			}

			ask := NewAllocationAsk(siAsk)
			if ask.priority != tt.out {
				t.Errorf("result %s, Want %v, got %v", tt.testMessage, tt.out, ask.priority)
			}
		})
	}
}
```

0yukali0:
Got it.

yangwwei:
@0yukali0  can u pls update the PR?

0yukali0:
Got it

Weiwei Yang ***@***.***> 於 2021年9月13日 週一 下午11:04 寫道：

0yukali0:
Hi @yangwwei , setPriority funtion  is not used in yunikorn-core now, should i add it into UT?
https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/scheduler/objects/allocation_ask.go#L134-L138
image

yangwwei (approved):
+1

PR #322, opened 2021-09-14
[YUNIKORN-834] Unify the "partition name" in REST APIs

chia7712:
### What is this PR for?
The partition name in response always contains the cluster id. By contrast, the partition name in URL does not include cluster id. That is weird and inconvenient to APIs users because it needs to parse partition name from the normalized partition name.
It seems to me the partition name in response should exclude cluster id. The cluster id should be a separate field in response.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-834

### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation.go
+++ b/pkg/scheduler/objects/allocation.go
@@ -118,6 +118,10 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
 	return NewAllocation(alloc.UUID, alloc.NodeID, ask)
 }
 
+func (a *Allocation) GetPartitionNameWithoutClusterID() string {
+	return common.GetPartitionNameWithoutClusterID(a.PartitionName)
+}
+
 // Convert the Allocation into a SI object. This is a limited set of values that gets copied into the SI.
 // We only use this to communicate *back* to the RM. All other fields are considered incoming fields from
 // the RM into the core.
--- a/pkg/scheduler/objects/allocation.go
+++ b/pkg/scheduler/objects/allocation.go
@@ -118,10 +118,6 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
 	return NewAllocation(alloc.UUID, alloc.NodeID, ask)
 }
 
-func (a *Allocation) GetPartitionNameWithoutClusterID() string {
-	return common.GetPartitionNameWithoutClusterID(a.PartitionName)
-}
-
 // Convert the Allocation into a SI object. This is a limited set of values that gets copied into the SI.
 // We only use this to communicate *back* to the RM. All other fields are considered incoming fields from
 // the RM into the core.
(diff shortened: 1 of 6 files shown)


pkg/scheduler/objects/allocation.go
```diff
@@ -118,6 +118,10 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
 	return NewAllocation(alloc.UUID, alloc.NodeID, ask)
 }
 
+func (a *Allocation) GetPartitionNameWithoutClusterID() string {
+	return common.GetPartitionNameWithoutClusterID(a.PartitionName)
```
manirajv06, line 122: Do we need this method? "common" scope is already available in handlers.go. Can we use it directly in handlers.go?

pkg/scheduler/objects/application.go
```diff
@@ -128,6 +128,10 @@ func (sa *Application) SetState(state string) {
 	sa.stateMachine.SetState(state)
 }
 
+func (sa *Application) GetPartitionNameWithoutClusterID() string {
+	return common.GetPartitionNameWithoutClusterID(sa.Partition)
```
manirajv06, line 132: Pls see above comment

pkg/scheduler/partition.go
```diff
@@ -483,11 +484,15 @@ func (pc *PartitionContext) GetQueueInfos() dao.QueueDAOInfo {
 	return pc.root.GetQueueInfos()
 }
 
+func (pc *PartitionContext) GetPartitionNameWithoutClusterID() string {
+	return common.GetPartitionNameWithoutClusterID(pc.Name)
```
manirajv06, line 488: Pls see above comment

manirajv06 (changes requested):

chia7712:
@manirajv06 thanks for your reviews. I have addressed your comment.

pkg/webservice/handlers.go
```diff
@@ -275,6 +276,7 @@ func getPartitionJSON(partition *scheduler.PartitionContext) *dao.PartitionDAOIn
 
 	queueDAOInfo := partition.GetQueueInfos()
 
+	// this method is used by a deprecated API, so it keeps returning normalized name
```
manirajv06, line 279: I am under the assumption that we would be doing follow up changes in web app/UI component as well similar to YUNIKORN-794. If it it so, then, should we take care of this place as well?

manirajv06 (changes requested):

pkg/webservice/handlers.go
chia7712: You are right. will copy that

manirajv06 (approved):
Looks good. Thanks.

PR #324, opened 2021-09-19
[YUNIKORN-854] node removal with inflight placeholder replacements

wilfred-s:
### What is this PR for?
If a node has a placeholder while the node is removed the cleanup leaves
inflight replacements untouched. This can lead to unlinked allocations
on a secondary node or asks that never get allocated.
Handling three cases for node removal linked to placeholders:
* the removed node contains both the placeholder and real allocations,
  the real allocation is retried (without placeholder)
* the removed node contains only the real allocation,
  placeholder is unlinked and will be removed after confirmation by shim
  the real allocation is retried (without placeholder)
* the removed node contains only the placeholder,
  the real allocation is confirmed on the 2nd node

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-854

Unit tests are added as part of the change

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -577,11 +577,14 @@ func (cc *ClusterContext) updateNodes(request *si.UpdateRequest) {
 		case si.UpdateNodeInfo_DECOMISSION:
 			// set the state to not schedulable then tell the partition to clean up
 			node.SetSchedulable(false)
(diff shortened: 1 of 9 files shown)


yangwwei (approved):
hi @wilfred-s , thanks for walking through the changes with me, +1. Feel free to commit this.

PR #328, opened 2021-10-02
[YUNIKORN-870] Enhance application-related REST APIs

chia7712:
### What is this PR for?
We are integrating yunikorn with monitor system through application REST APIs. The application rest APIs is useful but it lacks three important information.
1. elapsed time of a application
2. the allocations to use when running (the `usedResource` is reset to zero when the application is terminated)
3. user name (It can enable us to trace the application for specific user, but I am fire to remove this if it is a security issue.)
Also, the application REST APIs need to be more flexible. For example, it should support following queries:
    1. applicationState
    2. user name (will be added)

This issue is related to following REST APIs.
1. `/ws/v1/apps`
2. `/ws/v1/partition/{partition}/queue/{queue}/applications`

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Update web docs of application REST APIs

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-870

### Questions:


wilfred-s (changes requested):
Elapsed time feels ambiguous when displayed. It is not possible to tell from the value if the app is still running or not. You need to calculate that info, or combine it with the state field.

Can we not store the `FinishedTime` in the application object instead, which we initialise as Zero time. When we unset the queue we set it to `time.Now()` Then we have multiple things we can do with that we could return the the finished time by itself or return both finished time and total run time. Based on what is set the end user can easily determine what is going on.

wilfred-s:
@chia7712 any update on this one. We seem to be so close to getting this done.

chia7712:
The response of `ws/v1/apps` is shown below.


```json
[
    {
        "applicationID": "spark-513ec7450ab34c81957c7bea0d8c5d2a",
        "usedResource": "[]",
        "maxUsedResource": "[]",
        "partition": "default",
        "queueName": "root.infinite",
        "submissionTime": 1645871481994982370,
        "finishedTime": null,
        "allocations": [],
        "applicationState": "New",
        "user": "nobody"
    },
    {
        "applicationID": "spark-2f6d5babe1dd42478f57d9e965af2411",
        "usedResource": "[memory:0 vcore:0]",
        "maxUsedResource": "[memory:3222 vcore:5000]",
        "partition": "default",
        "queueName": "root.infinite",
        "submissionTime": 1645871347893002178,
        "finishedTime": 1645871391609017573,
        "allocations": [],
        "applicationState": "Completed",
        "user": "nobody"
    }
]
```

The first application has `null` finishedTime since it is not finished.

pkg/scheduler/objects/application.go
```diff
@@ -77,6 +78,7 @@ type Application struct {
 	execTimeout          time.Duration          // execTimeout for the application run
 	placeholderTimer     *time.Timer            // placeholder replace timer
 	gangSchedulingStyle  string                 // gang scheduling style can be hard (after timeout we fail the application), or soft (after timeeout we schedule it as a normal application)
+	finishedTime         *time.Time             // the time of finishing this application. the default value is nil
```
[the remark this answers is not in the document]
wilfred-s: see comment below

pkg/scheduler/objects/application.go
```diff
@@ -1193,6 +1202,19 @@ func (sa *Application) UnSetQueue() {
 	sa.Lock()
 	defer sa.Unlock()
 	sa.queue = nil
+	now := time.Now()
+	sa.finishedTime = &now
+}
+
+// FinishedTimeInUnixNano return the finished time if this application is terminated. Otherwise, it returns null
+func (sa *Application) FinishedTimeInUnixNano() *int64 {
```
wilfred-s, line 1210: We might want to support this for other time values also.
Can we generalise this into a utility function? Bases on a time object and use time.IsZero():
```
func ZeroTimeInUnixNano(t time.Time) *int64 {
	if t.IsZero() {
		return nil
	}
	tInt := t.UnixNano()
	return &tInt
}
```

wilfred-s (changes requested):
I think using the pointer to the int is the correct solution to display the not finished case. I think hiding the conversion in a utility function as part of `common/utils.go` make it more usable in the long run.
Can you also confirm that this works for the new API call path, I checked and it looks like it does just want a confirmation.

pkg/scheduler/objects/application.go
wilfred-s: the conversion into the `*int64` can then be made as part of the DAO creation

chia7712:
> Can you also confirm that this works for the new API call path, I checked and it looks like it does just want a confirmation.

I have checked related call path, and there is nothing weird.

PR #332, opened 2021-10-23
YUNIKORN-905: Core side changes of YUNIKORN-337

manirajv06:
### What is this PR for?

Core side changes of YUNIKORN-337

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-905

### Questions:


pkg/plugins/types.go
```diff
@@ -21,49 +21,11 @@ package plugins
...
-	// this might be called even the container scheduling state is unchanged
-	// the shim side cannot assume to only receive updates on state changes
-	// the shim side implementation must be thread safe
-	Update(request *si.UpdateContainerSchedulingStateRequest)
-}
-
-type ConfigurationPlugin interface {
-	UpdateConfiguration(args *si.UpdateConfigurationRequest) *si.UpdateConfigurationResponse
-}
+}
```
wilfred-s, line 31: nit: missing newline

manirajv06:
> 1. Need to discuss and see how we can break the dependency completely as there are some places in shim where code refers core directly.

Filed https://issues.apache.org/jira/browse/YUNIKORN-930 to handle this as part of 1.1.0.

pkg/scheduler/context.go
```diff
@@ -201,31 +201,106 @@ func (cc *ClusterContext) processRMConfigUpdateEvent(event *rmevent.RMConfigUpda
...
-		cc.updateNodes(request)
+func (cc *ClusterContext) processNodes(request *si.NodeRequest) {
+	// 1) Add / remove / update Nodes
+	if len(request.GetNodes()) > 0 {
+		acceptedNodes := make([]*si.AcceptedNode, 0)
+		rejectedNodes := make([]*si.RejectedNode, 0)
+		for _, nodeInfo := range request.GetNodes() {
+			switch nodeInfo.Action {
+			case si.NodeInfo_CREATE:
+				sn := objects.NewNode(nodeInfo)
```
wilfred-s, line 217: Can we move this create into its own method? Similar to the way `updateNode` is called for all update cases? Makes the code more readable:
```
// addNode adds a node to the cluster returns error if the action fails or nil on success
func (cc *ClusterContext) addNode(nodeInfo *si.NodeInfo) error {
}
```
If the response is a an error the error message can be used in the reject state otherwise it is an accepted node.

wilfred-s (changes requested):
Now that the SI has been checked in the dependency in `go.mod` needs to be updated to pull in the latest SI version.

wilfred-s (approved):
LGTM

wilfred-s:
Can you please have a look at the unit test failures, these might be just test implementation issues but the file has changed (all failures in recovery_test.go):
--- FAIL: TestFailReplacePlaceholder (0.08s)
    partition_test.go:1706: assertion failed: node-1 (alloc.NodeID string) != node-2 (nodeID2 string): should be allocated on node-2
--- FAIL: TestSchedulerRecovery (2.10s)
    mock_rm_callback_test.go:164: assertion failed: error is not nil: timeout waiting for condition: Failed to wait for node state to become accepted: node-1:1234, called from: TestSchedulerRecovery in recovery_test.go:283
--- FAIL: TestSchedulerRecovery2Allocations (1.37s)
    mock_rm_callback_test.go:164: assertion failed: error is not nil: timeout waiting for condition: Failed to wait for node state to become accepted: node-1:1234, called from: TestSchedulerRecovery2Allocations in recovery_test.go:454
--- FAIL: TestSchedulerRecoveryWithoutAppInfo (1.02s)
    mock_rm_callback_test.go:164: assertion failed: error is not nil: timeout waiting for condition: Failed to wait for node state to become accepted: node-1:1234, called from: TestSchedulerRecoveryWithoutAppInfo in recovery_test.go:579

manirajv06:
Fixed unit test failures. Most of the Node registration failures occured because app doesn't exist around node reg process happens. Ensure app existence by calling UpdateApplication before the UpdateNode fix these problems.

wilfred-s (approved):
LGTM

wilfred-s:
Based on the unit test failures we must make sure that the order in the shim is correct. First recover apps then the nodes.
Looking at the change we might have had an issue in the RMProxy for a long time. I do think that we need to add a retry in the node update when we recover the node. Even in the previous implementation there was no guarantee that all the application were added before a node was recovered. The tests in the unit tests used the order processing dependency to make sure it worked. There was _never_ an order requirements on the events send by a shim. An event to recover a node could be a separate UpdateRequest from the applications that should be recovered. That means we relied on the go routine ordering to hopefully do things correctly: i.e. events send by the shim to create new apps would be processed before node recovery started.
That is a dangerous assumption: filing a follow up jira.

manirajv06:
Yes, we will need to take care of the ordering while recovering happens in shim side atleast for now to ensure this whole major changes works end to end well. However, we will need to address YUNIKORN-936 as a long term fix. Hope we are on the same page. If so, can we commit this so that I can take care of go.mod changes in shim?

In fact, we skimmed through this particular ordering problem in one of our early offline discussions and decided to capture this in design doc as well. Since that it had come out very openly in test failures, stressed the same in this PR discussions.

wilfred-s:
Sorry for the delay in committing, all done now

PR #336, opened 2021-11-09
[YUNIKORN-940] Periodic & on-demand state dump in Yunikorn

pbacsko:
### What is this PR for?
Add new REST interfaces to the web service:
* request a full state dump from Yunikorn
* enable writing full state information periodically to an external file
* disable periodic state dumps

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-940

### Questions:

### DIFF
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -1386,13 +1386,12 @@ func prepareSchedulerContext(t *testing.T) *scheduler.ClusterContext {
 func waitForStateDumpFile(t *testing.T) {
 	for {
 		var attempts int
-		info, err := os.Stat(stateDumpFilePath)
-		if err != nil {
-			t.Fatal(err)
-		}
-		if info.Size() > 0 {
+		info, _ := os.Stat(stateDumpFilePath)
+
+		if info != nil && info.Size() > 0 {
 			break
 		}
+
 		if attempts++; attempts > 10 {
 			t.Fatal("Nothing has been written to the state dump file")
 		}
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -1327,7 +1327,7 @@ func TestEnableDisablePeriodicStateDump(t *testing.T) {
 	waitForStateDumpFile(t)
 	fileContents, err2 := ioutil.ReadFile(stateDumpFilePath)
 	if err2 != nil {
-		t.Fatal("unable to open state dump file")
+		t.Fatal(err2)
 	}
 	var aggregated AggregatedStateInfo
 	if err3 := json.Unmarshal(fileContents, &aggregated); err3 != nil {
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -1298,15 +1298,14 @@ func TestGetFullStateDump(t *testing.T) {
 	assert.Assert(t, len(receivedBytes) > 0, "json response is empty")
 	assert.Check(t, statusCode != http.StatusInternalServerError, "response status code")
 	var aggregated AggregatedStateInfo
-	if err := json.Unmarshal(receivedBytes, &aggregated); err != nil {
-		t.Fatal(err)
-	}
+	err = json.Unmarshal(receivedBytes, &aggregated)
+	assert.NilError(t, err)
 	verifyStateDumpJSON(t, &aggregated)
 }
 
 func TestEnableDisablePeriodicStateDump(t *testing.T) {
 	schedulerContext = prepareSchedulerContext(t)
-	defer deleteStateDumpFile()
+	defer deleteStateDumpFile(t)
 	defer terminateGoroutine()
 
 	imHistory = history.NewInternalMetricsHistory(5)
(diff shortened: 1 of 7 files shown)


kingamarton (changes requested):

pkg/webservice/state_dump.go
```diff
@@ -0,0 +1,207 @@
...
+	metrics2 "github.com/apache/incubator-yunikorn-core/pkg/metrics"
+	"github.com/apache/incubator-yunikorn-core/pkg/scheduler"
+	"github.com/apache/incubator-yunikorn-core/pkg/webservice/dao"
+
+	"github.com/gorilla/mux"
+	"go.uber.org/zap"
+)
+
+const (
+	defaultStateDumpPeriod = 60
```
kingamarton, line 41: From this line we cannot know what this 60 is? mins? seconds? Since this is a duration, please use Duration type instead of int, or if it is not possible please rename the variable to mirror the exact value.

pkg/webservice/state_dump.go
```diff
@@ -0,0 +1,207 @@
...
+	Partitions         []*dao.PartitionInfo
+	Applications       []*dao.ApplicationDAOInfo
+	AppHistory         []*dao.ApplicationHistoryDAOInfo
+	Nodes              []*dao.NodesDAOInfo
+	NodesUtilization   []*dao.NodesUtilDAOInfo
+	ClusterInfo        []*dao.ClusterDAOInfo
+	ClusterUtilization []*dao.ClustersUtilDAOInfo
+	ContainerHistory   []*dao.ContainerHistoryDAOInfo
+	Queues             []*dao.PartitionDAOInfo
+	SchedulerHealth    *dao.SchedulerHealthDAOInfo
```
kingamarton, line 63: I am not sure if the health status should be included or not. Now it is set as liveness probe, so when the scheduler becomes unhealthy, it is logged and also it will be restarted.

pkg/webservice/state_dump.go
kingamarton: Creating the state dump by default every minute seems a little too aggressive for me. It will make huge files. I think we should print the dumps less freqvently. What is the default value in yarn for the state dump creeation?

kingamarton (changes requested):

pkg/webservice/handlers_test.go
```diff
@@ -1279,3 +1282,159 @@ func TestGetNodesUtilization(t *testing.T) {
...
+	}
+	verifyStateDumpJSON(t, &aggregated)
+}
+
+func TestEnableDisablePeriodicStateDump(t *testing.T) {
+	schedulerContext = prepareSchedulerContext(t)
+	defer deleteStateDumpFile()
+	defer terminateGoroutine()
+
+	imHistory = history.NewInternalMetricsHistory(5)
```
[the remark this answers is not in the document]
pbacsko: Yes, we do. There is a part which turns to `imHistory`, so it has to be initialized.

pkg/webservice/handlers_test.go
```diff
@@ -1279,3 +1282,159 @@ func TestGetNodesUtilization(t *testing.T) {
...
+	}
+	req = mux.SetURLVars(req, vars)
+
+	assert.NilError(t, err)
+	resp := &MockResponseWriter{}
+
+	// enable state dump, check file contents
+	enablePeriodicStateDump(resp, req)
+	statusCode := resp.statusCode
+	assert.Check(t, statusCode != http.StatusInternalServerError, "response status code")
```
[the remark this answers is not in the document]
pbacsko: It does not stop the execution of the test if it fails. But we can switch back to the normal assert.

pkg/webservice/handlers_test.go
```diff
@@ -1279,3 +1282,159 @@ func TestGetNodesUtilization(t *testing.T) {
...
+}
+
+func waitForStateDumpFile(t *testing.T) {
+	for {
+		var attempts int
+
+		info, err := os.Stat(stateDumpFilePath)
+
+		// tolerate only "file not found" errors
+		if err != nil && !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
pbacsko: Other error mean different things (eg. permission denied) and it does not make sense to continue (I don't expect it to happen though). "File not found" is normal, we have to wait until it is created by the background goroutine.

pkg/webservice/state_dump.go
pbacsko: > Creating the state dump by default every minute seems a little too aggressive for me. It will make huge files. I think we should print the dumps less freqvently. What is the default value in yarn for the state dump creeation?

Fair Scheduler performs this step in every 25 `update()` call (update is called for every heartbeat). To me, a state dump per minute does not sound that aggressive.

pkg/webservice/routes.go
```diff
@@ -170,6 +170,24 @@ var webRoutes = routes{
...
+		"PUT",
+		"/ws/v1/enableperiodicstatedump/{period}",
+		enablePeriodicStateDump,
+	},
+	route{
+		"Scheduler",
+		"PUT",
+		"/ws/v1/disableperiodicstatedump",
+		disablePeriodicStateDump,
+	},
```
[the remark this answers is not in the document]
pbacsko: Any ideas? Like `"/ws/v1/statedump/{switch}/{period}" or what?

pkg/webservice/state_dump.go
pbacsko: Makes sense, removing it.

pkg/webservice/state_dump.go
```diff
@@ -0,0 +1,225 @@
...
+		for {
+			select {
+			case <-abort:
+				ticker.Stop()
+				file.Close()
+				return
+			case <-ticker.C:
+				if err := doStateDump(file); err != nil {
+					log.Logger().Error("state dump failed", zap.Error(err))
+					if err := stopBackGroundStateDump(); err != nil {
```
kingamarton, line 195: shouldn't we restart it in case of some error? If it failed once for some reason it is sure that it will fail for the next calls as well?

kingamarton (changes requested):

pkg/webservice/state_dump.go
pbacsko: This means a file I/O error, which is unlikely to be recoverable. I'd rather stop when that happens (that's most likely a full or failed disk).

kingamarton (approved):

PR #338, opened 2021-11-10
[YUNIKORN-839]getNodesUtilJSON is broken

HuangTing-Yao:
### What is this PR for?
Leverage resourceExist flag only in first for loop, and remove the resourceExist flag check in second for loop .
Also simplify the output, we get same output when input resource type isn't exist and the utilization bucket get no node.
Both situation return 0 number of node, and nodes Name is Null.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
(https://issues.apache.org/jira/browse/YUNIKORN-839)

### Questions:


wilfred-s (approved):
LGTM

PR #343, opened 2021-11-23
[YUNIKORN-952] nil pointer error in webservice.getApplicationJSON()

pbacsko:
### What is this PR for?
A slice is incorrectly created in `getApplicationsDAO()`. The initial length should be zero and the capacity should be `size`.
Extended an existing unit test for better coverage.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-952

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -774,8 +774,7 @@ func getApplicationsDAO(lists map[string]*scheduler.PartitionContext) []*dao.App
 	result := make([]*dao.ApplicationDAOInfo, 0, 32)
 
 	for _, partition := range lists {
-		size := partition.GetTotalCompletedApplicationCount() + partition.GetTotalApplicationCount()
-		appList := make([]*objects.Application, 0, size)
+		var appList []*objects.Application
 		appList = append(appList, partition.GetApplications()...)
 		appList = append(appList, partition.GetCompletedApplications()...)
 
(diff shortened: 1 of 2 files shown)


wilfred-s (changes requested):
It would be far easier to not specify a capacity and let the system handle the allocation of the array and re-slice. The speed gains is most likely minimal. Append has logic build in to re-allocate if the array is not big enough. Simply initialising the slice via:
```
appList := make([]*objects.Application, 0)
```
The other thing is relying on the fact that you can append to a nil slice:
```
var appList []*objects.Application
appList = append(appList, partition.GetApplications()...) 
```

pkg/webservice/handlers.go
```diff
@@ -774,8 +774,7 @@ func getApplicationsDAO(lists map[string]*scheduler.PartitionContext) []*dao.App
 	result := make([]*dao.ApplicationDAOInfo, 0, 32)
```
wilfred-s, line 774: Can we simplify this one too? The capacity is arbitrarily set.
The make in line 800 (new code) should also be cleaned up to get it inline with the rest of the make calls: should be similar to the make in line 790

wilfred-s (changes requested):
two small nits left to clean up

wilfred-s (approved):
LGTM, approved

kingamarton (approved):

PR #344, opened 2021-12-03
[YUNIKORN-963] Fix calculateNodesResourceUsage when node's resource is overallocated

ashutoshsingh85:
### What is this PR for?

When a node resource is over-allocated `calculateNodesResourceUsage` crashes as it does not expect any resource to be over-allocated. If a node is over-allocated because of any reason, it should be handled gracefully.
So this PR is updating `calculateNodesResourceUsage` so that it considers an over-allocated node as 100% used.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-963

Added UT

### DIFF
--- a/pkg/scheduler/partition_test.go
+++ b/pkg/scheduler/partition_test.go
@@ -373,14 +373,14 @@ func TestCalculateNodesResourceUsage(t *testing.T) {
 	alloc := objects.NewAllocation(allocID, nodeID1, newAllocationAsk("key", "appID", occupiedResources))
 	node.AddAllocation(alloc)
 	usageMap := partition.calculateNodesResourceUsage()
-	assert.Assert(t, node.GetAvailableResource().Resources["first"] == resources.Quantity(50))
+	assert.Equal(t, node.GetAvailableResource().Resources["first"], resources.Quantity(50))
 	assert.Equal(t, usageMap["first"][4], 1)
 
 	occupiedResources = resources.NewResourceFromMap(map[string]resources.Quantity{"first": 50})
 	alloc = objects.NewAllocation(allocID, nodeID1, newAllocationAsk("key", "appID", occupiedResources))
 	node.AddAllocation(alloc)
 	usageMap = partition.calculateNodesResourceUsage()
-	assert.Assert(t, node.GetAvailableResource().Resources["first"] == resources.Quantity(0))
+	assert.Equal(t, node.GetAvailableResource().Resources["first"], resources.Quantity(0))
 	assert.Equal(t, usageMap["first"][9], 1)
 
 	newCapacity := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 80})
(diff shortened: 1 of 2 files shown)


pkg/scheduler/partition_test.go
```diff
@@ -361,6 +361,35 @@ func TestRemoveNodeWithPlaceholders(t *testing.T) {
...
+	oldCapacity := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 100})
+	node := newNodeMaxResource(nodeID1, oldCapacity)
+	err = partition.AddNode(node, nil)
+	assert.NilError(t, err)
+
+	occupiedResources := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 50})
+	alloc := objects.NewAllocation(allocID, nodeID1, newAllocationAsk("key", "appID", occupiedResources))
+	node.AddAllocation(alloc)
+	usageMap := partition.calculateNodesResourceUsage()
+	assert.Assert(t, node.GetAvailableResource().Resources["first"] == 50)
```
wilfred-s, line 376: can we change this to an `assert.Equal`?
It will require replacing the 50 with `resources.Quantity(50)` i.e.:
```
assert.Equal(t, node.GetAvailableResource().Resources["first"], resources.Quantity(50))
```

pkg/scheduler/partition_test.go
```diff
@@ -361,6 +361,35 @@ func TestRemoveNodeWithPlaceholders(t *testing.T) {
...
+	assert.Equal(t, usageMap["first"][4], 1)
+
+	occupiedResources = resources.NewResourceFromMap(map[string]resources.Quantity{"first": 50})
+	alloc = objects.NewAllocation(allocID, nodeID1, newAllocationAsk("key", "appID", occupiedResources))
+	node.AddAllocation(alloc)
+	usageMap = partition.calculateNodesResourceUsage()
+	assert.Assert(t, node.GetAvailableResource().Resources["first"] == 0)
```
wilfred-s, line 383: same as above assert

pkg/scheduler/partition_test.go
```diff
@@ -361,6 +361,35 @@ func TestRemoveNodeWithPlaceholders(t *testing.T) {
...
+	assert.Equal(t, usageMap["first"][9], 1)
+
+	newCapacity := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 80})
+	node.SetCapacity(newCapacity)
+	usageMap = partition.calculateNodesResourceUsage()
+	assert.Assert(t, node.GetAvailableResource().HasNegativeValue() == true)
```
wilfred-s, line 389: This can be simplified by removing the `== true`. The assert checks for boolean to be true

wilfred-s (changes requested):
nits only around the use of the asserts

wilfred-s (approved):
LGTM
Thank you for the fix and your first contribution to Apache YuniKorn.

PR #349, opened 2021-12-15
[YUNIKORN-981] cleanup queue comments and logging

wilfred-s:
### What is this PR for?
Cleanup queue comments to comply with best practice.
Fix logging levels, add new logging calls where they are missing. Remove
logging an error recursively for each queue level.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-981

No testing added or changed, only logging and comment changes no logic changes

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -39,7 +39,7 @@ import (
 
 const AppTagNamespaceResourceQuota = "namespace.resourcequota"
 
-// Represents Queue inside Scheduler
+// Queue structure inside Scheduler
 type Queue struct {
 	QueuePath string // Fully qualified path for the queue
 	Name      string // Queue name as in the config etc.
@@ -72,6 +72,7 @@ type Queue struct {
 	sync.RWMutex
 }
 
+// newBlankQueue creates a new empty queue objects with all values initialised.
 func newBlankQueue() *Queue {
 	return &Queue{
 		children:              make(map[string]*Queue),
@@ -86,7 +87,7 @@ func newBlankQueue() *Queue {
 	}
 }
 
-// Create a new queue from scratch based on the configuration
+// NewConfiguredQueue creates a new queue from scratch based on the configuration
 // lock free as it cannot be referenced yet
 func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error) {
 	sq := newBlankQueue()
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -129,8 +130,7 @@ func NewDynamicQueue(name string, leaf bool, parent *Queue) (*Queue, error) {
 	}
 	// name might not be checked do it here
 	if !configs.QueueNameRegExp.MatchString(name) {
-		return nil, fmt.Errorf("invalid queue name %s, a name must only have alphanumeric characters,"+
-			" - or _, and be no longer than 64 characters", name)
+		return nil, fmt.Errorf("invalid queue name '%s', a name must only have alphanumeric characters, - or _, and be no longer than 64 characters", name)
```
chia7712, line 133: Is this line too long?

pkg/scheduler/objects/queue.go
```diff
@@ -783,17 +808,22 @@ func (sq *Queue) IncAllocatedResource(alloc *resources.Resource, nodeReported bo
...
 		}
 	}
 	// check the parent: need to pass before updating
 	if sq.parent != nil {
 		if err := sq.parent.IncAllocatedResource(alloc, nodeReported); err != nil {
-			log.Logger().Error("parent queue exceeds maximum resource",
-				zap.Any("allocationId", alloc),
-				zap.String("maxResource", sq.maxResource.String()),
-				zap.Error(err))
+			// only log the warning if we get to the leaf
```
chia7712, line 818: The main reason is that we don't want to see looped warnings, right? If so, could we add the main reason to the comment? Otherwise, it seems to me the comment is not more clear than the code below.

chia7712 (approved):
LGTM. two small questions below.

pkg/scheduler/objects/queue.go
yuchaoran2011: Unrelated to this PR, but how come we set the limit to 64? In Kubernetes, 63 is the limit (e.g. for labels)

wilfred-s: It is not a label value or name it is a queue name. K8s labels are different in a number of ways.  The major difference is that label values and names can contain dots, queue names cannot contain dots. It looks similar but it is not and we do not want to be either.

pkg/scheduler/objects/queue.go
wilfred-s: yes extended the comment text in both places

chia7712 (approved):
LGTM

PR #350, opened 2021-12-15
YUNIKORN-462: Streamline core to shim update on allocation change

manirajv06:
### What is this PR for?
New Allocation and Release allocations events has been changed from async to synchronous communication so that shim updates its own cache immediately followed up by pod bind/delete steps. With this change, ReSyncSchedulerCache plugin doesn't have any more role to play. Hence, cleaned up the related plugin code as well.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-462

### Questions:

### DIFF
--- a/pkg/rmproxy/rmevent/events.go
+++ b/pkg/rmproxy/rmevent/events.go
@@ -86,11 +86,6 @@ type RMRejectedAllocationAskEvent struct {
 type RMReleaseAllocationEvent struct {
 	RmID                string
 	ReleasedAllocations []*si.AllocationRelease
-}
-
-type RMReleaseAllocationSyncEvent struct {
-	RmID                string
-	ReleasedAllocations []*si.AllocationRelease
 	Channel             chan *Result
 }
 
--- a/pkg/rmproxy/rmevent/events.go
+++ b/pkg/rmproxy/rmevent/events.go
@@ -65,10 +65,10 @@ type Result struct {
 }
 
 // Outgoing events from the scheduler to the RM
-// These events communicate the response to the RMUpdateRequestEvent
 type RMNewAllocationsEvent struct {
 	RmID        string
 	Allocations []*si.Allocation
+	Channel     chan *Result
 }
 
 type RMApplicationUpdateEvent struct {
@@ -86,6 +86,12 @@ type RMRejectedAllocationAskEvent struct {
 type RMReleaseAllocationEvent struct {
 	RmID                string
 	ReleasedAllocations []*si.AllocationRelease
+}
+
+type RMReleaseAllocationSyncEvent struct {
+	RmID                string
+	ReleasedAllocations []*si.AllocationRelease
+	Channel             chan *Result
 }
 
 type RMReleaseAllocationAskEvent struct {
--- a/pkg/rmproxy/rmproxy.go
+++ b/pkg/rmproxy/rmproxy.go
@@ -98,6 +98,10 @@ func (rmp *RMProxy) processAllocationUpdateEvent(event *rmevent.RMNewAllocations
 	rmp.RLock()
 	defer rmp.RUnlock()
 	if len(event.Allocations) == 0 {
+		event.Channel <- &rmevent.Result{
+			Reason:    "RMNewAllocationsEvent has zero allocations"	,
+			Succeeded: false,
+		}
 		return
 	}
 	response := &si.AllocationResponse{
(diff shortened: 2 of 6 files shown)


wilfred-s:
Please check why the linter cannot build the code. The change seems to reference a non existing field. This might be a typo or wrong version of the interface etc.

manirajv06:
Have made changes in SI as well. If you can take a look at https://github.com/apache/incubator-yunikorn-scheduler-interface/pull/58 first before core and shim's pr, then whole changes gives us clear picture and easy for us to fix these lint issues.

pkg/rmproxy/rmevent/events.go
```diff
@@ -88,6 +88,12 @@ type RMReleaseAllocationEvent struct {
 	ReleasedAllocations []*si.AllocationRelease
 }
 
+type RMReleaseAllocationSyncEvent struct {
+	RmID                string
+	ReleasedAllocations []*si.AllocationRelease
+	Channel             chan *Result
+}
+
```
wilfred-s, line 96: Can we hide this in the logic? Have only one event (not a sync and async version).
For a sync event you add a channel in the caller and wait for the response on the channel. For the async processing we can pass in a nil channel. Make the proxy handle it. I think it will allow us to remove code duplication.

pkg/rmproxy/rmproxy.go
```diff
@@ -98,13 +98,21 @@ func (rmp *RMProxy) processAllocationUpdateEvent(event *rmevent.RMNewAllocations
 	rmp.RLock()
 	defer rmp.RUnlock()
 	if len(event.Allocations) == 0 {
+		event.Channel <- &rmevent.Result{
+			Reason:    "RMNewAllocationsEvent has zero allocations"	,
+			Succeeded: false,
+		}
```
wilfred-s, line 104: an empty list of allocations should not be a failure that changes the current behaviour we have now

wilfred-s (changes requested):

pkg/rmproxy/rmproxy.go
```diff
@@ -139,14 +140,18 @@ func (rmp *RMProxy) processApplicationUpdateEvent(event *rmevent.RMApplicationUp
 func (rmp *RMProxy) processRMReleaseAllocationEvent(event *rmevent.RMReleaseAllocationEvent) {
 	rmp.RLock()
 	defer rmp.RUnlock()
-	if len(event.ReleasedAllocations) == 0 {
-		return
-	}
```
wilfred-s, line 144: Same as above use the optimisation

pkg/rmproxy/rmproxy.go
```diff
@@ -139,14 +140,18 @@ func (rmp *RMProxy) processApplicationUpdateEvent(event *rmevent.RMApplicationUp
...
 	response := &si.AllocationResponse{
 		Released: event.ReleasedAllocations,
 	}
 	rmp.triggerUpdateAllocation(event.RmID, response)
 	metrics.GetSchedulerMetrics().AddReleasedContainers(len(event.ReleasedAllocations))
+
+	// Done, notify channel
+	if event.Channel != nil {
```
wilfred-s, line 150: Is it still the case that the channel can be nil after the updates we have discussed?

pkg/scheduler/context.go
```diff
@@ -840,6 +797,7 @@ func (cc *ClusterContext) notifyRMAllocationReleased(rmID string, released []*ob
 	releaseEvent := &rmevent.RMReleaseAllocationEvent{
 		ReleasedAllocations: make([]*si.AllocationRelease, 0),
 		RmID:                rmID,
+		Channel:             nil,
```
wilfred-s, line 800: This should no longer be the case based on the SI discussions.

wilfred-s (changes requested):
We need to update this based on the SI discussion. It would simplify the change too

pkg/rmproxy/rmproxy.go
```diff
@@ -97,14 +97,15 @@ func (rmp *RMProxy) handleUpdateResponseError(rmID string, err error) {
 func (rmp *RMProxy) processAllocationUpdateEvent(event *rmevent.RMNewAllocationsEvent) {
 	rmp.RLock()
 	defer rmp.RUnlock()
-	if len(event.Allocations) == 0 {
-		return
-	}
```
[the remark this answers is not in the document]
manirajv06: This has been already handled in caller side. For example notifyRMNewAllocation would be called only if it is necessary. Added extra check for the same in notifyRMNewAllocation to make it more safer. Also, sending it as success if case of no allocations gives wrong impression on the caller side. Anyways, I think caller side code is good enough.

wilfred-s (changes requested):

wilfred-s (approved):
LGTM

PR #352, opened 2021-12-21
[YUNIKORN-721]Improve YuniKorn core's queue-level and scheduler metrics

chenya-zhang:
### What is this PR for?

To improve YuniKorn core's queue-level and scheduler metrics:

- Differentiate queue level metrics with scheduler metrics, e.g. using "IncQueueApplicationsAccepted"
- Refactor related queue and scheduler metrics definitions
- Refactor related queue and scheduler metrics operation functions
- Update metrics naming and help messages 
- Update in-line comments and documentations

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [X] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-721

Existing tests

### Questions:

### DIFF
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -52,7 +52,7 @@ func InitQueueMetrics(name string) CoreQueueMetrics {
 			Subsystem: substituteQueueName(name),
 			Name:      "queue_resource",
 			Help:      "Queue resource metrics. State of the resource includes `guaranteed`, `max`, `used`, `pending`, `available`.",
-		}, []string{"state"})
+		}, []string{"state", "resource"})
 
 	var queueMetricsList = []prometheus.Collector{
 		q.appMetrics,
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -46,8 +46,8 @@ func forQueue(name string) CoreQueueMetrics {
 		prometheus.CounterOpts{
 			Namespace: Namespace,
 			Subsystem: substituteQueueName(name),
-			Name:      "app_metrics",
-			Help:      "Application Metrics",
+			Name:      "queue_app_metrics",
+			Help:      "Application metrics from a queue",
 		}, []string{"state"})
 
 	q.usedResourceMetrics = prometheus.NewGaugeVec(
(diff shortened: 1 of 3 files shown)


pkg/metrics/queue.go
```diff
@@ -28,60 +28,38 @@ import (
...
+// QueueMetrics to declare queue metrics
 type QueueMetrics struct {
-	// metrics related to app
-	appMetrics *prometheus.CounterVec
-
-	// metrics related to resource
-	usedResourceMetrics      *prometheus.GaugeVec
-	pendingResourceMetrics   *prometheus.GaugeVec
-	availableResourceMetrics *prometheus.GaugeVec
+	appMetrics      *prometheus.GaugeVec
```
yuchaoran2011, line 33: This should be a CounterVec

pkg/metrics/queue.go
```diff
@@ -28,60 +28,38 @@ import (
...
+	q.appMetrics = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: Namespace,
 			Subsystem: substituteQueueName(name),
-			Name:      "pending_resource",
-			Help:      "Queue pending resource",
-		}, []string{"resource"})
+			Name:      "queue_app",
+			Help:      "Queue application metrics. State of the application includes `running`.",
+		}, []string{"state"})
```
yuchaoran2011, line 47: Looks like other states like `accepted`, `rejected` and `completed` got removed and only `running` is retained. We should keep the other states too

yuchaoran2011 (changes requested):

pkg/metrics/queue.go
chenya-zhang: I think it is more meaningful to count the current running apps in a queue not all the apps that have run in a queue unless the scheduler restarts.

I checked the Prometheus definition on counter v.s. gauge.

> Do not use a counter to expose a value that can decrease. For example, do not use a counter for the number of currently running processes; instead use a gauge.

> Gauges are typically used for measured values like temperatures or current memory usage, but also "counts" that can go up and down, like the number of concurrent requests.

https://prometheus.io/docs/concepts/metric_types/

pkg/metrics/queue.go
chenya-zhang: Similar to the above, I think it is more meaningful to count the current running apps in a queue not all the apps that have run/accepted/rejected/completed in a queue unless the scheduler restarts. Not very sure what the business value is.

From the scheduler/cluster level, we have the metrics `applicationSubmission` and `application` to count apps that have run/accepted/rejected/completed. I think it should satisfy our operational needs since all queues share the same scheduler in a cluster.

One thing to note is that, on the queue level, these metrics are never implemented so it is not a backward incompatible change. I will help to implement them in future PRs.

yuchaoran2011 (approved):
@chenya-zhang Thanks for the clarifications. I understand the reasoning now. LGTM. @yangwwei @wilfred-s Any other questions/comments before it's merged?

yangwwei (approved):
LGTM, thank you @chenya-zhang , @yuchaoran2011

yangwwei:
@chenya-zhang feel free to get this merged ： ）

PR #353, opened 2021-12-22
YUNIKORN-985: Enforcing MaxRunningApps limit

manirajv06:
### What is this PR for?

Introduced a UserGroupManagement package to manage the user and group limits configured at queue level. With this help of package, Enforced MaxRunningApps limit as part of scheduler.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-985

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -24,6 +24,7 @@ import (
 	"sync"
 	"time"
 
+	ugm "github.com/apache/incubator-yunikorn-core/pkg/scheduler/usergroupmanagement"
 	"github.com/looplab/fsm"
 	"go.uber.org/zap"
 
@@ -68,6 +69,8 @@ type Queue struct {
 	stateMachine       *fsm.FSM            // the state of the queue for scheduling
 	stateTime          time.Time           // last time the state was updated (needed for cleanup)
 	template           *template.Template
+	limits			   []configs.Limit
+	userGroupManager   *ugm.UserGroupManager // UserGroupManager
 
 	sync.RWMutex
 }
@@ -111,6 +114,8 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
 		}
 		// pull the properties from the parent that should be set on the child
 		sq.mergeProperties(parent.getProperties(), conf.Properties)
+
+		sq.userGroupManager = ugm.UserGroupManager.NewUsersManager(sq, sq)
 	}
 	sq.UpdateSortType()
 
@@ -246,7 +251,12 @@ func (sq *Queue) applyConf(conf configs.QueueConfig) error {
 	}
 
 	sq.properties = conf.Properties
+	sq.limits = conf.Limits
 	return nil
+}
+
+func (sq *Queue) GetConfigLimits() []configs.Limit{
+	return sq.limits
 }
 
 // setResources sets the maxResource and guaranteedResource of the queue from the config.
@@ -1279,6 +1289,10 @@ func (sq *Queue) updateAllocatedResourceMetrics() {
 			metrics.GetQueueMetrics(sq.QueuePath).SetQueueAllocatedResourceMetrics(k, float64(v))
 		}
 	}
+}
+
+func (sq *Queue) GetUserGroupManager() *ugm.UserGroupManager {
+	return sq.userGroupManager
 }
 
 func (sq *Queue) String() string {
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -114,9 +114,13 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
 		}
 		// pull the properties from the parent that should be set on the child
 		sq.mergeProperties(parent.getProperties(), conf.Properties)
+	}
 
+	// UGM is required only for leaf queues
+	if ! conf.Parent {
 		sq.userGroupManager = ugm.UserGroupManager.NewUsersManager(sq, sq)
 	}
+
 	sq.UpdateSortType()
 
 	log.Logger().Info("configured queue added to scheduler",
@@ -255,7 +259,29 @@ func (sq *Queue) applyConf(conf configs.QueueConfig) error {
 	return nil
 }
 
-func (sq *Queue) GetConfigLimits() []configs.Limit{
+func (sq *Queue) GetConfigLimits() []configs.Limit {
+	var limit []configs.Limit
+	// get the limit for the parent first
+	if sq.parent != nil {
+		limit = sq.parent.GetConfigLimits()
+	}
+	return sq.internalGetConfigLimits(limit)
+}
+
+func (sq *Queue) internalGetConfigLimits(parentLimit []configs.Limit) []configs.Limit {
+	sq.RLock()
+	defer sq.RUnlock()
+	// no limit set in parent queue, not even for root
+	if parentLimit == nil {
+		if sq.limits == nil {
+			return nil
+		}
+		return sq.limits
+	}
+	// parent limit set, no queue limit return parent
+	if sq.limits == nil {
+		return parentLimit
+	}
 	return sq.limits
 }
 
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -68,7 +68,7 @@ type Queue struct {
 	stateMachine       *fsm.FSM            // the state of the queue for scheduling
 	stateTime          time.Time           // last time the state was updated (needed for cleanup)
 	template           *template.Template
-	limits			   []configs.Limit
+	limits             []configs.Limit
 	userGroupManager   *UserGroupManager // UserGroupManager
 
 	sync.RWMutex
@@ -116,7 +116,7 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
 	}
 
 	// UGM is required only for leaf queues
-	if ! conf.Parent {
+	if !conf.Parent {
 		sq.userGroupManager = NewUserGroupManager(sq)
 	}
 
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -24,7 +24,6 @@ import (
 	"sync"
 	"time"
 
-	ugm "github.com/apache/incubator-yunikorn-core/pkg/scheduler/usergroupmanagement"
 	"github.com/looplab/fsm"
 	"go.uber.org/zap"
 
@@ -70,7 +69,7 @@ type Queue struct {
 	stateTime          time.Time           // last time the state was updated (needed for cleanup)
 	template           *template.Template
 	limits			   []configs.Limit
-	userGroupManager   *ugm.UserGroupManager // UserGroupManager
+	userGroupManager   *UserGroupManager // UserGroupManager
 
 	sync.RWMutex
 }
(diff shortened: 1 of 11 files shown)


manirajv06:
This WIP patch lays the foundation for applying the User Group limits settings configured at Queue level. Released this PR mainly to review the overall approach as documented in https://docs.google.com/document/d/1Td01tG9AwRtYfmTQoXSqDn3G6xYtw53PqPDt2sP8mOs/edit?usp=sharing.

Need to do the following:

1. Inherit User and Group limit settings configured at Parent Queue level for Leaf Queue as well like any other properties.
2. Unit tests coverage

pkg/scheduler/usergroupmanagement/user.go
```diff
@@ -0,0 +1,57 @@
...
+func (u *User) IncRunningApplications() {
+	atomic.AddInt32(u.runningApplications, 1)
+}
+
+func (u *User) DecRunningApplications() {
+	atomic.AddInt32(u.runningApplications, -1)
+}
+
+func (u *User) CanRun() bool {
+	if atomic.LoadInt32(u.runningApplications) < u.maxApplications {
```
pbacsko, line 40: This part should be double-checked, because it feels racy. 

By the time `LoadInt()` completes, a separate goroutine might come along and increase `runningApplications`. This feels more sound:

```
	current := atomic.LoadInt32(&runningApplications)
	if atomic.CompareAndSwapInt32(&runningApplications, current, current) < u.maxApplications {
		// runningApplications was definitely not incremented if we end up here
	}
```

It's easy to get this wrong. Also, I don't think we deliver to 32-bit targets, so I'd recommend `int64` instead.

If performance is not a concern (and this does not look like a hot code path), then an ordinary `Mutex` is probably better because it's easiser to reason about. Atomic stuff is too low level.

pkg/scheduler/usergroupmanagement/user.go
```diff
@@ -0,0 +1,57 @@
...
+	"sync/atomic"
+
+	"github.com/apache/incubator-yunikorn-core/pkg/common/resources"
+)
+
+type User struct {
+	name	string // Name of the user
+	maxResources	*resources.Resource // Max Resource configured per user
+	maxApplications	int32 // Max Applications configured per user
+	runningApplications	*int32 // Running Applications
```
pbacsko, line 13: Comments can be removed, everything is self-explanatory here

pkg/scheduler/usergroupmanagement/user.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (u *User) CanRun() bool {
+	if atomic.LoadInt32(u.runningApplications) < u.maxApplications {
+		return true
+	} else {
+		return false
+	}
+}
+
+// SetUsedGroup A user may belongs to more than one group. In case of any group changes for
```
pbacsko, line 47: "belong"

pkg/scheduler/usergroupmanagement/user.go
```diff
@@ -0,0 +1,57 @@
...
+// any user while running applications, without this usedGroup info might lead
+// to confusion while doing any metrics calculation as and when user is done with
+// their activities. Also this info avoid parsing all mapped groups while doing calculations
```
pbacsko, line 50: "Also[,] this info avoid[s] parsing"

pkg/scheduler/usergroupmanagement/group.go
```diff
@@ -0,0 +1,44 @@
...
+	"sync/atomic"
+
+	"github.com/apache/incubator-yunikorn-core/pkg/common/resources"
+)
+
+type Group struct {
+	name	string // Name of the group
+	maxResources	*resources.Resource // Max Resource configured per group
+	maxApplications	int32 // Max Applications configured per group
+	runningApplications	*int32 // Running Applications
```
pbacsko, line 13: Comments can be removed

pkg/scheduler/partition.go
```diff
@@ -338,6 +338,32 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
+		}
+	}
+
+	for _, group := range app.GetUser().Groups {
+		// Is there any group (to which user belongs to) config has limit settings?
+		g := queue.GetUserGroupManager().GetGroup(group)
+		if g != nil && g.CanRun() {
+			g.IncRunningApplications()
+			if user != nil {
+				user.SetUsedGroup(g.GetName())
```
pbacsko, line 358: What if the user belongs to multiple groups? Can that happen?
If so, is the list of groups predictable?

pkg/scheduler/partition.go
```diff
@@ -380,7 +406,33 @@ func (pc *PartitionContext) removeApplication(appID string) []*objects.Allocatio
...
+		}
+
+		// Used when limit has been configured only for group, not for individual user
+		if ! updateGroupMetrics && len(app.GetUser().User) > 0 {
+			for _, group := range app.GetUser().Groups {
+				// Is there any group (to which user belongs to) config has limit settings?
+				g := queue.GetUserGroupManager().GetGroup(group)
+				if g != nil {
+					g.DecRunningApplications()
+					break
```
pbacsko, line 430: This is related to my comment above. If we our pick from multiple groups is deterministic, then it's fine. Otherwise we have to think about how to handle the multiple group case (I don't know how common that is, but `usergroup.UserGroup` is `string[]`, so at least it has theoretical possibility).

pkg/scheduler/usergroupmanagement/user.go
pbacsko: Another approach that might work, if you incorporate the increment inside `CanRun()`:

```
func (u *User) CanRun() bool {
  for {
    current := atomic.LoadInt32(&u.runningApplications)
    if current < u.maxApplications {
      return false
    }

    if atomic.CompareAndSwapInt32(&u.runningApplications, current, current + 1) {
      return true
    }
    
   // increment failed, so we re-attempt in an infinite loop
  }
}
```

This seems to be OK, too, but again, too low level.

manirajv06: I think LoadInt32 ensures that load happens in atomic way irrespective of the in-flight writes. 

Yes, it is low level but it is good enough to meet our needs. I think Mutex is more than what we need in this particular context. Also, atomic package has been used in some other places too in shim. Thoughts?

pkg/scheduler/partition.go
manirajv06: Yes, a user may belong to more than group. Added user.SetUsedGroup() mainly to reduce this kind of friction by storing the group and use the same in future as well. As and when configured group arrives while traversing groups of user, that group would be picked up and set. Also, setting group using user.SetUsedGroup() avoids parsing groups again while removing the application to larger extent.

manirajv06:
Completed:

1. Inherit limits from its parent queue hierarchy if limit is not configured for the queue where application has been dded. 
2. UGM is required only for leaf queues. So object initialisation happens accordingly.
3. Addressed review comments

To do:

1. Handle "*" cases configured for user and group
2. Unit tests coverage

pkg/scheduler/usergroupmanagement/user.go
pbacsko: I was thinking about this and I've become more confident that even my first suggestion is wrong. You can't separate the loading from the increment. The two operations are atomic, but their composition is not.

Consider the following state with multiple Goroutines running simultaneously:

maxApplications = 30
runningApplications = 29

**Scenario 1**

The first is incorrect because of:
Goruoutine 1:  loads current value ->  checks if 29 < 30 -> returns true -> increments runningApplications to 30
Goruoutine 2:  loads current value -> [execution suspended for a brief period] -> checks if 29 < 30 (current value is already 30) -> returns true -> increments runningApplications to 31

Goruoutine 2 should not return true, but it does.

**Scenario 2**

The first suggestion defends against this scenario 1, but not against this:
Goruoutine 1:  loads current value ->  checks if 29 < 30 -> returns true -> increments runningApplications to 30 or 31
Goruoutine 2:  loads current value ->  checks if 29 < 30 -> returns true -> increments runningApplications to 31 or 30

You can't have `CanRun()` and `IncRunningApplications()` in two separate calls. The check and increment must behave atomically. It would only work if the call site of these methods were written as atomic:
```
guard  *sync.Mutex
...
	if user != nil {
                guard.Lock()
		if user.CanRun() {
			user.IncRunningApplications()
		} else {
                ...
                }
                guard.Unlock()
```

But this makes the usage of atomics unnecessary.

pkg/scheduler/partition.go
pbacsko: Ok, this might be a nitpick, but I looked at the existing group cache code.

I think we might have an issue here. Let's say an user "john" who belongs to "qe" group submitted an application which runs for a while. We have limits set for this group. Then "john" is removed from "qe" and the group cache gets updated. The application finishes but `DecRunningApplications()` will not be called.

We need to store this information somewhere else, probably adding an extra field to `Application` and not relying on `app.GetUser().Groups`.

pbacsko: cc @wilfred-s @craigcondit

pkg/scheduler/usergroupmanagement/user.go
pbacsko: cc @wilfred-s @craigcondit

pkg/scheduler/partition.go
manirajv06: > We need to store this information somewhere else, probably adding an extra field to Application and not relying on app.GetUser().Groups.

Please see my earlier comment https://github.com/apache/incubator-yunikorn-core/pull/353#discussion_r780043936.  user.SetUsedGroup() precisely takes care of this issue only.

pbacsko: From what I understand, that only works if `updateGroupMetrics = true`. Maybe I'm still not reading it correctly, but if "limit has been configured only for group, not for individual user" (as the comment says), then a modified list of groups will be checked against that saved value. So in my example, "qe" will not be part of if.

manirajv06: Please see line no.415.

pbacsko: Talked to Wilfred, let's wait for his feedback. You guys will sync up soon.

pbacsko (changes requested):
@manirajv06 could you address possible gofmt / lintern / header stuff so that we have a clean build? Just to see where we're at with the current PR.

pkg/scheduler/objects/queue.go
```diff
@@ -246,9 +259,36 @@ func (sq *Queue) applyConf(conf configs.QueueConfig) error {
...
+		if sq.limits == nil {
+			return nil
+		}
+		return sq.limits
+	}
+	// parent limit set, no queue limit return parent
+	if sq.limits == nil {
+		return parentLimit
+	}
+	return sq.limits
```
pbacsko, line 289: Couldn't this be simplified?

Eg. 
```if sq.limits != nil {
  return. sq.limits
}

return parentLimit
```

So we return the current queues's limit if not nil, otherwise we just pass the parentLimit no matter what.

pbacsko (changes requested):
Overall it looks fine.

Please check the small example codes that I linked. I can see that you also created concurrent unit tests, which is a great idea. I'm assuming that problems do not show up when running those. It would be good to reason about why not, because I think it should be possible to reproduce the problem I demonstrated with unit tests as well.

wilfred-s:
A lot has changed since we started working on this. Based on our off-line discussion I don't think this PR is a good base for the implementation. Lets start fresh with the new design.

PR #356, opened 2022-01-10
[YUNIKORN-982] placeholder size checks

wilfred-s:
### What is this PR for?
Check placeholder during allocation cycle if placeholder is smaller than the real allocation the placeholder(s) get removed.
Check placeholder during replacement if placeholder is larger than the real allocation the node and queue usage is adjusted.
Additional unit test:
partition: rewrite of placeholder and real alloc tests
node: new test for delta processing on replace
application: test fixes only

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-982

New unit tests added for all use cases

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -827,6 +827,34 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator, 
 			if ph.released || request.taskGroupName != ph.taskGroupName {
 				continue
 			}
+			// before we check anything we need to check the resources equality
+			delta := resources.Sub(ph.AllocatedResource, request.AllocatedResource)
+			// Any negative value in the delta means that at least one of the requested resource in the real
+			// allocation is larger than the placeholder. We need to cancel the placeholder and check the next
+			// placeholder. This should trigger a cleanup of all the placeholder as a task group is always one size.
+			if delta.HasNegativeValue() {
+				log.Logger().Warn("releasing placeholder: real allocation is larger than placeholder",
+					zap.String("requested resource", request.AllocatedResource.String()),
+					zap.String("placeholderID", ph.UUID),
+					zap.String("placeholder resource", ph.AllocatedResource.String()))
+				// release the placeholder and tell the RM
+				ph.released = true
+				sa.notifyRMAllocationReleased(sa.rmID, []*Allocation{ph}, si.TerminationType_TIMEOUT, "cancel placeholder: resource incompatible")
+				// add an event on the app to show the release
+				if eventCache := events.GetEventCache(); eventCache != nil {
+					message := fmt.Sprintf("Task group '%s' in application '%s': allocation resources '%s' are not matching placeholder '%s' allocation with ID '%s'", ph.taskGroupName, sa.ApplicationID, request.AllocatedResource.String(), ph.AllocatedResource.String(), ph.AllocationKey)
+					if event, err := events.CreateRequestEventRecord(request.AllocationKey, sa.ApplicationID, "releasing placeholder: real allocation is larger than placeholder", message); err != nil {
+						log.Logger().Warn("Event creation failed",
+							zap.String("event message", message),
+							zap.Error(err))
+					} else {
+						eventCache.AddEvent(event)
+					}
+				}
+				continue
+			}
+			// placeholder is the same or larger continue processing and difference is handled when the placeholder
+			// is swapped with the real one.
 			if phFit == nil && reqFit == nil {
 				phFit = ph
 				reqFit = request
@@ -848,21 +876,6 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator, 
 				if err != nil {
 					log.Logger().Warn("ask repeat update failed unexpectedly",
 						zap.Error(err))
-				}
-				// Is real allocation matches with placeholder resource?
-				if !resources.Equals(alloc.AllocatedResource, ph.AllocatedResource) {
-					log.Logger().Warn("Real allocation is not matching with placeholder allocation", zap.String("Real alloc: ", alloc.AllocationKey), zap.String("Real alloc resource: ", alloc.AllocatedResource.DAOString()), zap.String("Placeholder alloc: ", ph.AllocationKey), zap.String("Placeholder alloc resource: ", ph.AllocatedResource.DAOString()))
-					// post scheduling events via the event plugin
-					if eventCache := events.GetEventCache(); eventCache != nil {
-						message := fmt.Sprintf("Real Pod %s allocation %s is not matching with placeholder %s allocation %s in application %s", request.AllocationKey, request.AllocatedResource.DAOString(), ph.AllocationKey, ph.AllocatedResource.DAOString(), request.ApplicationID)
-						if event, err := events.CreateRequestEventRecord(request.AllocationKey, request.ApplicationID, "Resource Allocation Mismatch between real and placeholder", message); err != nil {
-							log.Logger().Warn("Event creation failed",
-								zap.String("event message", message),
-								zap.Error(err))
-						} else {
-							eventCache.AddEvent(event)
-						}
-					}
 				}
 				return alloc
 			}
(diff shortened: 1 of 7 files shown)


pkg/scheduler/partition.go
```diff
@@ -704,6 +704,28 @@ func (pc *PartitionContext) removeNodeAllocations(node *objects.Node) ([]*object
...
 					// ignore the return as that is the same as alloc, the alloc is gone after this call
 					_ = app.ReplaceAllocation(allocID)
+					// we need to check the resources equality
+					delta := resources.Sub(release.AllocatedResource, alloc.AllocatedResource)
+					// Any negative value in the delta means that at least one of the requested resource in the
+					// placeholder is larger than the real allocation. The nodes are correct the queue needs adjusting.
+					// The reverse case is handled during allocation.
+					if delta.HasNegativeValue() {
+						// this looks incorrect but the delta is negative and the result will be a real decrease
+						err := app.GetQueue().IncAllocatedResource(delta, false)
```
manirajv06, line 714: Would this new queue adjustment cause any issues in conjunction with below adjustment in place? app.GetQueue().DecAllocatedResource(alloc.AllocatedResource)

Can we double check? May be unit test to removeNode with replacement on different node ensures this?

Other than this, overall changes looks good.

manirajv06 (changes requested):

pkg/scheduler/objects/application.go
```diff
@@ -827,13 +827,41 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator,
 			if ph.released || request.taskGroupName != ph.taskGroupName {
 				continue
 			}
+			// before we check anything we need to check the resources equality
+			delta := resources.Sub(ph.AllocatedResource, request.AllocatedResource)
```
chenya-zhang, line 831: Maybe we could replace "ph" (placeholder) with "hold" (v.s. "request") to make our codebase easier to understand. :) But if it appears in too many places, it's okay.

pkg/scheduler/objects/application.go
```diff
@@ -827,13 +827,41 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator,
...
+			// Any negative value in the delta means that at least one of the requested resource in the real
+			// allocation is larger than the placeholder. We need to cancel the placeholder and check the next
+			// placeholder. This should trigger a cleanup of all the placeholder as a task group is always one size.
```
chenya-zhang, line 834: > This should trigger a cleanup of all the placeholder as a task group is always one size.

It sounds to me that all other placeholders will be removed (= cleaned up)?

pkg/scheduler/objects/application.go
```diff
@@ -827,13 +827,41 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator,
...
+			if delta.HasNegativeValue() {
+				log.Logger().Warn("releasing placeholder: real allocation is larger than placeholder",
+					zap.String("requested resource", request.AllocatedResource.String()),
+					zap.String("placeholderID", ph.UUID),
+					zap.String("placeholder resource", ph.AllocatedResource.String()))
+				// release the placeholder and tell the RM
+				ph.released = true
+				sa.notifyRMAllocationReleased(sa.rmID, []*Allocation{ph}, si.TerminationType_TIMEOUT, "cancel placeholder: resource incompatible")
```
chenya-zhang, line 842: > cancel placeholder: resource incompatible

Will it be easier to understand if we say "requested resource larger than placeholder" directly? "resource incompatible" may also refer to resource type.

pkg/scheduler/objects/application.go
pbacsko: I would rephrase it as "task group has a size of one", sounds more correct.

pkg/scheduler/objects/application.go
pbacsko, line 842: `si.TerminationType_TIMEOUT` - can this be changed to something else or would we break API with that? Something like `si.TerminationType_RESOURCE_MISMATCH`.

pkg/scheduler/objects/node_test.go
```diff
@@ -488,6 +488,40 @@ func TestRemoveAllocation(t *testing.T) {
...
+	allocID := "real-1"
+	piece := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 25, "second": 50})
+	alloc := newAllocation(appID1, allocID, nodeID1, "queue-1", piece)
+	// calculate the delta: new allocation resource - placeholder (should be negative!)
+	delta := resources.Sub(piece, half)
+	assert.Assert(t, delta.HasNegativeValue(), "expected negative values in delta")
+	// swap and check the calculation
+	node.ReplaceAllocation(phID, alloc, delta)
+	if node.GetAllocation(allocID) == nil {
+		t.Fatal("failed to replace allocation: allocation not returned")
```
pbacsko, line 514: Nit: can't this be an assert as well?

pkg/scheduler/objects/node_test.go
```diff
@@ -488,6 +488,40 @@ func TestRemoveAllocation(t *testing.T) {
...
+	}
+	assert.Assert(t, resources.Equals(node.GetAllocatedResource(), piece), "allocated resource not set correctly %v got %v", piece, node.GetAllocatedResource())
+
+	// clean up all should be zero
+	if node.RemoveAllocation(allocID) == nil {
+		t.Error("allocation should have been removed but was not")
```
pbacsko, line 520: Same here

pkg/scheduler/objects/node.go
```diff
@@ -316,15 +316,17 @@ func (sn *Node) AddAllocation(alloc *Allocation) bool {
 	return false
 }
 
-// Replace the paceholder allocation on the node. No usage changes as the placeholder must
-// be the same size as the real allocation.
-func (sn *Node) ReplaceAllocation(uuid string, replace *Allocation) {
+// ReplaceAllocation replaces the placeholder with the real allocation on the node.
+// The delta passed in is the difference in resource usage between placeholder and real allocation.
+// It should always be a negative value (really a decreased usage).
```
pbacsko, line 321: If this is a strong requirement, then it might make sense to at least log a warning if this expectation (`delta.HasNegativeValue() == true`) is not met. This can lead to bugs that are hard to figure out.

pkg/scheduler/objects/application.go
wilfred-s: Updated comment to be clearer

pkg/scheduler/objects/application.go
wilfred-s: I did not introduce a new state as the processing for both would be exactly the same in both the shim and the core. It would also require a new SI and a new shim to make the same changes work.

wilfred-s: Correct because the resource could be incompatible. The real allocation has a reference to resource type GPU with a value of 1, but the placeholder does not which results in GPU becoming negative in the `delta`. The same real allocation could also ask for memory 2GB while the placeholder specifies 4GB. The "incompatible resource" covers these mixed cases better than just the larger than message.
The event generated,  which is what the end user sees, has all the detail needed.

pkg/scheduler/objects/node.go
wilfred-s: added log entry

pkg/scheduler/objects/node_test.go
wilfred-s: We should file a follow up for that, the node testing code has a large number of places where it should  be using asserts.

Nil object checks do look a bit weird in the assert, might not improve readability:
```
assert.Assert(t, node.GetAllocation(allocID) != nil, "failed to replace allocation: allocation not returned")
```

pkg/scheduler/partition.go
wilfred-s: The update of the queue is enclosed in a check for nodeID difference. The last line of that block is a continue call. when we get into the block we can never reach line 770 which does the queue update for the other cases.

pkg/scheduler/objects/application.go
wilfred-s: too big a change

pbacsko (approved):
LGTM +1

craigcondit (approved):
+1. LGTM.

chenya-zhang (approved):
+1 LGTM, thanks @wilfred-s!

PR #357, opened 2022-01-12
[YUNIKORN-960] Expose cluster utilization info in new REST calls

anuraagnalluri:
### What is this PR for?
Move cluster utilization info from `/ws/v1/clusters/utilization` to `/ws/v1/partitions`.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-960

Modified `handlers_test.go`

Screen Shot 2022-01-19 at 8 35 54 PM

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -670,8 +670,11 @@ func getPartitionInfoDAO(lists map[string]*scheduler.PartitionContext) []*dao.Pa
 		partitionInfo.LastStateTransitionTime = partitionContext.GetStateTime().String()
 
 		capacityInfo := dao.PartitionCapacity{}
-		capacityInfo.Capacity = partitionContext.GetTotalPartitionResource().DAOString()
-		capacityInfo.UsedCapacity = partitionContext.GetAllocatedResource().DAOString()
+		capacity := partitionContext.GetTotalPartitionResource()
+		usedCapacity := partitionContext.GetAllocatedResource()
+		capacityInfo.Capacity = capacity.DAOString()
+		capacityInfo.UsedCapacity = usedCapacity.DAOString()
+		capacityInfo.Utilization = resources.CalculateAbsUsedCapacity(capacity, usedCapacity).DAOString()
 		partitionInfo.Capacity = capacityInfo
 		partitionInfo.NodeSortingPolicy = dao.NodeSortingPolicy{
 			Type:            partitionContext.GetNodeSortingPolicyType().String(),
--- a/pkg/webservice/dao/partition_info.go
+++ b/pkg/webservice/dao/partition_info.go
@@ -37,6 +37,7 @@ type PartitionInfo struct {
 type PartitionCapacity struct {
 	Capacity     string `json:"capacity"`
 	UsedCapacity string `json:"usedCapacity"`
+	Utilization  string `json:"utilization"`
 }
 
 type NodeInfo struct {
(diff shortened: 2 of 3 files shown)


anuraagnalluri:
@wilfred-s I wanted to test locally, but the only way I know to bring up the Web UI is to run through K8s. But installing Yunikorn with the 0.12.1 helm chart doesn't work anymore due to the recent decoupling scheduler/admission-controller changes. The helm chart specifies a default override to run `admission_util.sh` on the scheduler, but this results in the following error now:

```
yunikorn-scheduler-5b558dff96-cqv45   1/2     PostStartHookError: command '/bin/sh /admission_util.sh create' exited with 2: /bin/sh: can't open '/admission_util.sh': No such file or directory
```
How do you recommend I proceed here? Is there a workaround I can use to test the REST calls?

wilfred-s:
Testing of the rest API does not require anything but a running scheduler. The rest API is exposed in the scheduler binary. Check the [build and run guide](https://yunikorn.apache.org/docs/developer_guide/build). Check out all repos, use the overrides as explained and you can the create a standalone scheduler binary using `make run-local`. Otherwise build a scheduler docker image and just start that. You have the full rest API available to test.
REST changes can also be unit tested via the `handler_test.go`

wilfred-s:
Testing of the rest API does not require anything but a running scheduler. The rest API is exposed in the scheduler binary. Check the [build and run guide](https://yunikorn.apache.org/docs/developer_guide/build). Check out all repos, use the overrides as explained and you can the create a standalone scheduler binary using `make run-local`. Otherwise build a scheduler docker image and just start that. You have the full rest API available to test.
REST changes can also be unit tested see for examples the `handler_test.go`

wilfred-s:
> ```
> yunikorn-scheduler-5b558dff96-cqv45   1/2     PostStartHookError: command '/bin/sh /admission_util.sh create' exited with 2: /bin/sh: can't open '/admission_util.sh': No such file or directory
> ```
> 
> How do you recommend I proceed here? Is there a workaround I can use to test the REST calls?

The current master code cannot be installed with an already released helm chart. You would need to build a helm chart locally. See my previous comment on how to test.

anuraagnalluri:
@wilfred-s Thank you, tested locally and attached screenshots showing the REST path change.

manirajv06 (changes requested):
looks good. Minor nit.

pkg/webservice/handlers.go
```diff
@@ -670,8 +670,11 @@ func getPartitionInfoDAO(lists map[string]*scheduler.PartitionContext) []*dao.Pa
 		partitionInfo.LastStateTransitionTime = partitionContext.GetStateTime().String()
 
 		capacityInfo := dao.PartitionCapacity{}
-		capacityInfo.Capacity = partitionContext.GetTotalPartitionResource().DAOString()
-		capacityInfo.UsedCapacity = partitionContext.GetAllocatedResource().DAOString()
+		capacity := partitionContext.GetTotalPartitionResource()
```
manirajv06, line 673: Can you avoid using "capacity" variable and set the value directly to capacityInfo.Capacity? or do you see any other benefit?

manirajv06 (changes requested):

pkg/webservice/handlers.go
anuraagnalluri: I set the `DAOString()` to `capacityInfo.Capacity` on 675 below. On 677, I use this calculation to get the abs used capacity. If I don't store the calculation itself in its own variable, I either have to convert the string to `*resource.Resource` or recompute the capacity as an arg to `CalculateAbsUsedCapacity`.

manirajv06 (approved):

PR #362, opened 2022-01-14
[YUNIKORN-958] Expose node utilisation info in new REST calls

anuraagnalluri:
### What is this PR for?
Introduce a field in the output of the new REST call -- `/ws/v1/partition/{partition}/nodes` that displays % utilization.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* Modify `node_tests.go` to introduce new test.

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-958

* `node_tests.go` (still need to implement).

Screen Shot 2022-01-14 at 2 04 17 PM

### DIFF
--- a/pkg/scheduler/objects/node.go
+++ b/pkg/scheduler/objects/node.go
@@ -239,6 +239,27 @@ func (sn *Node) GetAvailableResource() *resources.Resource {
 	return sn.availableResource.Clone()
 }
 
+// Get the utilized resource on this node.
+func (sn *Node) GetUtilizedResource() *resources.Resource {
+	total := sn.GetCapacity()
+	resourceAllocated := sn.GetAllocatedResource()
+	m := make(map[string]bool)
+	utilizedResource := make(map[string]resources.Quantity)
+
+	for name, v := range total.Resources {
+		if v > 0 {
+			m[name] = true
+		}
+	}
+
+	for name := range resourceAllocated.Resources {
+		if _, ok := m[name]; ok {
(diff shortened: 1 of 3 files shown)


anuraagnalluri:
@manirajv06 Attached screenshot of added `utilized` field. I didn't specify that the unit for this was `%` in the field name. Should I rename to `UtilizedPct` or something of the like?

pkg/scheduler/objects/node.go
```diff
@@ -239,6 +239,27 @@ func (sn *Node) GetAvailableResource() *resources.Resource {
...
+	total := sn.GetCapacity()
+	resourceAllocated := sn.GetAllocatedResource()
+	m := make(map[string]bool)
+	utilizedResource := make(map[string]resources.Quantity)
+
+	for name, v := range total.Resources {
+		if v > 0 {
+			m[name] = true
+		}
+	}
```
manirajv06, line 253: Can we avoid this entire for loop and check each res value of "total" directly in below for loop?

manirajv06 (changes requested):
Also, please ensure unit test has been covered in already existing test methods in handlers_test.go

manirajv06:
> @manirajv06 Attached screenshot of added `utilized` field. I didn't specify that the unit for this was `%` in the field name. Should I rename to `UtilizedPct` or something of the like?

I don't think this is required. Specifying 'utilized' filed as displayed in screenshot is good enough.

PR #365, opened 2022-01-21
[YUNIKORN-949] Location of the state dump file should be configurable

anuraagnalluri:
### What is this PR for?
Location of the state dump file should be configurable instead of hardcoded. PR also provides a default retention policy to ensure the file does not grow unbounded.   

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-949

* Modified tests in handlers_test.go and manual testing

Screen Shot 2022-01-23 at 3 12 16 PM
Screen Shot 2022-01-23 at 3 12 31 PM
Screen Shot 2022-01-23 at 3 13 13 PM

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,6 +46,7 @@ require (
 	golang.org/x/text v0.3.2 // indirect
 	golang.org/x/tools v0.0.0-20200415000939-92398ad77b89 // indirect
 	google.golang.org/grpc v1.26.0
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 	gopkg.in/yaml.v2 v2.2.8
 	gotest.tools v2.2.0+incompatible
 	honnef.co/go/tools v0.0.1-2020.1.3 // indirect
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -90,9 +90,22 @@ partitions:
     queues:
       - name: root
 `
+
 const configDefault = `
 partitions:
   - name: default
+    queues:
+      - name: root
+        submitacl: "*"
+        queues:
+          - name: default
+          - name: noapps
+`
+
+const configStateDumpFilePath = `
+partitions:
+  - name: default
+    statedumpfilepath: "tmp/non-default-yunikorn-state.txt"
     queues:
       - name: root
         submitacl: "*"
@@ -1326,8 +1339,9 @@ func TestGetNodesUtilization(t *testing.T) {
 	assert.Equal(t, len(nodesDao), 0)
 }
 
-func TestGetFullStateDump(t *testing.T) {
-	schedulerContext = prepareSchedulerContext(t)
+func TestGetFullStateDumpDefaultPath(t *testing.T) {
+	schedulerContext = prepareSchedulerContext(t, false)
+	defer deleteStateDumpFile(t, schedulerContext)
 
 	partitionContext := schedulerContext.GetPartitionMapClone()
 	context := partitionContext[normalizedPartitionName]
@@ -1342,19 +1356,54 @@ func TestGetFullStateDump(t *testing.T) {
 	resp := &MockResponseWriter{}
 
 	getFullStateDump(resp, req)
-	receivedBytes := resp.outputBytes
 	statusCode := resp.statusCode
-	assert.Assert(t, len(receivedBytes) > 0, "json response is empty")
+	fi, err := os.Stat(defaultStateDumpFilePath)
+	assert.NilError(t, err)
+	assert.Assert(t, fi.Size() > 0, "json response is empty")
 	assert.Check(t, statusCode != http.StatusInternalServerError, "response status code")
 	var aggregated AggregatedStateInfo
+	receivedBytes, err := ioutil.ReadFile(defaultStateDumpFilePath)
+	assert.NilError(t, err)
+	err = json.Unmarshal(receivedBytes, &aggregated)
+	assert.NilError(t, err)
+	verifyStateDumpJSON(t, &aggregated)
+}
+
+func TestGetFullStateDumpNonDefaultPath(t *testing.T) {
+	stateDumpFilePath := "tmp/non-default-yunikorn-state.txt"
+	defer deleteStateDumpDir(t)
+	schedulerContext = prepareSchedulerContext(t, true)
+	defer deleteStateDumpFile(t, schedulerContext)
+
+	partitionContext := schedulerContext.GetPartitionMapClone()
+	context := partitionContext[normalizedPartitionName]
+	app := newApplication("appID", normalizedPartitionName, "root.default", rmID)
+	err := context.AddApplication(app)
+	assert.NilError(t, err, "failed to add Application to partition")
+
+	imHistory = history.NewInternalMetricsHistory(5)
+	req, err2 := http.NewRequest("GET", "/ws/v1/getfullstatedump", strings.NewReader(""))
+	assert.NilError(t, err2)
+	req = mux.SetURLVars(req, make(map[string]string))
+	resp := &MockResponseWriter{}
+
+	getFullStateDump(resp, req)
+	statusCode := resp.statusCode
+	fi, err := os.Stat(stateDumpFilePath)
+	assert.NilError(t, err)
+	assert.Assert(t, fi.Size() > 0, "json response is empty")
+	assert.Check(t, statusCode != http.StatusInternalServerError, "response status code")
+	var aggregated AggregatedStateInfo
+	receivedBytes, err := ioutil.ReadFile(stateDumpFilePath)
+	assert.NilError(t, err)
 	err = json.Unmarshal(receivedBytes, &aggregated)
 	assert.NilError(t, err)
 	verifyStateDumpJSON(t, &aggregated)
 }
 
 func TestEnableDisablePeriodicStateDump(t *testing.T) {
-	schedulerContext = prepareSchedulerContext(t)
-	defer deleteStateDumpFile(t)
+	schedulerContext = prepareSchedulerContext(t, false)
+	defer deleteStateDumpFile(t, schedulerContext)
 	defer terminateGoroutine()
 
 	imHistory = history.NewInternalMetricsHistory(5)
@@ -1373,7 +1422,7 @@ func TestEnableDisablePeriodicStateDump(t *testing.T) {
 	assert.Equal(t, statusCode, 0, "response status code")
 
 	waitForStateDumpFile(t)
-	fileContents, err2 := ioutil.ReadFile(stateDumpFilePath)
+	fileContents, err2 := ioutil.ReadFile(defaultStateDumpFilePath)
 	assert.NilError(t, err2)
 	var aggregated AggregatedStateInfo
 	err3 := json.Unmarshal(fileContents, &aggregated)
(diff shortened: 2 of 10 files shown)


anuraagnalluri:
@yangwwei @pbacsko First screenshot is example of how state dump files look after scheduler was running idly for 3+ days. In this PR, we cap the number of max files at 10 consisting of 10MB each using a log rotator pkg called `Lumberjack`.

Second image is monitoring the file size of the current file we're writing to. We always write to the non-timestamped default `yunikorn-state.txt` or user supplied path + filename via configmap as Peter recommended. When we reach file size limit (10MB in this case), `Lumberjack` will timestamp the file, delete any old ones if the max file number limit is reached, and start logging to a new non-timestamped file.

Third image is just a limited view of the JSON output in one of these files.

go.mod
```diff
@@ -46,6 +46,7 @@ require (
 	golang.org/x/text v0.3.2 // indirect
 	golang.org/x/tools v0.0.0-20200415000939-92398ad77b89 // indirect
 	google.golang.org/grpc v1.26.0
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0
```
yangwwei, line 49: MIT license, good to go
Just did a little more research, golang doesn't have a very standard way to do rolling logs, this library seems to be a reasonable choice. Also noticed this lib was mentioned in Zap's doc: https://github.com/uber-go/zap/blob/master/FAQ.md#does-zap-support-log-rotation.

pkg/webservice/state_dump.go
```diff
@@ -157,7 +160,20 @@ func doStateDump(w io.Writer) error {
 	if err != nil {
 		return err
 	}
-	if _, err = w.Write(prettyJSON); err != nil {
+
+	stateLog := log.New(w, "", 0)
```
yangwwei, line 164: do we need to new a state log every time?
should we just reuse the same object and use it to write the state data over time?

pkg/webservice/state_dump.go
```diff
@@ -170,16 +186,21 @@ func startBackGroundStateDump(period time.Duration) error {
...
 		var errMsg = "state dump already running"
-		log.Logger().Error(errMsg)
+		yunikornLog.Logger().Error(errMsg)
 		return fmt.Errorf(errMsg)
 	}
 
+	stateDumpFilePath, exists := os.LookupEnv(stateDumpFilePathEnvKey)
+	if !exists {
+		stateDumpFilePath = defaultStateDumpFilePath
+	}
```
yangwwei, line 196: We should avoid dup code, this is the same as 166 - 169
if this is for dynamic reloading new ENV, can we have a convenient function `getStateDumpFilePath()`

yangwwei (changes requested):

yangwwei:
hi @anuraagnalluri thanks for the PR, I did a quick review. @pbacsko please help to review this patch.

>However, when I edit the configmap, I don't see the environment variable change in the scheduler pod even after 30 min. I read about config hot refresh here, but am missing a few details. Is the hot refresh enabled by default, and if so, how long does it typically take?

Yes, it is enabled by default. YK loads configs from the configmap, and it keeps watching the configmap object, if some fields changes, YK does a reload and load all new values to memory.  But here it seems to be a bit different. Because you are directly fetching ENV var using the os library, you can only expect that work unless the ENV updates got propagated to the docker container.  And that is very unlikely not the case, you can see a similar post [here](https://stackoverflow.com/questions/45050050/can-i-modify-containers-environment-variables-without-restarting-pod-using-kube). If we want this to be a config supports hot-refresh, it cannot be a ENV var, it can be a partition level config property in the configmap, and it needs a bit more effort to load/reload that property on changes.

anuraagnalluri:
@yangwwei Understood, thank you for the explanation. I will try to change this to a partition level config property. Since the code for YK's monitoring of configmap object already exists, I assume I'd just need to include additional logic to set/unset the env var for `STATE_DUMP_FILE_PATH` based on existence of the config property. I'll modify the PR to reflect that soon.

pkg/webservice/state_dump.go
anuraagnalluri: Good point -- this may be wasteful. However, the log object requires being constructed from the response writer which is why I had it in this method. I'll familiarize myself more with how the response writer gets passed from `routes.go` and see if I can change this.

pkg/webservice/state_dump.go
```diff
@@ -157,29 +160,48 @@ func doStateDump(w io.Writer) error {
...
+
+	stateLog := log.New(w, "", 0)
+	stateDumpFilePath := getStateDumpFilePath()
+	stateLog.SetOutput(&lumberjack.Logger{
+		Filename:   stateDumpFilePath,
+		MaxSize:    10,
+		MaxBackups: 10,
+	})
+
+	if err = stateLog.Output(2, string(prettyJSON)); err != nil {
```
pbacsko, line 172: 2 looks like a magic number, pls extract to const

pkg/webservice/state_dump.go
```diff
@@ -157,29 +160,48 @@ func doStateDump(w io.Writer) error {
...
 		return err
 	}
-	if _, err = w.Write(prettyJSON); err != nil {
+
+	stateLog := log.New(w, "", 0)
+	stateDumpFilePath := getStateDumpFilePath()
+	stateLog.SetOutput(&lumberjack.Logger{
+		Filename:   stateDumpFilePath,
+		MaxSize:    10,
+		MaxBackups: 10,
```
pbacsko, line 169: shouldn't these two be configurable as well?

pbacsko (changes requested):
Left two minor comments.

pkg/webservice/handlers_test.go
```diff
@@ -1342,19 +1343,54 @@ func TestGetFullStateDump(t *testing.T) {
...
+
+	imHistory = history.NewInternalMetricsHistory(5)
+	req, err2 := http.NewRequest("GET", "/ws/v1/getfullstatedump", strings.NewReader(""))
+	assert.NilError(t, err2)
+	req = mux.SetURLVars(req, make(map[string]string))
+	resp := &MockResponseWriter{}
+
+	getFullStateDump(resp, req)
+	statusCode := resp.statusCode
+	fi, err := os.Stat(stateDumpFilePath)
```
anuraagnalluri, line 1379: This line will fail since `tmp/yunikorn-state.txt` doesn't exist. The `stateDumpFilePath` needs to be reflected in the `context` this test is reading. Trying to figure out the best way to set `context.StateDumpFilePath` for this test.

pkg/webservice/state_dump.go
anuraagnalluri: I'd like to leave as hardcoded for now if that's ok and file a follow-up JIRA to make them configurable.

pkg/webservice/state_dump.go
```diff
@@ -157,29 +161,52 @@ func doStateDump(w io.Writer) error {
...
 }
 
+func getStateDumpFilePath(cc *scheduler.ClusterContext) string {
+	cc.RLock()
+	defer cc.RUnlock()
+
+	for _, partition := range cc.GetPartitionMapClone() {
+		if partition.GetStateDumpFilePath() != "" {
+			return partition.GetStateDumpFilePath()
+		}
```
anuraagnalluri, line 187: Since state dump returns info of all partitions, but the state dump output path is a partition-level config property, we iterate through partitions until we find first occurrence of the property. This first occurrence will be used as the output path if multiple values are configured for different partitions. If no occurrences, we default to `yunikorn-state.txt`.

anuraagnalluri:
@yangwwei As discussed, I made the state dump file path a partition-level config parameter. This works as expected during testing (hot refresh takes effect and location of the output file changes after 2 minutes).

pkg/webservice/state_dump.go
anuraagnalluri: @yangwwei Maybe I'm missing something obvious but I couldn't get this working after several hours. I tried to use `zapcore` directly in `webservices.go` so we only instantiate one logger in `NewWebApp` as per the [zap/lumberjack documentation](https://github.com/uber-go/zap/blob/master/FAQ.md#does-zap-support-log-rotation).

However, it seems zap must log the message string as a value with `MessageKey` as per its [EncoderConfig](https://pkg.go.dev/go.uber.org/zap/zapcore#EncoderConfig). Then, I tried to unmarshal the nested JSON to a nested struct (ex. outer struct containing [this struct](https://github.com/apache/incubator-yunikorn-core/blob/master/pkg/webservice/state_dump.go#L51)). Problem with this is that the logged message is still surrounded by `"`, so it doesn't unmarshal in to the appropriate type (only string).

Happy to take another stab at it if you have any pointers, or want me to send a commit with screenshots of the zap log output to get a better idea of the issue. Functionally, things are working fine as per [YUNIKORN-949](https://issues.apache.org/jira/browse/YUNIKORN-949).

pkg/webservice/state_dump.go
pbacsko: Fine, OK for me.

pbacsko (commented):
LGTM

Just have a question - this file path is independent of the partitions, so it feels more logical to place this in `SchedulerConfig` instead of `PartitionConfig`. I know @yangwwei recommended this as a partition-level setting, just wondering if lifting it one level further has more drawbacks.

anuraagnalluri:
@pbacsko Thanks for the recommendation. @yangwwei can you advise?

pkg/webservice/state_dump.go
yangwwei: > Just have a question - this file path is independent of the partitions, so it feels more logical to place this in `SchedulerConfig` instead of `PartitionConfig`. I know @yangwwei recommended this as a partition-level setting, just wondering if lifting it one level further has more drawbacks.

Right now, we do not have a global "schedulerConfig" field, all configs are partition-based. It is not a big problem now because we only support a single partition by far.  I agree it is more logical to have some global config to config this, but it is not something that should be done in this PR. So current approach looks good.

yangwwei:
> Just have a question - this file path is independent of the partitions, so it feels more logical to place this in `SchedulerConfig` instead of `PartitionConfig`. I know @yangwwei recommended this as a partition-level setting, just wondering if lifting it one level further has more drawbacks.

Right now, we do not have a global "schedulerConfig" field, all configs are partition-based. It is not a big problem now because we only support a single partition by far.  I agree it is more logical to have some global config to config this, but it is not something that should be done in this PR. So current approach looks good.

yangwwei (approved):
We need a follow-up task to track the new config, which should be documented properly. @anuraagnalluri can you help to follow up on that?
@pbacsko feel free to merge this when you think it is ready.

anuraagnalluri:
@yangwwei Thank you, yes I will create a JIRA for documenting new config and assign it to myself.

yangwwei:
hi @anuraagnalluri thanks. @pbacsko are we good to go for this one?

pbacsko:
Ah sorry guys, for some reason, I thought that Weiwei had already merged this.

+1

pbacsko (approved):

PR #366, opened 2022-02-01
[YUNIKORN-1065] panic on node update

wilfred-s:
### What is this PR for?
If a node is not found during the update logging causes a panic as it
refers to the wrong object. Fixed another possible panic in node
processing in the context. Unit tests added to cover the panic cases.

Fixed layout and formatting issues in error messages and comments

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1065

Unit tests cover all problem cases.

### DIFF
--- a/pkg/metrics/event.go
+++ b/pkg/metrics/event.go
@@ -86,6 +86,16 @@ func initEventMetrics() CoreEventMetrics {
 	return metrics
 }
 
+func (em *eventMetrics) Reset() {
+	em.totalEventsCollected.Set(0)
+	em.totalEventsCreated.Set(0)
+	em.totalEventsChanneled.Set(0)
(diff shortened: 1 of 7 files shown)


manirajv06 (approved):
Looks good.

HuangTing-Yao (approved):
LGTM, I think we can close it after fix the HealthCheck UT problem.

wilfred-s:
The unit test failure is caused by my change. The call to `metrics.Reset()` in the health tests does not work It is an empty function. The tests I added trigger a metrics update and thus cause the health tests to fail.
Implementing a proper rest of all metrics that can be reset.

craigcondit (approved):
+1 LGTM.

PR #367, opened 2022-02-01
[YUNIKORN-1054] Add more queue metrics to Yunikorn

pbacsko:
### What is this PR for?
Adding more queue metrics to Yunikorn.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1054

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application_state.go
+++ b/pkg/scheduler/objects/application_state.go
@@ -117,25 +117,21 @@ func NewAppState() *fsm.FSM {
 			},
 		},
 		fsm.Callbacks{
+			// The state machine is tightly tied to the Application object.
+			//
+			// The first argument must always be an Application and if there is a second,
+			// that must be a string. If this precondition is not met, a runtime panic
+			// will occur.
 			"enter_state": func(event *fsm.Event) {
-				app, ok := event.Args[0].(*Application)
-				if !ok {
-					log.Logger().Warn("The first argument is not an Application")
-					return
-				}
+				app := event.Args[0].(*Application) //nolint:errcheck
 				log.Logger().Info("Application state transition",
 					zap.String("appID", app.ApplicationID),
 					zap.String("source", event.Src),
 					zap.String("destination", event.Dst),
 					zap.String("event", event.Event))
 				if len(event.Args) == 2 {
-					eventInfo, ok := event.Args[1].(string)
-					if ok {
-						app.OnStateChange(event, eventInfo)
-					} else {
-						log.Logger().Warn("unable to get additional event info")
-						app.OnStateChange(event, "")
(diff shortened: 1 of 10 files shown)


pbacsko:
Still in WIP. 
Discuss testing & `RMReleaseAllocationEvent.QueueNames`.

pkg/scheduler/objects/application_state.go
```diff
@@ -143,49 +141,53 @@ func NewAppState() *fsm.FSM {
 				}
 			},
 			"leave_state": func(event *fsm.Event) {
-				event.Args[0].(*Application).clearStateTimer()
+				event.Args[0].(*Application).clearStateTimer() //nolint:errcheck
```
wilfred-s, line 144: Lets do this consistently:
* Remove all error checks on cast from all callback definitions including the string cast
* Add the nolint comment to all casts 
* Document at the start of the callback declaration the fact that arg 0 must be an app object and arg 1 a string. Doing something different will cause a run time panic.

wilfred-s (changes requested):
Can we instead of passing the queue info all the way back up rely on the `partitionContext` code to trigger the queue metrics changes?
Every add and remove will go through the partition and in the partition code we have access to the application and thus the queue. We need to either increase or decrease the allocated resources in the queues on add and remove of the alloc. Check for the `removeNodeAllocation()` and `removeAllocation()` code in the partition context. Inside those we know the number of containers removed as we call `DecAllocatedResource()` with the sum of resources. We thus also know the count.
Similarly we can do the add of allocations.

I think that will simplify the change.

pbacsko:
TODO:
* Pending apps per queue
* Killed apps per queue (needed?)

pkg/scheduler/partition.go
```diff
@@ -1314,6 +1317,8 @@ func (pc *PartitionContext) removeAllocation(release *si.AllocationRelease) ([]*
 				zap.String("appID", appID),
 				zap.String("allocationId", uuid),
 				zap.Error(err))
+		} else {
+			metrics.GetQueueMetrics(queue.QueuePath).IncReleasedContainer()
```
wilfred-s, line 1321: I think this should be a GetQueuePath() call and not a direct access to the QueuePath

wilfred-s (changes requested):
One minor nit: unlocked QueuePath access

wilfred-s (approved):
LGTM

PR #370, opened 2022-02-04
[YUNIKORN-1052] Add more cluster-wide metrics to Yunikorn

pbacsko:
### What is this PR for?
Add more cluster-wide metrics to Yunikorn.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1052

### Questions:

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -636,7 +636,7 @@ func (cc *ClusterContext) updateNode(nodeInfo *si.NodeInfo) {
 	case si.NodeInfo_UPDATE:
 		var newReadyStatus bool
 		var err error
-		if newReadyStatus, err = strconv.ParseBool(nodeInfo.Attributes["ready"]); err != nil {
+		if newReadyStatus, err = strconv.ParseBool(nodeInfo.Attributes[objects.ReadyFlag]); err != nil {
 			log.Logger().Error("Could not parse ready attribute, assuming true", zap.Any("attributes", nodeInfo.Attributes))
 			newReadyStatus = true
 		}
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -21,6 +21,7 @@ package scheduler
 import (
 	"fmt"
 	"math"
+	"strconv"
 	"sync"
 	"time"
 
@@ -553,6 +554,9 @@ func (cc *ClusterContext) removePartition(partitionName string) {
 // nil nodeInfo objects must be filtered out before calling this function
 func (cc *ClusterContext) addNode(nodeInfo *si.NodeInfo, nodeCount int) error {
 	sn := objects.NewNode(nodeInfo)
+	if !sn.IsReady() {
+		metrics.GetSchedulerMetrics().IncUnhealthyNodes()
+	}
 	// if this node is unlimited, check the following things:
 	// 1. if reservation is enabled, reject the node
 	// 2. any other nodes in the list, reject the node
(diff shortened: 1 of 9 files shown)


pkg/scheduler/objects/node.go
```diff
@@ -72,9 +81,9 @@ func NewNode(proto *si.NodeInfo) *Node {
 		allocations:       make(map[string]*Allocation),
 		schedulable:       true,
 		listeners:         make([]NodeListener, 0),
+		ready:             ready,
```
pbacsko, line 84: On this abstraction level, this field could have a different name like `acceptingContainers` or just `healthy`.

pkg/scheduler/objects/node.go
```diff
@@ -49,6 +50,7 @@ type Node struct {
 	allocations       map[string]*Allocation
 	schedulable       bool
 	unlimited         bool
+	ready             bool
```
pbacsko, line 53: On this abstraction level, this field could have a different name like `acceptingContainers` or just `healthy`.

pkg/scheduler/context.go
```diff
@@ -631,6 +635,23 @@ func (cc *ClusterContext) updateNode(nodeInfo *si.NodeInfo) {
 
 	switch nodeInfo.Action {
 	case si.NodeInfo_UPDATE:
+		var newReadyStatus bool
+		var err error
+		if newReadyStatus, err = strconv.ParseBool(nodeInfo.Attributes["ready"]); err != nil {
+			log.Logger().Error("Could not parse ready attribute, assuming true", zap.Any("attributes", nodeInfo.Attributes))
+			newReadyStatus = true
```
wilfred-s, line 642: Are we sure that we get this information from the k8shim?
Looking at what is send by the shim we seem to only get the resources and some fixed rack related info

pkg/scheduler/objects/node.go
wilfred-s: I think so too, isn't it up to us to pick one?

pkg/scheduler/context.go
pbacsko: Please check https://github.com/apache/incubator-yunikorn-k8shim/pull/362. This is the shim side change.

pkg/scheduler/objects/node.go
wilfred-s: Checking the shim side and K8s itself I think we need to stick with _Ready_ to not confuse things.

wilfred-s (approved):
LGTM
went to the code again, with the shim change it looks good

wilfred-s:
@pbacsko can I commit this I don't think we need anything more at the moment?

pbacsko:
> @pbacsko can I commit this I don't think we need anything more at the moment?

Yeah, it's in good shape. I'll do two minor cleanup-related things:
1) Replace "ready" literal with a constant 
2) Simplify tests a bit, there is some code duplication

I'll post an update soon.

pkg/scheduler/context.go
```diff
@@ -640,10 +661,21 @@ func (cc *ClusterContext) updateNode(nodeInfo *si.NodeInfo) {
 	case si.NodeInfo_DRAIN_NODE:
 		// set the state to not schedulable
 		node.SetSchedulable(false)
+		metrics.GetSchedulerMetrics().IncDrainingNodes()
```
wilfred-s, line 664: oops, I overlooked an error here I think: if the node is already draining we should not increment again and we do not need to change state:
```
if node.IsSchedulable() {
	node.SetSchedulable(false)
	metrics.GetSchedulerMetrics().IncDrainingNodes()
}
```
this is handled the other way around

wilfred-s (changes requested):
late incoming change

pkg/scheduler/context.go
pbacsko: Fixed

wilfred-s (approved):
LGTM

PR #372, opened 2022-02-08
[YUNIKORN-961] Expose build and version information in a REST call

steinsgateted:
### What is this PR for?
in k8shim:
save go build version and build date, and pass to core via scheduler-interface.

in core:
save k8shim's information and the time the scheduler was started.
then the Rest call pushes all information out.

in scheduler-interface:
the message RegisterResourceManagerRequest is extended with the build data.
add a map<string, string> buildInfo = 4 to the message.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-961

### Questions:


wilfred-s:
The go.mod file needs an update to pick up the latest version of the scheduler interface. Please follow these steps to pick up the new scheduler interface changes:

1. Make sure you are in the directory where the scheduler interface code is located.
2. Update the local checked out code of the scheduler interface for the master branch to get the latest commit:
```git checkout master; git pull; git status```
The status should show up to date with the origin from where it was cloned.
1. Run the following command to get the pseudo version:
```TZ=UTC git --no-pager show --quiet --abbrev=12 --date='format-local:%Y%m%d%H%M%S'  --format="v0.0.0-%cd-%h"```
This command will print a line like this:
```v0.0.0-20220214023648-0de002ac41d3```
1. Update the go.mod file for the core repository
      -   Copy the whole generated version reference
      -   Open the go.mod file
      -   Replace the scheduler interface version reference with the one generated in step 5.

Save the go.mod file and commit/push the change to add it to this PR.

**Comment by wilfred-s** [comment]
Fixed the merge conflict caused by YUNIKORN-1073 to get the build going

**Comment by wilfred-s** [comment]
The solution works but would break when there is more than one RM. We need to prevent that.

We need to move the registration of the data into the `ClusterContext.processRMRegistrationEvent()` and track it for that specific RM. Since a RM can have more than one partition we should not store it in the partition (partition map uses the normalised name which is "[RM]partition"). A new field in the `ClusterContext` struct:
```rmInfos map[string]*rmInformation```
The _rmInformation_ is a simple struct. It should contain all the data from the build info (simple map). The registration or start date should be set when the object is created. We can add more details to that in the next steps. We do need to make sure that the fields are not ambiguous: Date could be anything, it is much clearer to use buildDate
Access to the _rmInformation_ would go through the cluster like the partition map cloning.

For making the data available on rest we should use a similar object to the rmInformation object in the dao.
I think we need to expose it as part of the cluster details and not as a separate endpoint as we could have multiple RMs etc.

File: pkg/webservice/dao/cluster_info.go [hunk 1]
```diff
@@ -18,18 +18,22 @@
...
+	TotalContainers       string              `json:"totalContainers"`
+	FailedContainers      string              `json:"failedContainers"`
+	PendingContainers     string              `json:"pendingContainers"`
+	RunningContainers     string              `json:"runningContainers"`
+	ActiveNodes           string              `json:"activeNodes"`
+	TotalNodes            string              `json:"totalNodes"`
+	FailedNodes           string              `json:"failedNodes"`
 }
+
+var ScheduleStartDate string
```
wilfred-s, line 39: Move this as a variable of type _time.Time_ with the name startTime ​into the `ClusterContext` object.
Set it as part of the `NewClusterContext` and `newClusterContext` calls.
Add a locked method to the clustercontext `GetStartTime() time.Time` to retrieve the value. 

That should also remove the dao dependency from the entrypoint which is much cleaner

pkg/webservice/handlers.go
```diff
@@ -200,25 +193,13 @@ func buildJSONErrorResponse(w http.ResponseWriter, detail string, code int) {
...
-
-	return &dao.BuildInfo{
-		ScheduleStartDate: dao.ScheduleStartDate,
-		RMBuildInfo:       rmBuildInfos,
-	}
-}
-
 func getClusterJSON(partition *scheduler.PartitionContext) *dao.ClusterDAOInfo {
 	clusterInfo := &dao.ClusterDAOInfo{}
+	clusterInfo.ScheduleStartDate = dao.ScheduleStartDate
```
wilfred-s, line 198: change to schedulerContext.GetStartTime and format it nicely here.

wilfred-s (changes requested):
Small change: the dao should only describe declare the object not include new variables. I should have picked that up in the earlier review.

wilfred-s:
This is looking good now, that side is good to go.

> Are we going to add some tests?

If you can then yes that would be the preferred option. We do have the limitation that we can only register one RM at the moment we need to keep that in mind. We can "fake" multiple RMs for the build info...
Simple test that we should consider:
- context test: register a RM with a couple of values, check that the cluster context has the correct details
- web service test: extend the existing TestGetClusterUtilJSON and check the RM details are there

steinsgateted:
Hi @wilfred-s,
Thanks for your explanation.
I would like some clarifications.
`context_test.go`: add `TestContext_AddRMBuildInformation` func.
Put the buildInfo of the two RMs into the `clusterContext`.
`handlers_test.go`: In `TestGetClusterUtilJSON` func, test some get functions. In order to test get functions in `handlers_test.go`, add `SetRMInfos` in `context.go`.

pkg/scheduler/context_test.go
```diff
@@ -255,6 +255,49 @@ func TestContext_AddNode(t *testing.T) {
 	}
 }
 
+func TestContext_AddRMBuildInformation(t *testing.T) {
+	context := createTestContext(t, pName)
+
+	rmID1 := "myCluster1"
+	BuildInfoMap1 := make(map[string]string)
```
wilfred-s, line 262: NIT: We do not have to export this `buildInfoMap` is enough.

pkg/scheduler/context_test.go
```diff
@@ -255,6 +255,49 @@ func TestContext_AddNode(t *testing.T) {
...
+	BuildInfoMap1["buildDate"] = "2006-01-02T15:04:05-0700"
+	BuildInfoMap1["buildVersion"] = "latest"
+	BuildInfoMap1["isPluginVersion"] = "false"
+	rm1 := &si.RegisterResourceManagerRequest{
+		RmID:        rmID1,
+		PolicyGroup: "policygroup",
+		Version:     "0.0.2",
+		BuildInfo:   BuildInfoMap1,
+	}
```
wilfred-s, line 271: Leverage the `SetRMInfos()` call that you have added. Create the test context as you did. Call `SetRMInfos()` with the buildInfoMap created here.
The first one should give a length of 1 etc (asserts like you have)
Create a second buildInfoMap object call `SetRMInfos()` again make sure that you have two objects in the list.
Update a value in one of the two maps (like buildVersion is 1.0.0) Call `SetRMInfos()` again with the updated map and make sure it is replaced as expected.
That will give us a full test of the code

pkg/webservice/handlers_test.go
```diff
@@ -607,6 +607,21 @@ func TestGetClusterUtilJSON(t *testing.T) {
 	assert.NilError(t, err, "Error when load clusterInfo from config")
 	assert.Equal(t, 1, len(schedulerContext.GetPartitionMapClone()))
 
+	// check build information of RM
+	BuildInfoMap := make(map[string]string)
```
wilfred-s, line 611: NIT: We do not have to export this `buildInfoMap` is enough.

pkg/scheduler/context.go
```diff
@@ -164,6 +173,15 @@ func (cc *ClusterContext) processRMRegistrationEvent(event *rmevent.RMRegistrati
...
 
+	// store the build information of RM
+	if cc.rmInfos == nil {
+		cc.rmInfos = make(map[string]*RMInformation)
+	}
+	cc.rmInfos[rmID] = &RMInformation{
+		RMBuildInformation: event.Registration.BuildInfo,
+	}
+	cc.rmInfos[rmID].RMBuildInformation["rmId"] = rmID
+
```
wilfred-s, line 184: Instead of duplicating this from `SetRMInfos()` we should call it here.

wilfred-s (changes requested):
Some code simplification to make. We're almost there :-)

steinsgateted:
Modified. Thanks.

wilfred-s (approved):
LGTM

PR #376, opened 2022-02-24
[YUNIKORN-1092] reservation matching on full names only

wilfred-s:
### What is this PR for?
Reservation matching should match the name of the app or node up until
the separator. Similar names would otherwise cause false positives while
being checked: i.e. node-1 and node-10, node-1 matches node-10's
reservations.

### What type of PR is it?
* [X ] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1092

unit tests are added to test the change


craigcondit (approved):
+1 LGTM.

PR #377, opened 2022-03-01
[YUNIKORN-1099] Fix deadlock in Application.notifyRMAllocationReleased()

craigcondit:
### What is this PR for?
Fixes a deadlock in Application.notifyRMAlloocationReleased(), and adds missing AllocationKey entry.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1099

Updated tests and deadlock no longer occurs when run in real scheduler.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1432,20 +1432,30 @@ func (sa *Application) notifyRMAllocationReleased(rmID string, released []*Alloc
 	if len(released) == 0 || sa.rmEventHandler == nil {
 		return
 	}
+	c := make(chan *rmevent.Result)
 	releaseEvent := &rmevent.RMReleaseAllocationEvent{
 		ReleasedAllocations: make([]*si.AllocationRelease, 0),
 		RmID:                rmID,
+		Channel:             c,
 	}
 	for _, alloc := range released {
 		releaseEvent.ReleasedAllocations = append(releaseEvent.ReleasedAllocations, &si.AllocationRelease{
 			ApplicationID:   alloc.ApplicationID,
 			PartitionName:   alloc.PartitionName,
+			AllocationKey:   alloc.AllocationKey,
 			UUID:            alloc.UUID,
 			TerminationType: terminationType,
 			Message:         message,
 		})
 	}
 	sa.rmEventHandler.HandleEvent(releaseEvent)
+	// Wait from channel
+	result := <-c
+	if result.Succeeded {
+		log.Logger().Debug("Successfully synced shim on released allocations. response: " + result.Reason)
+	} else {
+		log.Logger().Info("failed to sync shim on released allocations")
+	}
 }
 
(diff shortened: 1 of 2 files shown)


craigcondit:
@manirajv06 can you take a look? This appears to have been a miss in YUNIKORN-462.

manirajv06 (commented):
@craigcondit Thanks. I am wondering why ReSyncSchedulerCache was not being called in this context earlier? But I think it should have been called even before. Right? Did we missed this even before? @wilfred-s 

Also, Should we need to clear cache in case of RMReleaseAllocationAskEvent events (application#notifyRMAllocationAskReleased() )?

Changes look good. Should we create a separate jira to refactor (or avoid duplicating) this method?

craigcondit:
> @craigcondit Thanks. I am wondering why ReSyncSchedulerCache was not being called in this context earlier? But I think it should have been called even before. Right? Did we missed this even before? @wilfred-s

It's possible this was a miss before, but the new logic after the core->shim refactoring creates a deadlock and stops the scheduler cold. That's what this PR addresses.

> Also, Should we need to clear cache in case of RMReleaseAllocationAskEvent events (application#notifyRMAllocationAskReleased() )?

I think given the scheduler interface is meant to be generic, it's important to let the shim process these events, whether or not it requires them. We don't really "clear the cache" per se on any of them, we just update the state to reflect what's in the core.

> Changes look good. Should we create a separate jira to refactor (or avoid duplicating) this method?

I'm not 100% sure a refactoring is needed, but if so, let's do a follow-up JIRA. If the changes look good, can you add a +1?

Thanks!

wilfred-s:
>I am not sure whether I communicated clearly. Just to make it explicit, I meant, application#notifyRMAllocationReleased() and context#notifyRMAllocationReleased() methods.

The two duplicate functions cannot be avoided. The application is in a sub package from the scheduler, objects. The scheduler imports that package. That means the application cannot contain the context. To do that would mean importing the scheduler package. That is a circular dependency.
That dependency is worked around by keeping a reference to the RMEventHandler in the application.

> Also, Should we need to clear cache in case of RMReleaseAllocationAskEvent events (application#notifyRMAllocationAskReleased() )?

An ask is a pod that has not been scheduled yet. So it is a "new" pod that was created. That pod might or might not be in the cache just yet. It surely is not linked to a node. The only way that a pod is normally removed in that state is if it comes from the shim to the core. The only exception is a placeholder timeout. A placeholder that has not been placed on a node.
We should not be leaving anything around at the moment in the cache. Even for placeholders that timeout we clean up including the caches through the placeholder manager.

wilfred-s (approved):
LGTM

manirajv06:
Got it. Thanks.

PR #380, opened 2022-03-06
[Yunikorn 1093] Track rejected applications

steinsgateted:
### What is this PR for?
Display rejected applications information in REST.
### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1093

### Questions:


steinsgateted:
If the app is rejected, display this message in /ws/v1/apps :
```
[
  {
    "applicationID": "application-sleep-0003",
    "usedResource": "[]",
    "maxUsedResource": "[]",
    "partition": "default",
    "queueName": "",
    "submissionTime": 1646541113515060500,
    "finishedTime": 1646541113515105300,
    "allocations": [],
    "applicationState": "Rejected",
    "user": "nobody",
    "rejectionMessage": "failed to place application application-sleep-0003: application rejected: no placment rule matched"
  }
]
```
Because the rejected app was not added to the queue, there is no information in /ws/v1/partition/{partitioName}/queue/{queueName}/applications.

pkg/scheduler/context.go
```diff
@@ -503,6 +503,11 @@ func (cc *ClusterContext) handleRMUpdateApplicationEvent(event *rmevent.RMUpdate
 				ApplicationID: app.ApplicationID,
 				Reason:        err.Error(),
 			})
+			rejectedApp := objects.NewApplication(app, ugi, cc.rmEventHandler, request.RmID)
+			stateChangeErr := partition.addRejectedApplication(rejectedApp, err.Error())
```
wilfred-s, line 507: We do not need to store this in a local var, just pass in directly into the call.
Also see comment below around the signature for `addRejectedApplication`

pkg/scheduler/objects/application.go
```diff
@@ -1507,3 +1509,22 @@ func (sa *Application) IsAllocationAssignedToApp(alloc *Allocation) bool {
...
+func (sa *Application) SetRejectionMessage(rejectionMessage string) {
+	sa.Lock()
+	defer sa.Unlock()
+	sa.rejectionMessage = rejectionMessage
+}
+
+func (sa *Application) GetRejectionMessage() string {
+	sa.RLock()
+	defer sa.RUnlock()
+	return sa.rejectionMessage
```
wilfred-s, line 1522: Should add test coverage, specially when setting the message from the state change

pkg/scheduler/objects/application.go
```diff
@@ -1507,3 +1509,22 @@ func (sa *Application) IsAllocationAssignedToApp(alloc *Allocation) bool {
...
+}
+
+// only for rejected application
+func (sa *Application) SetFinishedTime() {
+	sa.Lock()
+	defer sa.Unlock()
+	sa.finishedTime = time.Now()
```
wilfred-s, line 1529: Add test coverage in `TestFinishedTime()`

pkg/scheduler/partition.go
```diff
@@ -443,6 +444,13 @@ func (pc *PartitionContext) getApplication(appID string) *objects.Application {
 	return pc.applications[appID]
 }
 
+func (pc *PartitionContext) getRejectedApplication(appID string) *objects.Application {
+	pc.RLock()
+	defer pc.RUnlock()
+
+	return pc.rejectedApplications[appID]
+}
+
```
wilfred-s, line 453: This is just used once in a test, in the line above in the test we directly check the map. Accessing the map in the test directly should be OK.

pkg/scheduler/partition.go
```diff
@@ -1445,3 +1490,19 @@ func (pc *PartitionContext) hasUnlimitedNode() bool {
...
 	return false
 }
+
+func (pc *PartitionContext) addRejectedApplication(rejectedApplication *objects.Application, rejectionMessage string) error {
+	if err := rejectedApplication.HandleApplicationEvent(objects.RejectApplication); err != nil {
+		log.Logger().Warn("Application state not changed to Rejected",
+			zap.String("currentState", rejectedApplication.CurrentState()),
+			zap.Error(err))
+		return err
+	}
```
wilfred-s, line 1500: Why are you returning this error? There is nothing that we can do with the error outside of this function. It also looks like it break processing of the message in the calling function.
This change should also never fail. Change this to something like `log.Logger().Warn("BUG: Unexpected failure...` and continue processing. Do not return the error.

NOTE: calling the event system will also call `OnStateChange()` as we enter a new state. That triggers an event to the shim to be created. That event has the wrong info in it as we do not want pass it back as an updated application. We need to filter out RejectApplication state changes.

pkg/scheduler/partition.go
```diff
@@ -1445,3 +1490,19 @@ func (pc *PartitionContext) hasUnlimitedNode() bool {
...
+	rejectedApplication.SetFinishedTime()
+	rejectedApplication.SetRejectionMessage(rejectionMessage)
```
wilfred-s, line 1502: Why do we not wrap this up in the state change call?
You already set the state timer in that call why not add these two to that action.
We also would not need functions to set these values in that case as the application object can not be accessed by anything yet.

pkg/webservice/handlers.go
```diff
@@ -129,6 +129,9 @@ func getApplicationsInfo(w http.ResponseWriter, r *http.Request) {
 		for _, app := range partition.GetCompletedApplications() {
 			addDao(app)
 		}
+		for _, app := range partition.GetRejectedApplications() {
+			addDao(app)
+		}
```
wilfred-s, line 134: Can we add a test in the `handler_test` to cover this?

wilfred-s (changes requested):

steinsgateted:
Thanks for your explanation.

wilfred-s (approved):
LGTM
thank you for your contribution

PR #384, opened 2022-03-10
[YUNIKORN-790] Implement MaxApplications enforcement

chungen0126:
### What is this PR for?
Add the variable for running application and max running application, and make sure that running application will not more than  max running application.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:


pbacsko:
You accidentally committed an empty file `yunikorn_state.txt`. That should be removed from the PR.

pkg/scheduler/objects/application_state.go
```diff
@@ -142,6 +143,9 @@ func NewAppState() *fsm.FSM {
 			},
 			fmt.Sprintf("enter_%s", Starting.String()): func(event *fsm.Event) {
 				app := event.Args[0].(*Application) //nolint:errcheck
+				if app.queue != nil {
+					app.queue.runningApps++
+				}
```
pbacsko, line 148: We have to update the counter in the parent as well.

pkg/scheduler/objects/application_state.go
```diff
@@ -142,6 +143,9 @@ func NewAppState() *fsm.FSM {
 			},
 			fmt.Sprintf("enter_%s", Starting.String()): func(event *fsm.Event) {
 				app := event.Args[0].(*Application) //nolint:errcheck
+				if app.queue != nil {
+					app.queue.runningApps++
```
pbacsko, line 147: We have to update the counter in the parent as well, all the way up to root.

So we need an `incrementRunningApps()` method like:
```
func (sq *Queue) IncrementRunningApps() {
	sq.Lock()
	defer sq.Unlock()
	sq.runningApps++
	if sq.parent != nil {
		sq.parent.IncrementRunningApps()		
	}
}
```

pkg/scheduler/objects/application_state.go
```diff
@@ -171,6 +175,9 @@ func NewAppState() *fsm.FSM {
 			},
 			fmt.Sprintf("leave_%s", Running.String()): func(event *fsm.Event) {
 				app := event.Args[0].(*Application) //nolint:errcheck
+				if app.queue != nil {
+					app.queue.runningApps--
```
pbacsko, line 179: Same here, we need a new `DecRunningApps()`

```
func (sq *Queue) DecRunningApps() {
	sq.Lock()
	defer sq.Unlock()
	sq.runningApps--
	if sq.parent != nil {
		sq.parent.DecRunningApps()
	}
}
```

pkg/scheduler/objects/application.go
```diff
@@ -1264,6 +1264,9 @@ func (sa *Application) AddAllocation(info *Allocation) {
 // Add the Allocation to the application
 // No locking must be called while holding the lock
 func (sa *Application) addAllocationInternal(info *Allocation) {
+	if sa.queue != nil && sa.queue.maxRunningApps != 0 && sa.stateMachine.Is(Accepted.String()) && sa.queue.maxRunningApps <= sa.queue.runningApps {
+		return
```
pbacsko, line 1268: I don't think we can just return.

The application cannot continue its transition because maxRunningApps is exceeded.

Therefore, it has to be saved to a slice along with the current timestamp so that when `queue.runningApps < queue.maxRunningApps` we can manually trigger the state transition.

The timestamp is needed because we want to trigger the app which has been waiting for the longest amount of time.

pkg/scheduler/objects/application_state.go
pbacsko: At this point, we have to update the pending applications and start the one which was blocked first.

Just a very POC-like code:
```
pendingApps []*Applications
...
        // TODO: not optimal, a SubmissionTime-based min-Heap is better
	sort.Slice(pendingApps, func(i, j int) bool {
		if pendingApps[i].SubmissionTime.Before(pendingApps[j].SubmissionTime) {
			return true
		}
		return false
	})

	selected := pendingApps[0]
	selected.stateMachine.Event(RunApplication.String(), selected) // trigger app to run
...
```

This is really just a rough idea, need to be tested thoroughly.

pbacsko (changes requested):
We need much more comprehensive logic to achieve what is described in YUNIKORN-790.
Please look at my comments.

1. Add the missing stuff
2. I think direct field access should be avoided. We just don't know which goroutines trigger state transitions. We really must use accessor methods to manipulate queue fields (runningApps), but every mutation should be carefully considered.
3. Write at least some basic tests

pkg/scheduler/objects/application.go
pbacsko: As @wilfred-s mentioned in the JIRA, we have to have a `CanRun()` method in the queue, something like:

```
func (sq *Queue) CanRun() {
	sq.RLock()
	defer sq.RUnlock()
	
	ok := sq.runningApps < sq.maxRunningApps
	if sq.parent != nil {
		return ok && sq.parent.CanRun()
	} else {
		return ok
	}
}
```

pkg/scheduler/objects/application_state.go
chungen0126: I think that it just needs to be blocked and doesn't need to start here. It will still be allocated first by the rmEventHandler.

pbacsko: Have you tried it?

I think these changes first need to be tested on sample setup with KIND/Minikube to see what happens.

Could you do that? I would like to know what happens.

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1299,35 @@ func (sq *Queue) String() string {
...
+
+func (sq *Queue) canRun() bool {
+	sq.RLock()
+	defer sq.RUnlock()
+
+	if sq.maxRunningApps == 0 {
+		return true
+	}
+	ok := sq.runningApps < sq.maxRunningApps
+	if sq.parent != nil {
```
pbacsko, line 1329: Just an idea:
If `ok` is false, it means that the limit is reached and printing a message is useful.
I suggest this change:

```
ok := sq.runningApps < sq.maxRunningApps
if !ok {
  log.Logger().Info("Maximum number of runnable apps reached", zap.String("queue", sq.QueuePath))
}
```

or maybe we can return earlier:
```
ok := sq.runningApps < sq.maxRunningApps
if !ok {
  log.Logger().Info("Maximum number of runnable apps reached", zap.String("queue", sq.QueuePath))
  return false
}
if sq.parent != nil {
  return sq.parent.canRun()
}
return true
```

pbacsko (changes requested):
Please test the changes and share the results with us.

pkg/scheduler/objects/application.go
```diff
@@ -1264,6 +1264,9 @@ func (sa *Application) AddAllocation(info *Allocation) {
 // Add the Allocation to the application
 // No locking must be called while holding the lock
 func (sa *Application) addAllocationInternal(info *Allocation) {
```
pbacsko, line 1266: @chungen0126 

Today I had a discussion and code-checking session with @wilfred-s and our conclusion is that this method is not the right place for this check.

We think that you should put this into `Queue.TryAllocate()` like that:

```
func (sq *Queue) TryAllocate(iterator func() NodeIterator) *Allocation {
        if !sq.CanRun() {   // maxApplications reached
            return
        }
	if sq.IsLeafQueue() {
		// get the headroom
		headRoom := sq.getHeadRoom()
		// process the apps (filters out app without pending requests)
      }
...
```
If you can't run at any level, you just leave and return immediately. That way you won't calculate allocations and update data structures.

However, this has an effect on `CanRun()` because in `TryAllocate()` we're walking down the hierarchy from "root", rather than checking from a particular leaf queue.

So I think we need something like:
```
func (sq *Queue) CanRun() bool {
	sq.RLock()
	defer sq.RUnlock()

	if sq.maxRunningApps == 0 {
		return true
	}
	
	return sq.runningApps < sq.maxRunningApps
}
```

So we don't need to check the parent all the way back to "root", because at any given level, we already know that. Eg if we're at "root.users.dev", we already know that "root.users" can accept new applications, because we're updating the counters constantly.

I hope this makes sense.

chungen0126: I add another condition to make sure it will still allocate the running application.

pkg/scheduler/objects/application_state.go
```diff
@@ -143,6 +143,9 @@ func NewAppState() *fsm.FSM {
 			},
 			fmt.Sprintf("enter_%s", Starting.String()): func(event *fsm.Event) {
 				app := event.Args[0].(*Application) //nolint:errcheck
+				if app.queue != nil {
```
wilfred-s, line 146: The app must have a queue set if it is in starting state, there is no need for a nil check

pkg/scheduler/objects/application_state.go
```diff
@@ -172,6 +175,9 @@ func NewAppState() *fsm.FSM {
 			},
 			fmt.Sprintf("leave_%s", Running.String()): func(event *fsm.Event) {
 				app := event.Args[0].(*Application) //nolint:errcheck
+				if app.queue != nil {
```
wilfred-s, line 178: The app must have a queue set if it is in running state, there is no need for a nil check

pkg/scheduler/objects/queue.go
```diff
@@ -1067,6 +1073,11 @@ func (sq *Queue) TryAllocate(iterator func() NodeIterator) *Allocation {
 		headRoom := sq.getHeadRoom()
 		// process the apps (filters out app without pending requests)
 		for _, app := range sq.sortApplications(true) {
+			if !app.IsRunning() && !app.queue.canRun() {
```
wilfred-s, line 1076: You are in the queue, should not access the app for this info. It also does not change while walking over the application list. You can pre-calculate the value and use it.

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1303,42 @@ func (sq *Queue) String() string {
 	return fmt.Sprintf("{QueuePath: %s, State: %s, StateTime: %x, MaxResource: %s}",
 		sq.QueuePath, sq.stateMachine.Current(), sq.stateTime, sq.maxResource)
 }
+
+func (sq *Queue) incRunningApps() {
```
wilfred-s, line 1307: Should this increment and decrement below be using a lock on update?
Do need to make sure we do not lock the whole hierarchy, see incPendingResource for example.

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1303,42 @@ func (sq *Queue) String() string {
...
+func (sq *Queue) decRunningApps() {
+	sq.runningApps--
+	if sq.parent != nil {
+		sq.parent.decRunningApps()
+	}
+}
+
+func (sq *Queue) canRun() bool {
+	sq.RLock()
+	defer sq.RUnlock()
```
wilfred-s, line 1323: This will lock the current queue and then walk up the tree, locking all parents.
Please follow the same structure as we have for all over: first walk up the tree, check the top and then when unwinding check the current queue. Locking only when the current queue is checked (see getHeadRoom as an example)

pkg/scheduler/objects/utilities_test.go
```diff
@@ -35,25 +35,27 @@ const (
...
 // Create the root queue, base for all testing
 func createRootQueue(maxRes map[string]string) (*Queue, error) {
-	return createManagedQueueWithProps(nil, "root", true, maxRes, nil)
+	return createManagedQueueWithProps(nil, "root", true, maxRes, nil, zero)
 }
 
 // wrapper around the create call using the one syntax for all queue types
-func createManagedQueue(parentSQ *Queue, name string, parent bool, maxRes map[string]string) (*Queue, error) {
-	return createManagedQueueWithProps(parentSQ, name, parent, maxRes, nil)
+func createManagedQueue(parentSQ *Queue, name string, parent bool, maxRes map[string]string, maxApps uint64) (*Queue, error) {
```
wilfred-s, line 47: Instead of changing this signature add a new function that allows setting max apps and leave this as is. i.e. `createManagedQueueMaxApps(....)` calling the next one with a 0 value

pkg/scheduler/objects/utilities_test.go
```diff
@@ -35,25 +35,27 @@ const (
...
+	return createManagedQueueWithProps(parentSQ, name, parent, maxRes, nil, maxApps)
 }
 
 // create managed queue with props set
-func createManagedQueueWithProps(parentSQ *Queue, name string, parent bool, maxRes, props map[string]string) (*Queue, error) {
+func createManagedQueueWithProps(parentSQ *Queue, name string, parent bool, maxRes, props map[string]string, maxApps uint64) (*Queue, error) {
```
wilfred-s, line 52: Instead of changing this signature add a new function that allows setting max apps and leave this as is. i.e. `createManagedQueuePropsMaxApps(....)`
Move the real code there and make `createManagedQueueWithProps()` call that one

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1303,42 @@ func (sq *Queue) String() string {
...
+
+func (sq *Queue) GetRunningApps() uint64 {
+	sq.RLock()
+	defer sq.RUnlock()
+	return sq.runningApps
+}
+
+func (sq *Queue) SetMaxRunningApps(maxApps uint64) {
+	sq.RLock()
+	defer sq.RUnlock()
```
wilfred-s, line 1342: This is a read lock securing a write action...
Also please comment that this is test only as it is not called from anywhere else.

wilfred-s (changes requested):

wilfred-s:
please rebase so we can move this forward.

wilfred-s:
Superseded by #429

PR #386, opened 2022-03-15
[YUNIKORN-1120] Allow unit specifiers in core configs

craigcondit:
### What is this PR for?
This PR updates the scheduler core to allow unit specifiers in configuration.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1120

Unit tests added / updated

### Questions:
* [x] - There is breaking changes for older versions.


pkg/common/resources/quantity.go
```diff
@@ -0,0 +1,103 @@
...
+	"errors"
+	"math/big"
+	"regexp"
+	"strconv"
+)
+
+// This code handles parsing of SI units in quantities:
+// <quantity>     ::= <signedNumber><suffix>
+// <signedNumber> ::= <digits> | <sign><digits>
+// <sign>         ::= "+" | "-"
```
wilfred-s, line 31: In the config(s) we do not allow negative values. I don't think we should allow that here.

pkg/common/resources/quantity.go
```diff
@@ -0,0 +1,103 @@
...
+// <digit>        ::= 0 | 1 | ... | 9
+// <digits>       ::= <digit> | <digit><digits>
+// <suffix>       ::= <binarySI> | <decimalSI>
+// <binarySI>     ::= Ki | Mi | Gi | Ti | Pi | Ei
+// <decimalSI>    ::= "" | k | K | M | G | T | P | E
+// Additionally, ParseVCore supports decimalSI of 'm' to indicate millicore.
+
+var whitespace = regexp.MustCompile(`\s+`)
```
wilfred-s, line 39: Use a `strings.TrimSpace()` instead of a regexp. We do not want to allow "1 000 000" as a value which we do with the current setup. If we allow spaces between value and suffix we need to add it in the regexp as below.

pkg/common/resources/quantity.go
```diff
@@ -0,0 +1,103 @@
...
+var legal = regexp.MustCompile(`^(?P<Number>[+-]?[0-9]+)(?P<Suffix>[A-Za-z]*)$`)
```
wilfred-s, line 40: Narrow the regexp for the parts further:
```
var legal = regexp.MustCompile(`^(?P<Number>[+-]?[0-9]+)\s*(?P<Suffix>[[mkKMGTPE]i?]*)$`)
```
That complies better to the described suffixes. The only extra suffixes that are not filtered out are K, ki and mi

pkg/common/resources/quantity.go
```diff
@@ -0,0 +1,103 @@
...
+	return parse(value, true)
+}
+
+func parse(value string, milli bool) (Quantity, error) {
+	value = whitespace.ReplaceAllLiteralString(value, "")
+
+	parts := legal.FindStringSubmatch(value)
+	if parts == nil || len(parts) != 3 {
+		return 0, errors.New("invalid quantity")
+	}
```
wilfred-s, line 77: FindStringSubmatch returns nil if it does not match. In all other cases it returns a fixed slice of 3 parts.
Simplify to `len(parts) == 0`

wilfred-s (changes requested):

pkg/common/resources/quantity.go
craigcondit: Removed negative support.

pkg/common/resources/quantity.go
craigcondit: Done.

pkg/common/resources/quantity.go
craigcondit: Done.

pkg/common/resources/quantity.go
craigcondit: Done.

wilfred-s (approved):
LGTM

PR #387, opened 2022-03-17
[YUNIKORN-1124] Avoid passing empty nodeAttributes in UpdateNode request

manirajv06:
### What is this PR for?

Core initialise nodeAttributes map and set the default partition when shim is not passing nodeAttributes as part of UpdateRequest (for delete, restore & decommissioning actions).

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1124

### Questions:

### DIFF
--- a/pkg/rmproxy/rmproxy.go
+++ b/pkg/rmproxy/rmproxy.go
@@ -365,6 +365,9 @@ func (rmp *RMProxy) UpdateNode(request *si.NodeRequest) error {
 	go func() {
 		if len(request.Nodes) > 0 {
 			for _, node := range request.Nodes {
+				if len(node.GetAttributes()) == 0 {
+					node.Attributes = map[string]string{}
+				}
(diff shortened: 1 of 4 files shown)


manirajv06:
Notes:

1. Leaving "yunikorn.apache.org/nodeType" change as is for now. @craigcondit can comment on this.
2. I think we should handle "Ready" attribute also as any other node attributes. Managing each attribute in a separate place doesn't seem to be clean and organised, may didn't get our attention at times. Yes, it may change during UpdateNode as well but still we should handle it in `initializeAttribute`, `GetAttribute` methods using appropriate locks like how listeners are being handled.

wilfred-s (approved):
LGTM

wilfred-s:
@manirajv06 can you log jiras as follow up for the issues described under point 2?

PR #397, opened 2022-03-29
[YUNIKORN-1163] Return history of application state transitions in REST API

craigcondit:
### What is this PR for?
Expose application state transitions in the REST API as part of the application objects.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1163

Unit tests updated to verify event logs.

stateLog

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -53,6 +53,11 @@ const (
 	AppTagStateAwareDisable string = "application.stateaware.disable"
 )
 
+type EventLogEntry struct {
+	EventTime        time.Time
+	ApplicationState string
+}
+
 type Application struct {
 	ApplicationID  string
 	Partition      string
@@ -80,6 +85,7 @@ type Application struct {
 	gangSchedulingStyle  string                 // gang scheduling style can be hard (after timeout we fail the application), or soft (after timeeout we schedule it as a normal application)
 	finishedTime         time.Time              // the time of finishing this application. the default value is zero time
 	rejectedMessage      string                 // If the application is rejected, save the rejected message
+	eventLog             []*EventLogEntry       // event log for this application
 
 	rmEventHandler     handler.EventHandler
 	rmID               string
@@ -106,6 +112,7 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
 		placeholderAsk:       resources.NewResourceFromProto(siApp.PlaceholderAsk),
 		finishedTime:         time.Time{},
 		rejectedMessage:      "",
+		eventLog:             make([]*EventLogEntry, 0),
 	}
 	placeholderTimeout := common.ConvertSITimeout(siApp.ExecutionTimeoutMilliSeconds)
 	if time.Duration(0) == placeholderTimeout {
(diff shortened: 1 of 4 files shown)


yangwwei (approved):
LGTM, this will be very useful, thank you @craigcondit !

craigcondit:
@wilfred-s I know you had some ideas surrounding this as well. Is this what you had in mind?

pkg/scheduler/objects/application.go
```diff
@@ -53,6 +53,11 @@ const (
 	AppTagStateAwareDisable string = "application.stateaware.disable"
 )
 
+type EventLogEntry struct {
+	EventTime        time.Time
+	ApplicationState string
+}
```
wilfred-s, line 59: This should be called StateLogEntry, we log state changes not every event.

pkg/scheduler/objects/application.go
```diff
@@ -80,6 +85,7 @@ type Application struct {
 	gangSchedulingStyle  string                 // gang scheduling style can be hard (after timeout we fail the application), or soft (after timeeout we schedule it as a normal application)
 	finishedTime         time.Time              // the time of finishing this application. the default value is zero time
 	rejectedMessage      string                 // If the application is rejected, save the rejected message
+	eventLog             []*EventLogEntry       // event log for this application
```
wilfred-s, line 88: This should be called stateLog

pkg/scheduler/objects/application.go
```diff
@@ -142,6 +149,20 @@ func (sa *Application) SetState(state string) {
 	sa.stateMachine.SetState(state)
 }
 
+func (sa *Application) recordEvent(time time.Time, appState string, prevState string) {
```
wilfred-s, line 152: signature cleanup:
* time is always passed in as time.Now(), we can set it here. The nano second differences are not important.
* prevState is not used as the ordering provides the previous state already

pkg/scheduler/objects/application.go
```diff
@@ -281,6 +303,9 @@ func (sa *Application) timeoutStateTimer(expectedState string, event application
 				sa.notifyRMAllocationReleased(sa.rmID, toRelease, si.TerminationType_TIMEOUT, "releasing placeholders on app complete")
 				sa.clearStateTimer()
 			} else {
+				// HandleApplicationEvent assumes the lock to be held
+				sa.Lock()
+				defer sa.Unlock()
```
wilfred-s, line 308: We have two choices: 
1. move adding this lock and a lock review of `timeoutStateTimer()` to a new jira., or
2. fix all locking issues here: `getPlaceholderAllocations()` accesses the allocations map unlocked which might cause issues. `clearStateTimer()` in all cases but here is called under lock.
We discussed the `clearStateTimer()` call yesterday and I am not sure it causes a problem but we might as well be careful.

wilfred-s (changes requested):

pkg/scheduler/objects/application.go
craigcondit: Updated in new commit.

pkg/scheduler/objects/application.go
craigcondit: Updated in new commit.

pbacsko (approved):
+1 LGTM

PR #399, opened 2022-04-01
[YUNIKORN-1164] Fix inconsistent lock behavior in application.go

craigcondit:
### What is this PR for?
Fixes inconsistent locking in application.go:

* Change QueuePath to queuePath and use accessor methods from external code
* Remove QueuePath from String() as it is accessed from both locked and unlocked contexts
* Introduce RejectApplication() and FailApplication() methods to handle unlocked usage of HandleApplicationWithEventInfo()
* Update timeoutStateTimer() to acquire lock as it is run from a new goroutine
* Ensure getOutstandingRequests() calls sortRequests() under write lock

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1164

Core unit tests as well as k8shim unit and e2e tests compiled against this version.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -61,10 +61,10 @@ type StateLogEntry struct {
 type Application struct {
 	ApplicationID  string
 	Partition      string
-	QueuePath      string
 	SubmissionTime time.Time
 
 	// Private fields need protection
+	queuePath            string
 	queue                *Queue                    // queue the application is running in
 	pending              *resources.Resource       // pending resources from asks for the app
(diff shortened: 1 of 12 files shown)


craigcondit:
Codecov decrease is due to no logic changes being introduced, but some additional lock calls are now present.

yuchaoran2011 (approved):
Nice fixes, thanks!

PR #402, opened 2022-04-06
[YUNIKORN-1155] Placeholder timeout should not be initialized to the default value during recovery

pbacsko:
### What is this PR for?
When Yunikorn is restarted, the placeholder timeout is reset to the original value, regardless of how much time the placeholder had already spent in pending/running state. This PR fixes this issue by checking the creationTime attribute of the pod and adjusting it if necessary.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1155

### Questions:


pkg/scheduler/objects/application.go
```diff
@@ -114,10 +115,19 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
...
 	placeholderTimeout := common.ConvertSITimeout(siApp.ExecutionTimeoutMilliSeconds)
 	if time.Duration(0) == placeholderTimeout {
 		placeholderTimeout = defaultPlaceholderTimeout
 	}
+
+	var timeoutReached bool
+	timeoutReached, placeholderTimeout = app.adjustPlaceholderTimeout(placeholderTimeout, siApp)
+	if timeoutReached {
+		// Yunikorn was restarted and the placeholder timed out in the meantime
+		placeholderTimeout = time.Millisecond
```
pbacsko, line 128: 1 msec is basically instant timeout. There is a guard against the value 0 somewhere else, so this is the smallest allowed value.

pkg/scheduler/objects/application.go
```diff
@@ -137,6 +147,37 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
 	return app
 }
 
+// Adjust the placeholder timeout if "creationTime" is defined. It's used during Yunikorn restart,
+// in order to properly track how long a placeholder pod should be in "Running" state.
+func (sa *Application) adjustPlaceholderTimeout(placeholderTimeout time.Duration, siApp *si.AddApplicationRequest) (bool, time.Duration) {
+	creationTimeTag := siApp.Tags["creationTime"]
```
pbacsko, line 153: As I mentioned in https://github.com/apache/yunikorn-k8shim/pull/405/, this is a tag, but `AddApplicationRequest` can be extended with a `creationTime` field.

pkg/scheduler/objects/application.go
```diff
@@ -137,6 +147,37 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
...
+	var creationTime int64
+	if creationTimeTag != "" {
+		var err error
+		creationTime, err = strconv.ParseInt(creationTimeTag, 10, 64)
+		if err != nil {
```
wilfred-s, line 158: Move to `common.ConvertSITimestamp(ts string) time.Time` for the conversion.

Does this work as a simpler option to do the calculation:
```
created := common.ConvertSITimestamp(creationTimeTag)
newPT := created.Add(placeholderTimeout).sub(time.Now())
```

wilfred-s (changes requested):

pkg/common/utils.go
```diff
@@ -115,6 +120,54 @@ func ConvertSITimeout(millis int64) time.Duration {
 	return time.Duration(result)
 }
 
+// ConvertSITimeoutWithAdjustment Similar to ConvertSITimeout, but this function also adjusts the timeout if
```
craigcondit, line 123: Can we get unit tests for these please?

craigcondit (changes requested):
Looks good, but let's get some unit tests around these new functions.

craigcondit (approved):
+1 LGTM.

pkg/common/utils.go
```diff
@@ -30,6 +30,14 @@ import (
 
 	"github.com/apache/yunikorn-core/pkg/log"
 	interfaceCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
+const CreationTime = "CreationTime"
```
wilfred-s, line 36: This should be moved to the SI as it used in both shim and core. Needs to be added to YUNIKORN-1122 like you did with `ready`.

pkg/common/utils.go
```diff
@@ -30,6 +30,14 @@ import (
...
+
+var (
+	currentTime = time.Now
+	Undefined   = time.Time{}
```
wilfred-s, line 40: I don't think we need these, see other comments.

pkg/common/utils.go
```diff
@@ -115,6 +123,56 @@ func ConvertSITimeout(millis int64) time.Duration {
...
+}
+
+func adjustTimeout(timeout time.Duration, siApp *si.AddApplicationRequest) time.Duration {
+	creationTimeTag := siApp.Tags[interfaceCommon.DomainYuniKorn+CreationTime]
+	if creationTimeTag == "" {
+		return timeout
+	}
+
+	created := ConvertSITimestamp(creationTimeTag)
+	if created == Undefined {
```
wilfred-s, line 143: use `created.IsZero()` for this instead of using a variable.

pkg/common/utils.go
```diff
@@ -115,6 +123,56 @@ func ConvertSITimeout(millis int64) time.Duration {
...
+		return timeout
+	}
+	now := currentTime()
```
wilfred-s, line 146: Testing and all other cases should never need to override `time.Now()`. We do not care if it is exactly 1 or 10 minutes as long as it is less than what is set as the timeout.

pkg/common/utils.go
```diff
@@ -115,6 +123,56 @@ func ConvertSITimeout(millis int64) time.Duration {
 	return time.Duration(result)
 }
 
+// ConvertSITimeoutWithAdjustment Similar to ConvertSITimeout, but this function also adjusts the timeout if
+// "creationTime" is defined. It's used during Yunikorn restart, in order to properly track how long a placeholder pod should
+// be in "Running" state.
+func ConvertSITimeoutWithAdjustment(siApp *si.AddApplicationRequest) time.Duration {
+	result := ConvertSITimeout(siApp.ExecutionTimeoutMilliSeconds)
```
wilfred-s, line 130: What happens if I do not have a timeout set? Setting the default is part of the application code which means we'll break at this point. Setting the default should be moved into the `ConvertSITimeout()` call.

pkg/common/utils_test.go
```diff
@@ -133,3 +135,55 @@ func TestGetRequiredNodeFromAsk(t *testing.T) {
 	nodeName = GetRequiredNodeFromTag(tag)
 	assert.Equal(t, nodeName, "Node2")
 }
+
+func TestConvertSITimeoutWithAdjustment(t *testing.T) {
```
wilfred-s, line 139: Missing test cases with no timeout set and using the default.

pkg/common/utils.go
```diff
@@ -115,6 +123,56 @@ func ConvertSITimeout(millis int64) time.Duration {
...
+
+	log.Logger().Info("Adjusting placeholder timeout",
+		zap.Duration("timeout duration", timeout),
+		zap.Time("creation time", created),
+		zap.Time("expected timeout", expectedTimeout))
+
+	return expectedTimeout.Sub(now)
+}
+
+func ConvertSITimestamp(ts string) time.Time {
```
wilfred-s, line 165: Design decision to be made: do we support negative numbers in the timestamp? i.e. dates before `1970-01-01 00:00:00`
Also not sure if you want to print a warning when the string passed in is empty. Pollutes the logs and would happen often.

wilfred-s:
Looking through the change the default timeout handling is broken for this change. Checking if the default timeout needs to be used is done after the calculation. Which means that old placeholders using the default timeout do not get adjusted.

pkg/common/utils.go
pbacsko: Fixed

pkg/common/utils.go
pbacsko: Fixed.

pkg/common/utils.go
pbacsko: Fixed

wilfred-s (approved):
LGTM defaults work again

PR #404, opened 2022-04-10
[YUNIKORN-522] smoke test for dynamic queue clean up

steinsgateted:
### What is this PR for?
Add a `TestDynamicQueueCleanUp` function to `smoke_test.go`.
When app is Completed, partition manager should be able to clean up the dynamically created queue.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-522

### Questions:


craigcondit (changes requested):
According to the pre-commit test run, this adds an additional 41 seconds to the total test runtime. We need to structure this in such a way that we can customize the delay before queues are cleaned up during testing. Ideally, this entire test should execute in a few seconds max.

steinsgateted:
@craigcondit Thank you for your review.
Add a `SetCompletingTimeout` function in `pkg/scheduler/objects/application.go`, change 30 seconds to 100 milliseconds.

craigcondit (approved):
+1. Will merge shortly.

PR #407, opened 2022-04-25
[YUNIKORN-1183] pre-select node for daemon set pod

manirajv06:
### What is this PR for?

Use the "required node" being sent as part of daemonset pod spec straight away to avoid traversing all nodes for allocation. Do the allocation if there is a fit in the node otherwise reserve the same straight away. This change improves the allocation cycle significantly for daemon set pods.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1183

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -664,6 +664,14 @@ func (sa *Application) IsReservedOnNode(nodeID string) bool {
 // If the reservation fails the function returns false, if the reservation is made it returns true.
 // If the node and ask combination was already reserved for the application this is a noop and returns true.
 func (sa *Application) Reserve(node *Node, ask *AllocationAsk) error {
+	sa.Lock()
+	defer sa.Unlock()
+	return sa.ReserveInternal(node, ask)
+}
+
+// Unlocked version for Reserve that really does the work.
+// Must only be called while holding the application lock.
+func (sa *Application) ReserveInternal(node *Node, ask *AllocationAsk) error {
 	// create the reservation (includes nil checks)
 	nodeReservation := newReservation(node, sa, ask, true)
 	if nodeReservation == nil {
@@ -855,7 +863,7 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
 					zap.String("allocation key", request.AllocationKey))
 				return alloc
 			}
-			if err := sa.Reserve(node, request); err != nil {
+			if err := sa.ReserveInternal(node, request); err != nil {
 				log.Logger().Warn("Failed to reserve the required node",
 					zap.String("required node", node.NodeID),
 					zap.String("allocation key", request.AllocationKey),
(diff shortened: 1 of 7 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -663,8 +663,6 @@ func (sa *Application) IsReservedOnNode(nodeID string) bool {
 // If the reservation fails the function returns false, if the reservation is made it returns true.
 // If the node and ask combination was already reserved for the application this is a noop and returns true.
 func (sa *Application) Reserve(node *Node, ask *AllocationAsk) error {
-	sa.Lock()
-	defer sa.Unlock()
```
wilfred-s, line 667: We should create a reserveInternal that is unlocked and leave Reserve locked calling reserveInternal.
In the allocation track (part of the change call reserveInternal as you already have the lock.

pkg/scheduler/objects/node.go
```diff
@@ -412,12 +412,6 @@ func (sn *Node) IsValidFor(ask *AllocationAsk) error {
 		}
 		return nil
 	}
-	// ask has required node, just check if the node is with the expected ID
-	// ignore the unschedulable flag in this case
-	if ask.requiredNode != sn.NodeID {
-		return fmt.Errorf("ask %s is restricted to node %s, skipping node %s",
-			ask.AllocationKey, ask.requiredNode, sn.NodeID)
-	}
```
wilfred-s, line 420: We can do further cleanup here: IsValidFor only checks IsSchedulable. The places that we call this should never be hit by daemon set pods and so we don't need to special case and can just call `node.IsSchedulable()` instead of `node.IsValidFor()`

wilfred-s (changes requested):

pkg/scheduler/objects/application.go
```diff
@@ -665,6 +665,12 @@ func (sa *Application) IsReservedOnNode(nodeID string) bool {
 func (sa *Application) Reserve(node *Node, ask *AllocationAsk) error {
 	sa.Lock()
 	defer sa.Unlock()
+	return sa.ReserveInternal(node, ask)
+}
+
+// Unlocked version for Reserve that really does the work.
+// Must only be called while holding the application lock.
+func (sa *Application) ReserveInternal(node *Node, ask *AllocationAsk) error {
```
pbacsko, line 673: Nit: internal functions start with lower case

pkg/scheduler/objects/application.go
```diff
@@ -816,7 +822,7 @@ func (sa *Application) getOutstandingRequests(headRoom *resources.Resource, tota
 
 // Try a regular allocation of the pending requests
 // This includes placeholders
-func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator func() NodeIterator) *Allocation {
+func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator func() NodeIterator, getnode func(string) *Node) *Allocation {
```
pbacsko, line 825: nit: `getNode` or `nodeFn` instead of `getnode`

pkg/scheduler/objects/application.go
```diff
@@ -844,6 +850,27 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
 			}
 			continue
 		}
+
+		requiredNode := request.GetRequiredNode()
+		// does request (daemon set pods?) has any constraint to run on specific node?
+		if requiredNode != "" {
+			node := getnode(requiredNode)
+			alloc := sa.tryNode(node, request)
```
pbacsko, line 858: What about calling `node.IsSchedulable()` here? Wouldn't that make sense?

pbacsko (changes requested):
I've had two minor observations and a question

pkg/scheduler/objects/application.go
manirajv06: No, We shouldn't call the method as it is not required for daemon set pod and this `if` block is meant for daemon set pod only. Hence the removal of `isValidFor` method. There could be a situation wherein running a type of daemon set pods itself is mandatory to make the node as "schedulable/ready".

wilfred-s (approved):
LGTM my comments have been resolved

pbacsko:
+1

PR #408, opened 2022-05-15
[YUNIKORN-1129]cleanup yunikorn-state.txt file in tests

HuangTing-Yao:
### What is this PR for?
remove the txt file, which generated by test function.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1129

### Questions:

### DIFF
--- a/pkg/common/configs/config_test.go
+++ b/pkg/common/configs/config_test.go
@@ -659,8 +659,7 @@ partitions:
 	stateDumpFilePath := "yunikorn-state.txt"
 	assert.Equal(t, conf.Partitions[0].StateDumpFilePath, stateDumpFilePath)
 	assert.Equal(t, conf.Partitions[1].StateDumpFilePath, "")
-	err = os.Remove(stateDumpFilePath)
-	assert.Equal(t, err, nil)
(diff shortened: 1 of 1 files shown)


pkg/common/configs/config_test.go
```diff
@@ -659,6 +659,8 @@ partitions:
 	stateDumpFilePath := "yunikorn-state.txt"
 	assert.Equal(t, conf.Partitions[0].StateDumpFilePath, stateDumpFilePath)
 	assert.Equal(t, conf.Partitions[1].StateDumpFilePath, "")
+	err = os.Remove(stateDumpFilePath)
```
pbacsko, line 662: This should be in a `defer` call. 
Also, just use `assert.NilError()`.

pbacsko (changes requested):
Need a small change.

pkg/common/configs/config_test.go
HuangTing-Yao: done.

PR #411, opened 2022-05-18
[YUNIKORN-1218] Scheduler crashed with concurrent map access error in…

yangwwei:
With the latest 1.0 version, we observed that occasionally scheduler crashed and restarted. This was due to the health checker having unsafe access to the application map of the partition context, a concurrent read/write error may occur when an app gets added or deleted in/from the partition at the same time. We need to use a read lock to access the map.

### What is this PR for?
Fix a bug in the health checker code.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1218

A patch to reproduce this issue has been attached in JIRA https://issues.apache.org/jira/browse/YUNIKORN-1218. Without the patch, it panics after running a while; with the patch, the crash won't happen.


craigcondit (approved):
+1. Will merge shortly.

PR #413, opened 2022-05-26
[YUNIKORN-1184] [Core] preempt pods based on the priorities

manirajv06:
### What is this PR for?

[YUNIKORN-1183](https://issues.apache.org/jira/browse/YUNIKORN-1183) ensures required node has been used straight away and reservation happens only in case of lack of resources in that particular required node. To make use of reservations in next cycle, need to pre empt the pods based on the pod priority to free up space in the same node.  Pre emption process tries to find the victims to meet the resource requirements of DS asks based on the design discussed in jira. Based on the pre emption results (nothing but the bunch of asks running in the required node), core communicates to shim for releases of those victims using appropriate termination type.

Some "good to have" features discussed in the design has been planned to take up in future phases and scope has been limited to rollout out the whole DS scheduling feature as early as possible in a usable way with minimal functionality.

Filtering ensures daemon set pod hasn't been added to the victims list. Sorting has been based on the following multiple criteria one by one in the specified order:

1. By type (regular pods, opted out pods, driver/owner pods),
2. By priority (least priority ask placed first),
3. By Create time or age of the ask (younger ask placed first),
4. By resource (ask with lesser allocated resources placed first)

Above sorting algorithms ensures driver pods has been always used as a last resort. In each above sorting step, same asks (for example, asks of same type) could be lined up and those asks could be sorted in the next successive steps based on criteria defined in each step. For example, In case of same type, priority criteria would be applied. In case of same priority, age criteria would be applied and so on..

After sorting, victims selection process supports only "Multiple" for now. It traverses the list of asks from beginning of the sorted list and goes on till sum of the collected asks resources matches the expected resource requirement of the daemon ask. "Single" vs "Multiple" based on the asks type etc can be taken up later in future phases.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1184

### Questions:


pkg/scheduler/objects/application.go
```diff
@@ -1050,26 +1049,54 @@ func (sa *Application) tryReservedAllocate(headRoom *resources.Resource, nodeIte
...
+			}
+		} else if reserve.node.GetCapacity().FitInMaxUndef(ask.GetAllocatedResource()) {
+			log.Logger().Info("Triggering preemption process for daemon set ask",
+				zap.String("ds allocation key", ask.AllocationKey))
+
+			// try preemption and see if we can free up resource
+			preemptor := NewSimplePreemptor(reserve.node)
+
+			// Are there any victims/asks to preempt?
+			victims := preemptor.GetVictims(ask)
```
wilfred-s, line 1076: We only need the `AllocatedResource` size of the ask, we work on just one repeat always.

pkg/scheduler/objects/simplepreemptor.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+)
+
+type PreemptionContext struct {
+	Node        *Node
+	allocations []*AllocationAsk
+
+	sync.RWMutex
+}
```
wilfred-s, line 33: Do we need a context here or can we just pass in a node + resource size and handle it internally?

pkg/scheduler/objects/simplepreemptor.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (p *PreemptionContext) filterAllocations() {
+	p.Lock()
+	defer p.Unlock()
+	for _, allocation := range p.Node.GetAllAllocations() {
+		// skip daemon set pods
+		if allocation.Ask.GetRequiredNode() != "" {
+			continue
+		}
+		p.allocations = append(p.allocations, allocation.Ask)
```
wilfred-s, line 52: Allocations not Asks

pkg/scheduler/objects/utilities_test.go
```diff
@@ -106,11 +105,11 @@ func newApplicationWithTags(appID, partition, queueName string, tags map[string]
 	return NewApplication(siApp, user, nil, "")
 }
 
-func newApplicationWithHandler(appID, partition, queueName string) (*Application, *appEventHandler) {
+func NewApplicationWithHandler(appID, partition, queueName string) (*Application, *rmproxy.MockedRMProxy) {
```
wilfred-s, line 108: No need to export

wilfred-s (changes requested):

wilfred-s (approved):
LGTM, all comments resolved
one nit left on the imports in the simplepreemptor_test.go which I will resolve on commit.

PR #415, opened 2022-06-10
[YUNIKORN-1235] pass priority on pods from shim into the core

manirajv06:
### What is this PR for?
 Core just needs to read `priority` as integer directly and use it. Earlier, conversion happens between class name and value before using the actual integer value. Now, not required.

### What type of PR is it?
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1235

### Questions:


wilfred-s (approved):
LGTM

PR #417, opened 2022-06-22
[YUNIKORN-1242] gang: delay in scheduling after using all placeholders

wilfred-s:
### What is this PR for?
In the tryAllocate a check fails to take into account that an
application could request more real allocations than defined in the gang
request.
If multiple gang types are used this could cause a delay in scheduling
the "extra" requests for that gang type. Worst case scenario that could
be the placeholder timeout.

### What type of PR is it?
* [X ] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1242

unit tests are added


manirajv06 (approved):
looks good.

PR #418, opened 2022-06-22
[YUNIKORN-1246] release reservation of required node in tryAllocate

wilfred-s:
### What is this PR for?
In the tryAllocate release the reservation if an ask with a requiredNode
specified is allocated after being reserved on the node.
Handle the case that the specified node is not known in the core.

Cleanup: required node tests

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1246

unit tests cover the use case


manirajv06 (approved):
lgtm

wilfred-s:
codecov is not picking up the fact that the new code is only tested from the partition_test.go We should probably add tests for the allocate cycle into the objects for `queue` and `application`

PR #419, opened 2022-06-24
[YUNIKORN-1233] REST api shows negative request time for some allocations

pbacsko:
### What is this PR for?
After restarting Yunikorn, negative RequestTime values are shown in the REST API for allocations.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1233

### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation_test.go
+++ b/pkg/scheduler/objects/allocation_test.go
@@ -28,6 +28,7 @@ import (
 	"gotest.tools/assert"
 
 	"github.com/apache/yunikorn-core/pkg/common/resources"
+	siCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/allocation_test.go
```diff
@@ -129,6 +130,8 @@ func TestNewAllocFromNilSI(t *testing.T) {
 func TestNewAllocFromSI(t *testing.T) {
 	res, err := resources.NewResourceFromConf(map[string]string{"first": "1"})
 	assert.NilError(t, err, "Resource creation failed")
+	tags := make(map[string]string)
+	tags[siCommon.CreationTime] = "100"
```
wilfred-s, line 134: can we define a fixed time in the past for the test,  and use that to make it simple to test and check:
```
const past = 1640995200 // 2022-1-1 00:00:00
tags[siCommon.CreationTime] =  strconv.Itoa(past)
```
Would also be good to see the negative test case tested: no time set and get a -1 back.

wilfred-s (changes requested):
minor nit on testing

wilfred-s (approved):
LGTM +1

PR #421, opened 2022-07-03
[YUNIKORN-1252] expose timed out placeholders count in application REST info

steinsgateted:
### What is this PR for?
add timed out placeholders count in application REST info and add some placeholderData tests

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1252

### Questions:

### DIFF
--- a/pkg/webservice/dao/application_info.go
+++ b/pkg/webservice/dao/application_info.go
@@ -48,4 +48,5 @@ type PlaceholderDAOInfo struct {
 	Count         int64            `json:"count"`
 	MinResource   map[string]int64 `json:"minResource"`
 	Replaced      int64            `json:"replaced"`
+	Timeout       int64            `json:"timeout"`
 }
--- a/pkg/webservice/dao/application_info.go
+++ b/pkg/webservice/dao/application_info.go
@@ -48,5 +48,5 @@ type PlaceholderDAOInfo struct {
 	Count         int64            `json:"count"`
 	MinResource   map[string]int64 `json:"minResource"`
 	Replaced      int64            `json:"replaced"`
-	Timeout       int64            `json:"timeout"`
+	TimedOut      int64            `json:"timedout"`
 }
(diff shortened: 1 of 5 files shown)


pkg/webservice/dao/application_info.go
```diff
@@ -48,4 +48,5 @@ type PlaceholderDAOInfo struct {
 	Count         int64            `json:"count"`
 	MinResource   map[string]int64 `json:"minResource"`
 	Replaced      int64            `json:"replaced"`
+	Timeout       int64            `json:"timeout"`
```
wilfred-s, line 51: Missing "d" in the name: Time**d**Out Also capitalise the `O`
json field should be timedout also.

pkg/webservice/handlers.go
```diff
@@ -240,6 +240,7 @@ func getApplicationJSON(app *objects.Application) *dao.ApplicationDAOInfo {
 			Count:         taskGroup.Count,
 			MinResource:   taskGroup.MinResource.DAOMap(),
 			Replaced:      taskGroup.Replaced,
+			Timeout:       taskGroup.Timedout,
```
wilfred-s, line 243: Missing "d" in the name: Time**d**Out Also capitalise the `O`

wilfred-s (changes requested):
I do not think we really test if the timeout counter is updated and then correctly exposed.
If you rely on the counter to be correct and that is covered in other tests than OK, however seeing a handler test that covers it would be good.

pkg/webservice/dao/application_info.go
steinsgateted: Done.Thanks.

pkg/webservice/handlers.go
steinsgateted: Done.Thanks.

steinsgateted:
@wilfred-s Thanks for your review.
In `application.go`, add a `SetTimedOutPlaceholder` func for handler test.

wilfred-s (approved):
LGTM +1

PR #423, opened 2022-07-15
[YUNIKORN-1263] Add pending requests to applciation object in REST API

craigcondit:
### What is this PR for?
The REST API responses for application objects include allocations, but not requests. Pending requests (along with diagnostic information related to predicate failure) would be very useful.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1263

Added unit tests for log functionality, verified that new response objects appear as expected.

Example addition to the application object:

```
"requests": [
      {
        "allocationKey": "f137fab6-3cfa-4536-93f7-bfff92689382",
        "allocationTags": {
          "kubernetes.io/label/app": "pause",
          "kubernetes.io/label/applicationId": "pod-with-node-affinity-01",
          "kubernetes.io/label/queue": "root.sandbox",
          "kubernetes.io/meta/namespace": "default",
          "kubernetes.io/meta/podName": "with-node-affinity"
        },
        "requestTime": 1657917851845743911,
        "resource": {
          "memory": 500000000,
          "vcore": 100
        },
        "pendingCount": 1,
        "priority": "0",
        "queueName": "root.default",
        "requiredNodeId": "",
        "applicationId": "pod-with-node-affinity-01",
        "partition": "default",
        "placeholder": false,
        "placeholderTimeout": 0,
        "taskGroupName": "",
        "allocationLog": [
          {
            "message": "node(s) didn't match Pod's node affinity, node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate",
            "lastOccurrence": 1657917910919138494,
            "count": 81
          },
          {
            "message": "node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate, node(s) didn't match Pod's node affinity",
            "lastOccurrence": 1657917911531769848,
            "count": 504
          },
          {
            "message": "node(s) didn't match Pod's node affinity",
            "lastOccurrence": 1657917911532076451,
            "count": 1170
          }
        ]
      }
    ],
```
### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation_ask.go
+++ b/pkg/scheduler/objects/allocation_ask.go
@@ -20,7 +20,6 @@ package objects
 
 import (
 	"fmt"
-	"sort"
 	"sync"
 	"time"
 
@@ -58,9 +57,9 @@ type AllocationAsk struct {
 }
 
 type AllocationLogEntry struct {
-	Message   string
-	Timestamp time.Time
-	Count     int32
+	Message        string
+	LastOccurrence time.Time
+	Count          int32
 }
 
 func NewAllocationAsk(ask *si.AllocationAsk) *AllocationAsk {
@@ -197,7 +196,7 @@ func (aa *AllocationAsk) GetAllocatedResource() *resources.Resource {
 	return aa.AllocatedResource
 }
 
-// LogSchedulingFailure keeps track of preconditions not being met for an allocation
+// LogAllocationFailure keeps track of preconditions not being met for an allocation
 func (aa *AllocationAsk) LogAllocationFailure(message string, allocate bool) {
 	// for now, don't log reservations
 	if !allocate {
@@ -214,11 +213,11 @@ func (aa *AllocationAsk) LogAllocationFailure(message string, allocate bool) {
 		}
 		aa.allocLog[message] = entry
 	}
-	entry.Timestamp = time.Now()
+	entry.LastOccurrence = time.Now()
 	entry.Count++
 }
 
-// GetSchedulingLog returns a list of log entries corresponding to allocation preconditions not being met
+// GetAllocationLog returns a list of log entries corresponding to allocation preconditions not being met
 func (aa *AllocationAsk) GetAllocationLog() []*AllocationLogEntry {
 	aa.RLock()
 	defer aa.RUnlock()
@@ -227,15 +226,12 @@ func (aa *AllocationAsk) GetAllocationLog() []*AllocationLogEntry {
 	i := 0
 	for _, log := range aa.allocLog {
 		res[i] = &AllocationLogEntry{
-			Message:   log.Message,
-			Timestamp: log.Timestamp,
-			Count:     log.Count,
+			Message:        log.Message,
+			LastOccurrence: log.LastOccurrence,
+			Count:          log.Count,
 		}
 		i++
 	}
-	sort.SliceStable(res, func(i, j int) bool {
-		return res[i].Timestamp.Before(res[j].Timestamp)
-	})
 	return res
 }
 
--- a/pkg/scheduler/objects/allocation_ask.go
+++ b/pkg/scheduler/objects/allocation_ask.go
@@ -20,6 +20,7 @@ package objects
 
 import (
 	"fmt"
+	"sort"
 	"sync"
 	"time"
 
@@ -51,8 +52,15 @@ type AllocationAsk struct {
 	requiredNode     string
 	allowPreemption  bool
 	originator       bool
+	allocLog         map[string]*AllocationLogEntry
 
 	sync.RWMutex
+}
+
+type AllocationLogEntry struct {
+	Message   string
+	Timestamp time.Time
+	Count     int32
 }
 
 func NewAllocationAsk(ask *si.AllocationAsk) *AllocationAsk {
@@ -72,6 +80,7 @@ func NewAllocationAsk(ask *si.AllocationAsk) *AllocationAsk {
 		requiredNode:      common.GetRequiredNodeFromTag(ask.Tags),
 		allowPreemption:   common.GetPreemptionFromTag(ask.Tags),
 		originator:        ask.Originator,
+		allocLog:          make(map[string]*AllocationLogEntry),
 	}
 	// this is a safety check placeholder and task group name must be set as a combo
 	// order is important as task group can be set without placeholder but not the other way around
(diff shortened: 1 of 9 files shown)


pkg/scheduler/objects/allocation_ask.go
```diff
@@ -51,10 +52,17 @@ type AllocationAsk struct {
...
 	allowPreemption  bool
 	originator       bool
+	allocLog         map[string]*AllocationLogEntry
 
 	sync.RWMutex
 }
 
+type AllocationLogEntry struct {
+	Message   string
+	Timestamp time.Time
```
wilfred-s, line 62: Timestamp is ambiguous. If count > 1 the timestamp is linked to the latest occurrence. That either needs to show in the name or we need two: first  occurrence and most recent occurrence timestamp.

pkg/scheduler/objects/allocation_ask.go
```diff
@@ -188,6 +197,48 @@ func (aa *AllocationAsk) GetAllocatedResource() *resources.Resource {
 	return aa.AllocatedResource
 }
 
+// LogSchedulingFailure keeps track of preconditions not being met for an allocation
+func (aa *AllocationAsk) LogAllocationFailure(message string, allocate bool) {
```
wilfred-s, line 201: Comment differs from real function name

pkg/scheduler/objects/allocation_ask.go
```diff
@@ -188,6 +197,48 @@ func (aa *AllocationAsk) GetAllocatedResource() *resources.Resource {
...
+			Message: message,
+		}
+		aa.allocLog[message] = entry
+	}
+	entry.Timestamp = time.Now()
+	entry.Count++
+}
+
+// GetSchedulingLog returns a list of log entries corresponding to allocation preconditions not being met
+func (aa *AllocationAsk) GetAllocationLog() []*AllocationLogEntry {
```
wilfred-s, line 222: Comment differs from real function name

pkg/scheduler/objects/allocation_ask.go
```diff
@@ -188,6 +197,48 @@ func (aa *AllocationAsk) GetAllocatedResource() *resources.Resource {
...
+	for _, log := range aa.allocLog {
+		res[i] = &AllocationLogEntry{
+			Message:   log.Message,
+			Timestamp: log.Timestamp,
+			Count:     log.Count,
+		}
+		i++
+	}
+	sort.SliceStable(res, func(i, j int) bool {
+		return res[i].Timestamp.Before(res[j].Timestamp)
```
wilfred-s, line 237: Can we sort outside of the lock? This could even be postponed to the web handling when it is all done out of band.

pkg/scheduler/objects/application.go
```diff
@@ -1387,6 +1386,17 @@ func (sa *Application) getPlaceholderAllocations() []*Allocation {
...
+// get a copy of all requests of the application
+func (sa *Application) GetAllRequests() []*AllocationAsk {
+	sa.RLock()
+	defer sa.RUnlock()
+	var requests []*AllocationAsk
+	for _, req := range sa.requests {
+		requests = append(requests, req)
+	}
+	return requests
+}
```
wilfred-s, line 1398: Code duplication we already have an unlocked `getAllReqeusts()` This locked version should be a wrapper around the unlocked version.

pkg/webservice/dao/allocation_ask_info.go
```diff
@@ -0,0 +1,42 @@
...
+
+type AllocationAskDAOInfo struct {
+	AllocationKey      string                    `json:"allocationKey"`
+	AllocationTags     map[string]string         `json:"allocationTags"`
+	RequestTime        int64                     `json:"requestTime"`
+	ResourcePerAlloc   map[string]int64          `json:"resource"`
+	PendingCount       int32                     `json:"pendingCount"`
+	Priority           string                    `json:"priority"`
+	QueueName          string                    `json:"queueName"`
+	RequiredNodeID     *string                   `json:"requiredNodeId"`
```
wilfred-s, line 35: Why handle this one different than other optional values in objects?
A simple empty string would suffice, unless we go back and do this for all optional values.

pkg/webservice/handlers.go
```diff
@@ -224,6 +226,42 @@ func getApplicationJSON(app *objects.Application) *dao.ApplicationDAOInfo {
 		}
 		allocationInfo = append(allocationInfo, allocInfo)
 	}
+	for _, req := range requests {
```
wilfred-s, line 229: Move all this processing into its own function. Do the sorting in there as per comment earlier around unlocked sorting.

pkg/webservice/handlers.go
```diff
@@ -224,6 +226,42 @@ func getApplicationJSON(app *objects.Application) *dao.ApplicationDAOInfo {
...
+		count := req.GetPendingAskRepeat()
+		if count > 0 {
+			nodeID := req.GetRequiredNode()
+			nodePtr := &nodeID
+			if nodeID == "" {
+				nodePtr = nil
+			}
```
wilfred-s, line 236: See comment above on other optional values.

wilfred-s (changes requested):
As a generic comment: we need to assess direct access to fields inside the allocationAsk. The Ask fields are almost all write only on create and read only but we have introduced a mutex for access. The Tags for instance are directly referenced as `allocation.Tags`. The `ask.AllocatedResource` is also directly referenced throughout the scheduling cycle.

craigcondit:
> As a generic comment: we need to assess direct access to fields inside the allocationAsk. The Ask fields are almost all write only on create and read only but we have introduced a mutex for access. The Tags for instance are directly referenced as `allocation.Tags`. The `ask.AllocatedResource` is also directly referenced throughout the scheduling cycle.

Agreed. I'll open a follow-up JIRA for the ask / allocation refactoring.

pkg/scheduler/objects/allocation_ask.go
craigcondit: Renamed to `LastOccurrence`.

pkg/scheduler/objects/allocation_ask.go
craigcondit: Fixed.

pkg/scheduler/objects/application.go
craigcondit: Fixed, also renamed to `getAllRequestsInternal()` to match `addAllocationInternal()`.

pkg/webservice/dao/allocation_ask_info.go
craigcondit: Fixed.

pkg/webservice/handlers.go
craigcondit: Done.

pkg/webservice/handlers.go
craigcondit: Fixed.

wilfred-s (approved):
LGTM

PR #428, opened 2022-08-09
[YUNIKORN-1276] REST API for specific application

manirajv06:
### What is this PR for?
Exposed an REST API to fetch the specific application details.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1276

### Questions:


pkg/webservice/handlers.go
```diff
@@ -677,6 +677,62 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	queueErr := validateQueue(queueName)
+	if queueErr != nil {
+		buildJSONErrorResponse(w, queueErr.Error(), http.StatusBadRequest)
+		return
+	}
+	if len(vars) != 3 {
+		buildJSONErrorResponse(w, "Incorrect URL path. Please check the usage documentation", http.StatusBadRequest)
```
wilfred-s, line 706: We need to be consistent with these checks. I would expect a standard ordering:
1.  the number of arguments
2. the existence of each argument by name
3. the content of the argument syntax check (if required)

We do not seem to do that consistently in any of the calls. It means we do too much processing before the eventual failure.

pkg/webservice/handlers.go
```diff
@@ -677,6 +677,62 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getApplication(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	writeHeaders(w)
+	partition, partitionExists := vars["partition"]
+	if !partitionExists {
+		buildJSONErrorResponse(w, "Partition is missing in URL path. Please check the usage documentation", http.StatusBadRequest)
```
wilfred-s, line 685: Need to make this constants or something not add the same text in multiple places in the handler, same for all error messages for parameter checks.

pkg/webservice/handlers.go
```diff
@@ -677,6 +677,62 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	app := queue.GetApplication(application)
+	if app == nil {
+		buildJSONErrorResponse(w, "Application not found", http.StatusBadRequest)
+		return
+	}
+
+	appDao := make([]*dao.ApplicationDAOInfo, 0, 1)
```
wilfred-s, line 728: Do we really want to return an array of objects? We know we only have 1 application that we're returning.

wilfred-s (changes requested):

pkg/webservice/handlers.go
manirajv06: We followed the same order while redesigned the whole REST API's. But then realised that doing 2 before 1 gives clear and concise err message by doing simple check on request parameters. For example, Pointing out specific param is missing is very clear instead of saying generic incorrect url path or something like that. We had discussion around this as part of design and through jira comments as well. Finally, we landed up in the current order. If you still think, may be we can handle this new jira as this require changes in all API's.

wilfred-s: Dug in a bit further: the URL matching inside gorilla already filters out anything malformed. The definition of the route makes sure that we always get the values. As per the [examples](https://github.com/gorilla/mux#examples) in the docs:
```
Paths can have variables. They are defined using the format {name} or {name:pattern}. If a regular expression pattern is not defined, the matched variable will be anything until the next slash.
```
The mux code does a regexp on the URL to see if it matches the route. If that regexp does not match a 404 is returned. I have tried to craft a request that will call the handler which does not have the parameters set and I cannot. When I get to the point that the handler is called I have the values in the map. The content might be garbage but I have content. I was not able to match the route in any other way.

In conclusion: anything that does not map to a route will give back a 404. Check 1 and 2 are thus completely unneeded.

wilfred-s (approved):
LGTM
cleaned up nicely, really simple to read now

PR #429, opened 2022-08-11
[YUNIKORN-790] Implement MaxApplications enforcement

lixmgl:
### What is this PR for?
Adding maxApplication implementation for Yunikorn queue.
This PR includes and updates existing wip https://github.com/apache/yunikorn-core/pull/384 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?

1. local test passed by running "make test" in Yunikorn-core and Yunikorn-shim repos(refer to screenshot)

Screen Shot 2022-09-01 at 10 30 33 PM
Screen Shot 2022-09-01 at 10 32 28 PM

2.Local test result: (launched 1000 Spark jobs to 5 Yunikorn queues on k8s cluster which has 85 nodes)
Screen Shot 2022-08-30 at 10 27 58 PM

Total duration and Performance looks good.
Each Yunikorn queue set maxApplication to 12, I queried running jobs number every minute, didn't exceed the limit (12*5 queues= 60).  
[load test result@08_30.pdf](https://github.com/apache/yunikorn-core/files/9464459/load.test.result%4008_30.pdf)

### Questions:
NA


pkg/scheduler/objects/application_state.go
```diff
@@ -143,6 +143,7 @@ func NewAppState() *fsm.FSM {
 			},
 			fmt.Sprintf("enter_%s", Starting.String()): func(event *fsm.Event) {
 				app := event.Args[0].(*Application) //nolint:errcheck
+				app.queue.incRunningApps()
```
wilfred-s, line 146: This is the correct point (see the [usage tracking doc](https://docs.google.com/document/d/19FYOL1Uw7tSf74qyuVN40Fv-GdWd2KrBcHJpUQLTBZ4/edit#heading=h.481iug6rohb3)) but we need to make sure the metrics are updated in the same place as the queue itself.

pkg/scheduler/objects/queue.go
```diff
@@ -657,6 +660,9 @@ func (sq *Queue) addChildQueue(child *Queue) error {
 	if sq.IsDraining() {
 		return fmt.Errorf("cannot add a child queue when queue is marked for deletion: %s", sq.QueuePath)
 	}
+	if sq.maxRunningApps != 0 && sq.maxRunningApps < child.maxRunningApps {
+		return fmt.Errorf("parent maxRunningApps must be larger than child maxRunningApps")
+	}
```
wilfred-s, line 665: This belongs in the config validations check, like we do for the "other" resources. Otherwise we could fail during the config change and leave the system in a strange state.

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1304,67 @@ func (sq *Queue) String() string {
...
+		sq.parent.incRunningApps()
+	}
+	sq.internalIncRunningApps()
+}
+
+func (sq *Queue) internalIncRunningApps() {
+	sq.Lock()
+	defer sq.Unlock()
+	sq.runningApps++
+}
```
wilfred-s, line 1322: This can be folded into the `incRunningApps()`

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1304,67 @@ func (sq *Queue) String() string {
...
+		sq.parent.decRunningApps()
+	}
+	sq.internalDecRunningApps()
+}
+
+func (sq *Queue) internalDecRunningApps() {
+	sq.Lock()
+	defer sq.Unlock()
+	sq.runningApps--
+}
```
wilfred-s, line 1338: This can be folded into the `decRunningApps()`

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1304,67 @@ func (sq *Queue) String() string {
...
+
+func (sq *Queue) internalCanRun(ok bool) bool {
+	sq.RLock()
+	defer sq.RUnlock()
+	if sq.maxRunningApps == 0 {
+		return true && ok
+	}
+	return ok && sq.runningApps < sq.maxRunningApps
+}
```
wilfred-s, line 1358: I had to really look twice to grasp what was going on with the `ok` flag. Isn't this a much simpler flow and faster, as we fail fast if a parent has failed:
```
func (sq *Queue) canRun() bool {
	if sq == nil {
		return false
	}
	// check the parent(s)
	if sq.parent != nil {
		ok := sq.parent.canRun()
		if !ok {
			return false
		}
	}
	// check this queue only if all parents allowed it
	sq.RLock()
	defer sq.RUnlock()
	if sq.maxRunningApps == 0 {
		return true
	}
	return sq.runningApps < sq.maxRunningApps
}
```

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1304,67 @@ func (sq *Queue) String() string {
...
+
+func (sq *Queue) GetRunningApps() uint64 {
+	sq.RLock()
+	defer sq.RUnlock()
+	return sq.runningApps
+}
+
+func (sq *Queue) SetMaxRunningApps(maxApps uint64) {
```
wilfred-s, line 1366: Add a comment to the exported function to show it is test only

wilfred-s (changes requested):
configuration check needs to move to the configvalidator code.

pkg/scheduler/objects/application_state.go
lixmgl: Updated in https://github.com/apache/yunikorn-core/pull/435

pkg/scheduler/objects/queue.go
lixmgl: Can you point me where is other resources config validations check?
If you are referring this method: https://github.com/apache/yunikorn-core/blob/branch-1.0/pkg/scheduler/objects/queue.go#L207
I don't think it's a good idea to add this child maxRunningApps check into this method since we need to iterate each child.

pkg/scheduler/objects/queue.go
lixmgl: The reason I separated this logic into a new method instead of merging into incRunningApps() is: 
Based on my understanding, instruction order is not ensured in golang even with lock  (refer: https://stackoverflow.com/questions/60340382/can-i-use-lock-to-ensure-instruction-order)
If we merge into incRunningApps(), sq.Lock() could get executed before this recursion. Each parent could be locked during recursion.  
if sq.parent != nil {
		ok = sq.parent.canRun()
}

pkg/scheduler/objects/queue.go
lixmgl: same reason as before.

pkg/scheduler/objects/queue.go
lixmgl: same reason as before, also I updated logic in canRun https://github.com/apache/yunikorn-core/pull/435

pkg/scheduler/objects/queue.go
lixmgl: deleted this method since no longer needed.

wilfred-s:
I am not sure which PR is the one you are working on. The email pointed back to this PR:
config validation code for the resources of a queue compared to its parent is here https://github.com/apache/yunikorn-core/blob/master/pkg/common/configs/configvalidator.go#L79

pkg/scheduler/objects/application_state.go
lixmgl: Updated

lixmgl:
> I am not sure which PR is the one you are working on. The email pointed back to this PR: config validation code for the resources of a queue compared to its parent is here https://github.com/apache/yunikorn-core/blob/master/pkg/common/configs/configvalidator.go#L79

I merged two PRs together. Please only review this one.

lixmgl:
> 



> 

updated, please review.

pkg/scheduler/objects/queue.go
pbacsko: I want to have some deeper understanding with this.

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1306,83 @@ func (sq *Queue) String() string {
...
+		return false
+	}
+	ok := true
+	if sq.parent != nil {
+		ok = sq.parent.incAllocatingAcceptedAppsIfCanRun(sa)
+	}
+	return sq.InternalIncAllocatingAcceptedAppsIfCanRun(ok, sa)
+}
+
+func (sq *Queue) InternalIncAllocatingAcceptedAppsIfCanRun(ok bool, sa *Application) bool {
```
pbacsko, line 1353: Can we have something else instead of "ok"?

"ok" is basically an idiomatic variable for various things such as map lookup results.

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1306,83 @@ func (sq *Queue) String() string {
...
+	sq.Lock()
+	defer sq.Unlock()
+	if sq.maxRunningApps == 0 {
+		return true && ok
+	}
+	result := ok && (sq.runningApps+uint64(len(sq.allocatingAcceptedApps)) < sq.maxRunningApps)
+	if result {
+		_, contains := sq.allocatingAcceptedApps[sa.ApplicationID]
```
pbacsko, line 1361: Here, `contains` is typically named `ok`.

pkg/scheduler/objects/application_state.go
```diff
@@ -143,6 +143,10 @@ func NewAppState() *fsm.FSM {
...
+				app.queue.decAllocatingAcceptedApps(app)
+				metrics.GetQueueMetrics(app.queuePath).IncQueueApplicationsRunning()
```
pbacsko, line 148: We have this call at two places now: `enter_Starting` and `enter_Running`.

pkg/scheduler/objects/application_state.go
```diff
@@ -143,6 +143,10 @@ func NewAppState() *fsm.FSM {
...
+				metrics.GetSchedulerMetrics().IncTotalApplicationsRunning()
```
pbacsko, line 149: Same for this, line 174 and 175.

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1306,83 @@ func (sq *Queue) String() string {
...
+		ok = sq.parent.incAllocatingAcceptedAppsIfCanRun(sa)
+	}
+	return sq.InternalIncAllocatingAcceptedAppsIfCanRun(ok, sa)
+}
+
+func (sq *Queue) InternalIncAllocatingAcceptedAppsIfCanRun(ok bool, sa *Application) bool {
+	sq.Lock()
+	defer sq.Unlock()
+	if sq.maxRunningApps == 0 {
+		return true && ok
```
pbacsko, line 1357: just `return ok`

pkg/scheduler/objects/utilities_test.go
```diff
@@ -49,11 +49,20 @@ func createManagedQueue(parentSQ *Queue, name string, parent bool, maxRes map[st
 
 // create managed queue with props set
 func createManagedQueueWithProps(parentSQ *Queue, name string, parent bool, maxRes, props map[string]string) (*Queue, error) {
+	return createManagedQueuePropsMaxApps(parentSQ, name, parent, maxRes, props, uint64(0))
```
pbacsko, line 52: Does `maxApps` have to be an `uint64`? I mean I understand that it's more logical, but now we have these weird casts here and in line queue.go:1359. Simple `int` is the most portable data type.

pbacsko, line 52: Does `maxApps` have to be an `uint64`? I mean I understand that it's more logical, but now we have these weird casts here and in queue.go:1359. Simple `int` is the most portable data type.

pbacsko, line 52: Does `maxApps` have to be an `uint64`? I mean I understand that it's more logical, but now we have these weird casts here and in line queue.go:1359. Simple `int` is the most portable data type.

pbacsko (changes requested):
I have some concerns regarding some pieces of the code, please take a look at those, thanks.

pkg/scheduler/objects/utilities_test.go
pbacsko, line 52: Does `maxApps` have to be an `uint64`? I mean I understand that it's more logical, but now we have these weird casts here and in line queue.go:1359. Simple `int` is the most portable data type.

pbacsko, line 52: Does `maxApps` have to be an `uint64`? I mean I understand that it's more logical, but now we have these weird casts here and in queue.go:1359. Simple `int` is the most portable data type.

pbacsko (changes requested):
I have some concerns regarding some pieces of the code, please take a look at those, thanks.

pkg/scheduler/objects/application_state.go
lixmgl: good catch, removed the one in enter_Starting, kept the one in enter_Running

pkg/scheduler/objects/application_state.go
lixmgl: same

pkg/scheduler/objects/queue.go
lixmgl: updated

pkg/scheduler/objects/queue.go
lixmgl: updated, using success instead of ok

pkg/scheduler/objects/queue.go
lixmgl: updated, using ok instead of contains

pkg/scheduler/objects/utilities_test.go
lixmgl: updated to int64

lixmgl: same

lixmgl: same

lixmgl:
> I have some concerns regarding some pieces of the code, please take a look at those, thanks.

Thanks for reviewing the code. I have updated them and addressed all your comments. Please take another look.

pkg/scheduler/objects/queue.go
pbacsko: @lixmgl I read the stackoverflow topic and that's related to goroutine scheduling. If what you're saying is correct that would mean that sequantial consistency is broken. What you're doing here is single threaded, meaning everything runs on a single goroutine. To me it would be extremely suprirising if merging `internalIncRunningApps()` into `incRunningApps()` would result in incorrect behavior.

lixmgl: @pbacsko what i did is first find the parent recursively [method](https://github.com/apache/yunikorn-core/pull/429/files#diff-27632d48eb925e150a33bc92370ceaa66c31048018d11ca7a53a0b50ab7250acR1310 ) 
Then lock this parent sq only and incRunningApps. It's not single threaded. 
This way we can avoid locking all the parents during the recursion. 
I recall @wilfred-s pointed out this issue here https://github.com/apache/yunikorn-core/pull/384/files#r843616529 for combining into one method.

pkg/scheduler/objects/utilities_test.go
lixmgl: reverted to unsigned int type. cc @wilfred-s 
related discussion is here: https://github.com/apache/yunikorn-site/pull/188#discussion_r990446069

lixmgl: reverted to unsigned int type. cc @wilfred-s
related discussion is here: https://github.com/apache/yunikorn-site/pull/188#discussion_r990446069

lixmgl: same reason as before

lixmgl:
Hi @manirajv06 Can you take a look? I saw @wilfred-s requested a review from you 11 days ago.

pkg/scheduler/objects/queue.go
craigcondit: Inlining `internalIncRunningApps()` will not change the behavior in any way, because the unlock will happen prior to returning from `incRunningApps()` in either case. They are functionally equivalent. However, do be sure that the lock and defer still come immediately before incrementing `runningApps`. They should not be relocated to the top of `incRunningApps()`

pkg/scheduler/objects/queue.go
craigcondit: And as above, this is unnecessary, and these can be combined.

craigcondit (changes requested):
I think we're pretty close. All I see is the merge of some of the inline functions. There's no thread-safety issues with doing so, and it makes the code clearer.

Provided that is completed, I think we're ready for a merge.

pkg/scheduler/objects/queue.go
lixmgl: @craigcondit Thanks for the info, it make sense.
I updated by merging them into one method. Please review.

pkg/scheduler/objects/queue.go
lixmgl: Updated.

pkg/scheduler/objects/queue.go
lixmgl: @craigcondit Actually I removed canRun() method, this file is outdated. 
Current canRun logic is inside this method: https://github.com/apache/yunikorn-core/pull/429/commits/030c1b44a69ec8302fc566f57af7489b26d5cc0e#diff-27632d48eb925e150a33bc92370ceaa66c31048018d11ca7a53a0b50ab7250acR1345

Please review and lmk if you have any comments.

lixmgl:
> 

Thanks @craigcondit for reviewing the PR.
I updated your comments about combining the methods. 
_canRun_ code is outdate, please review _incAllocatingAcceptedAppsIfCanRun_ method

craigcondit (changes requested):
Minor nit on log statement and one more function to combine, otherwise looks ok. @wilfred-s, @pbacsko can you take one last look at this as well?

pkg/scheduler/objects/queue.go
```diff
@@ -1292,3 +1306,71 @@ func (sq *Queue) String() string {
...
+		return false
+	}
+	success := true
+	if sq.parent != nil {
+		success = sq.parent.incAllocatingAcceptedAppsIfCanRun(sa)
+	}
+	return sq.InternalIncAllocatingAcceptedAppsIfCanRun(success, sa)
+}
+
+func (sq *Queue) InternalIncAllocatingAcceptedAppsIfCanRun(success bool, sa *Application) bool {
```
craigcondit, line 1345: This can be combined into incAllocatingAppsIfCanRun() as well, since it is a tail call. I think that improves the visibility.

pkg/scheduler/objects/queue.go
```diff
@@ -1164,13 +1175,16 @@ func (sq *Queue) TryReservedAllocate(iterator func() NodeIterator) *Allocation {
...
-				if alloc != nil {
-					log.Logger().Debug("reservation found for allocation found on queue",
-						zap.String("queueName", sq.QueuePath),
-						zap.String("appID", appID),
-						zap.String("allocation", alloc.String()))
-					return alloc
+				if app.IsStarting() || app.IsRunning() || app.IsAccepted() {
+					alloc := app.tryReservedAllocate(headRoom, iterator)
+					if alloc != nil {
+						log.Logger().Debug("Rainie test accepted apps: reservation found for allocation found on queue",
```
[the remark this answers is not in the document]
lixmgl: good catch, removed :)

pkg/scheduler/objects/queue.go
lixmgl: Sure, combined.

lixmgl:
> 

Thank you @craigcondit. I addressed comments.
@wilfred-s @pbacsko please take a look when you have time, thanks!

craigcondit (approved):
+1 looks good to me.

wilfred-s:
@lixmgl / @pbacsko / @manirajv06 / @craigcondit 
Please check PR #455. I ported the changes to master and moved the logic into the queue code.
I added unit tests and one full cycle test into the partition.
I think I have covered all the cases.

lixmgl:
> @lixmgl / @pbacsko / @manirajv06 / @craigcondit Please check PR #455. I ported the changes to master and moved the logic into the queue code. I added unit tests and one full cycle test into the partition. I think I have covered all the cases.

PR https://github.com/apache/yunikorn-core/pull/455 LGTM.
Thanks for the refactoring and full cycle tests @wilfred-s 
Happy late Thanksgiving :)
Can you merge my PR first then add your PR?

wilfred-s:
> Can we merge this PR to master first then add your PR?

This PR formed the basis under my port. We can merge the ported PR in one go. @craigcondit will handle the proper attribution to you when we commit the change. I just did some cleanup and the port :-)

lixmgl:
> > Can we merge this PR to master first then add your PR?
> 
> This PR formed the basis under my port. We can merge the ported PR in one go. @craigcondit will handle the proper attribution to you when we commit the change. I just did some cleanup and the port :-)

I see, sounds good. Thanks @wilfred-s and @craigcondit !

craigcondit:
Closed in favor of #455.

PR #430, opened 2022-08-18
[YUNIKORN-1282] Resource metrics are not tracked correctly for all queue types

ycr-oss:
### What is this PR for?
* Track resource usage in metrics for non-leaf queues
* Update resource metrics during config and node updates

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1282

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -266,17 +266,17 @@ func (sq *Queue) setResources(resource configs.Resources) error {
 
 	if resources.StrictlyGreaterThanZero(maxResource) {
 		sq.maxResource = maxResource
+		sq.updateMaxResourceMetrics()
 	} else {
 		log.Logger().Debug("max resources setting ignored: cannot set zero max resources")
 	}
-	sq.updateMaxResourceMetrics()
 
 	if resources.StrictlyGreaterThanZero(guaranteedResource) {
 		sq.guaranteedResource = guaranteedResource
+		sq.updateGuaranteedResourceMetrics()
 	} else {
 		log.Logger().Debug("guaranteed resources setting ignored: cannot set zero max resources")
 	}
-	sq.updateGuaranteedResourceMetrics()
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -269,12 +269,14 @@ func (sq *Queue) setResources(resource configs.Resources) error {
 	} else {
 		log.Logger().Debug("max resources setting ignored: cannot set zero max resources")
 	}
+	sq.updateMaxResourceMetrics()
```
wilfred-s, line 272: I think we need this inside the `if` clause with the check for `StrictlyGreaterThanZero`. If there is no max set the object is a non nil pointer with no content.  We do not allow 0 values without at least 1 value larger than 0. That means this call would not result in any metrics being generated. The for loop inside the  `updateMaxResourceMetrics` never executes. It thus is a call to nowhere.

Same for guaranteed resources.

pkg/scheduler/objects/queue.go
```diff
@@ -1017,6 +1019,8 @@ func (sq *Queue) SetMaxResource(max *resources.Resource) {
 		zap.String("current max", sq.maxResource.String()),
 		zap.String("new max", max.String()))
 	sq.maxResource = max.Clone()
+	sq.updateGuaranteedResourceMetrics()
```
wilfred-s, line 1022: We only set max so we should not set the guaranteed metric.

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
ycr-oss: Good catch. Removed

pkg/scheduler/objects/queue.go
ycr-oss: I agree. Updated

wilfred-s (approved):
LGTM

PR #439, opened 2022-09-23
[YUNIKORN-1326] Develop tracker module.

manirajv06:
### What is this PR for?
Initial implementation of the tracker module.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1326

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/group_tracker.go
+++ b/pkg/scheduler/ugm/group_tracker.go
@@ -22,18 +22,19 @@ import (
 	"sync"
 
 	"github.com/apache/yunikorn-core/pkg/common/resources"
+	"github.com/apache/yunikorn-core/pkg/webservice/dao"
 )
 
 type GroupTracker struct {
-	groupName    string
-	applications map[string]bool
-	queueTracker *QueueTracker
+	groupName    string          // Name of the group for which usage is being tracked upon
+	applications map[string]bool // Hold applications currently run by all users belong to this group
+	queueTracker *QueueTracker   // Holds the actual resource usage of queue path where application run
 
 	sync.RWMutex
 }
 
 func NewGroupTracker(group string) *GroupTracker {
-	queueTracker := NewQueueTracker("root")
+	queueTracker := NewQueueTracker()
 	groupTracker := &GroupTracker{
 		groupName:    group,
 		applications: make(map[string]bool),
@@ -64,23 +65,32 @@ func (gt *GroupTracker) getTrackedApplications() map[string]bool {
 	return gt.applications
 }
 
+func (gt *GroupTracker) getGroupResourceUsageDAOInfo(queueTracker *QueueTracker) *dao.GroupResourceUsageDAOInfo {
+	groupResourceUsage := &dao.GroupResourceUsageDAOInfo{
+		Applications: []string{},
+	}
+	groupResourceUsage.GroupName = gt.groupName
+	for app, active := range gt.applications {
+		if active {
+			groupResourceUsage.Applications = append(groupResourceUsage.Applications, app)
+		}
+	}
+	groupResourceUsage.Queues = gt.queueTracker.getResourceUsageDAOInfo("root", "root", gt.queueTracker)
+	return groupResourceUsage
+}
+
 // GetResource only for tests
 func (gt *GroupTracker) GetResource() map[string]*resources.Resource {
 	resources := make(map[string]*resources.Resource)
-	return gt.internalGetResource("root", "root", gt.queueTracker, resources)
+	usage := gt.getGroupResourceUsageDAOInfo(gt.queueTracker)
+	return gt.internalGetResource(usage.Queues, resources)
 }
 
-func (gt *GroupTracker) internalGetResource(parentQueuePath string, queueName string, queueTracker *QueueTracker, resources map[string]*resources.Resource) map[string]*resources.Resource {
-	fullQueuePath := ""
-	if queueName == "root" {
-		fullQueuePath = parentQueuePath
-	} else {
-		fullQueuePath = parentQueuePath + "." + queueTracker.queueName
-	}
-	resources[fullQueuePath] = queueTracker.resourceUsage
-	if len(queueTracker.childQueueTrackers) > 0 {
-		for childQueueName, childQueueTracker := range queueTracker.childQueueTrackers {
-			gt.internalGetResource(fullQueuePath, childQueueName, childQueueTracker, resources)
+func (gt *GroupTracker) internalGetResource(usage *dao.ResourceUsageDAOInfo, resources map[string]*resources.Resource) map[string]*resources.Resource {
+	resources[usage.QueueName] = usage.ResourceUsage
+	if len(usage.Children) > 0 {
+		for _, resourceUsage := range usage.Children {
+			gt.internalGetResource(resourceUsage, resources)
 		}
 	}
 	return resources
--- a/pkg/scheduler/ugm/group_tracker.go
+++ b/pkg/scheduler/ugm/group_tracker.go
@@ -27,7 +27,7 @@ import (
 type GroupTracker struct {
 	groupName    string
 	applications map[string]bool
-	queue        *QueueTracker
+	queueTracker *QueueTracker
 
 	sync.RWMutex
 }
@@ -37,7 +37,7 @@ func NewGroupTracker(group string) *GroupTracker {
 	groupTracker := &GroupTracker{
 		groupName:    group,
 		applications: make(map[string]bool),
-		queue:        queueTracker,
+		queueTracker: queueTracker,
 	}
 	return groupTracker
 }
@@ -46,7 +46,7 @@ func (gt *GroupTracker) increaseTrackedResource(queuePath, applicationID string,
 	gt.Lock()
 	defer gt.Unlock()
 	gt.applications[applicationID] = true
-	return gt.queue.increaseTrackedResource(queuePath, applicationID, usage)
+	return gt.queueTracker.increaseTrackedResource(queuePath, applicationID, usage)
 }
 
 func (gt *GroupTracker) decreaseTrackedResource(queuePath, applicationID string, usage *resources.Resource, removeApp bool) error {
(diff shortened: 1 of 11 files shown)


pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+	"github.com/apache/yunikorn-core/pkg/common/security"
+	"github.com/apache/yunikorn-core/pkg/log"
+)
+
+type Manager struct {
+	users  map[string]*UserTracker
+	groups map[string]*GroupTracker
+	sync.RWMutex
```
pbacsko, line 35: You only use `Lock()`, but not `RLock()`, so it's simpler to just use `Mutex`.

pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,176 @@
...
+	"go.uber.org/zap"
+
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+	"github.com/apache/yunikorn-core/pkg/common/security"
+	"github.com/apache/yunikorn-core/pkg/log"
+)
+
+type Manager struct {
+	users  map[string]*UserTracker
+	groups map[string]*GroupTracker
```
pbacsko, line 34: Since this is a mapping, I'd be happier with names like `userTrackers` and `groupTrackers`.

pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	} else {
+		log.Logger().Info("group tracker must be available in groups map",
+			zap.String("group", m.getGroup(user)))
+		return fmt.Errorf("group tracker for %s is missing in groups map", m.getGroup(user))
+	}
+	return nil
+}
+
+func (m *Manager) checkAppGroup(applicationID string, user *security.UserGroup) {
```
pbacsko, line 117: To me, it looks like this method makes sure that each group in the manager will have a `GroupTracker`. How about calling it `ensureGroupTrackerForApp()` or something similar?

pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,176 @@
...
+
+// getGroup Based on the current limitations, group name and username is same. hence, using username as group name.
+// It would be changed in future based on user group resolution, limit configuration processing etc
+func (m *Manager) getGroup(user *security.UserGroup) string {
+	return user.User
+}
+
+// cleaner Auto wakeup go routine to remove the user and group based on applications being tracked upon, its root queue usage etc
+// todo
+func (m *Manager) cleaner() {
```
pbacsko, line 140: Seems to be unused, will there be a code in the future which calls this?

pkg/scheduler/ugm/group_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"sync"
+
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+)
+
+type GroupTracker struct {
+	groupName    string
+	applications map[string]bool
+	queue        *QueueTracker
```
pbacsko, line 30: Nit: `queueTracker` seems to be a better name

pkg/scheduler/ugm/group_tracker.go
```diff
@@ -0,0 +1,87 @@
...
+	gt.applications[applicationID] = true
+	return gt.queue.increaseTrackedResource(queuePath, applicationID, usage)
+}
+
+func (gt *GroupTracker) decreaseTrackedResource(queuePath, applicationID string, usage *resources.Resource, removeApp bool) error {
+	if removeApp {
+		gt.Lock()
+		defer gt.Unlock()
+		delete(gt.applications, applicationID)
+	}
```
pbacsko, line 57: In this code, we only do `gt.Lock()`, if `removeApp` is true. This is not the case with the increase pair, where we call `gt.queue.increaseTrackedResource()` within a lock. Could this be an issue? At least for consistency reasons, I think the behavior should be the same.

pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			return err
+		}
+		if removeApp {
+			if m.isUserRemovable(userTracker) {
+				delete(m.users, user.User)
+			}
+		}
+	} else {
+		log.Logger().Info("user tracker must be available in users map",
```
pbacsko, line 93: Not `Logger().Error()`?

pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			return err
+		}
+		if removeApp {
+			if m.isGroupRemovable(groupTracker) {
+				delete(m.groups, m.getGroup(user))
+			}
+		}
+	} else {
+		log.Logger().Info("group tracker must be available in groups map",
```
pbacsko, line 110: Not Logger().Error()?

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/apache/yunikorn-core/pkg/common/configs"
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+	"github.com/apache/yunikorn-core/pkg/log"
+)
+
+type QueueTracker struct {
+	queueName           string
+	resourceUsage       *resources.Resource
+	runningApplications map[string]bool
+	childQueues         map[string]*QueueTracker
```
pbacsko, line 36: `childQueueTrackers` ?

pkg/scheduler/ugm/user_tracker.go
```diff
@@ -0,0 +1,99 @@
...
+	"sync"
+
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+	"github.com/apache/yunikorn-core/pkg/common/security"
+)
+
+type UserTracker struct {
+	userName string
+	group    map[string]*GroupTracker
+	queue    *QueueTracker
```
pbacsko, line 31: `queueTracker` ?

wilfred-s (changes requested):
The objects and implemented functions do not line up with the design that was made. Moving away from the designed interface needs discussion before we do this.
The changes also mean that the behaviour as described in the design is not implemented.

manirajv06:
@wilfred-s and myself reviewed the changes together in a call. Made changes based on the comments.

pkg/scheduler/ugm/group_tracker.go
```diff
@@ -0,0 +1,97 @@
...
+	return gt.applications
+}
+
+func (gt *GroupTracker) getGroupResourceUsageDAOInfo(queueTracker *QueueTracker) *dao.GroupResourceUsageDAOInfo {
+	groupResourceUsage := &dao.GroupResourceUsageDAOInfo{
+		Applications: []string{},
+	}
+	groupResourceUsage.GroupName = gt.groupName
+	for app, active := range gt.applications {
+		if active {
```
wilfred-s, line 74: active is always true so this check is redundant.

pkg/scheduler/ugm/group_tracker.go
```diff
@@ -0,0 +1,97 @@
...
+
+// GetResource only for tests
+func (gt *GroupTracker) GetResource() map[string]*resources.Resource {
+	resources := make(map[string]*resources.Resource)
+	usage := gt.getGroupResourceUsageDAOInfo(gt.queueTracker)
+	return gt.internalGetResource(usage.Queues, resources)
+}
+
+func (gt *GroupTracker) internalGetResource(usage *dao.ResourceUsageDAOInfo, resources map[string]*resources.Resource) map[string]*resources.Resource {
+	resources[usage.QueueName] = usage.ResourceUsage
```
wilfred-s, line 90: this breaks when you have two queue with the same name at different levels or under different parents

pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,273 @@
...
+	sync.RWMutex
+}
+
+func Init() {
+	once.Do(func() {
+		m = NewManager()
+	})
+}
+
+func NewManager() *Manager {
```
wilfred-s, line 49: I can call this and re-init the manager which can cause issues. This needs to be shielded by the onceDo()

pkg/scheduler/ugm/user_tracker.go
```diff
@@ -0,0 +1,114 @@
...
+	// Group is not fixed for user unlike other systems and would be selected based on queue limit config processing flow and may vary for different applications.
+	// Hence, group tracker object may vary for same user running different applications linked through this map with key as application id
+	// and group tracker object as value.
+	appGroupTrackers map[string]*GroupTracker
+	queueTracker     *QueueTracker // Holds the actual resource usage of queue path where application runs
+
+	sync.RWMutex
+}
+
+func NewUserTracker(user security.UserGroup) *UserTracker {
```
wilfred-s, line 42: should not export this

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+type QueueTracker struct {
+	queueName           string
+	resourceUsage       *resources.Resource
+	runningApplications map[string]bool
+	childQueueTrackers  map[string]*QueueTracker
+}
+
+func NewQueueTracker() *QueueTracker {
```
wilfred-s, line 40: should not export this

pkg/scheduler/ugm/group_tracker.go
```diff
@@ -0,0 +1,97 @@
...
+
+type GroupTracker struct {
+	groupName    string          // Name of the group for which usage is being tracked upon
+	applications map[string]bool // Hold applications currently run by all users belong to this group
+	queueTracker *QueueTracker   // Holds the actual resource usage of queue path where application run
+
+	sync.RWMutex
+}
+
+func NewGroupTracker(group string) *GroupTracker {
```
wilfred-s, line 36: should not export this

wilfred-s (changes requested):
We need to make sure that the coverage for the new code is above 80% at least.

wilfred-s (approved):
LGTM, change looks good
we'll get he base in and add further tests etc in follow up changes.

PR #441, opened 2022-10-06
[YUNIKORN-1328] Handle application state changes and trigger tracker interfaces

manirajv06:
### What is this PR for?

Integrated ugm module with application state transition change blocks and other places.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1328

### Questions:


manirajv06:
**Notes:**

Why increase resource usage in app#AddAllocationInternal instead of enter_starting app transition code block?

1. In case of a gang app, app state would be changed to "starting" only after all ph's have been allocated. But we need to update the user's resource usage during this ph's allocation stage itself.
2. In case of a non - gang app, after 1st allocation (starting->running), the app would be in running state for future allocations (from 2nd allocation onwards) too. So these future allocations need to be accounted for in the user's resource usage.

Hence, app#AddAllocationInternal is the right place to increase the resource usage. However, With this change, queue metrics and scheduler metrics to increase app running would be moved from enter_running app transition code block to enter_starting app transition code block.

Similarly, app#removeAllocationInternal would be used to decrease the resource usage but not remove the app by calling the decrease method with removeApp as false.

Finally, when the app enters “completing” state, the app would be removed by calling the decrease method with removeApp as true. With this change, even queue metrics and scheduler metrics to do decrease operations would be taken care too.

Even in case of "resuming" (gang app start to work as non gang app because all ph timeouts and gang scheduling fallback style is soft) state, app would be removed as anyway it would be added again when it start to function as non gang app). This activity happens in enter_resuming state transition code block. In case "hard" gang scheduling style, app would be moved to failed state when all ph's timeout happens, hence app would be removed by calling the decrease method with removeApp as true. This activity happens in enter_failing state transition code block. With this change, even queue metrics and scheduler metrics to do decrease operations would be taken care too.


**Stack traces of app#AddAllocationInternal**

Node Registration

p#addAllocation -> app#AddAllocation -> app#AddAllocationInternal

Alloc releases

context#processAllocationReleases -> p#removeAllocation -> app#ReplaceAllocation -> app#AddAllocationInternal

Node Removal

p#removeNodeAllocations -> app#ReplaceAllocation -> app#AddAllocationInternal

Regular scheduling core path

app#tryAllocate or app#tryNodes or app#tryNodesNoReserve or app#tryReservedAllocate -> app#AddAllocationInternal


**Stack traces of app#removeAllocationInternal**

Normal Alloc releases

context#processAllocationReleases -> p#removeAllocation -> app#RemoveAllocation -> app#removeAllocationInternal

Normal Alloc removal during Node Removal 

p#removeNodeAllocations -> app#RemoveAllocation -> app#removeAllocationInternal

ph Alloc releases

context#processAllocationReleases -> p#removeAllocation -> app#ReplaceAllocation -> 
app#removeAllocationInternal

Ph Alloc removal during Node Removal

p#removeNodeAllocations -> app#ReplaceAllocation -> 
app#removeAllocationInternal


**User resource usage enforcement and update**

User quota would be enforced in app#tryAllocate just before the queue headroom check to ensure user quota is not exceeded.

}

After these above check, app#tryAllocate calls app#tryNode to Node pre allocate checks, adding allocation into node, incrementing queue allocated resource for resource updates etc and finally calls app#addAllocationInternal. As described above, app#addAllocationInternal does the actual user’s resource usage updates.

pkg/scheduler/ugm/manager.go
```diff
@@ -0,0 +1,288 @@
...
+			groupTracker = m.groupTrackers[group]
+		}
+		userTracker.setGroupForApp(applicationID, groupTracker)
+	}
+	return nil
+}
+
+// getGroup Based on the current limitations, username and group name is same. Groups[0] is always set and same as username.
+// It would be changed in future based on user group resolution, limit configuration processing etc
+func (m *Manager) getGroup(user security.UserGroup) (string, error) {
```
pbacsko, line 243: Question: what if, for whatever reason, we don't have groups for a certain user? Can that happen? Eg. we disable the user-group retrieval in the AC, what happens here?

wilfred-s (changes requested):
Based on the discussion: we need to be careful when we remove a placeholder as it could be a timed out placeholder on an app that already has other allocations. So going to zero for placeholders can never (?) trigger an app usage tracking removal. The other thing that we need to check is the size difference between placeholder and allocation. If the placeholder is larger than the real one do we account correctly?

wilfred-s:
We seem to have a deadlock in the code. One of the tests times out which normally means we have locked ourselves out.
Can you please check?

craigcondit:
@manirajv06 this PR seems to have quite a bit of unnecessary bits from other JIRAs included. Can you clean this up and rebase / squash?

manirajv06:
@craigcondit @wilfred-s Cleaned up unwanted old commits (included as part of dependent jira rebase).

pkg/scheduler/partition_test.go
```diff
@@ -2007,21 +2100,28 @@ func TestPlaceholderBiggerThanReal(t *testing.T) {
...
+}
+
+func assertUserGroupResource(t *testing.T, expected *resources.Resource) {
+	ugm := ugm.GetUserManager()
+	userResource, err := ugm.GetUserResources(security.UserGroup{User: "testuser", Groups: []string{"testgroup"}})
+	assert.NilError(t, err, "User should have been tracked by this time and available in user trackers map")
+	groupResource, err := ugm.GetGroupResources("testgroup")
+	assert.NilError(t, err, "Group should have been tracked by this time and available in group trackers map")
+	assert.Equal(t, userResource.String(), expected.String())
+	assert.Equal(t, groupResource.String(), expected.String())
```
wilfred-s, line 2113: Same as for the other instance of this check we need to make it user independent, and move it to utilities_test.go

pkg/webservice/handlers_test.go
```diff
@@ -468,7 +468,7 @@ func TestGetClusterUtilJSON(t *testing.T) {
 	assert.Equal(t, partitionName, partition.Name)
 	// new app to partition
 	appID := "appID-1"
-	app := newApplication(appID, partitionName, queueName, rmID, security.UserGroup{})
+	app := newApplication(appID, partitionName, queueName, rmID, security.UserGroup{User: "testuser", Groups: []string{"testgroup"}})
```
wilfred-s, line 471: `newApplication` is a wrapper defined in this file. A simpler change would be to check if and empty `security.UserGroup{}` is passed in in that implementation and then override with a test user.

pkg/scheduler/ugm/manager.go
```diff
@@ -193,20 +193,26 @@ func (m *Manager) DecreaseTrackedResource(queuePath string, applicationID string
 	return nil
 }
 
-func (m *Manager) GetUserResources(user security.UserGroup) *resources.Resource {
-	return nil
+func (m *Manager) GetUserResources(user security.UserGroup) (*resources.Resource, error) {
+	if m.userTrackers[user.User] != nil {
```
wilfred-s, line 197: map are not concurrent safe access needs to be under lock

pkg/scheduler/ugm/manager.go
```diff
@@ -193,20 +193,26 @@ func (m *Manager) DecreaseTrackedResource(queuePath string, applicationID string
...
+	if m.groupTrackers[group] != nil {
+		return m.groupTrackers[group].queueTracker.resourceUsage, nil
+	}
+	return nil, fmt.Errorf("group %s is not available in group trackers map", group)
 }
 
-func (m *Manager) GetUsersResources() []*UserTracker {
-	return nil
+func (m *Manager) GetUsersResources() map[string]*UserTracker {
+	return m.userTrackers
```
wilfred-s, line 211: This could cause race conditions. The caller gets that map and can manipulate the map.
maps are not concurrency safe and need accessing under lock. We should pass back a copy of the content.
That is why the `[]*userTracker` was chosen

pkg/scheduler/ugm/manager.go
```diff
@@ -279,10 +285,18 @@ func (m *Manager) isGroupRemovable(gt *GroupTracker) bool {
...
-func (m *Manager) getGroupTrackers() map[string]*GroupTracker {
+func (m *Manager) GetGroupTrackers() map[string]*GroupTracker {
 	return m.groupTrackers
 }
+
+func (m *Manager) ClearUserTrackers() {
+	m.userTrackers = make(map[string]*UserTracker)
+}
+
+func (m *Manager) ClearGroupTrackers() {
```
wilfred-s, line 300: race condition needs to happen under lock

wilfred-s (changes requested):

manirajv06:
@wilfred-s and I had one more round of review mostly around simplifying test related helper methods, clean up etc. Incorporated the changes.

PR #446, opened 2022-11-08
[YUNIKORN-1366] Read core configuration from shim

craigcondit:
### What is this PR for?
Reads the core scheduler configuration from the shim instead of the filesystem. Removes config watching and update logic as this is now the responsibility of the shim. Allows log configuration to be updated dynamically.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1366

Updated unit tests.

### Questions:


pkg/entrypoint/entrypoint.go
```diff
@@ -55,6 +55,10 @@ func StartAllServicesWithLogger(logger *zap.Logger, zapConfigs *zap.Config) *Ser
 	return StartAllServices()
 }
 
+func InitializeLogger(logger *zap.Logger, zapConfigs *zap.Config) {
+	log.InitializeLogger(logger, zapConfigs)
+}
+
```
wilfred-s, line 61: This is never called, the `StartAllServicesWithLogger()` is called from the k8shim entrypoint we never just want to call this as part of the entry

pkg/log/logger.go
```diff
@@ -23,50 +23,63 @@ import (
...
+	"sync/atomic"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
 var once sync.Once
-var logger *zap.Logger
-var config *zap.Config
+var holderValue atomic.Value
```
wilfred-s, line 33: We do not need an atomic here. The logger variable is protected by the `sync.Once`. The logger internally protects itself. We cannot build a logger without a config. Neither must change after creation.
On read we just need to return the global. There is nothing that warrants the overhead of the load and the cast of the atomic value.

pkg/log/logger.go
```diff
@@ -23,50 +23,63 @@ import (
...
-
-	return logger
 }
 
 func InitializeLogger(log *zap.Logger, zapConfig *zap.Config) {
-	once.Do(func() {
-		logger = log
-		config = zapConfig
-		logger.Info("Using an already initialized logger")
+	once.Do(initLogger)
```
wilfred-s, line 74: This changes the init for the logger from the shim into the core. The `InitializeLogger()` passes in a pre-build config from the k8shim and makes sure that the shim and the core run with *the exact same* config object. Without that changing the log level via the rest call is only going to change the core log level.

It also does the work twice in this case: we have a logger and config we just want to set it. We can change that in a later jira to decouple things but not as part of this change.

pkg/log/logger.go
```diff
@@ -116,14 +126,17 @@ func createConfig() *zap.Config {
...
+
 	oldLevel := config.Level.String()
 
 	// noop if the input is the same as what is set
 	if newLevel == oldLevel {
 		return nil
 	}
 
-	logger.Info("Updating log level",
+	Logger().Info("Updating log level",
```
wilfred-s, line 139: consistency: either use  `Logger()` or use the local var logger throughout the function:  line 129, 139 & 147

pkg/log/logger_test.go
```diff
@@ -62,8 +63,9 @@ func TestIsDebugEnabled(t *testing.T) {
 		Encoding: "console",
 	}
 	var err error
-	logger, err = zapConfig.Build()
+	logger, err := zapConfig.Build()
 	assert.NilError(t, err, "debug level logger create failed")
+	InitializeLogger(logger, &zapConfig)
```
wilfred-s, line 68: These tests explicitly did not trigger the `once.Do()` to make sure we re-try things.

pkg/log/logger_test.go
```diff
@@ -119,9 +121,9 @@ func TestInitializeLogger(t *testing.T) {
 
 	InitializeLogger(localLogger, &zapConfig)
 	assert.Equal(t, Logger(), localLogger)
-	// second initialization should not do anything
+	// second initialization should update
 	InitializeLogger(localLogger2, &zapConfig)
-	assert.Equal(t, Logger(), localLogger)
+	assert.Equal(t, Logger(), localLogger2)
```
wilfred-s, line 126: See comment earlier: this should not change anything as it can break the link between the k8shim and the core logger which breaks the ability to update the k8shim log level.

pkg/scheduler/context.go
```diff
@@ -313,14 +333,14 @@ func (cc *ClusterContext) UpdateSchedulerConfig(conf *configs.SchedulerConfig) e
 // Locked version of the configuration update called outside of event system.
 // Updates the current config via the config loader.
 // Used in test only, normal updates use the internal call and the webservice must use the UpdateSchedulerConfig
```
wilfred-s, line 335: NIT: Can we update this comment top remove the web service reference?

wilfred-s (changes requested):

pkg/entrypoint/entrypoint.go
craigcondit: It gets called in the updated shim code when the configuration changes, so it's needed here.

pkg/log/logger.go
craigcondit: This is here to allow dynamic reconfiguration of the logger. The overhead is minimal (probably a few machine ops per logged statement).

pkg/log/logger.go
craigcondit: initLogger is used now only to provide a "fallback" logger in case something is logged early. Once the scheduler is initialized propery, the dynamic configuration will replace the logger with the real, configured one. This is all part of allowing dynamic reconfiguration via the configmap.

pkg/log/logger.go
craigcondit: good catch, this should be logger from Line 129.

pkg/log/logger_test.go
craigcondit: re-init is no longer a problem as we handle this case cleanly.

pkg/log/logger_test.go
craigcondit: As above, InitializeLogger() is how the shim will reconfigure the logger when the configmap changes. I've tested this locally in my shim, and it works as expected.

pkg/entrypoint/entrypoint.go
craigcondit: Per offline discussions, reverting all logging changes.

pkg/log/logger.go
craigcondit: Per offline discussions, reverting all logging changes.

pkg/log/logger_test.go
craigcondit: Per offline discussions, reverting all logging changes.

pkg/log/logger.go
craigcondit: Per offline discussions, reverting all logging changes.

pkg/log/logger_test.go
craigcondit: Removed logging changes.

pbacsko (commented):
Have you looked at codecov complaints? Anything to do to make it green?

craigcondit:
> Have you looked at codecov complaints? Anything to do to make it green?

That's due to all the code which is being deleted. The number of lines is actually increasing, but the overall delta is slightly negative. All the new code is being tested.

pkg/common/configs/configs.go
```diff
@@ -53,3 +52,51 @@ func (ctx *SchedulerConfigContext) Get(policyGroup string) *SchedulerConfig {
...
+	configMapLock.Lock()
+	defer configMapLock.Unlock()
+	configMapCallbacks[id] = callback
+}
+
+func RemoveConfigMapCallback(id string) {
+	configMapLock.Lock()
+	defer configMapLock.Unlock()
+	delete(configMapCallbacks, id)
+}
```
pbacsko, line 66: 2 things:
* Comments are missing
* When are these functions called? I can see only one reference from the unit test.

pbacsko (commented):
I have one question regarding callbacks.

pkg/common/configs/configs.go
craigcondit: I'll add the comments.

Currently, nothing calls these, as we don't have any config vars specific to the core. See comments on https://issues.apache.org/jira/browse/YUNIKORN-1213 for the first likely candidate (controlling health check interval).

The idea is that code which can respond to config changes will get a callback when new values are loaded, and can update internal state.

wilfred-s (changes requested):
One last change to make sure the default config works

craigcondit:
> One last change to make sure the default config works

@wilfred-s I don't see a specific code comment here. What do we need to change?

pkg/common/configs/config.go
```diff
@@ -227,5 +193,14 @@ func GetConfigurationString(requestBytes []byte) string {
...
+// DefaultSchedulerConfig contains the default scheduler configuration; used if no other is provided
+var DefaultSchedulerConfig = `
+partitions:
+  - name: default
+    placementrules:
+      - name: tag
+        value: namespace
+        create: true
+    queues:
+      - name: root
```
wilfred-s, line 205: We are missing the following line from the code:
```
          submitacl: '*'
```

craigcondit: Updated.

wilfred-s (approved):
LGTM
filed new jiras for missing unit test coverage (not new as part of this change)

PR #447, opened 2022-11-10
[YUNIKORN-1395] Account for preemptions in the placeholder data

wilfred-s:
### What is this PR for?
If we preempt a placeholder to make room for a daemon set pod we do not update the counter inside the application placeholder data object. This causes the scheduler to think it needs to replace a placeholder but never finds one to replace causing the real allocation to never succeed

Same fix for placeholders that get removed as part of a node removal.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1395

manual testing, unit tests will follow


craigcondit (approved):
+1 LGTM.

PR #450, opened 2022-11-16
[YUNIKORN-1402] add unit test for nil safety of NewResourceFromProto

0yukali0:
### What is this PR for?
UT of NewResourceFromProto does not include the nil input
res1 := NewResourceFromMap(nil)
res2 := NewResourceFromProto(res1.ToProto())
res1.Resources is not equally to res2.Resources because the front one has address and the last one does not has address.

So, I add the nil case to UT in this commit and NewResourceFromMap would return a default empty resource when the input is nil.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring
* [x] UT

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1402

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -63,6 +63,9 @@ func NewResourceFromProto(proto *si.Resource) *Resource {
 }
 
 func NewResourceFromMap(m map[string]Quantity) *Resource {
+    if m == nil {
+        return NewResource()
+    }
 	return &Resource{Resources: m}
 }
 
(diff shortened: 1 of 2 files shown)


craigcondit (approved):
As far as I can tell, NewResourceFromMap() is only called from test code, and never with a nil value, so we're in effect "testing our tests". That said, I don't see the harm.

+1, will merge shortly.

PR #452, opened 2022-11-17
[YUNIKORN-1398] Remove non daemonset reservation from node before all ocating daemonset pod

manirajv06:
### What is this PR for?
Cancel non daemon set reservations (if any) on specific node to give preference to daemon sets allocation during upscaling

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1398

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -221,6 +221,15 @@ func (m *Manager) GetUsersResources() []*UserTracker {
 	return userTrackers
 }
 
+func (m *Manager) GetUserTracker(user string) *UserTracker {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+	if m.userTrackers[user] != nil {
+		return m.userTrackers[user]
+	}
+	return nil
+}
+
 func (m *Manager) GetGroupsResources() []*GroupTracker {
 	m.lock.RLock()
 	defer m.lock.RUnlock()
@@ -229,6 +238,15 @@ func (m *Manager) GetGroupsResources() []*GroupTracker {
 		groupTrackers = append(groupTrackers, tracker)
 	}
 	return groupTrackers
+}
+
+func (m *Manager) GetGroupTracker(group string) *GroupTracker {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+	if m.groupTrackers[group] != nil {
+		return m.groupTrackers[group]
+	}
+	return nil
 }
 
 func (m *Manager) ensureGroupTrackerForApp(queuePath string, applicationID string, user security.UserGroup) error {
--- a/pkg/scheduler/ugm/manager_test.go
+++ b/pkg/scheduler/ugm/manager_test.go
@@ -73,6 +73,8 @@ func TestAddRemoveUserAndGroups(t *testing.T) {
 	assert.Equal(t, false, manager.isUserRemovable(userTracker))
 	assert.Equal(t, false, manager.isGroupRemovable(groupTracker))
 	assertUGM(t, user, usage1, 1)
+	assert.Equal(t, user.User, manager.GetUserTracker(user.User).userName)
+	assert.Equal(t, user.Groups[0], manager.GetGroupTracker(user.Groups[0]).groupName)
 
 	err = manager.IncreaseTrackedResource(queuePath1, TestApp1, usage1, user)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -886,6 +886,40 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+					reservedKeys, releasedAsks := node.UnReserveApps()
+					if len(reservedKeys) == 0 || len(releasedAsks) == 0 {
+						log.Logger().Warn("Unable to cancel reservations on node",
+							zap.String("application ID", sa.ApplicationID),
+							zap.String("allocationKey", request.GetAllocationKey()),
+							zap.String("required node", requiredNode),
+							zap.Int("reservedKeys count", len(reservedKeys)),
+							zap.Int("releasedAsks count", len(releasedAsks)))
+					} else {
+						log.Logger().Debug("Cancelled reservation on required node",
```
craigcondit, line 913: I think we should log this at Info() as it's rather important.

craigcondit (changes requested):
Minor nit on logging. Otherwise looks ok. @wilfred-s can you review as well? Want to be sure we're not missing anything.

manirajv06:
Code coverage complaint can be ignored as specific code path has been covered as part of unit test cases. Please validate this too as part of  review.

pkg/scheduler/objects/application.go
```diff
@@ -886,6 +886,40 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
 				return nil
 			}
+
+			// Are there any non daemon set reservations on specific required node?
+			// Cancel those reservations to run daemon set pods
+			reservations := node.getReservations()
+			if len(reservations) > 0 {
+				unreserve := true
+				for _, res := range node.getReservations() {
+					// skip the node
```
wilfred-s, line 896: nit: can be fixed on commit, explain why we skip the node. Even I had to think twice why we would skip.

pkg/scheduler/objects/application.go
```diff
@@ -886,6 +886,40 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+					if res.ask.GetRequiredNode() != "" {
+						unreserve = false
+						break
+					}
+				}
+				if unreserve {
+					// un reserve all the apps that were reserved on the node
+					reservedKeys, releasedAsks := node.UnReserveApps()
+					if len(reservedKeys) == 0 || len(releasedAsks) == 0 {
```
wilfred-s, line 905: The comments on the `node.UnReserveApps()` method are wide ranging: we do not lock the node or the app inside this method. It also does a callback into the application which is unlocked. This can cause data races. We *must* not use this method.

The implementation should, if we do not skip the node, iterate over the list of reservations returned and call `app.UnReserve(node, ask)` to make sure we do the correct locking etc.
Logging is simply handed for each returning call.

pkg/scheduler/objects/node_test.go
```diff
@@ -231,12 +231,14 @@ func TestNodeReservation(t *testing.T) {
 	if node.IsReserved() && !node.isReservedForApp(appID1) {
 		t.Errorf("node should have reservations for app-1")
 	}
+	assert.Equal(t, 1, len(node.getReservations()), "node should have 1 reservations")
```
wilfred-s, line 234: nit: move up to line 225 directly after the error check

pkg/scheduler/partition_test.go
```diff
@@ -361,6 +361,7 @@ func TestRemoveNodeWithPlaceholders(t *testing.T) {
 	assert.Equal(t, 0, len(allocs), "expected no allocations for the app")
 	assert.Assert(t, resources.Equals(app.GetPendingResource(), appRes), "app should have updated pending resources")
 	// check the interim state of the placeholder involved
+	// check the interim state of the placeholder involved
```
wilfred-s, line 364: merge error? remove duplicated line

wilfred-s (changes requested):
node.UnReserveApps() method cannot be used for the cleanup

pkg/scheduler/objects/application.go
```diff
@@ -886,6 +886,47 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+							log.Logger().Info("Cancelled reservation on required node",
+								zap.String("affected application ID", res.app.ApplicationID),
+								zap.String("affected allocationKey", res.getKey()),
+								zap.String("required node", requiredNode),
+								zap.Int("reservations count", num),
+								zap.String("application ID", sa.ApplicationID),
+								zap.String("allocation key", request.GetAllocationKey()))
+						}
+						// remove the reservation of the queue
+						res.app.queue.UnReserve(res.app.ApplicationID, num)
```
manirajv06, line 925: @wilfred-s like above, do we also need to take care of reservation count in partition?

manirajv06:
Moved the cleanup code into a new method and handled different versions of same method using app id check for now. Will file a new jira to explore tryLock in Go.

wilfred-s (changes requested):

wilfred-s:
@manirajv06, I think the rebased change did not get rebased correctly. Please check that.
I have also left a review on the change itself.

manirajv06:
Some problem in rebasing, hence created new pr - https://github.com/apache/yunikorn-core/pull/470

PR #455, opened 2022-11-24
[YUNIKORN-790] Implement MaxApplications for queues

wilfred-s:
### What is this PR for?
Applications are considered running when they have allocations assigned. The first state considered is the Starting state. Applications during gang placeholder allocations stay in Accepted state but are counted as part of the running count.
Max applications is configurable in the hierarchy. Configuration checks are added and enforced. Tests are added for the configuration checks.

Logic for the tracking, updating and enforcement is located inside the queue code. Unit tests are added for the logic.
Allocation cycle test is added as part of the partition tests.

### What type of PR is it?
* [X] - Feature

### What is the Jira issue?

Unit tests and one full cycle test is added.

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -55,17 +55,20 @@ type Queue struct {
 	// The queue properties should be treated as immutable the value is a merge of the
 	// parent properties with the config for this queue only manipulated during creation
 	// of the queue or via a queue configuration update.
-	properties         map[string]string
-	adminACL           security.ACL        // admin ACL
-	submitACL          security.ACL        // submit ACL
-	maxResource        *resources.Resource // When not set, max = nil
-	guaranteedResource *resources.Resource // When not set, Guaranteed == 0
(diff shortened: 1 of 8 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -1255,3 +1299,75 @@ func (sq *Queue) String() string {
...
+func (sq *Queue) incRunningApps(appID string) {
+	if sq == nil {
+		return
+	}
+	if sq.parent != nil {
+		sq.parent.incRunningApps(appID)
+	}
+	sq.Lock()
+	defer sq.Unlock()
+	delete(sq.allocatingAcceptedApps, appID)
```
craigcondit, line 1315: We need to be sure that when an application goes to a terminal state that we also remove from sq.allocatingAcceptedApps, otherwise we potentially leak slots if some placeholders exist but we kill the app.

craigcondit (changes requested):
Mostly good, found one subtle bug relating to killing apps that haven't hit RUNNING state yet.

PR #78, opened 2019-12-19
add instructions for debug setup

yangwwei:
this will be useful for developers to debug issues locally

PR #457, opened 2022-11-28
[YUNIKORN-1430] change go version check in Makefile

wusamzong:
### What is this PR for?
Point `MOD_VERSION` to `.go_version`, which will cause the make to fail if go is too old
I am going to update to all repos
[yunikorn-core #457](https://github.com/apache/yunikorn-core/pull/457)
[yunikorn-k8shim #491](https://github.com/apache/yunikorn-k8shim/pull/491)
[yunikorn-scheduler-interface #78](https://github.com/apache/yunikorn-scheduler-interface/pull/78)

### What type of PR is it?
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1430

`make`
Here is the result I use go1.16
image

Here is the result I use go1.18
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,13 +17,13 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 2
-      - name: Check license
-        run: make license-check
-        if: ${{ github.event_name == 'pull_request' }}
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
           go-version-file: .go_version
+      - name: Check license
+        run: make license-check
+        if: ${{ github.event_name == 'pull_request' }}
       - name: Go lint
         run: make lint
         if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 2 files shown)


craigcondit (changes requested):
Need to move the go setup task prior to the license check task.

wusamzong:
> Need to move the go setup task prior to the license check task.

I’m not sure I understand what you mean.
If it should be changed to the following?
```
.PHONY: run
run: license-check build
	@echo "running scheduler locally"
	...
.PHONY: run_plugin
run_plugin: license-check build_plugin
	@echo "running scheduler plugin locally"
	...
```
Put the `license-check` phase before the `build` phase?

wusamzong:
> Need to move the go setup task prior to the license check task.

Do you mean that if an error occurs during the `go build` phase, `Build should be run with at least go $(MOD_VERSION) or later, found $(GO_VERSION)` will be displayed?

craigcondit:
> Do you mean that if an error occurs during the `go build` phase, `Build should be run with at least go $(MOD_VERSION) or later, found $(GO_VERSION)` will be displayed?

No, the build fails because the license check GitHub action comes before the step which installs the correct version of go. You need to update the GitHub workflow file to reverse this order.

wilfred-s (approved):
LGTM

PR #458, opened 2022-11-29
[YUNIKORN-1434] Mark allocation as being preempted

pbacsko:
### What is this PR for?
Avoid multiple preemption for the same Ask, otherwise the output is flooded by preemption attempt messages.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1434

### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation_ask_test.go
+++ b/pkg/scheduler/objects/allocation_ask_test.go
@@ -205,9 +205,3 @@ func sortedLog(ask *AllocationAsk) []*AllocationLogEntry {
 	})
 	return log
 }
-
-func (aa *AllocationAsk) SetTriggeredPreemption(triggered bool) {
-	aa.Lock()
-	defer aa.Unlock()
-	aa.preemptionTriggered = triggered
-}
--- a/pkg/scheduler/partition_test.go
+++ b/pkg/scheduler/partition_test.go
@@ -1364,7 +1364,6 @@ func setupPreemptionForRequiredNode(t *testing.T) (*PartitionContext, *objects.A
 	}
 	releases, _ := partition.removeAllocation(release)
 	assert.Equal(t, 1, len(releases), "unexpected number of allocations released")
-	ask2.SetTriggeredPreemption(false)
 	assertUserGroupResource(t, getTestUserGroup(), resources.NewResourceFromMap(map[string]resources.Quantity{"vcore": 0}))
 	return partition, app
 }


pkg/scheduler/objects/application.go
```diff
@@ -1093,6 +1097,9 @@ func (sa *Application) tryReservedAllocate(headRoom *resources.Resource, nodeIte
 
 		// Do we need a specific node?
 		if ask.GetRequiredNode() != "" {
+			if ask.HasTriggeredPreemption() || time.Since(ask.GetLastPreemptionAttempt()) < sa.preemptionAttemptInterval {
```
pbacsko, line 1100: This check also defends against the scenario when there's no victim and we want to re-attempt preemption (I picked an arbitrary value of 5 seconds). I don't know if this is a legit case or not, if not, then it can be removed.

craigcondit: I'm not sure we want to do this. Once we have triggered preemption, the pods will be scheduled for termination, but may not yet have terminated (either gracefully or not). If graceful termination takes longer than 5 seconds, we're going to attempt to trigger again, which could result in evicting even more pods. In the worst case scenario, we could kill everything else running on the node.

pkg/scheduler/objects/application.go
```diff
@@ -1742,3 +1754,10 @@ func (sa *Application) SetTimedOutPlaceholder(taskGroupName string, timedOut int
 		sa.placeholderData[taskGroupName].TimedOut = timedOut
 	}
 }
+
+// test only
+func (sa *Application) SetPreemptionAttemptInterval(interval time.Duration) {
```
craigcondit, line 1759: Again, I think we need to reconsider this. It's the RM's responsibility to kill tasks when we ask it to, and we need to trust that it will occur. However, expecting that this happens in some predetermined time is asking for trouble. By preempting once (only) and reserving the node, we are reserving space until the requisite tasks are terminated.

The only thing we might want to do is keep track of whether or not all the tasks have been preempted yet for a node. If they have, and we are still unschedulable, that likely indicates that another scheduler got in line ahead of us and we need to try preempting again.

craigcondit (changes requested):
Looks good for the most part, but I'm very concerned about making multiple preemption attempts while tasks are still outstanding. I think that's going to cause catastrophic failures in some cases.

pkg/scheduler/objects/application.go
pbacsko: The is exactly what this change is intended to prevent, ie. once preemption has been triggered for a given ask, it should not be re-tried again - that's why we mark requests. Currently, it's possible that we try to terminate the same set of allocation a dozen times in a second.

pkg/scheduler/objects/application.go
pbacsko: The interval is only considered if there was a preemption attempt for an ask, but no victim allocations were found. In this case, we probably want to do it again, but not immediately, because we'd flood the logs. If the preemption request was successful, then this interval doesn't matter.

pbacsko: Maybe the naming `PreemptionRetryInterval` would be better?

pkg/scheduler/objects/application.go
craigcondit: Ok, I think I understand now.

pkg/scheduler/objects/application.go
craigcondit: Yes, I like the rename idea, I think that's clearer.

craigcondit (changes requested):
Rename PreemptionAttemptInterval to PreemptionRetryInterval.

pkg/scheduler/objects/application.go
wilfred-s: I think @craigcondit was right: once we triggered preemption the kills will happen and cannot be reversed. We do not re-trigger *ever* in any case not even after a short period of time.
The idea is to find a victim list and if we have a list:
-  mark the ask as preemption triggered 
- kill all victims
- mark all victims as preempted

Next scheduling cycle if the ask is still there and has the preemption triggered flag skip it and do not re-trigger. The node is reserved for the ask so it will get scheduled on that node. It could take up to the grace time for the pod to shutdown, which could be loooong in scheduling times.

pkg/scheduler/objects/application.go
wilfred-s: See earlier comment I think this is not needed at all.

pkg/scheduler/objects/application.go
```diff
@@ -1128,6 +1135,7 @@ func (sa *Application) tryReservedAllocate(headRoom *resources.Resource, nodeIte
 }
 
 func (sa *Application) tryPreemption(reserve *reservation, ask *AllocationAsk) bool {
+	ask.SetLastPreemptionAttempt(time.Now())
```
wilfred-s, line 1138: We should not set this unless we have really preempted otherwise it looks like we did but we really did not. See comment above.

wilfred-s (changes requested):
End goal is to not trigger preemption twice in any case. I think we can simplify the change to just do that.

pkg/scheduler/objects/application.go
manirajv06: Yes, ensuring above three steps is good enough. If we anticipate any failures after marking victim as preempted (third step), we could break this into 1. victim_marked_for_preemption 2. victim_preempted (after core receive the terminated/killed confirmation from shim). This breakage would help us to retry only the victim staying in "victim_marked_for_preemption" state for so long.

wilfred-s: The delete of the pod should never fail. If it does there is a problem on the node. A pod that stays in terminating state  for instance will not get removed if we just retry the removal. Sending a second delete will just be ignored by the API server as it is already processing a request like it.

pkg/scheduler/objects/allocation_ask.go
```diff
@@ -245,3 +247,31 @@ func (aa *AllocationAsk) GetAllocationLog() []*AllocationLogEntry {
...
+
+// HasTriggeredPreemption returns whether this ask has triggered preemption
+func (aa *AllocationAsk) HasTriggeredPreemption() bool {
+	aa.RLock()
+	defer aa.RUnlock()
+	return aa.preemptionTriggered
+}
+
+// SetLastPreemptionAttempt sets the time which tells when the last preemption attempt was performed for this ask.
+func (aa *AllocationAsk) SetLastPreemptionAttempt(attempt time.Time) {
```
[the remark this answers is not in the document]
pbacsko: Removed.

pkg/scheduler/objects/application.go
pbacsko: Removed.

pkg/scheduler/objects/allocation_ask.go
```diff
@@ -262,16 +261,9 @@ func (aa *AllocationAsk) HasTriggeredPreemption() bool {
 	return aa.preemptionTriggered
 }
 
-// SetLastPreemptionAttempt sets the time which tells when the last preemption attempt was performed for this ask.
-func (aa *AllocationAsk) SetLastPreemptionAttempt(attempt time.Time) {
+// test only
```
craigcondit, line 264: This is only called from the objects pkg, so can be moved to a *_test.go file. No need to pollute the main package with test code.

pkg/scheduler/objects/application.go
```diff
@@ -1147,13 +1140,13 @@ func (sa *Application) tryPreemption(reserve *reservation, ask *AllocationAsk) b
 	// Are there any victims/asks to preempt?
 	victims := preemptor.GetVictims()
 	if len(victims) > 0 {
+		log.Logger().Info("Found victims for daemon set ask preemption ",
```
craigcondit, line 1143: Given we've renamed this feature to 'required node preemption', can we change this message to reflect that? Maybe "Found victims for required node preemption "

craigcondit (changes requested):
Nearly there, some minor comments.

pkg/scheduler/objects/application.go

pkg/scheduler/objects/allocation_ask.go
pbacsko: Done

craigcondit (approved):
+1 will merge shortly.

PR #459, opened 2022-11-30
[YUNIKORN-1435] Expose RM state dump in REST API

craigcondit:
### What is this PR for?
When an RM has registered as a `StateDumpPlugin`, its output will be included in the full state dump under the `RMDiagnostics` entry.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1435

Unit tests have been added.

### Questions:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ go 1.16
 
 require (
 	github.com/HdrHistogram/hdrhistogram-go v1.0.1 // indirect
-	github.com/apache/yunikorn-scheduler-interface v0.0.0-20221110194141-115505199f57
+	github.com/apache/yunikorn-scheduler-interface v0.0.0-20221130170804-42d2286739d8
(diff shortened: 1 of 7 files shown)


manirajv06 (changes requested):
changes looks good. test cases needs to be added.

craigcondit:
> changes looks good. test cases needs to be added.

Existing test cases already cover this.

manirajv06:
I can see asserts for some properties like Partitions, Nodes, ClusterInfo, Queues etc but for not for Applications, AppHistory, ContainerHistory and recently added RMDiagnostics. I think we should file a jira to cover the left out properties.

PR #463, opened 2022-12-01
[YUNIKORN-1440] Cleanup terminated applications once expired

bgrams:
### What is this PR for?

Allows terminated applications to be cleaned up once marked as expired. The `completedApplications` map otherwise grows indefinitely and may lead to memory pressure.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1440

Unit test added

### Questions:

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -1077,6 +1077,18 @@ func (pc *PartitionContext) GetCompletedApplications() []*objects.Application {
 	return appList
 }
 
+func (pc *PartitionContext) GetCompletedAppsByState(state string) map[string]*objects.Application {
+	pc.RLock()
+	defer pc.RUnlock()
+	appMap := make(map[string]*objects.Application)
+	for k, app := range pc.completedApplications {
+		if app.CurrentState() == state {
+			appMap[k] = app
+		}
+	}
+	return appMap
+}
+
 func (pc *PartitionContext) GetRejectedApplications() []*objects.Application {
 	pc.RLock()
 	defer pc.RUnlock()
@@ -1433,7 +1445,7 @@ func (pc *PartitionContext) cleanupExpiredApps() {
 		delete(pc.rejectedApplications, app.ApplicationID)
 		pc.Unlock()
 	}
-	for k, app := range pc.completedApplications {
+	for k, app := range pc.GetCompletedAppsByState(objects.Expired.String()) {
 		pc.Lock()
 		if app.IsExpired() {
 			delete(pc.completedApplications, k)
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -1445,11 +1445,9 @@ func (pc *PartitionContext) cleanupExpiredApps() {
 		delete(pc.rejectedApplications, app.ApplicationID)
 		pc.Unlock()
 	}
-	for k, app := range pc.GetCompletedAppsByState(objects.Expired.String()) {
+	for k := range pc.GetCompletedAppsByState(objects.Expired.String()) {
 		pc.Lock()
-		if app.IsExpired() {
-			delete(pc.completedApplications, k)
-		}
+		delete(pc.completedApplications, k)
 		pc.Unlock()
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/scheduler/partition.go
```diff
@@ -1433,6 +1433,13 @@ func (pc *PartitionContext) cleanupExpiredApps() {
 		delete(pc.rejectedApplications, app.ApplicationID)
 		pc.Unlock()
 	}
+	for k, app := range pc.completedApplications {
+		pc.Lock()
+		if app.IsExpired() {
+			delete(pc.completedApplications, k)
+		}
+		pc.Unlock()
+	}
```
wilfred-s, line 1442: This will cause a data race. The completedApplications map can be manipulated from a different go routine. Maps are not go routine safe. You will need to introduce a `pc.GetCompletedAppsByState()` method to do the filtering under a lock like we do for rejected apps.
The delete will still need to be within the lock.

wilfred-s (changes requested):
Looking at the TestCleanupCompletedApps it is even more broken than what you have fixed.

The proper test should have two apps in the `completedApplications` map. One is in the Completed state, and the second one is in the Expired state. Running the cleanupExpiredApps() should remove one and not the other.
Both apps should have moved from Completing to Completed using the event. The `partition.applications` list should be 2 before the events and 0 after the events. Vice versa for the `partition.completedApplications` list

pkg/scheduler/partition.go
bgrams: Thanks Wilfred, addressed in [209b0dc](https://github.com/apache/yunikorn-core/pull/463/commits/209b0dc7d4822d3e8cea3d3571eb13121ba39d17)

bgrams:
> Looking at the TestCleanupCompletedApps it is even more broken than what you have fixed.
> 
> The proper test should have two apps in the `completedApplications` map. One is in the Completed state, and the second one is in the Expired state. Running the cleanupExpiredApps() should remove one and not the other. Both apps should have moved from Completing to Completed using the event. The `partition.applications` list should be 2 before the events and 0 after the events. Vice versa for the `partition.completedApplications` list

Addressed in [26e126a](https://github.com/apache/yunikorn-core/pull/463/commits/26e126a649331bac065b146e3bd08c9793e7f8ac)

pkg/scheduler/partition_test.go
```diff
@@ -1681,39 +1698,62 @@ func TestCompleteApp(t *testing.T) {
...
+
 	// mark the app for removal
-	completedApp.SetState(objects.Expired.String())
+	completedApp1.SetState(objects.Expired.String())
 	partition.cleanupExpiredApps()
-	assert.Assert(t, len(partition.applications) == 1, "the partition should have 1 app")
-	assert.Assert(t, partition.getApplication(completedApp.ApplicationID) == nil, "completed application should have been deleted")
-	assert.Assert(t, partition.getApplication(newApp.ApplicationID) != nil, "new application should still be in the partition")
-	assert.Assert(t, len(partition.GetAppsByState(objects.Completed.String())) == 0, "the partition should have 0 completed app")
-	assert.Assert(t, len(partition.GetAppsByState(objects.Expired.String())) == 0, "the partition should have 0 expired app")
```
bgrams, line 1716: Noting that the removal of these is covered by the check in [L.1749](https://github.com/apache/yunikorn-core/pull/463/files#diff-2edfc3fe8b3fca0a3a2a8a3924a0c5a078e3f8e9046748067d65248b458aae6aR1749) as well as the refactor of [`TestCleanupApp`](https://github.com/apache/yunikorn-core/pull/463/files#diff-2edfc3fe8b3fca0a3a2a8a3924a0c5a078e3f8e9046748067d65248b458aae6aR1707) with a similar test structure. I'm open to feedback on this refactor.

wilfred-s (changes requested):
some minor points to clean up

craigcondit (approved):
+1 Will merge shortly.

PR #464, opened 2022-12-01
[YUNIKORN-1441] Data race in TestTimeoutPlaceholderAllocReleased

pbacsko:
### What is this PR for?
A data race can occur in a unit test which uses `Application.addPlaceholderData()`.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1441

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1782,10 +1782,3 @@ func (sa *Application) SetTimedOutPlaceholder(taskGroupName string, timedOut int
 		sa.placeholderData[taskGroupName].TimedOut = timedOut
 	}
 }
-
-// test only
-func (sa *Application) addPlaceholderDataWithLocking(ask *AllocationAsk) {
-	sa.Lock()
-	defer sa.Unlock()
-	sa.addPlaceholderData(ask)
-}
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -1782,3 +1782,10 @@ func (sa *Application) SetTimedOutPlaceholder(taskGroupName string, timedOut int
 		sa.placeholderData[taskGroupName].TimedOut = timedOut
 	}
 }
+
+// test only
+func (sa *Application) addPlaceholderDataWithLocking(ask *AllocationAsk) {
+	sa.Lock()
+	defer sa.Unlock()
+	sa.addPlaceholderData(ask)
+}
```
craigcondit, line 1791: Move this function into application_test.go since it is test-only. It's in the same package so will still compile.

craigcondit (changes requested):
Code location change.

pkg/scheduler/objects/application.go

PR #466, opened 2022-12-05
[YUNIKORN-1445] partition cleanup: refactor code

wilfred-s:
### What is this PR for?
The cleanup code needs a "cleanup"
Refactoring of cleanupExpiredApps to rely on just application ID lists. Make GetAppsByState, GetRejectedAppsByState & GetCompletedAppsByState internal to the partition and change the return value

Remove GetAppsInTerminatedState and GetTotalCompletedApplicationCount as they are unused or only for test and not needed.
Cleanup tests to use assert.Equal instead of asser.Assert

### What type of PR is it?
* [X] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1445

Covered by existing unit test.
Code coverage should increase slightly as checks are run via methods instead of direct value access

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -1039,12 +1039,6 @@ func (pc *PartitionContext) GetTotalApplicationCount() int {
 	return len(pc.applications)
 }
 
-func (pc *PartitionContext) GetTotalCompletedApplicationCount() int {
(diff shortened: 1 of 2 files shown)


pbacsko (approved):
+1 LGTM

PR #472, opened 2022-12-11
[YUNIKORN-1428] Add unit tests to ensure multiple users belongs to single group and vice versa cases

zhuqi-lucas:
### What is this PR for?
We need to add testing: to make sure multiple users with the same group trigger the correct updates etc.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1428

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application_test.go
+++ b/pkg/scheduler/objects/application_test.go
@@ -1465,6 +1465,54 @@ func TestIncAndDecUserResourceUsage(t *testing.T) {
 	assertUserGroupResource(t, getTestUserGroup(), resources.NewResourceFromMap(map[string]resources.Quantity{"first": 0}))
 }
 
+func TestIncAndDecUserResourceUsageInSameGroup(t *testing.T) {
+	res := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 5})
+	testgroups := []string{"testgroup"}
+	app := newApplicationWithUserGroup(appID1, "default", "root.unknown", "testuser", testgroups)
+	if app == nil || app.ApplicationID != appID1 {
+		t.Fatalf("app create failed which should not have %v", app)
+	}
+	app2 := newApplicationWithUserGroup(appID2, "default", "root.unknown", "testuser2", testgroups)
+	if app2 == nil || app2.ApplicationID != appID2 {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/application_test.go
```diff
@@ -90,15 +90,15 @@ func TestNewApplication(t *testing.T) {
 	if app.rmEventHandler == nil {
 		t.Fatal("non nil handler was not set in the new app")
 	}
-	assertUserGroupResource(t, getTestUserGroup(), nil)
+	assertUserGroupResource(t, getTestUserGroup("testuser", []string{"testgroup"}), nil, nil, 0)
```
wilfred-s, line 93: Same as before: if you need the ability to check user and group separately you add a new function for it instead of changing 50+ callers.

pkg/scheduler/objects/utilities_test.go
```diff
@@ -224,14 +224,14 @@ func newAllocationAskAll(allocKey, appID, taskGroup string, res *resources.Resou
 	return NewAllocationAskFromSI(ask)
 }
 
-func getTestUserGroup() security.UserGroup {
-	return security.UserGroup{User: "testuser", Groups: []string{"testgroup"}}
+func getTestUserGroup(userName string, groupNameList []string) security.UserGroup {
```
wilfred-s, line 227: Same here: use a separate function `getUserGroup(userName string, groupNameList []string) security.UserGroup` You can even make `getTestUserGroup()` call that one with the default test user details: `"testuser", []string{"testgroup"}`

wilfred-s (changes requested):

zhuqi-lucas:
Thank you @wilfred-s for review, addressed comments in latest PR.

pkg/scheduler/objects/utilities_test.go
zhuqi-lucas: Addressed in latest PR.

wilfred-s (approved):
LGTM, one typo will fix on commit

PR #471, opened 2022-12-09
[YUNIKORN-1440][FOLLOWUP] Remove expired apps from queue during cleanup

bgrams:
### What is this PR for?
Completed applications are stored permanently in 2 places - the PartitionContext and the application Queue. #463 ensured that these are permanently removed from the PC. This PR ensures that the Queue is considered in the cleanup loop as well.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1440

### How should this be tested?
Tests added

### Questions:
* [ ] - The licenses files need update.
* [ ] - There is breaking changes for older versions.
* [ ] - It needs documentation.

PR #474, opened 2022-12-12
[YUNIKORN-1468] Remove completedApplications from Queue

bgrams:
### What is this PR for?
As described in the Jira, this is a more comprehensive fix for the issues identified via YUNIKORN-1440 and discussed in #471. Proposed changes ensure that app history and cleanup mechanisms are solely managed at the Partition level and unlinked from the Queue. Queue history is also removed from the `/ws/v1/partition/{partition}/queue/{queue}/applications` REST endpoint.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1468

Tests modified but largely unchanged.

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -535,7 +535,14 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	apps := queue.GetCopyOfApps()
-	completedApps := queue.GetCopyOfCompletedApps()
+
+	completedApps := make(map[string]*objects.Application)
+	for _, app := range partitionContext.GetCompletedApplications() {
+		if app.GetQueuePath() == queueName {
+			completedApps[app.ApplicationID] = app
+		}
+	}
+
 	appsDao := make([]*dao.ApplicationDAOInfo, 0, len(apps)+len(completedApps))
 	for _, app := range apps {
 		appsDao = append(appsDao, getApplicationJSON(app))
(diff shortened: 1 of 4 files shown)


pkg/webservice/handlers.go
```diff
@@ -534,15 +534,18 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
 	}
-	for _, app := range completedApps {
-		appsDao = append(appsDao, getApplicationJSON(app))
+
+	for _, app := range partitionContext.GetCompletedApplications() {
+		if app.GetQueuePath() == queueName {
+			appsDao = append(appsDao, getApplicationJSON(app))
+		}
 	}
+
```
wilfred-s, line 548: I would suggest that we remove that and add a release note for the same to YuniKorn 1.2 explaining the reasoning behind the change.

wilfred-s (changes requested):

pkg/webservice/handlers.go
craigcondit: Agreed. We can deprecate the field in the 1.2 docs, and remove in 2.0. It will just be empty until then.

craigcondit (changes requested):
Looks good, we should just remove the iteration through apps for the queue REST API.

pkg/webservice/handlers.go
bgrams: SGTM. Addressed in latest commits.

craigcondit (approved):
+1 LGTM.

PR #476, opened 2022-12-13
[YUNIKORN-1464] Add new queue configuration options

pbacsko:
### What is this PR for?
Add new queue properties "preemption.policy" and "preemption.delay"

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1464

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -51,7 +51,8 @@ type Queue struct {
 	reservedApps          map[string]int          // applications reserved within this queue, with reservation count
 	parent                *Queue                  // link back to the parent in the scheduler
 	pending               *resources.Resource     // pending resource for the apps in the queue
-
+	preemptionFenced      bool
+	preemptionDelay       time.Duration
 	// The queue properties should be treated as immutable the value is a merge of the
 	// parent properties with the config for this queue only manipulated during creation
 	// of the queue or via a queue configuration update.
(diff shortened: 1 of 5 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -296,39 +297,64 @@ func (sq *Queue) setTemplate(conf configs.ChildTemplate) error {
...
 }
 
+func isFencingEnabled(str string) (bool, error) {
+	switch str {
+	// fifo is the default policy when not set
+	case "default":
+		return false, nil
+	case "fence":
+		return true, nil
+	default:
```
craigcondit, line 353: We need to also add "disabled" to this as that's going to be how we disable preemption for a queue. Also, we should probably have these be constants.

pkg/scheduler/objects/queue.go
```diff
@@ -296,39 +297,64 @@ func (sq *Queue) setTemplate(conf configs.ChildTemplate) error {
...
+				if err != nil {
+					log.Logger().Debug("queue fencing property configuration error",
+						zap.Error(err))
+				}
+			case configs.PreemptionDelay:
+				sq.preemptionDelay, err = time.ParseDuration(value)
+				if err != nil {
+					log.Logger().Debug("preemption delay property configuration error",
+						zap.Error(err))
+				}
```
craigcondit, line 331: We need to ensure that preemptionDelay is positive, and default to 30s if not specified.

craigcondit (changes requested):
Some minor behavioral changes.

pkg/scheduler/objects/queue.go
craigcondit: Also, preemption policy needs to be supported on parent queues as well, so we can't just apply these in leaf queues.

PR #478, opened 2022-12-14
[YUNIKORN-1465] Track dynamic preemption priority of applications

pbacsko:
### What is this PR for?
We want to track the lowest priority allocation (running pod) so that we can always return this value without unecessary calculations.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1465

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1510,8 +1510,9 @@ func (sa *Application) addAllocationInternal(info *Allocation) {
 		sa.allocatedResource = resources.Add(sa.allocatedResource, info.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedResource, sa.maxAllocatedResource)
 	}
-	if info.GetPriority() < sa.allocMinPriority {
-		sa.allocMinPriority = info.GetPriority()
+	priority := info.GetPriority()
+	if priority < sa.allocMinPriority {
+		sa.allocMinPriority = priority
 	}
 	sa.allocations[info.GetUUID()] = info
 }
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -1508,6 +1510,9 @@ func (sa *Application) addAllocationInternal(info *Allocation) {
 		sa.allocatedResource = resources.Add(sa.allocatedResource, info.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedResource, sa.maxAllocatedResource)
 	}
+	if info.GetPriority() < sa.allocMinPriority {
```
craigcondit, line 1513: add a variable here so we don't call GetPriority() multiple times (this locks the object repeatedly)

pkg/scheduler/objects/application.go
```diff
@@ -1640,9 +1645,32 @@ func (sa *Application) removeAllocationInternal(uuid string, releaseType si.Term
 		}
 	}
 	delete(sa.allocations, uuid)
+	if alloc.GetPriority() == sa.allocMinPriority {
+		sa.updateAllocationMinPriority(alloc.GetPriority())
+	}
 	return alloc
 }
 
+func (sa *Application) updateAllocationMinPriority(allocPriority int32) {
```
craigcondit, line 1654: This can just be a simple loop, tracking the lowest priority value we find. The extra minPriorityCount stuff doesn't save any time and adds complexity.

craigcondit (changes requested):
Logically good, but can be simplified.

craigcondit (approved):
+1 LGTM.

PR #479, opened 2022-12-14
[YUNIKORN-1466] Track dynamic preemption priority for queues

pbacsko:
### What is this PR for?
We want to track the lowest priority application so that we can always return this value without unecessary calculations.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1466

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -697,21 +697,26 @@ func (sq *Queue) RemoveApplication(app *Application) {
 		}
 	}
 
-	sq.Lock()
-	delete(sq.applications, appID)
-	delete(sq.appPriorities, appID)
-	delete(sq.appPreemptionPriorities, appID)
-	delete(sq.allocatingAcceptedApps, appID)
-	priority := sq.recalculatePriority()
-	preemptionPriority := sq.recalculatePreemptionPriority()
-	sq.Unlock()
-
+	priority, preemptionPriority := sq.internalRemoveApp(appID)
 	sq.parent.UpdateQueuePriority(sq.Name, priority)
 	sq.parent.UpdateQueuePreemptionPriority(sq.Name, preemptionPriority)
 
 	log.Logger().Info("Application completed and removed from queue",
 		zap.String("queueName", sq.QueuePath),
 		zap.String("applicationID", appID))
+}
+
+func (sq *Queue) internalRemoveApp(appID string) (int32, int32) {
+	sq.Lock()
+	defer sq.Unlock()
+	delete(sq.applications, appID)
+	delete(sq.appPriorities, appID)
+	delete(sq.appPreemptionPriorities, appID)
+	delete(sq.allocatingAcceptedApps, appID)
+	priority := sq.recalculatePriority()
+	preemptionPriority := sq.recalculatePreemptionPriority()
+
+	return priority, preemptionPriority
 }
 
 // GetCopyOfApps gets a shallow copy of all non-completed apps holding the lock
@@ -1535,7 +1540,7 @@ func priorityValueByPolicy(policy policies.PriorityPolicy, offset int32, priorit
 }
 
 func (sq *Queue) UpdateApplicationPriority(applicationID string, priority int32) {
-	if sq == nil || !sq.isLeaf {
+	if sq == nil || !sq.IsLeafQueue() {
 		return
 	}
 	value := sq.updateApplicationPriorityInternal(applicationID, priority)
@@ -1556,7 +1561,7 @@ func (sq *Queue) updateApplicationPriorityInternal(applicationID string, priorit
 }
 
 func (sq *Queue) UpdateQueuePriority(queueName string, priority int32) {
-	if sq == nil || sq.isLeaf {
+	if sq == nil || sq.IsLeafQueue() {
 		return
 	}
 	value := sq.updateQueuePriorityInternal(queueName, priority)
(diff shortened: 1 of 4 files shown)


craigcondit (changes requested):
This is a good start, but we need to update the priorities when app priorities (for leaf queues) or subqueue priorities (for parent queues) change. Additionally, this needs to propagate up the tree and take into account priority offsets as well.

pbacsko:
> This is a good start, but we need to update the priorities when app priorities (for leaf queues) or subqueue priorities (for parent queues) change. Additionally, this needs to propagate up the tree and take into account priority offsets as well.

Updated the PR with the new approach, no tests.

wilfred-s:
@pbacsko Please fix the conflicts so we can do a final review on the change.

pkg/scheduler/objects/queue.go
```diff
@@ -628,6 +635,8 @@ func (sq *Queue) AddApplication(app *Application) {
 	defer sq.Unlock()
 	sq.applications[app.ApplicationID] = app
 	sq.appPriorities[app.ApplicationID] = app.GetAskMaxPriority()
+	sq.appPreemptionPriorities[app.ApplicationID] = app.GetAllocationMinPriority()
+
```
wilfred-s, line 639: I did not see this when we added priorities earlier: when an application gets added to a queue there are no asks or allocations yet. The `queue.AddApplication()` call is triggered from a new application request from a shim and that request cannot have an ask/allocation. The init can be a simple default value for both.

pkg/scheduler/objects/queue.go
```diff
@@ -691,11 +700,14 @@ func (sq *Queue) RemoveApplication(app *Application) {
 	sq.Lock()
 	delete(sq.applications, appID)
 	delete(sq.appPriorities, appID)
+	delete(sq.appPreemptionPriorities, appID)
 	delete(sq.allocatingAcceptedApps, appID)
 	priority := sq.recalculatePriority()
+	preemptionPriority := sq.recalculatePreemptionPriority()
 	sq.Unlock()
```
wilfred-s, line 707: Can we refactor all these calls into a separate locked function `internalRemoveApp(appID string) int32` makes it much cleaner and a defer can be used to guarantee the unlock even if we add things later.

pkg/scheduler/objects/queue.go
```diff
@@ -1571,3 +1593,84 @@ func (sq *Queue) recalculatePriority() int32 {
 	sq.currentPriority = curr
 	return priorityValueByPolicy(sq.priorityPolicy, sq.priorityOffset, curr)
 }
+
+func (sq *Queue) UpdateApplicationPreemptionPriority(applicationID string, priority int32) {
+	if sq == nil || !sq.isLeaf {
```
wilfred-s, line 1598: there is a locked method to check `sq.isLeaf` should be using that one

pkg/scheduler/objects/queue.go
```diff
@@ -1571,3 +1593,84 @@ func (sq *Queue) recalculatePriority() int32 {
...
+		return
+	}
+	value := sq.updateApplicationPreemptionPriorityInternal(applicationID, priority)
+	sq.parent.UpdateQueuePreemptionPriority(sq.Name, value)
+}
+
+func (sq *Queue) UpdateQueuePreemptionPriority(queueName string, priority int32) {
+	if sq == nil || sq.isLeaf {
```
wilfred-s, line 1606: there is a locked method to check `sq.isLeaf` should be using that one

pkg/scheduler/objects/queue.go
```diff
@@ -1571,3 +1593,84 @@ func (sq *Queue) recalculatePriority() int32 {
...
+		log.Logger().Debug("Unknown application", zap.String("applicationID", applicationID))
+		return sq.preemptionPriority
+	}
+
+	sq.appPreemptionPriorities[applicationID] = priority
+	return sq.recalculatePreemptionPriority()
+}
+
+func (sq *Queue) recalculatePreemptionPriority() int32 {
+	var items *map[string]int32
```
wilfred-s, line 1640: I am not sure that I understand this indirection. If I say:
```
items = sq.appPreemptionPriorities
```
both will point to the same data structure it will not copy the data.

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
pbacsko: Technically it's true, I still prefer this approach because it slightly reduces coupling. cc @craigcondit

pkg/scheduler/objects/queue.go
pbacsko: Done

pkg/scheduler/objects/queue.go
pbacsko: Done

pkg/scheduler/objects/queue.go
pbacsko: Done

pkg/scheduler/objects/queue.go
pbacsko: I guess the idea here is that by using a pointer it's a tad cheaper to iterate over the map because we avoid copying.

wilfred-s: There is no copy when you do the assign, see this example code:
https://go.dev/play/p/qcJ9P-4ZCdB
Both point to one and the same data structure

pkg/scheduler/objects/queue.go
wilfred-s: The application object cannot be accessed by anything until the end of this function itself.  The queue is locked.
The overhead is 2 function calls, and 2 lock acquire and release actions. Locking cannot be contested as the application object cannot be accessed yet. That overhead is less than the Tag check...
OK to leave it.

pkg/scheduler/objects/queue.go
pbacsko: Oh, that's great. Makes sense to change then.

craigcondit (approved):
+1 LGTM.

PR #489, opened 2022-12-24
[YUNIKORN-1487] Add test for NewConfiguredQueue()

wusamzong:
### What is this PR for?
I add a test for NewConfiguraedQueue(). 
The test contains three part:
1. check simple variable assignment
2. adding leaf to queue
3. adding non-leaf to queue

### What type of PR is it?
* [x] - Task

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1487/

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue_test.go
+++ b/pkg/scheduler/objects/queue_test.go
@@ -1549,6 +1549,10 @@ func TestNewConfiguredQueue(t *testing.T) {
 	// check variable assignment
 	properties := getProperties()
 	resourceConf := getResourceConf()
+	// turn resouce config into resource struct
+	resourceStruct, err := resources.NewResourceFromConf(resourceConf)
+	assert.NilError(t, err, "failed to create new resource from config: %v", err)
+
 	parentConfig := configs.QueueConfig{
 		Name:            "PARENT_QUEUE",
 		Parent:          true,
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/queue_test.go
```diff
@@ -1545,6 +1545,67 @@ func TestApplyConf(t *testing.T) {
...
+	parent, err := NewConfiguredQueue(parentConfig, nil)
+	assert.NilError(t, err, "failed to create queue: %v", err)
+	assert.Equal(t, parent.Name, "parent_queue")
+	assert.Equal(t, parent.QueuePath, "parent_queue")
+	assert.Equal(t, parent.isManaged, true)
+	assert.Equal(t, parent.maxRunningApps, uint64(32))
+	assert.Assert(t, reflect.DeepEqual(properties, parent.template.GetProperties()))
+	// turn resouce config into resource struct
+	resourceStruct, err := resources.NewResourceFromConf(resourceConf)
+	assert.NilError(t, err, "failed to create new resource from config: %v", err)
```
wilfred-s, line 1573: move with the line that creates the resource conf (1551) for clarity

pkg/scheduler/objects/queue_test.go
```diff
@@ -1545,6 +1545,67 @@ func TestApplyConf(t *testing.T) {
...
+			Guaranteed: getResourceConf(),
+		},
+	}
+	childLeaf, err := NewConfiguredQueue(leafConfig, parent)
+	assert.NilError(t, err, "failed to create queue: %v", err)
+	assert.Equal(t, childLeaf.QueuePath, "parent_queue.leaf_queue")
+	assert.Assert(t, childLeaf.template == nil)
+	assert.Assert(t, reflect.DeepEqual(childLeaf.properties, parent.template.GetProperties()))
+	assert.Assert(t, reflect.DeepEqual(childLeaf.maxResource, parent.template.GetMaxResource()))
+	assert.Assert(t, reflect.DeepEqual(childLeaf.guaranteedResource, parent.template.GetGuaranteedResource()))
```
wilfred-s, line 1593: use `resources.Equals()` for the check. If by coincidence we get a zero value for a quantity the config and the set value will not be the same as we ignore zero max and guaranteed. It is safer in the long run as it tests expected behaviour.

wilfred-s (changes requested):
two minor points:

wilfred-s (approved):
LGTM

PR #492, opened 2023-01-02
[YUNIKORN-1446] Adopting table-driven style in the acl unit test

0yukali0:
### What is this PR for?
Current ACL unit tests do not adopt table-driven style, so I rewrite the ACL unit test file.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1446

make test in the core rep

### Questions:

### DIFF
--- a/pkg/common/security/acl_test.go
+++ b/pkg/common/security/acl_test.go
@@ -232,7 +232,10 @@ func TestACLAccess(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(fmt.Sprintf("vistor %v, acl %s", tt.visitor, tt.acl), func(t *testing.T) {
-			acl, _ := NewACL(tt.acl)
+			acl, err := NewACL(tt.acl)
+			if err != nil {
+				t.Error("the number of space should not be more than 2 because the number of categories only include users and groups")
+			}
 			if pass := acl.CheckAccess(tt.visitor); pass != tt.expected {
 				t.Errorf("allow expect:%v, got %v", tt.expected, pass)
 			}
--- a/pkg/common/security/acl_test.go
+++ b/pkg/common/security/acl_test.go
@@ -56,202 +56,79 @@ func IsSameACL(got, expected ACL) error {
 
 func TestACLCreate(t *testing.T) {
 	tests := []struct {
-		caseName string
 		input    string
 		expected ACL
 	}{
 		{
-			"Empty",
 			"",
-			ACL{
-				allAllowed: false,
-			},
-		},
+			ACL{allAllowed: false}},
 		{
-			"A space",
 			" ",
-			ACL{
-				users:      make(map[string]bool),
-				groups:     make(map[string]bool),
-				allAllowed: false,
-			},
-		},
+			ACL{users: make(map[string]bool), groups: make(map[string]bool), allAllowed: false}},
 		{
-			"One user",
 			"user1",
-			ACL{
-				users: map[string]bool{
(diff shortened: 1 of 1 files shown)


pkg/common/security/acl_test.go
```diff
@@ -19,119 +19,206 @@
...
+	if expected.users != nil && got.users != nil {
+		for username, expectedAllow := range expected.users {
+			if gotAllow, ok := got.users[username]; !ok {
+				return fmt.Errorf("username %s does not exist", username)
+			} else if expectedAllow != gotAllow {
+				return fmt.Errorf("username %s is not same", username)
+			}
+		}
+	} else if (expected.users != nil && got.users == nil) || (expected.users == nil && got.users != nil) {
+		return errors.New("users of an acl is not nil, expect one and got one should be same")
```
wilfred-s, line 40: You are not catching the case that `got` has more users than `expected`
Simplify and change check ordering:
1. check if length of both maps are equal: if not return
2. check equality of the members: let the range handle nil maps. 
3. Combine the user not found and the allow flag difference in one check:
```
gotAllow, ok := got.users[username]; !ok || gotAllow != expectedAllow
```

pkg/common/security/acl_test.go
```diff
@@ -19,119 +19,206 @@
...
 	}
+
+	if expected.groups != nil && got.groups != nil {
```
wilfred-s, line 43: Same as for users

pkg/common/security/acl_test.go
```diff
@@ -19,119 +19,206 @@
...
+		},
+		{
+			"*",
+			UserGroup{User: "user1", Groups: []string{"group1"}},
+			true,
+		},
+		{
+			acl:      "",
+			visitor:  UserGroup{User: "", Groups: nil},
+			expected: false,
```
wilfred-s, line 205: nit: remove struct field names

pkg/common/security/acl_test.go
```diff
@@ -19,119 +19,206 @@
...
+			"",
+			UserGroup{User: "user1", Groups: []string{"group1"}},
+			false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("vistor %v, acl %s", tt.visitor, tt.acl), func(t *testing.T) {
+			acl, err := NewACL(tt.acl)
+			if err != nil {
+				t.Error(err.Error())
```
wilfred-s, line 217: nit: explain that parse error is unexpected as you did before

wilfred-s (changes requested):

pkg/common/security/acl_test.go
0yukali0: Hi @wilfred-s 
Understood, thank you for comments.
I am going to update the commit later.

pkg/common/security/acl_test.go
0yukali0: I would remove the err later because the acl string is in the create unit tests.

0yukali0: Oops, I forgot that lint would request developer to handle the err in the UT.
So I updated the message of newACL error in the latest commit.

wilfred-s (approved):
LGTM

PR #498, opened 2023-01-31
[YUNIKORN-1549][core] Introduce a config to specify what attribute to use as instance type

zhuqi-lucas:
### What is this PR for?
We add a new config in https://github.com/apache/yunikorn-k8shim/pull/518 , the core side will use the extraconfig.
And we need to parse the instanceType in core side.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1549

### Questions:


craigcondit:
Please add a constant to the SI for the configuration key and default value. There is NO need to change the shim at all - every configuration setting in the configmap is passed along to the core as of v1.2.0.

wilfred-s:
-1 see the remarks in the K8shim PR.
First fix the attributes implementation and use a standardised label name.

zhuqi-lucas:
Hi @wilfred-s ,@craigcondit , https://github.com/apache/yunikorn-k8shim/pull/519 has been merged, and i have changed the code now.

And i will add the default value settings to yunikorn-site after this merged.

craigcondit:
@zhuqi-lucas, I'm now -1 on this approach. As I'm looking at this I see a problem. The core should not have any knowledge of the shim -- it is meant to be platform-agnostic, so we shouldn't have a Kubernetes-specific key or value present here. We already have an `InstanceType` constant in the SI that resolves to `si/instance-type`.  The kubernetes-specific code should go in the shim, and be used to determine which node label or annotation should be used to populate the `si/instance-type` attribute. This way, no matter what implementation the shim uses, the core can always rely on a node's instance type (if specified) being present in that attribute.

zhuqi-lucas:
Thank you @craigcondit for review and clarify, i change the core side code to use the SI value:
'''
InstanceType        = "si/instance-type"
'''

And the shim side will fill this value, if i make sense right?

craigcondit:
> Thank you @craigcondit for review and clarify, i change the core side code to use the SI value: InstanceType = "si/instance-type"
> 
> And the shim side will fill this value, if i make sense right?

Yes, the shim side code will need to be written yet, but this looks better.

craigcondit (approved):
+1, but would like @wilfred-s to weigh in as well.

zhuqi-lucas:
Thank you @craigcondit for quick checking, and the shim side change i submitted a new PR:
https://github.com/apache/yunikorn-k8shim/pull/531

yzhangal:
Thanks @craigcondit for approving the PR. HI @wilfred-s, would you please help reviewing again? thanks. FYI @zhuqi-lucas

PR #505, opened 2023-02-14
[YUNIKORN-1567] Do not attempt to schedule tasks if partition update is in progress

pbacsko:
### What is this PR for?
We should block config updates while we're scheduling allocation asks. If the queue hierarchy changes, this can result in an incorrect, undeterministic behavior.
`ClusterContext.Lock()` is too coarse, `PartitionContext.Lock()` is also problematic for this purpose.
For each partition, a mutex is created which is locked during scheduling and config update.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1567

### Questions:

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -44,7 +44,6 @@ const disableReservation = "DISABLE_RESERVATION"
 
 type ClusterContext struct {
 	partitions     map[string]*PartitionContext
-	updateLocks    map[string]*sync.Mutex
 	policyGroup    string
 	rmEventHandler handler.EventHandler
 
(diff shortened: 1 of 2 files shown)


craigcondit:
This looks overly complex. What is the problem with using a partition lock? At the very least, the lock should live in the partition itself and not in a separate map...

pbacsko:
> This looks overly complex. What is the problem with using a partition lock? At the very least, the lock should live in the partition itself and not in a separate map...

Sorry, I forgot to answer this. Partition Lock is not hold, but there are various calls when it's hold then released for a change. I think scheduling it's designed to be "no-lock".

This solution might also raise some eyebrows, but as I went through various codepaths, it looks safe as long as we maintain certain invariants.

craigcondit:
> Sorry, I forgot to answer this. Partition Lock is not hold, but there are various calls when it's hold then released for a change. I think scheduling it's designed to be "no-lock".

I still think we can take the partition context lock (for reads) during the scheduling cycle, rather than introducing a new one. This still lets us (eventually) process partitions in parallel without worrying about structural changes happening during scheduling. @wilfred-s what do you think?

PR #506, opened 2023-02-15
[YUNIKORN-1578][core] Use zap.Stringer instead of calling String on object

zhuqi-lucas:
### What is this PR for?
Throughout the code we have lots of places where we use the following call structure:

zap.String("text here", object.String()) 
This causes the object.String() call to happen before we even consider if it needs to be logged. Zap has a lazy way of doing this that removes the call to create the string  until it is needed, after the level checks etc. The above call becomes:

zap.Stringer("text here", object) 
Affect core and k8shim

### What type of PR is it?
* [x] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1578

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -870,7 +870,7 @@ func CalculateAbsUsedCapacity(capacity, used *Resource) *Resource {
 		log.Logger().Debug("Cannot calculate absolute capacity because of missing capacity or usage")
 		return absResource
 	}
-	var missingResources strings.Builder
+	missingResources := &strings.Builder{}
 	for resourceName, availableResource := range capacity.Resources {
 		var absResValue int64
 		if usedResource, ok := used.Resources[resourceName]; ok {
@@ -909,7 +909,7 @@ func CalculateAbsUsedCapacity(capacity, used *Resource) *Resource {
 	}
 	if missingResources.Len() != 0 {
 		log.Logger().Debug("Absolute usage result is missing resource information",
-			zap.Stringer("missing resource(s)", &missingResources))
+			zap.Stringer("missing resource(s)", missingResources))
 	}
 	return absResource
 }
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -909,7 +909,7 @@ func CalculateAbsUsedCapacity(capacity, used *Resource) *Resource {
 	}
 	if missingResources.Len() != 0 {
 		log.Logger().Debug("Absolute usage result is missing resource information",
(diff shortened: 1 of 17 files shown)


manirajv06 (changes requested):
Changes looks good. Minor nits.

pkg/common/resources/resources.go
```diff
@@ -909,7 +909,7 @@ func CalculateAbsUsedCapacity(capacity, used *Resource) *Resource {
 	}
 	if missingResources.Len() != 0 {
 		log.Logger().Debug("Absolute usage result is missing resource information",
-			zap.String("missing resource(s)", missingResources.String()))
+			zap.Stringer("missing resource(s)", &missingResources))
```
manirajv06, line 912: Can you check the effect of & ?

pkg/scheduler/objects/allocation.go
```diff
@@ -116,7 +116,7 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
 	// order is important as task group can be set without placeholder but not the other way around
 	if alloc.Placeholder && alloc.TaskGroupName == "" {
 		log.Logger().Debug("Allocation cannot be a placeholder without a TaskGroupName",
-			zap.String("SI alloc", alloc.String()))
+			zap.Stringer("SI alloc", alloc),
```
manirajv06, line 119: Fix syntax issue

manirajv06 (changes requested):

pkg/common/resources/resources.go
zhuqi-lucas: Thank you @manirajv06 for review, I have checked the code here:
`
// String returns the accumulated string.
func (b *Builder) String() string {
	return *(*string)(unsafe.Pointer(&b.buf))
}
`

pkg/scheduler/objects/allocation.go
zhuqi-lucas: Addressed it.

pkg/common/resources/resources.go
wilfred-s: A much cleaner solution would be to change the init of the variable to:
```
missingResources := &strings.Builder{}
```
All functions of the Builder use pointer receivers and this is consistent for all calls.

zhuqi-lucas: Good point @wilfred-s , addressed it now.

wilfred-s (approved):
+1 LGTM
@manirajv06 please check

manirajv06 (approved):
Looks good. Merging..

PR #507, opened 2023-02-17
[YUNIKORN-1385] To provide visibility to application's aggregated res…

yzhangal:
…ource consumption

### What is this PR for?

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:


yzhangal:
I have done local test with docker desktop and real cluster testing, the with ./deployments/examples/sleep/sleeppods.yaml,

It can report the following:
{“appId”:“application-sleep-0002-test13”,“submissionTime”:“1675494465465”,“startTime”:“1675494467467”,“finishTime”:“1675494529037”,“user”:“kubernetes-admin”,“queue”:“root.sandbox”,“state”:“Completed”,“rmID”:“mycluster”,“resourceUsage”:{“”:{“memorySeconds”:15500000000,“vcoreSeconds”:3100}}}
{“appId”:“application-sleep-0001-test13”,“submissionTime”:“1675494465465”,“startTime”:“1675494467466”,“finishTime”:“1675494529563”,“user”:“kubernetes-admin”,“queue”:“root.sandbox”,“state”:“Completed”,“rmID”:“mycluster”,“resourceUsage”:{“”:{“memorySeconds”:16000000000,“vcoreSeconds”:3200}}}
{“appId”:“application-sleep-0003-test13”,“submissionTime”:“1675494466465”,“startTime”:“1675494468467”,“finishTime”:“1675494529576”,“user”:“kubernetes-admin”,“queue”:“root.sandbox”,“state”:“Completed”,“rmID”:“mycluster”,“resourceUsage”:{“”:{“memorySeconds”:15500000000,“vcoreSeconds”:3100}}}

The two level map introduced by this diff will have one entry at the top level map for each instance type used by the application, then one entry at the second level for each resource type (CPU, memory etc) used. 

Number of top level buckets: the number of instance types
Second level, for each instance type: the number of resource types, currently vcore and Memory

I did basic memory foot print testing with pprof, due to the small scale of apps that can run in parallel, the resource usage data added to Application struct is not noticeable. Based on the analysis done in https://levelup.gitconnected.com/memory-allocation-and-performance-in-golang-maps-b267b5ad9217, 

"A map in Go is a hash table that stores its key/value pairs into buckets. Each bucket is an array that holds up to 8 entries. The default number of buckets is 1. Once the number of entries across each bucket reaches an average load of buckets (aka load factor), the map gets bigger by doubling the number of buckets. Every time a map grows, it allocates memory for the new-coming entries. In practice, every time the load of the buckets reaches 6.5 or more, the map grows. This value is hardcoded and was chosen to optimize memory usage."

Then I did a measurement with a standalone program, by creating 1000 such maps, with 5 instance types, and 2 resource types each instance type. The memory consumption is measured to be about 256 bytes per map (or per Yunikorn application with this PR). I used "top" command to observe the memory usage change.

yzhangal:
This PR is meant to be added on top of https://issues.apache.org/jira/browse/YUNIKORN-1549. But I hope to get some review of the first version while we are waiting for YUNIKORN-1549.

Hi @yangwwei @wilfred-s @craigcondit @zhuqi-lucas would you please help taking a look? thanks a lot.

yzhangal:
The 256 bytes per map is higher than what I expected so I'm still taking further look to see if it can be reduced. But Would appreciate if you can take a look at the PR. Thanks @wilfred-s @yangwwei @craigcondit @zhuqi-lucas

pkg/scheduler/objects/application_test.go
```diff
@@ -1061,6 +1064,125 @@ func TestCompleted(t *testing.T) {
...
+	//fmt.Print("resourceMap: ")
+	//fmt.Println(resourceUsageMap)
+
+	itypeResource, ok := resourceUsageMap["itype-1"]
+	if !ok {
+		assert.Assert(t, memorySeconds == -1 || vcoresSecconds == -1, "no resource usage")
+		return
+	}
+
+	//fmt.Print("itypeResource: ")
```
zhuqi-lucas, line 1086: Remove debug print.

pkg/scheduler/objects/node.go
```diff
@@ -117,6 +121,15 @@ func (sn *Node) GetAttribute(key string) string {
 	return sn.attributes[key]
 }
 
+// Get inst type from attributes passed to core from shim.
+func (sn *Node) GetInstType() string {
+	itype := sn.GetAttribute(YunikornInstanceType)
```
zhuqi-lucas, line 126: Change after instanceType merged.

pkg/scheduler/objects/application_test.go
```diff
@@ -1061,6 +1064,125 @@ func TestCompleted(t *testing.T) {
...
+	var jsonMap map[string]interface{}
+	var resource interface{}
+	var detailedResource map[string]interface{}
+
+	json.Unmarshal([]byte(appSummary), &jsonMap)
+
+	resource = jsonMap["resourceUsage"]
+
+	resourceUsageMap := resource.(map[string]interface{})
+	//fmt.Print("resourceMap: ")
```
zhuqi-lucas, line 1077: Remove debug print.

pkg/scheduler/objects/application.go
```diff
@@ -93,6 +97,7 @@ type Application struct {
 	execTimeout          time.Duration               // execTimeout for the application run
 	placeholderTimer     *time.Timer                 // placeholder replace timer
 	gangSchedulingStyle  string                      // gang scheduling style can be hard (after timeout we fail the application), or soft (after timeeout we schedule it as a normal application)
+	startTime            time.Time                   // the time that the application starts running. Default is zero.
```
zhuqi-lucas, line 100: One question, when we are starting state the app for gang scheduler enabled, it means the gang scheduler placeholder allocations are all allocated, if these placeholder allocations need to calculate the resource usage, because before the startTime here, the gang scheduler allocations are all bound in k8s shim side.
And if the gang scheduler timeout, if the placeholder allocations need to calculate even the app failed, because some placeholder allocations have been occupied resources for some time, and the default timeout is 15 mins, it's not a short time for resources occupied.

cc @wilfred-s @craigcondit

zhuqi-lucas (commented):
Thank a lot @jyz0309 for this work, left some comments.

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,55 @@ func (q Quantity) string() string {
...
+}
+
+// Aggregate the resource usage to UsedResourceMap[instType]
+// The time the given resource used is the delta between the resource createTime and currentTime
+func (urt *UsedResourceTracker) AggregateUsedResource(instType string,
+	resource *Resource, createTime time.Time) {
+	var aggregatedResourceTime map[string]int64
+	var timeDiff int64
+	var releaseTime time.Time
+	var resourceSecondsKey string
```
pbacsko, line 67: Declare the variables close to their usage, increasing readability. Most of the time you can use `:=`.
Eg.
```
releaseTime := time.Now()
timeDiff := int64(releaseTime.Sub(createTime).Seconds())
resourceSecondsKey := key + "Seconds"
```
It's preferred to start the method with Lock/Unlock pair if there's no reason to do otherwise.

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,55 @@ func (q Quantity) string() string {
...
+			curUsage = 0
+		}
+		curUsage += int64(element) * timeDiff  //resource size times timeDiff
+		aggregatedResourceTime[resourceSecondsKey] = curUsage
+	}
+	urt.UsedResourceMap[instType] = aggregatedResourceTime
+}
+
+func (urt *UsedResourceTracker) GetResourceUsageSummary() string {
+	jsonStr, err := json.Marshal(urt.UsedResourceMap)
```
pbacsko, line 90: `jsonStr` suggests that it's a string, but it's a byte array.

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,55 @@ func (q Quantity) string() string {
...
+	if err != nil {
+		return string(err.Error())
+	} else {
+		return string(jsonStr)
+	}
+}
```
pbacsko, line 96: Simplify:
```
if err != nil {
  return err.Error()   // already a string
}
return string(jsonStr)
```

pkg/scheduler/objects/application.go
```diff
@@ -1920,6 +1947,26 @@ func (sa *Application) cleanupAsks() {
...
+	res += "\"submissionTime\":\"" + strconv.FormatInt(sa.SubmissionTime.UnixMilli(), 10) + "\","
+	res += "\"startTime\":\"" + strconv.FormatInt(sa.StartTime().UnixMilli(), 10) + "\","
+	res += "\"finishTime\":\"" + strconv.FormatInt(sa.FinishedTime().UnixMilli(), 10) + "\","
+	res += "\"user\":\"" + sa.GetUser().User + "\","
+	res += "\"queue\":\"" + sa.GetQueuePath() + "\","
+	res += "\"state\":\"" + sa.stateMachine.Current() + "\","
+	res += "\"rmID\":\"" + rmID + "\","
+	res += "\"resourceUsage\":" + sa.usedResourceTracker.GetResourceUsageSummary()
+	res += "}"
+	return res
```
pbacsko, line 1967: It's weird to see all these formatting here. I do believe the `Application` object should not be concerned about formatting,  that should be delegated to the caller.

One approach is to just call the Get* methods to retrieve the relevant piece of data. Missing getters can be added.

If it's necessary to return a snapshot of these data guarded by locks, you can create an `AppSummary` type which contains all these information and return that.

pkg/metrics/init.go
```diff
@@ -229,3 +236,7 @@ func formatMetricName(metricName string) string {
 	}
 	return string(newBytes)
 }
+
+func (m *Metrics) LogAppSummary(appSummary string) {
+	log.Println(AppSummaryHeader + appSummary)
+}
```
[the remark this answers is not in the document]
pbacsko: This method is called from  `PartitionContext`. It has a very specific purpose, it only prints something else to the console which is completely unrelated to metrics. Should be moved.

pbacsko (changes requested):
Added various comments.

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,55 @@ func (q Quantity) string() string {
 	return strconv.FormatInt(int64(q), 10)
 }
 
+// Util struct to keep track of application resource usage
+type UsedResourceTracker struct {
+	// Two level map for aggregated resource usage
+	// With instance type being the top level key, the mapped value is a map:
+	//   resource type (CPU, memory etc) -> the aggregated used time (in seconds) of the resource type
+	//
+	UsedResourceMap map[string]map[string]int64
```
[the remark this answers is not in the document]
yzhangal: Thanks for the review Qi, very good feedback, will be addressed in new rev.

pkg/common/resources/resources.go
yzhangal: Thanks for the review and very good feedback Pbacsko! Go language is new to me and I'm still learning, will address in next rev.

pkg/common/resources/resources.go
yzhangal: Good, will be addressed in next rev.

pkg/common/resources/resources.go
yzhangal: Thanks, to be addressed in next rev.

pkg/metrics/init.go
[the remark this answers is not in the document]
yzhangal: Hi Qi, Println itself will append newline. Thanks.

HI Pbacsko, I did spend good amount of time to identify a suitable place for this, I put it here thinking that appsummary is part of the application (aggregated) metrics, but if you have other place to recommend, I will be happy to give it a try. Thanks.

pkg/scheduler/objects/application.go
yzhangal: Hi Qi, thanks for the thoughts. My understanding is that placeholder allocations as treated as being used, see 
    func (sa *Application) removeAllocationInternal(
only when the placeholder allocations are replaced with "real" allocations, they will be considered used.
I may have misunderstanding here, further comments are welcome.
Thanks,

pkg/scheduler/objects/application.go
yzhangal: Thanks, will be addressed in next rev.

pkg/metrics/init.go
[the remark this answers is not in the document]
craigcondit: This needs to go through the zap Logger or we risk corrupting output. The standard log package goes directly to stdout/stderr and does not use a consistent format. The AppSummaryHeader should be a zap.String("key", "value") as well, this allows for JSON-filtering of the resulting message.

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,52 @@ func (q Quantity) string() string {
...
+			curUsage = 0
+		}
+		curUsage += int64(element) * timeDiff  // resource size times timeDiff
+		aggregatedResourceTime[resourceSecondsKey] = curUsage
+	}
+	urt.UsedResourceMap[instType] = aggregatedResourceTime
+}
+
+func (urt *UsedResourceTracker) GetResourceUsageSummary() string {
+	urt.Lock()
```
craigcondit, line 86: This should be RLock() / RUnlock()

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,52 @@ func (q Quantity) string() string {
...
+	defer urt.Unlock()
+	urm, err := json.Marshal(urt.UsedResourceMap)
```
craigcondit, line 88: We should not be marshalling into JSON here, as the zap logger can do that for us. Instead, we should log individual fields.

craigcondit (changes requested):
I think there's still some fundamental issues here to be addressed.

yzhangal:
> I think there's still some fundamental issues here to be addressed.

Thanks a lot for the detailed review. Please see my reply/comments below.

pkg/common/resources/resources.go
yzhangal: Good point, will address in next rev.

pkg/common/resources/resources.go
yzhangal: HI Craig, that means we need to create a snapshot (deep copy) of the UsedResourceMap and return this snapshot here instead of the marshalled string, and then have zap logger report the map as json.  That's what you suggested right?

 @wilfred-s raised the concern that map uses more memory than one would expect earlier, and I tested it, it's indeed a lot of more memory consumption than I expected. Plus the runtime to do deep copy. I wonder if we really want to go this route now, or we can defer to later when there is a consumer requiring map instead of string? This is internal to yunikorn-core right now and the app summary is only used for the logging added by this ticket. 

Thanks.

pkg/metrics/init.go
```diff
@@ -166,6 +169,10 @@ type CoreEventMetrics interface {
 	Reset()
 }
 
+func (m *Metrics) LogAppSummary(appSummary string) {
```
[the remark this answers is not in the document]
yzhangal: I log out json format (like YARN does) in one line for each app for the ease of fluentBit filtering. Would you please explain why we have to report separate fields? To me, json is more compact, and fluentBit filtering will be much more harder with individually reported fields. Thanks.

pkg/scheduler/objects/application.go
```diff
@@ -107,6 +112,50 @@ type Application struct {
...
+	StartTime time.Time
+	FinishTime time.Time
+	User string
+	Queue string
+	State string
+	RmID string
+	ResourceUsage string
+}
+
+func (as *ApplicationSummary) GetReport() string {
```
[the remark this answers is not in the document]
yzhangal: See my comment above about json or separated fields, thanks.

pkg/scheduler/objects/application.go
```diff
@@ -107,6 +112,50 @@ type Application struct {
...
+	appSummary := &ApplicationSummary{
+		ApplicationID: sa.ApplicationID,
+		SubmissionTime: sa.SubmissionTime,
+		StartTime: sa.startTime,
+		FinishTime: sa.finishedTime,
+		User: sa.user.User,
+		Queue: sa.queuePath,
+		State: sa.stateMachine.Current(),
+		RmID: rmID,
+		ResourceUsage: sa.usedResourceTracker.GetResourceUsageSummary(),
```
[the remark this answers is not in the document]
yzhangal: This structure is meant to produce a string app summary, there is no other tooling downstream except for the one line logging I added with this ticket.  One can always get the resource usage structure from the application struct. And based on the map memory footprint concern @wilfred-s raised, can we use string for now, and change to map later when we really need a map?

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,52 @@ func (q Quantity) string() string {
...
+	defer urt.Unlock()
+
+	releaseTime := time.Now()
+	timeDiff := int64(releaseTime.Sub(createTime).Seconds())
+	aggregatedResourceTime, ok := urt.UsedResourceMap[instType]
+	if !ok {
+		aggregatedResourceTime = map[string]int64{}
+	}
+	for key, element := range resource.Resources {
+		resourceSecondsKey := key + "Seconds"
```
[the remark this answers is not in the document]
yzhangal: Thanks for your detailed review. 

What custom resource types you were referring too? are they not going to be based on seconds? 

To explain why I did this:

I'm trying to make it easy for viewer to grasp the meaning of the numbers reported when one looks at the app summary reported.  

Plus, the measurement of resource usage should be "resourceAmount * timeLength", not just "resourceAmount". Without a time unit ("Seconds" here) in the name, one can misread it as amount of resource, instead of "amount of resource * timeLength". If you notice, current yunikorn UI does report memory/vcore as "resources used", which is a misnomer in the UI, because what the UI displayed is a snapshot of currently used resourceAmount, not "resourceAmount * timeLength". 

In addition, having the "Seconds" string in it makes it easy to adapt to since we are migrating from YARN to yunikorn, as YARN does report memorySeconds and vcoreSeconds too. 

Wonder with the above explanation, would you still suggest removing the "Seconds"? 

Thanks.

pkg/common/resources/resources.go
craigcondit: This also needs to be able to be turned off via zap logger levels. Right now, every user will pay the cost of marshalling this complex object every time, even if logging of these events is disabled. That's part of what the zap logger framework gets you. It also standardizes representation of objects in the log so that the log formats are consistent throughout the application. I'll say again: This MUST not be a string returned here.

pkg/metrics/init.go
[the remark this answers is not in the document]
craigcondit: As before, this needs to be consistent with other logging used throughout the system, and be configurable via log levels in zap. Doing it this way forces ALL users of logging to have to deal with a log message that is formatted completely differently from the rest of the application log messages.

pkg/scheduler/objects/application.go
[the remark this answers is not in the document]
craigcondit: As this will be output to the same destination as other logs, it MUST be formatted consistently. This is not negotiable.

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,52 @@ func (q Quantity) string() string {
...
+// Aggregate the resource usage to UsedResourceMap[instType]
+// The time the given resource used is the delta between the resource createTime and currentTime
+func (urt *UsedResourceTracker) AggregateUsedResource(instType string,
+	resource *Resource, createTime time.Time) {
+	urt.Lock()
+	defer urt.Unlock()
+
+	releaseTime := time.Now()
+	timeDiff := int64(releaseTime.Sub(createTime).Seconds())
+	aggregatedResourceTime, ok := urt.UsedResourceMap[instType]
```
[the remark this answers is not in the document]
craigcondit: The core alone is needed here. We need to record a timestamp into the allocation at the same time the usage is added to the queue (i.e. when a node is assigned).

pkg/common/resources/resources.go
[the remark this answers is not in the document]
craigcondit: Keep the names consistent as this way a downstream user doesn't need to subtract off "Seconds" to get the original resource type. It's simply unnecessary, and adds additional complexity.

pkg/common/resources/resources.go
yzhangal: HI @craigcondit , wonder if this would work:

I renamed UsedResourceTracker to UsedResource here, and added Clone() method in UsedResource,

func (ur *UsedResource) Clone() *UsedResource {
	if ur == nil:
		return nil

	ret := NewUsedResource()
	ur.RLock()
	defer ur.RUnlock()
	for k, v := range ur.UsedResourceMap {
		sourceEntry := map[string]int64(v)
		destEntry := make(map[string]int64)
		for key, element := range sourceEntry {
			destEntry[key] = element
		}
		ret.UsedResourceMap[k] = destEntry
	}
	return ret
}

func (sa *Application) GetApplicationSummary(rmID string) *ApplicationSummary {
	sa.RLock()
	defer sa.RUnlock()
	appSummary := &ApplicationSummary{
		ApplicationID: sa.ApplicationID,
		SubmissionTime: sa.SubmissionTime,
		StartTime: sa.startTime,
		FinishTime: sa.finishedTime,
		User: sa.user.User,
		Queue: sa.queuePath,
		State: sa.stateMachine.Current(),
		RmID: rmID,
		**ResourceUsage: sa.usedResource.Clone(),**
	}
	return appSummary
}

func (as *ApplicationSummary) GetReport() string {
	res := "{"
	res += "\"appId\":\"" + as.ApplicationID + "\","
	res += "\"submissionTime\":\"" + strconv.FormatInt(as.SubmissionTime.UnixMilli(), 10) + "\","
	res += "\"startTime\":\"" + strconv.FormatInt(as.StartTime.UnixMilli(), 10) + "\","
	res += "\"finishTime\":\"" + strconv.FormatInt(as.FinishTime.UnixMilli(), 10) + "\","
	res += "\"user\":\"" + as.User + "\","
	res += "\"queue\":\"" + as.Queue + "\","
	res += "\"state\":\"" + as.State + "\","
	res += "\"rmID\":\"" + as.RmID + "\","
	**res += "\"resourceUsage\":" + as.ResourceUsage.GetResourceUsageJson()**
	res += "}"
	return res
}
Above I introduced a GetResourceUsageJson() because I was doing different processing with the different fields you can see in the above method, which abstract what information we need to report in what format.

func (sa *Application) GetAppSummary(rmID string) string {
	appSummary := sa.GetApplicationSummary(rmID)
	ret := appSummary.GetReport()
	appSummary.ResourceUsage = nil
	return ret

And the final report is created with:

	log.logger().Info(
		zap.String("YK_APP_SUMMARY", app.GetAppSummary(pc.RmID)))

inside  func (pc *PartitionContext) moveTerminatedApp(appID string) {

If user chooses to use log level higher than info, then, s/he may not get the report. 

To be honest, we actually hope we can always report this without being impacted by the log level.

Thanks

craigcondit: This is not YARN, we don't do the same things. We have to play nice within the Kubernetes ecosystem, where logging all goes through a single output (stdout). Because of this, we MUST be consistent in how we log, or we make it difficult for ALL downstream consumers of log events. We can't just change things up for the convenience of your consumer code. 

As for this:

`log.logger().Info(zap.String("YK_APP_SUMMARY", app.GetAppSummary(pc.RmID)))`

You will end up evaluating `app.GetSummary(pc.RmID)` whether or not logging is enabled. To avoid this, one method would be to use `Stringer()` and provide a deferred function that retrieves the app summary as a string. However, this breaks the structured logging as you are logging a complex object. It's better to log this as individual fields so that the zap log line can be parsed appropriately.

pkg/scheduler/objects/application.go
yzhangal: Thanks @craigcondit 

I have some questions about the startTime.

Currently I implemented the startTime in the app state machine as: 

```
fmt.Sprintf("enter_%s", Starting.String()): func(_ context.Context, event *fsm.Event) {
				app.startTime = time.Now()
```
				
I guess the Start state in this diagram (https://yunikorn.apache.org/docs/next/design/scheduler_object_states/) means the scheduler started working on allocation for the app. Wonder if it makes sense to have a state in the diagram to represent the first pod binding? If there is such a state, we can set the startTime in the state machine when entering this state.

Otherwise, I studied the code and saw that we can possibly have a field bindingTime in the Allocation structure to represent the binding time, and set it in the following method:

`func (sa *Application) tryNodes(ask *AllocationAsk, iterator NodeIterator) *Allocation {`

would you please confirm?

Thanks.

craigcondit: If you're going to account for usage accurately, you need to track each task from start to finish, not at the app level. If you have 1000 tasks, but can only run 10 at a time, you're going to have a stair-step chart as the first batch of 10 run and complete, then the next 10, etc. If each task takes 1 minute to complete, an accurate assessment of total time should be 1000 minutes. But tracking each task from the start of the application, you will account for 1 minute for each of the first 10 pods, 2 minutes for each of the next 10, etc. The final batch of 10 (100 batches total) will be tracked as 100 minutes each. On average, 50 minutes will be accounted to each task, so you will over-report actual usage by a factor of 50.

craigcondit: The right way to do this would be to track the timestamp per allocation when the allocation is associated with a node, and when an allocation terminates, report that usage as ((allocationEndTime - allocationStartTime) * resources.

yzhangal: Thanks @craigcondit , that boils down to the last question in my previous comment:

Otherwise, I studied the code and saw that we can possibly have a field bindingTime in the Allocation structure to represent the binding time, and set it in the following method:

`func (sa *Application) tryNodes(ask *AllocationAsk, iterator NodeIterator) *Allocation {`

I think the allocationEndTime of your term is the time when the allocation is released, which I currently use and the pod finish time.

Would you please advice?

Thanks.

craigcondit: Yes, tracking the time (per allocation) between allocated and released will work. The application doesn't really impact this at all.

craigcondit: A future enhancement could be to track total accumulated resources per unit time in the application for released allocations, and then provide an API to also get "in-progress" allocations. This could be used to get better granularity than task termination for long-running applications.

yzhangal: Thanks, agreed.

yzhangal: Hi Craig,

I found that we can not recover the bindTime for recovered allocation because si.Allocation come from the shim, unless we change the protocol of for this structure:

// Create a new Allocation from a node recovered allocation.
// Also creates an AllocationAsk to maintain backward compatible behaviour
// This returns a nil Allocation on nil input or errors
func NewAllocationFromSI(alloc *si.Allocation, instType string) *Allocation {

Or we simply set the current time to bind time for this kind of allocation? It seems that if the recovered allocation is because of the yunikorn restart, then we do need to let the shim to pass this info.

Thanks.

yzhangal: Hi @craigcondit Another question, I saw that all callers of the following method pass "Reserved" as the value for the first parameter, wonder if it's intended to pass a different value at all, given that it's meant to be a reserved allocation.

```
func newReservedAllocation(result AllocationResult, nodeID string, instType string, ask *AllocationAsk) *Allocation {
	alloc := NewAllocation("", nodeID, instType, ask)
```
Thanks.

pkg/common/resources/resources.go
```diff
@@ -44,6 +46,59 @@ func (q Quantity) string() string {
...
+
+func (ur *UsedResource) Clone() *UsedResource {
+	if ur == nil {
+		return nil
+	}
+	ret := NewUsedResource()
+	ur.RLock()
+	defer ur.RUnlock()
+	for k, v := range ur.UsedResourceMap {
+		sourceEntry := map[string]int64(v)
```
pbacsko, line 68: If I'm reading this correctly, this is completely unnecessary and very complicated read. The variable itself and the type conversion too. Just range through "v".

pkg/scheduler/objects/application.go
```diff
@@ -107,6 +111,48 @@ type Application struct {
...
+	sa.RLock()
+	defer sa.RUnlock()
+	appSummary := &ApplicationSummary{
+		ApplicationID: sa.ApplicationID,
+		SubmissionTime: sa.SubmissionTime,
+		StartTime: sa.startTime,
+		FinishTime: sa.finishedTime,
+		User: sa.user.User,
+		Queue: sa.queuePath,
+		State: sa.stateMachine.Current(),
```
pbacsko, line 149: This is guarded by a read lock inside the state machine. It's most likely not an issue, but it's best not to have multiple locks grabbed this way. So move this out before `sa.Lock()` and save it to a variable.

pbacsko (changes requested):
I found two minor issues, maybe I'll take a deeper look next time.

pkg/scheduler/objects/node.go
```diff
@@ -33,6 +33,10 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
+const (
+	UnknownInstanceType = "unknown"
```
craigcondit, line 37: Maybe we want unknown nodes to stand out a bit more -- perhaps "OTHER" in uppercase?

pkg/scheduler/objects/preemption.go
```diff
@@ -658,6 +660,10 @@ func (pcr *predicateCheckResult) populateVictims(victimsByNode map[string][]*All
 	for i := 0; i <= pcr.index; i++ {
 		victim := victimList[i]
 		pcr.victims = append(pcr.victims, victim)
+		// Take the instance type info from the first victim Allocation of this given nodeID
```
craigcondit, line 663: This is just incorrect. Just because preemptions happened on a node does not ensure that node will ultimately be where the new allocation goes.

pkg/scheduler/objects/preemption.go
```diff
@@ -349,6 +349,7 @@ func (p *Preemptor) checkPreemptionPredicates(predicateChecks []*si.PreemptionPr
 		result := &predicateCheckResult{
 			allocationKey: check.AllocationKey,
 			nodeID:        check.NodeID,
+			instType:      UnknownInstanceType,
```
craigcondit, line 352: Let's keep the instance type out of the preemption code. It's only needed on allocations. Once a node is chosen for the allocation we can apply the instance type.

craigcondit (changes requested):
A few minor points, a few questions, plus some refactoring around instance type needed.

yzhangal (commented):
HI @craigcondit , thanks much for your review, would you please see my responses here when you get chance? thanks.

pkg/scheduler/objects/allocation.go
```diff
@@ -99,16 +103,17 @@ func NewAllocation(uuid, nodeID string, ask *AllocationAsk) *Allocation {
 	}
 }
 
-func newReservedAllocation(result AllocationResult, nodeID string, ask *AllocationAsk) *Allocation {
-	alloc := NewAllocation("", nodeID, ask)
+func newReservedAllocation(result AllocationResult, nodeID string, instType string, ask *AllocationAsk) *Allocation {
```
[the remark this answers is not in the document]
craigcondit: What I meant was have you done an exhaustive check for all code paths that use the allocation?

pkg/scheduler/objects/preemption.go
craigcondit: The instance type of the node used for the NEW allocation is what's needed. This has nothing at all to do with the preemption logic. Please fix this.

pkg/scheduler/objects/preemption.go
craigcondit: As I indicated above: Only the instance type of the node assigned to the NEW allocation matters -- this has nothing to do with the preemption logic. Refactor this so that the instance type is handled outside the preemption code.

pkg/scheduler/objects/allocation.go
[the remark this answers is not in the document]
yzhangal: Thanks @craigcondit  yes I did. What I did was, I made sure all code paths that set an allocation to Allocated/AllocatedReserved/Replaced also set the bindTime.

pkg/scheduler/objects/node.go
yzhangal: Sure, will change in next rev.

pkg/scheduler/objects/preemption.go
yzhangal: Sorry I did not make it clear previously, 

the `tryNodes()` function below is called by 

```
func (p *Preemptor) TryPreemption() (*Allocation, bool) {
...
        // try to find a node to schedule on and victims to preempt
	nodeID, instType, victims, ok := p.tryNodes() // The nodeId, instType info returned here is used to construct
                                                                                     //  a new allocation below
	if !ok {
		// no preemption possible
		return nil, false
	}
...
	if p.headRoom.FitInMaxUndef(p.ask.GetAllocatedResource()) {
		log.Logger().Info("Reserving node for ask after preemption",
			zap.String("allocationKey", p.ask.GetAllocationKey()),
			zap.String("nodeID", nodeID),
			zap.Int("victimCount", len(victims)))
		return newReservedAllocation(Reserved, nodeID, instType, p.ask), true
	}

...

```

which in turn is called by

```
// tryAllocate will perform a regular allocation of a pending request, includes placeholders.
func (sa *Application) tryAllocate(headRoom *resources.Resource, preemptionDelay time.Duration, preemptAttemptsRemaining *int, nodeIterator func() NodeIterator, fullNodeIterator func() NodeIterator, getNodeFn func(string) *Node) *Allocation {

```
which returns Allocation object, which need the instance type info.

```
func (p *Preemptor) tryNodes() (string, string, []*Allocation, bool) {
	// calculate victim list for each node
	predicateChecks := make([]*si.PreemptionPredicatesArgs, 0)
	victimsByNode := make(map[string][]*Allocation)
	for nodeID, nodeAvailable := range p.nodeAvailableMap {
		allocations, ok := p.allocationsByNode[nodeID]
		if !ok {
			// no allocations present, but node may still be available for scheduling
			allocations = make([]*Allocation, 0)
		}
		// identify which victims and in which order should be tried
		if idx, victims := p.calculateVictimsByNode(nodeAvailable, allocations); victims != nil {
			victimsByNode[nodeID] = victims
			keys := make([]string, 0)
			for _, victim := range victims {
				keys = append(keys, victim.GetAllocationKey())
			}
			// only check this node if there are victims or we have not already tried scheduling
			if len(victims) > 0 || !p.nodesTried {
				predicateChecks = append(predicateChecks, &si.PreemptionPredicatesArgs{
					AllocationKey:         p.ask.GetAllocationKey(),
					NodeID:                nodeID,
					PreemptAllocationKeys: keys,
					StartIndex:            int32(idx),
				})
			}
		}
	}
	// call predicates to evaluate each node
	result := p.checkPreemptionPredicates(predicateChecks, victimsByNode)
	if result != nil && result.success {
		return result.nodeID, result.instType, result.victims, true
	}

	return "", UnknownInstanceType, nil, false
}
```

which calls

```
	// call predicates to evaluate each node
	result := p.checkPreemptionPredicates(predicateChecks, victimsByNode)
	if result != nil && result.success {
		return result.nodeID, result.instType, result.victims, true
	}
```

My understanding is that we choose an candidate node (after identifying preemption vicitims) for new allocation, so I need to make 

`         result := p.checkPreemptionPredicates(predicateChecks, victimsByNode)
`
populate the instanceType of the successfully chosen node candidate. 

Wonder if we can have a screen share session for us to walk through a bit. 

Thanks a lot.

craigcondit: The Preemptor stores mappings between node and various information needed for preemption. See initWorkingState() -- you can add a nodeInstanceTypeMap there and populate it. Then when a node is chosen for the allocation, look it up in that map. That's far simpler than what you have done here.

yzhangal: It was via a discussion with @wilfred-s @yangwwei that we decided to add an instanceType in Allocation object, because we don't have a global map between nodeId and instanceType.  We not only need instanceType for preemption code (which is new), but also need instanceType for all nodes.  I wish we have a global map, but we think it's more of a management hassle, so we decided to have the instType info included together with nodeId when creating Allocation object.

yzhangal: The down side is with my current implementation is that whenever we create an Allocation object, we need to have both the nodeId and instType information available.

craigcondit: We do not need the instance type for all nodes, only those that might be candidates. The code in initWorkingState() that can be updated:

```
allocationsByNode := make(map[string][]*Allocation)
queueByAlloc := make(map[string]*QueuePreemptionSnapshot)
nodeAvailableMap := make(map[string]*resources.Resource)
nodeInstanceTypeMap := make(map[string]string) // <-- TODO add this

// build a map from NodeID to allocation and from allocationID to queue capacities
for _, victims := range p.allocationsByQueue {
	for _, allocation := range victims.PotentialVictims {
		nodeID := allocation.GetNodeID()
		allocations, ok := allocationsByNode[nodeID]
		if !ok {
			allocations = make([]*Allocation, 0)
		}
		allocationsByNode[nodeID] = append(allocations, allocation)
		queueByAlloc[allocation.GetAllocationKey()] = victims
		nodeInstanceTypeMap[nodeID] = node.GetInstanceType() // <-- TODO add this
	}
}
```

yzhangal: Thanks, when I said "all nodes", I meant outside the preemption code, when we do an allocation, we need similar information. Doing a partial map for preemption code makes it a bit different than the handling outside the preemption code. If you think this is what we should do, I will give it a try.

Do you think it's worth to have a global nodeId to instanceType mapping? we discussed this before and chose to use the alternative solution that's currently in the PR.

Thanks.

craigcondit: I think the solution I suggested should be sufficient for now. A global mapping further complicates the context, and requires additional locking at multiple levels, which the node iterator prevents currently. This is something that could be revisited in the future, but that's where we are at today.

yzhangal: Thanks a lot @craigcondit just updated with new rev. In the code change you suggested, I had to do:

nodeInstanceTypeMap[nodeID] = allocation.GetInstanceType()

instead of
 
nodeInstanceTypeMap[nodeID] = node.GetInstanceType() 

Would you please take a look again when you get chance? thanks

pkg/scheduler/objects/preemption.go
```diff
@@ -505,7 +507,7 @@ func (p *Preemptor) tryNodes() (string, string, []*Allocation, bool) {
 	// call predicates to evaluate each node
 	result := p.checkPreemptionPredicates(predicateChecks, victimsByNode)
 	if result != nil && result.success {
-		return result.nodeID, result.instType, result.victims, true
+		return result.nodeID, p.nodeInstanceTypeMap[result.nodeID], result.victims, true
```
craigcondit, line 510: Always check the result of a map lookup, just in case. If it's not found, log a BUG message (see elsewhere for examples).

craigcondit: Use unknown if not found.

pkg/scheduler/objects/preemption.go
```diff
@@ -153,6 +155,7 @@ func (p *Preemptor) initWorkingState() {
 			}
 			allocationsByNode[nodeID] = append(allocations, allocation)
 			queueByAlloc[allocation.GetAllocationKey()] = victims
+			nodeInstanceTypeMap[nodeID] = allocation.GetInstanceType()
```
craigcondit, line 158: This *should* work in theory, however, it's probably best to check that allocation.GetInstanceType() was in fact populated. If not, ignore it and move on. We can use unknown if the map entry ends up not being populated.

craigcondit (changes requested):
A few more issues.

pkg/scheduler/objects/preemption.go
craigcondit: Added review comments - I think this can work, but we need to add some defensive checks in case the instance type is somehow not populated.

pkg/scheduler/objects/preemption.go
```diff
@@ -174,6 +177,20 @@ func (p *Preemptor) initWorkingState() {
 	p.allocationsByNode = allocationsByNode
 	p.queueByAlloc = queueByAlloc
 	p.nodeAvailableMap = nodeAvailableMap
+	p.nodeInstanceTypeMap = nodeInstanceTypeMap
+}
+
+// getInstanceType retrieves instance type from the nodeInstanceTypeMap for given nodeId
+// if not found, issue error log and return UnknownInstanceType
+func (p *Preemptor) getInstanceType(nodeId string) string {
```
craigcondit, line 185: This doesn't really need its own method.

pkg/scheduler/objects/preemption.go
craigcondit: We should only set the node value if the allocation was populated. If the instance type is empty, we might clobber a valid instance type from another allocation.

craigcondit (changes requested):
See review comments.

pkg/scheduler/objects/preemption.go
yzhangal: Many thanks @craigcondit good point, I just updated the PR with a few changes, including your above comment,  assigning the map

`p.nodeInstanceTypeMap = nodeInstanceTypeMap`

which I forgot in last rev, and using "UNKNOWN", 

Thanks again for taking further look!

pkg/scheduler/objects/preemption.go
yzhangal: Very much appreciated @craigcondit uploaded new rev again to address both latest points you made. Thanks

pkg/scheduler/objects/preemption.go
```diff
@@ -155,7 +155,10 @@ func (p *Preemptor) initWorkingState() {
 			}
 			allocationsByNode[nodeID] = append(allocations, allocation)
 			queueByAlloc[allocation.GetAllocationKey()] = victims
-			nodeInstanceTypeMap[nodeID] = allocation.GetInstanceType()
+			instType := allocation.GetInstanceType()
+			if instType != UnknownInstanceType {
```
craigcondit, line 159: Just to code defensively, I would skip the update if the allocation type is either unknown *or* empty. This should ensure we don't clobber the instance type. I know we're confident all the allocation instance types have been set, but there may be edge cases (recovery for instance).

craigcondit (changes requested):
One more minor change, and I think we're there.

pkg/scheduler/objects/preemption.go
yzhangal: Thanks a lot @craigcondit I'm working on a new rev.

yzhangal: Hi @craigcondit new rev uploaded.  Many thanks again!

craigcondit (approved):
+1 LGTM.

yzhangal:
Thank you so much @craigcondit @pbacsko @wilfred-s  @yangwwei @zhuqi-lucas and other community folks! I learned a lot working with you on adding this feature. Look forward to more collaboration moving forward!

PR #511, opened 2023-02-24
[YUNIKORN-1467] [WIP] Implement priority-based victim selection algorithm

craigcondit:
- Update to latest SI
- Implement getGuaranteedHeadRoom()
- Rename tryPreemption() -> tryRequiredNodePreemption()
- Update ask with both allowPreemptSelf / allowPreemptOther policies
- Remove preemption from partition conf
- Track last attempted preemption time on ask
- Walk preemption fence subtree to find potential victims
- Filter out allocations which don't help meet shortfall
- Filter out allocations which would reduce queue below guaranteed amounts
- Filter out nodes which have no allocations or cannot satisfy request
- Sort nodes by number of tasks which can be preempted
- Call shim to check predicates
- Limit number of preemption attempts to 10 requests per queue
- Revert YUNIKORN-1465 as we no longer need per-app preemption priorities
- Revert YUNIKORN-1466 as we no longer need per-queue preemption priorities

TODO:
- Unit testing (minimal unit tests present, not yet complete)
- E2E testing in shim (not started)
- Queue doesn't yet track resources for in-flight preemptions. This is necessary to avoid race conditions as we might preempt too much in a queue.
- Currently, preemption will only be triggered if there are no nodes capable of running a task. It will NOT be triggered in the case where only queue limits apply, as the initial check that a task is within limits prior to scheduling can't detect things like a sibling queue that is over guaranteed limits eating up a parent's max resources. Preemption should be allowed in this case, but the current node-centric algorithm doesn't handle this. We need some more design work on this scenario.

Issues:
- It appears that we get a reservation for a node even though that node is not a fit, resulting in the node being excluded from consideration due to the node iterator skipping reserved nodes.
- Configuration currently disallows parent guaranteed limits to be below the sum of child queue guarantees. Is this something we want to keep?

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1467

TODO 

### Questions:


pkg/scheduler/objects/application.go
```diff
@@ -961,11 +972,22 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+			if alloc := sa.tryNodes(request, iterator); alloc != nil {
+				// have a candidate return it
 				return alloc
 			}
+
+			// no nodes qualify, attempt preemption
+			if *preemptAttemptsRemaining > 0 {
+				*preemptAttemptsRemaining--
+				fullIterator := fullNodeIterator()
+				if fullIterator != nil {
```
pbacsko, line 984: When is `fullIterator == nil`? Is this a legit case? I don't see coverage in the unit tests.

pkg/scheduler/objects/application.go
```diff
@@ -903,8 +901,21 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
 		}
-		// resource must fit in headroom otherwise skip the request
+
+		// resource must fit in headroom otherwise skip the request (unless preemption could help)
 		if !headRoom.FitInMaxUndef(request.GetAllocatedResource()) {
+			// attempt preemption
+			if *preemptAttemptsRemaining > 0 {
+				*preemptAttemptsRemaining--
+				fullIterator := fullNodeIterator()
+				if fullIterator != nil {
```
pbacsko, line 911: When is `fullIterator == nil`? Is this a legit case? I don't see coverage in the unit tests.

pkg/scheduler/objects/preemption.go
```diff
@@ -0,0 +1,875 @@
...
+	Parent             *QueuePreemptionSnapshot // snapshot of parent queue
+	QueuePath          string                   // fully qualified path to queue
+	AllocatedResource  *resources.Resource      // allocated resources
+	PreemptingResource *resources.Resource      // resources currently flagged for preemption
+	MaxResource        *resources.Resource      // maximum resources for this queue
+	GuaranteedResource *resources.Resource      // guaranteed resources for this queue
+	PotentialVictims   []*Allocation            // list of allocations which could be preempted
+}
+
+// NewPreemptor creates a new preemptor. The preemptor itself is not thread safe, and assumes the application lock is held.
```
pbacsko, line 73: Nit: we already have `RequiredNodePreemptor`, could we call this `GenericPreemptor` to emphasize the difference?

pkg/metrics/queue.go
```diff
@@ -131,3 +131,11 @@ func (m *QueueMetrics) SetQueuePendingResourceMetrics(resourceName string, value
 func (m *QueueMetrics) AddQueuePendingResourceMetrics(resourceName string, value float64) {
 	m.ResourceMetrics.With(prometheus.Labels{"state": "pending", "resource": resourceName}).Add(value)
 }
+
+func (m *QueueMetrics) SetQueuePreemptingResourceMetrics(resourceName string, value float64) {
+	m.ResourceMetrics.With(prometheus.Labels{"state": "preempting", "resource": resourceName}).Set(value)
+}
+
+func (m *QueueMetrics) AddQueuePreemptingResourceMetrics(resourceName string, value float64) {
```
pbacsko, line 139: This method doesn't seem to be called from anywhere

pkg/scheduler/objects/preemption.go
```diff
@@ -0,0 +1,875 @@
...
+			wg.Wait()
+			close(ch)
+		}()
+		for result := range ch {
+			// if result is successful, keep track of it
+			if result.success {
+				if bestResult == nil {
+					bestResult = result
+				} else if result.betterThan(bestResult, p.allocationsByNode) {
+					bestResult = result
```
pbacsko, line 383: codecov says this branch is not covered, is this relevant to cover?

pbacsko (changes requested):
OK, I performed a first round of review, great work in general!

I identified two things that need to be addressed:
1) Feature toggle for preemption, I can't find it anywhere, I think it's very important to have.
2) Missing tests in `partition_test.go`, those are crucial to have to see how the preemption works accross multiple scheduling cycles.

I added some minor comments and questions as well.

pkg/scheduler/objects/application.go
craigcondit: The node iterator will return nil if there are no nodes to iterate. This is simply defensive coding to guard against that case. DIdn't seem worthy of a unit test.

pkg/scheduler/objects/application.go
craigcondit: Same as above.

pkg/scheduler/objects/preemption.go
craigcondit: I'm hesitant to do this. In fact, there was a point during the development cycle where I considered folding the required node preemptor back into this one (making it the singular preemptor). I'm not generally a fan of longer names when shorter ones will do. The required node preemptor is of course the odd one out now, so the longer name makes sense.

pkg/scheduler/objects/preemption.go
```diff
@@ -0,0 +1,875 @@
...
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
+var (
+	preemptAttemptFrequency        = 15 * time.Second
+	preemptCheckConcurrency        = 10
+	scoreFitMax             uint64 = 1 << 32
+	scoreOriginator         uint64 = 1 << 33
+	scoreNoPreempt          uint64 = 1 << 34
+	scoreUnfit              uint64 = 1 << 35
```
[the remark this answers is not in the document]
craigcondit: In go, consts are untyped, and we need these to be uint64 to work properly and avoid unnecessary casts everywhere. As for 1<<32, this leaves a full 32-bits for the priority value to slot into. So things get scored by whether they have opted out of preemption, whether they are the app originator, and finally by priority.

pkg/metrics/queue.go
craigcondit: Correct, it was added for parity with the other metrics methods. It seemed appropriate to match the existing patterns.

pkg/scheduler/objects/application.go
```diff
@@ -903,8 +901,21 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, nodeIterator fu
...
+					if alloc, ok := sa.tryPreemption(headRoom, preemptionDelay, request, fullIterator, false); ok {
```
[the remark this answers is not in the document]
craigcondit: There are actually several ways (and why there is no global flag). First, preemption will never occur if guaranteed resources are not set. That is typically not done, so most existing users won't see preemption occur. Second, making each queue a preemption fence effectively disables preemption, as does setting guaranteed == max (or unset). Preemption can also be disabled on any queue by setting the policy to disabled.

pkg/scheduler/objects/queue.go
```diff
@@ -41,29 +40,32 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 )
 
+var (
+	maxPreemptionsPerQueue = 10 // maximum number of asks to attempt to preempt for in a single queue
```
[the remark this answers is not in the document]
craigcondit: I made it a var because it's possible we may configure it later. I didn't want to do so right away, as I think this will work fine as a default until proven otherwise.

pkg/scheduler/partition_test.go
```diff
@@ -123,7 +123,6 @@ func TestNewWithPlacement(t *testing.T) {
 			},
```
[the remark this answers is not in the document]
craigcondit: I'll look into this, but the further up the chain we go, the more difficult the setup gets as more of the environment needs to be mocked. This gets even more difficult in the case of preemption due to the predicate checks. It may not be a worthwhile effort.

pkg/scheduler/objects/preemption.go
craigcondit: Added additional test in latest commit.

pkg/scheduler/objects/queue.go
```diff
@@ -41,29 +40,32 @@ import (
...
+	sortType            policies.SortPolicy       // How applications (leaf) or queues (parents) are sorted
+	children            map[string]*Queue         // Only for direct children, parent queue only
+	childPriorities     map[string]int32          // cached priorities for child queues
+	applications        map[string]*Application   // only for leaf queue
+	appPriorities       map[string]int32          // cached priorities for application
+	reservedApps        map[string]int            // applications reserved within this queue, with reservation count
+	parent              *Queue                    // link back to the parent in the scheduler
+	pending             *resources.Resource       // pending resource for the apps in the queue
+	allocatedResource   *resources.Resource       // allocated resource for the apps in the queue
+	preemptingResource  *resources.Resource       // preempting resource for the apps in the queue
```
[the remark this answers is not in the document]
craigcondit: Added preemption-centric test with checks on preempting resources.

manirajv06 (commented):
Great work @craigcondit. 

While going through these big chunk of changes, felt that there is a lot of room for improvement from code simplification perspective. Since effort is big, it is not possible to expect it out from the first cut itself, but I do expect that it will happen over the future iterations. 

A) For example, findEligiblePreemptionVictims, calculateAdditionalVictims and tryNodes methods can be refactored to simplify it further to 

1. Possibly finalising the victims as early as possible rather than taking some of the decisions at later stages and rejecting it later.
2. to avoid redundant checks (seems like? but need to go in detail to confirm this), etc

thereby increase the overall efficiency of preemption cycle.

B) Is there a way to avoid doing preemptPredicateCheck for all batches? Can we avoid it doing for rest of the batches if we are done with the current batch being processed?

So, things like this..all these can taken up later as future enhancements if it makes sense.

Thanks @craigcondit  for this big effort.

pkg/scheduler/objects/preemption.go
```diff
@@ -0,0 +1,875 @@
...
+	batches := batchPreemptionChecks(predicateChecks, preemptCheckConcurrency)
+	var bestResult *predicateCheckResult = nil
+	for _, batch := range batches {
+		var wg sync.WaitGroup
+		ch := make(chan *predicateCheckResult, len(batch))
+		expected := 0
+		for _, args := range batch {
+			// add goroutine for checking preemption
+			wg.Add(1)
+			expected++
```
manirajv06, line 369: Are we using this?

pbacsko (approved):
+1 from my part - pending on Mani's questions

pkg/scheduler/objects/preemption.go
craigcondit: Yes this is used.

craigcondit:
>
> While going through these big chunk of changes, felt that there is a lot of room for improvement from code simplification perspective. Since effort is big, it is not possible to expect it out from the first cut itself, but I do expect that it will happen over the future iterations. 

The code is optimized for efficiency as we take great pains to avoid expensive computations until later in the cycle. This code is very much on the hot path of the scheduler so we make every attempt to abort early if preemption cannot be done. Combining the methods you cited would require more work to be done upfront and very likely make the entire process less performant -- remember that the normal case is that we do NOT find a preemption solution. Preemption itself is rare and so we want to short-circuit out as soon as we can, even at the expense of a preemption taking longer. 

> For example, findEligiblePreemptionVictims, calculateAdditionalVictims and tryNodes methods can be refactored to simplify it further to possibly finalizing the victims as early as possible rather than taking some of the decisions at later stages and rejecting it later?

This is by design as we cannot know until predicate checks are run which solutions (if any) are viable. We gather all the possible victims in each node first as we might need to check more pods than expected due to predicate constraints. 

> To avoid redundant checks (seems like? but need to go in detail to confirm this), etc thereby increase the overall efficiency of preemption cycle.

The checks may seem redundant but this is to give us every opportunity to short-circuit out early. They are structured to avoid doing expensive computations early in the process. 

> Is there a way to avoid doing preemptPredicateCheck for all batches? Can we avoid it doing for rest of the batches if we are done with the current batch being processed?

We already do this. Once a batch is processed we evaluate the solutions found (if any) and stop if a preemption solution that is "good enough" is discovered. We only continue if there were no optimal solutions found. 

Additionally to avoid repeated expensive computations, we only ever try to preempt for a single pod once every 15 seconds, and only check a maximum of 10 pods in any one queue.

PR #513, opened 2023-02-27
[YUNIKORN-1603] Configuration processing to support limit new wildcard interpretation syntax checking

zhuqi-lucas:
### What is this PR for?
Wildcard interpretation
Wildcard can only be used as the only entry in the limit object according to the documentation. There can be multiple limit objects in the overall limits for a queue. As part of this design the interpretation of the wildcard needs to be defined. Wildcards interpretation is related to the ordering of the limit objects in the overall limits object. The usage and interpretation will be different for users and groups. 

In general terms: allowing a wildcard in the user or group list only as part of the last entry of the limits list. After the wildcard has been added for either the user or group list we do not allow a user or group list with a non wildcard entry. This is especially important for the group resolution. It does allow specifying a user and group wildcard with different limits set. In all cases, there will only be a match using the wildcard if none of the earlier limit entries match.

As part of the already existing queue configuration processing the syntax for the configuration is checked. This check is triggered also for a reload of the file, via the config map. Rudimentary limit object checking is implemented as part of the config check.

 

The above described changes around the wildcards and what is allowed is not part of the configuration validation and must be added. 

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse//YUNIKORN-1603

### Questions:

### DIFF
--- a/pkg/common/configs/config_test.go
+++ b/pkg/common/configs/config_test.go
@@ -1175,6 +1175,130 @@ partitions:
 	if err == nil {
 		t.Errorf("limit parsing should have failed group @: %v", conf)
 	}
+
+	data = `
+partitions:
+  - name: default
+    limits:
+      - limit: dot user
+        users:
+        - user.lastname
+        maxapplications: 1
+      - limit: "@ user"
+        users:
+        - user@domain
+        maxapplications: 1
+      - limit: wildcard user
+        users:
+        - "*"
+        maxapplications: 1
+      - limit: wildcard group
+        groups:
+        - "*"
+        maxapplications: 1
+      - limit: error no wildcard user after wildcard user
+        users:
+        - "user1"
+        maxapplications: 1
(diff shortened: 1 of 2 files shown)


0yukali0 (approved):
LGTM.

0yukali0:
BTW, I think that we need to add an issue to describe writing style of placement rules.
The user and group limit object using wildcard must be the last two in the limits object.

zhuqi-lucas:
> 

Thank you @0yukali0 for good suggestion , i agree with you that we also need corresponding doc changes for this! cc @wilfred-s If you have any other suggestions for this PR, thanks!

PR #515, opened 2023-03-06
[YUNIKORN-1606] Configuration processing should support limit quota check at the same level.

zhuqi-lucas:
### What is this PR for?
No checks are performed for the limits against the queue quotas. This means that as part of the current checks a limit can be set that is higher than the queue would allow. There is also no hierarchical check in place. This means that a child could set higher limits than is possible based on the parent settings.

 

The configuration validation processing must be extended to the limit objects. At each level in the configuration the resource quota assigned in a limit must be smaller than the maximum allowed at that level for the queue. The queue quota processing already has this builtin and the limit should be checked against the quota.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:


pkg/common/configs/configvalidator.go
```diff
@@ -262,13 +285,32 @@ func checkLimit(limit Limit) error {
...
+
+	if queue.MaxApplications != 0 && (limit.MaxApplications > queue.MaxApplications || limit.MaxApplications == 0) {
+		return fmt.Errorf("invalid MaxApplications settings for limit user names '%s' groups %s exeecd current the queue MaxApplications", limit.Users, limit.Groups)
+	}
+
+	// If queue is RootQueue, the queue.Resources.Max will be null, we don't need to check for root queue
+	// But we may need to check the root resource during loading the config and after partition resource loading when update node
+	if queue.Name != RootQueue {
+		queueMaxResource, err := resources.NewResourceFromConf(queue.Resources.Max)
+		if err != nil {
```
0yukali0, line 299: Hi @zhuqi-lucas. 
I think we can add a unit test  to improve the coverage.

zhuqi-lucas: Thank you @0yukali0 for review, i will add a unit test for this!

zhuqi-lucas:
Hi @0yukali0 , i have improved unit test to increase the test coverage now, thanks a lot!

pkg/common/configs/configvalidator.go
```diff
@@ -357,13 +357,32 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
...
+			log.Logger().Debug("resource parsing failed",
+				zap.Error(err))
+			return fmt.Errorf("parse queue max resource failed: %s", err.Error())
```
manirajv06, line 374: Can we have queue name in this error message?

manirajv06 (changes requested):
looks good. Minor nit.

pkg/common/configs/configvalidator.go
zhuqi-lucas: Sure @manirajv06 , thank you for review, addressed in latest PR.

pkg/common/configs/configvalidator.go
```diff
@@ -357,13 +357,32 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
 	}
 	// at least some resource should be not null
 	if limit.MaxApplications == 0 && resources.IsZero(limitResource) {
-		return fmt.Errorf("invalid resource combination for limit names '%s' all resource limits are null", limit.Users)
+		return fmt.Errorf("invalid resource combination for limit user names '%s' groups %s all resource limits are null", limit.Users, limit.Groups)
```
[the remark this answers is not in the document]
zhuqi-lucas: Good suggestion @mitdesai , thank you review!

pkg/common/configs/configvalidator.go
```diff
@@ -357,13 +357,32 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
...
 	}
+
+	if queue.MaxApplications != 0 && (limit.MaxApplications > queue.MaxApplications || limit.MaxApplications == 0) {
+		return fmt.Errorf("invalid MaxApplications settings for limit user names '%s' groups %s exeecd current the queue MaxApplications", limit.Users, limit.Groups)
```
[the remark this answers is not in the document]
zhuqi-lucas: Done

pkg/common/configs/configvalidator.go
```diff
@@ -357,13 +357,32 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
...
+	// But we may need to check the root resource during loading the config and after partition resource loading when update node
+	if queue.Name != RootQueue {
+		queueMaxResource, err := resources.NewResourceFromConf(queue.Resources.Max)
+		if err != nil {
+			log.Logger().Debug("resource parsing failed",
+				zap.Error(err))
+			return fmt.Errorf("parse queue max resource failed: %s", err.Error())
+		}
+		if !queueMaxResource.FitInMaxUndef(limitResource) {
+			return fmt.Errorf("invalid MaxResources settings for limit user names '%s'  groups %s exeecd current the queue MaxResources", limit.Users, limit.Groups)
```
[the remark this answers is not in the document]
zhuqi-lucas: Done!

mitdesai:
Thanks for taking care of that @zhuqi-lucas. Looks good to me otherwise.

Thanks
Mit

pkg/common/configs/configvalidator.go
```diff
@@ -357,13 +357,32 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
 	}
 	// at least some resource should be not null
 	if limit.MaxApplications == 0 && resources.IsZero(limitResource) {
-		return fmt.Errorf("invalid resource combination for limit names '%s' all resource limits are null", limit.Users)
+		return fmt.Errorf("invalid resource combination for limit %s all resource limits are null", limit.Limit)
 	}
+
+	if queue.MaxApplications != 0 && (limit.MaxApplications > queue.MaxApplications || limit.MaxApplications == 0) {
+		return fmt.Errorf("invalid MaxApplications settings for limit %s exeecd current the queue MaxApplications", limit.Limit)
```
0yukali0, line 364: In the website, the limits description is following.
```
maxapplications is an unsigned integer value, larger than 1, which allows you to limit the number of running applications for the configured user or group. Specifying a zero maximum applications limit is not allowed as it would implicitly deny access. Denying access must be handled via the ACL entries.
```

so I think that the error description is not suitable when conditions (queue.MaxApplications != 0 && limit.MaxApplications == 0) happens.
I think this part will be like this.

if queue.MaxApplications != 0 {
  if limit.MaxApplications > queue.MaxApplications  {
    return fmt.Errorf("invalid MaxApplications settings for limit %s exeecd current the queue MaxApplications", limit.Limit)
  }
  if limit.MaxApplications == 0 {
    return fmt.Errorf("MaxApplications is 0 in limit name %s, it should be 1 ~ %d", limit.Limit, queue.MaxApplications)
  }
}

0yukali0 (changes requested):
Almost finished.

pkg/common/configs/configvalidator.go
zhuqi-lucas: Hi @0yukali0 , thank you for review, but if we don't set the MaxApplications, it will have a default value, it will be 0. 
If we want to change to init the MaxApplications to an unlimited value rather than 0?

0yukali0: No,  the default value of MaxApplications in the limit is 0.
The message hints that admin should set the MaxApplications in the limit when the MaxApplications in the queue is set.
Because users will not know that setting the MaxApplications in limit is necessary after setting the MaxApplications in the queue.

zhuqi-lucas: Good catch @0yukali0 , it makes sense to me now!

zhuqi-lucas: Addressed in latest PR now.

0yukali0 (approved):
LGTM

pkg/common/configs/configvalidator.go
0yukali0: Hi @zhuqi-lucas, I will merge this to the master.
Thanks for  the contribution.

PR #516, opened 2023-03-10
[YUNIKORN-1568] Placement rules should be validated during config update

pbacsko:
### What is this PR for?
Validate the placement rules on a best effort basis to see whether it refers to queues which are not parents or not available.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1568

### Questions:

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -32,6 +32,7 @@ import (
 	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/apache/yunikorn-core/pkg/common/security"
 	"github.com/apache/yunikorn-core/pkg/log"
+	"github.com/apache/yunikorn-core/pkg/scheduler/placement/types"
 	"github.com/apache/yunikorn-core/pkg/scheduler/policies"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -169,9 +185,65 @@ func checkPlacementRules(partition *PartitionConfig) error {
...
+	if len(conf) == 0 {
+		if !create {
+			return nonExistingQueue
+		}
+		return checkOK
+	}
+
+	var queueConf *QueueConfig
+	for _, queue := range conf {
+		queue := queue
```
0yukali0, line 220: Duplicate assignment.
variable "queue" already get the queue from the QueueConfig in each loop.

0yukali0 (commented):
I think that adding some unit tests shows the expected results and improves the coverage.
1. adding illegal examples in the checkACL unit tests returns the queueNotParent message or the nonExistingQueue message.
2. adding unit test of getLongestStaticPath
3. checkQueueHierarchyForPlacement

pkg/common/configs/configvalidator.go
pbacsko: It's to avoid gosec linter error, it says "implicit memory aliasing in for loop" G601. This seems to be an idiomatic solution.

0yukali0: Understood.

0yukali0 (approved):
LGTM

PR #517, opened 2023-03-21
[YUNIKORN-1645] Add base type to UserTracker and GroupTracker to reduce code duplications

pbacsko:
### What is this PR for?
`UserTracker` and `GroupTracker` share some code in common. Let's add a base type which reduces code duplication. This also helps YUNIKORN-1566 which will add another method for moving a running application.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1645

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/base_tracker.go
+++ b/pkg/scheduler/ugm/base_tracker.go
@@ -1,0 +1,44 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


wilfred-s:
I am not sure about this. It reduces duplication of code at the cost of: 
- ease of debugging, the inlined function is more difficult to track.
- extra code lines, even after I remove the license header the new solution uses 10 more lines of code,
- overhead, the two extra function calls add extra work at runtime.

It might also need an unwind as soon as we start looking at enforcing the quotas set and not just do the tracking in [YUNIKORN-1573](https://issues.apache.org/jira/browse/YUNIKORN-1573) That requires more work to be done under the lock which would be different for the group and user tracker

pbacsko:
Closing this for now. We can reopen it later.

PR #520, opened 2023-03-22
[YUNIKORN-1478] Move to go v1.20

wilfred-s:
### What is this PR for?
Update to latest go release

### What type of PR is it?
* [X] - Feature

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1478

standard build will run with new go and linter


Makefile
```diff
@@ -171,5 +171,5 @@ fsm_graph: clean
 .PHONY: clean
 clean:
 	@echo "cleaning up caches and output"
-	go clean -cache -testcache -r -x ./... 2>&1 >/dev/null
+	go clean -cache -testcache -r
```
pbacsko, line 174: We might want to keep the `/dev/null` redirect, this command can generate a lot of output which doesn't seem to be useful.

craigcondit (approved):
+1 with restoration of go clean output > /dev/null. Will update during merge.

PR #521, opened 2023-03-23
[YUNIKORN-1559] Fix gang scheduling for missing resource from queue

elihschiff:
### What is this PR for?

In YUNIKORN-1559 I am trying to add a `pods=1` resource to all pods during scheduling. During gang scheduling, for resources that do not have a value set explicitly on the queue, yunikorn fails to schedule them correctly.

Imagine a queue with vcore set as 100 and memory set as 100Gi. Now imagine a cluster where there are more than enough running nodes to fit pods up to the queue limit. If we try and schedule 10 pods with vcore=10, memory=10Gi, and pods=1. While the queue does not explicitly have a `pods` resource limit set, yunikorn uses this code to ignore that and schedule all 10 based on the fact that the vcore/memory are below the limit. https://github.com/apache/yunikorn-core/blob/1e7d249631bab0f91438d87d3d9efb40a3e2a364/pkg/common/resources/resources.go#L435-L438

However, if there are not enough nodes yet, we see the problem this PR is trying to address. Yunikorn does take into account the `pods` value when determining if a pod should be set to tell the cluster autoscaler to scale up. This means that pods will never scale up new nodes.

One other note, this issue can be seen without the need for a `pods` resource. For example, create a queue with just `vcore` set and no value for `memory` set. Then try and schedule more resource that nodes that exist in a cluster, but fewer resources that the vcore limit on that queue. Those pods will not scale up new resources (unless you apply this PR's change first)

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1559

I attempted to test this in my k8s cluster along with my yunikorn-k8shims change https://github.com/apache/yunikorn-k8shim/pull/558, that said I dont know if there are potential other issues I dont know about. I am still a beginner with yunikorn's codebase and I very likely could have missed something important.

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -166,12 +166,18 @@ func (r *Resource) AddTo(add *Resource) {
 // Should be used by temporary computation only
 // A nil base resource does not change
 // A nil passed in resource is treated as a zero valued resource and leaves the base unchanged.
-func (r *Resource) SubFrom(sub *Resource) {
+func (r *Resource) SubFrom(sub *Resource, skipUndef bool) {
 	if r != nil {
 		if sub == nil {
 			return
 		}
 		for k, v := range sub.Resources {
+			_, ok := r.Resources[k]
+			// skip if not defined (queue quota checks: undefined resources are considered max)
+			if skipUndef && !ok {
+				continue
+			}
+
 			r.Resources[k] = subVal(r.Resources[k], v)
 		}
 	}
--- a/pkg/common/resources/resources_test.go
+++ b/pkg/common/resources/resources_test.go
@@ -889,7 +889,6 @@ func TestSub(t *testing.T) {
 	}
 }
 
-// TODO test `true`
 func TestSubFromNil(t *testing.T) {
 	// make sure we're nil safe IDE will complain about the non nil check
 	defer func() {
--- a/pkg/common/resources/resources_test.go
+++ b/pkg/common/resources/resources_test.go
@@ -889,6 +889,7 @@ func TestSub(t *testing.T) {
 	}
 }
 
+// TODO test `true`
 func TestSubFromNil(t *testing.T) {
 	// make sure we're nil safe IDE will complain about the non nil check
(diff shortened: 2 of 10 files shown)


pbacsko:
Added @craigcondit and @wilfred-s as reviewers.

craigcondit:
There are two phases to scheduling. In the first phase, we only look at whether or not an ask fits into a queue -- in this case, the existing logic should be correct, as if a queue doesn't specify a limit, it is in effect unlimited. This goes for any resource type. The second phase deals with placement on nodes. Once pods:1 is added to each request, this should get handled properly by the node placement logic (as resources will not fit in that case).

YuniKorn will only mark a pod as unschedulable (and therefore eligible for autoscaling) if it would fit in a queue, but NOT on any available node. TL;DR: I don't think this is the correct fix, as the existing logic should already do the right thing.

@wilfred-s do I have this right?

wilfred-s:
As @craigcondit mentioned the normal allocation cycle properly handles any type of resource. If you set a limit on pods in the queue pods get limited after the change in apache/yunikorn-k8shim#558 
The scale up issue is limited to the way we calculate the outstanding requests that fit in the headroom. I have been able to reproduce the issue in tests by simply adding pods that request a resource type that is not limited. That causes the headroom to go negative for certain resource types and nothing fits.
I opened a [new bug](https://issues.apache.org/jira/browse/YUNIKORN-1658) to fix that side of the calculations and ignore unlimited resources as we should. Similar to the way we ignore them when we look at the fit on a node or in a queue. The k8shim side change for BestEffort pods should be tested on top of that change.

This PR contains changes that could break current accounting of resources. The only points that can ignore undefined resources are in the `queue.internalHeadRoom()` and `application. getOutstandingRequests` as they look at quotas which are infinite if undefined

PR #522, opened 2023-03-24
[YUNIKORN-1641] Queue tracker object is not cleaned up when a queue is deleted

zhuqi-lucas:
### What is this PR for?
Currently, every queue has its own QueueTracker object which is managed by ugm.Manager.

But when the queue is deleted either by a config change or automatically by the partition manager, the tracker object stays in memory, causing a memory leak.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1641

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/group_tracker.go
+++ b/pkg/scheduler/ugm/group_tracker.go
@@ -50,7 +50,7 @@ func (gt *GroupTracker) increaseTrackedResource(queuePath, applicationID string,
 	return gt.queueTracker.increaseTrackedResource(queuePath, applicationID, usage)
 }
 
-func (gt *GroupTracker) decreaseTrackedResource(queuePath, applicationID string, usage *resources.Resource, removeApp bool) error {
+func (gt *GroupTracker) decreaseTrackedResource(queuePath, applicationID string, usage *resources.Resource, removeApp bool) (bool, error) {
 	gt.Lock()
 	defer gt.Unlock()
 	if removeApp {
--- a/pkg/scheduler/ugm/group_tracker_test.go
+++ b/pkg/scheduler/ugm/group_tracker_test.go
@@ -132,13 +132,23 @@ func TestGTDecreaseTrackedResource(t *testing.T) {
 	assert.Equal(t, "map[mem:60000000 vcore:60000]", actualResources1["root.parent.child1"].String(), "wrong resource")
 	assert.Equal(t, "map[mem:10000000 vcore:10000]", actualResources1["root.parent.child2"].String(), "wrong resource")
 
-	err = groupTracker.decreaseTrackedResource(queuePath1, TestApp1, usage1, true)
+	usage4, err := resources.NewResourceFromConf(map[string]string{"mem": "60M", "vcore": "60"})
+	if err != nil {
+		t.Errorf("new resource create returned error or wrong resource: error %t, res %v", err, usage3)
+	}
+
+	err = groupTracker.decreaseTrackedResource(queuePath1, TestApp1, usage4, true)
 	if err != nil {
 		t.Fatalf("unable to decrease tracked resource: queuepath %s, app %s, res %v, error %t", queuePath1, TestApp1, usage1, err)
 	}
 	assert.Equal(t, 1, len(groupTracker.getTrackedApplications()))
 
-	err = groupTracker.decreaseTrackedResource(queuePath2, TestApp2, usage2, true)
+	usage5, err := resources.NewResourceFromConf(map[string]string{"mem": "10M", "vcore": "10"})
+	if err != nil {
+		t.Errorf("new resource create returned error or wrong resource: error %t, res %v", err, usage3)
+	}
+
+	err = groupTracker.decreaseTrackedResource(queuePath2, TestApp2, usage5, true)
 	if err != nil {
 		t.Fatalf("unable to decrease tracked resource: queuepath %s, app %s, res %v, error %t", queuePath2, TestApp2, usage2, err)
 	}
--- a/pkg/scheduler/ugm/group_tracker_test.go
+++ b/pkg/scheduler/ugm/group_tracker_test.go
@@ -117,14 +117,18 @@ func TestGTDecreaseTrackedResource(t *testing.T) {
 	if err != nil {
 		t.Errorf("new resource create returned error or wrong resource: error %t, res %v", err, usage3)
 	}
-	err = groupTracker.decreaseTrackedResource(queuePath1, TestApp1, usage3, false)
+	removeQT, err := groupTracker.decreaseTrackedResource(queuePath1, TestApp1, usage3, false)
 	if err != nil {
 		t.Fatalf("unable to decrease tracked resource: queuepath %s, app %s, res %v, error %t", queuePath1, TestApp1, usage3, err)
 	}
-	err = groupTracker.decreaseTrackedResource(queuePath2, TestApp2, usage3, false)
+	assert.Equal(t, removeQT, false, "wrong remove queue tracker value")
+
+	removeQT, err = groupTracker.decreaseTrackedResource(queuePath2, TestApp2, usage3, false)
 	if err != nil {
 		t.Fatalf("unable to decrease tracked resource: queuepath %s, app %s, res %v, error %t", queuePath2, TestApp2, usage3, err)
 	}
+	assert.Equal(t, removeQT, false, "wrong remove queue tracker value")
+
 	actualResources1 := getGroupResource(groupTracker)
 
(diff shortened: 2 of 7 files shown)


pkg/scheduler/ugm/group_tracker.go
```diff
@@ -78,3 +78,7 @@ func (gt *GroupTracker) GetGroupResourceUsageDAOInfo() *dao.GroupResourceUsageDA
 	groupResourceUsage.Queues = gt.queueTracker.getResourceUsageDAOInfo("")
 	return groupResourceUsage
 }
+
+func (gt *GroupTracker) CleanDeletedQueueTracker(queuePath string) {
```
pbacsko, line 82: Just call it simply `DeleteQueueTracker()`.

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -151,3 +151,23 @@ func (qt *QueueTracker) getResourceUsageDAOInfo(parentQueuePath string) *dao.Res
...
+	idx := strings.Index(queuePath, configs.DOT)
+	childQueuePath := ""
+	if idx != -1 {
+		childQueuePath = queuePath[idx+1:]
+	}
+	childIndex := strings.Index(childQueuePath, configs.DOT)
+	immediateChildQueueName := childQueuePath
+	if childIndex != -1 {
+		immediateChildQueueName = childQueuePath[:childIndex]
+	}
```
pbacsko, line 165: I just added a function `getChildQueuePath()` which returns two string, essentially this part can be replaced with that call.

pkg/scheduler/ugm/queue_tracker_test.go
```diff
@@ -155,3 +155,52 @@ func getQTResource(qt *QueueTracker) map[string]*resources.Resource {
 	usage := qt.getResourceUsageDAOInfo("")
 	return internalGetResource(usage, resources)
 }
+
+func TestQTCleanDeletedQueueTracker(t *testing.T) {
```
pbacsko, line 159: `TestQTDeleteQueueTracker()`

pkg/scheduler/ugm/user_tracker.go
```diff
@@ -97,3 +97,7 @@ func (ut *UserTracker) GetUserResourceUsageDAOInfo() *dao.UserResourceUsageDAOIn
 	userResourceUsage.Queues = ut.queueTracker.getResourceUsageDAOInfo("")
 	return userResourceUsage
 }
+
+func (ut *UserTracker) CleanDeletedQueueTracker(queuePath string) {
```
pbacsko, line 101: Same here, `DeleteQueueTracker()`

pkg/scheduler/partition_manager.go
```diff
@@ -120,6 +121,12 @@ func (manager *partitionManager) cleanQueues(queue *objects.Queue) {
 				log.Logger().Debug("unexpected failure removing the queue",
 					zap.String("partitionName", manager.pc.Name),
 					zap.String("queue", queue.QueuePath))
+			} else {
+				// we should clean the queue tracker for ths deleted queue
+				log.Logger().Debug("removing queueTracker",
+					zap.String("queueName", queue.QueuePath),
+					zap.String("partitionName", manager.pc.Name))
+				ugm.GetUserManager().CleanDeletedQueueTrackers(queue.QueuePath)
```
pbacsko, line 129: Minor simplification:
```
		if queue.IsEmpty() {
			// all OK update the queue hierarchy and partition
			if !queue.RemoveQueue() {
				log.Logger().Debug("unexpected failure removing the queue",
					zap.String("partitionName", manager.pc.Name),
					zap.String("queue", queue.QueuePath))
                                return
			}
			log.Logger().Debug("removing queueTracker",
				zap.String("queueName", queue.QueuePath),
				zap.String("partitionName", manager.pc.Name))
                        ugm.GetUserManager().CleanDeletedQueueTrackers(queue.QueuePath)
		} ...
```

pbacsko (changes requested):
Added some comments.

pkg/scheduler/ugm/group_tracker.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

pkg/scheduler/ugm/manager.go
```diff
@@ -304,6 +304,17 @@ func (m *Manager) cleaner() {
 	}
 }
 
+func (m *Manager) CleanDeletedQueueTrackers(queuePath string) {
```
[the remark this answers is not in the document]
zhuqi-lucas: Good suggestion, addressed in latest PR.

pkg/scheduler/ugm/queue_tracker.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

pkg/scheduler/ugm/queue_tracker_test.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

pkg/scheduler/ugm/user_tracker.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

pkg/scheduler/partition_manager.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

zhuqi-lucas:
> Added some comments.

Thank you @pbacsko for review and good suggestions, addressed in latest PR.

wilfred-s:
See the comments in the jira I would prefer a far lighter approach based on zero tracked usage in the `decreaseTrackedResource`.

At this point a -1 on the approach

zhuqi-lucas:
> 

Thank you @wilfred-s , i will address the comments, and submit a new PR.

pkg/scheduler/ugm/group_tracker_test.go
```diff
@@ -132,13 +132,23 @@ func TestGTDecreaseTrackedResource(t *testing.T) {
...
 	}
 	assert.Equal(t, 1, len(groupTracker.getTrackedApplications()))
 
-	err = groupTracker.decreaseTrackedResource(queuePath2, TestApp2, usage2, true)
+	usage5, err := resources.NewResourceFromConf(map[string]string{"mem": "10M", "vcore": "10"})
+	if err != nil {
+		t.Errorf("new resource create returned error or wrong resource: error %t, res %v", err, usage3)
+	}
+
+	err = groupTracker.decreaseTrackedResource(queuePath2, TestApp2, usage5, true)
```
pbacsko, line 151: We need tests for:
1) Leaf reaches zero, check that it gets removed
2) Set resources of parent to zero while still having leaf trackers to see that Error object is returned

pbacsko (changes requested):
Please add more tests

pkg/scheduler/ugm/group_tracker_test.go
zhuqi-lucas: Thank you @pbacsko for review!
Add more unit tests now:
1. Has been covered in queuetrackertest: 
       // Make sure childQueueTracker cleaned
	assert.Equal(t, len(queueTracker.childQueueTrackers["parent"].childQueueTrackers), 1)
	
2. Added in latest PR.

zhuqi-lucas:
> Please add more tests

Added now, thank you @pbacsko !

wilfred-s (changes requested):
We do not want walk over all child trackers again when we clean up. The child knows if it can be removed when it returns from the decrease call. 
Simplification:
* change the signature of decreaseTrackedResource to return extra boolean (3 signatures)
* boolean is the removal trigger (i.e. removeQT)
* boolean value is defined on the queue tracker as
```
 len(qt.childQueueTrackers) == 0 && len(qt.runningApplications) == 0  && resources.IsZero(tracker.resourceUsage)
```
* if removeQT is true remove the child from the list
* after that action calculate for the current queue and return.

The user and group removal can also be triggered on this same remove flag as the root queue removal should be seen as a sign to remove the user or group. Replacing the current `isUserRemovable` and `isGroupRemovable` checks in the manager.

zhuqi-lucas:
Thank you @wilfred-s for review, i have addressed the comments in latest PR.

wilfred-s (approved):
thanks for sticking with me,
approved, LGTM

PR #523, opened 2023-03-28
[YUNIKORN-1658] autoscaling support with unlimited resources

wilfred-s:
### What is this PR for?
* internalHeadRoom of the queue calculates incorrectly if allocations have an unlimited resource type as the unlimited resource turns negative in the queue headroom.
* getOutstandingRequests from the application breaks if a request has a resource type that is not limited in the queue hierarchy. This will cause the "fit in" check to fail as the unlimited resource turns negative in the headroom.

Both issues result in a scale up failure.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1658

unit tests added
in e2e scenario: add a unlimited resource type to a request

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -353,7 +353,21 @@ func Sub(left, right *Resource) *Resource {
 	return out
 }
 
-// Subtract resource returning a new resource with the result. A nil resource is considered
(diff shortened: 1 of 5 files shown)


craigcondit (approved):
+1 LGTM.

PR #525, opened 2023-03-30
[YUNIKORN-1668] Check there must be one more limit setting for group with specific name defined in the config when use wildcard group limit

zhuqi-lucas:
### What is this PR for?
Specifying a wildcard for the group limit sets a cumulative limit for all users in that queue. If there is no specific group mentioned the wildcard group limit would thus be the same as the queue limit. For that reason we do not allow specifying only one group limit that is using the wildcard. There must be at least one limit with a group name defined.  

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1668

### Questions:

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -89,8 +89,6 @@ type placementPathCheckResult int
 type LimitValidator struct {
 	existedUserName  map[string]bool
 	existedGroupName map[string]bool
-	userWildCardIdx  int
-	groupWildCardIdx int
 }
 
 const (
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -86,6 +86,14 @@ type placementStaticPath struct {
 
 type placementPathCheckResult int
 
+type LimitCheckHelper struct {
+	existedUserName  map[string]bool
+	existedGroupName map[string]bool
+	userWildCardIdx  int
+	groupWildCardIdx int
+	groupSpecificIdx int
+}
+
 const (
 	checkOK placementPathCheckResult = iota
 	nonExistingQueue
@@ -307,7 +315,7 @@ func checkPlacementFilter(filter Filter) error {
 }
 
 // Check a single limit entry
-func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int, groupSpecificIdx *int) error {
+func checkLimit(limit Limit, currIdx int, limitCheckHelper *LimitCheckHelper) error {
 	if len(limit.Users) == 0 && len(limit.Groups) == 0 {
 		return fmt.Errorf("empty user and group lists defined in limit '%v'", limit)
 	}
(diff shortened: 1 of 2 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -86,6 +86,14 @@ type placementStaticPath struct {
 
 type placementPathCheckResult int
 
+type LimitCheckHelper struct {
```
manirajv06, line 89: LimitValidator? or something like that

pkg/common/configs/configvalidator.go
```diff
@@ -320,13 +328,17 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
...
-			*userWildCardIdx = currIdx
-		} else if *userWildCardIdx != -1 && currIdx > *userWildCardIdx {
+			limitCheckHelper.userWildCardIdx = currIdx
+		} else if limitCheckHelper.userWildCardIdx != -1 && currIdx > limitCheckHelper.userWildCardIdx {
 			return fmt.Errorf("should not set no wildcard user %s after wildcard user limit", name)
 		}
+		if limitCheckHelper.existedUserName[name] {
+			return fmt.Errorf("duplicated user name %s , already existed", name)
+		}
+		limitCheckHelper.existedUserName[name] = true
```
manirajv06, line 341: Should we move this block to the beginning? as duplicate check is expected to happen first ideally

pkg/common/configs/configvalidator.go
```diff
@@ -336,14 +348,30 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
...
+
+	// Specifying a wildcard for the group limit sets a cumulative limit for all users in that queue.
+	// If there is no specific group mentioned the wildcard group limit would thus be the same as the queue limit.
+	// For that reason we do not allow specifying only one group limit that is using the wildcard.
+	// There must be at least one limit with a group name defined.
+	if limitCheckHelper.groupWildCardIdx != -1 && limitCheckHelper.groupSpecificIdx == -1 {
+		return fmt.Errorf("should not specify only one group limit that is using the wildcard. " +
+			"There must be at least one limit with a group name defined ")
+	}
+
```
manirajv06, line 374: Do we really need groupWildCardIdx? Can we do something like this?

if existedGroupName has *  and len of existedGroupName is 1 then return error.

manirajv06 (changes requested):

pkg/common/configs/configvalidator.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

pkg/common/configs/configvalidator.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

pkg/common/configs/configvalidator.go
zhuqi-lucas: Good suggestion, addressed in latest PR.

zhuqi-lucas:
Thank you @manirajv06 for review and good suggestions, addressed in latest PR.

pkg/common/configs/configvalidator.go
```diff
@@ -316,34 +321,46 @@ func checkLimit(limit Limit, currIdx int, userWildCardIdx, groupWildCardIdx *int
 		if name != "*" && !UserRegExp.MatchString(name) {
 			return fmt.Errorf("invalid limit user name '%s' in limit definition", name)
 		}
+
+		if limitValidator.existedUserName[name] {
+			return fmt.Errorf("duplicated user name %s , already existed", name)
+		}
+		limitValidator.existedUserName[name] = true
```
manirajv06, line 328: Instead of having limitValidator as struct, we can have local variables (say users, groups etc) of type map to do this as anyways removed all other variables in above said struct. This way, we can remove the struct altogether.

Other than this, we are good.

manirajv06 (changes requested):

pkg/common/configs/configvalidator.go
zhuqi-lucas: Thank you for this good suggestion, addressed in latest PR!

manirajv06 (commented):
Looks good. One final question:

I think we should allow same user and/or group name under different queues with different limit settings. Is it happening? Can we add test case to validate this?

zhuqi-lucas:
> 

Hi @manirajv06 , the checkLimits function are only in the same queue level, so different queues are not affected. Also add the unit test for double check, thanks!

manirajv06 (approved):

PR #526, opened 2023-04-05
[YUNIKORN-1679] Re-write event storage from map to slice

pbacsko:
### What is this PR for?
Replace the map to a slice inside `defaultEventStore` so we don't lose events that belong to the same ObjectID.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1679

### Questions:


pbacsko:
Current solution is complicated. Will rewrite a simpler data structure.

wilfred-s (approved):
LGTM

PR #530, opened 2023-04-17
[YUNIKORN-1472] Replacement for gorilla/mux

zhuqi-lucas:
### What is this PR for?

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1472

### Questions:


pkg/webservice/handlers.go
```diff
@@ -481,9 +482,9 @@ func getPartitions(w http.ResponseWriter, r *http.Request) {
 }
 
 func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
-	vars := mux.Vars(r)
+	vars := httprouter.ParamsFromContext(r.Context())
```
wilfred-s, line 485: httprouter.ParamsFromContext can return nil: https://github.com/julienschmidt/httprouter/blob/master/router.go#L118-L119
Should we not handle that or are we 100% sure that can never happen?

pkg/webservice/webservice.go
```diff
@@ -41,26 +42,29 @@ type WebService struct {
...
-			Name(webRoute.Name).
-			Handler(handler)
+		switch webRoute.Method {
+		case "GET":
+			router.HandlerFunc(http.MethodGet, webRoute.Pattern, handler)
+		case "PUT":
+			router.HandlerFunc(http.MethodPut, webRoute.Pattern, handler)
+		case "POST":
+			router.HandlerFunc(http.MethodPost, webRoute.Pattern, handler)
+		}
```
wilfred-s, line 56: Unneeded switch, http.Method* are simple strings use:
```
router.Handle(webroute.Method, webroute.Pattern, handler)
```

pkg/webservice/webservice.go
```diff
@@ -41,26 +42,29 @@ type WebService struct {
...
 }
 
-func loggingHandler(inner http.Handler, name string) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+func loggingHandler(inner http.Handler, name string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
 		start := time.Now()
 		inner.ServeHTTP(w, r)
 		log.Logger().Debug(fmt.Sprintf("%s\t%s\t%s\t%s",
 			r.Method, r.RequestURI, name, time.Since(start)))
```
wilfred-s, line 66: Remove the Sprintf. Building the string causes overhead we do not want/need. Move to standard logging call using zap.String(name, value) for each, ordering: name, method, URI and time.Since(start)

wilfred-s (changes requested):

pkg/webservice/webservice.go
zhuqi-lucas: Good suggestion!

pkg/webservice/webservice.go
zhuqi-lucas: My bad, this is more reasonable! Thank you @wilfred-s !

pkg/webservice/handlers.go
zhuqi-lucas: Right, we can't 100% sure, i will address this change.

zhuqi-lucas:
Addressed the comments, thank you @wilfred-s for review.

craigcondit (approved):
+1 LGTM.

PR #533, opened 2023-04-28
[YUNIKORN-1709] Add event streaming logic

pbacsko:
### What is this PR for?
Core implementation of event streaming. Clients are updated real time on the REST interface. A persistent HTTP connection is maintained for them.

Consumers can request a new stream and receive core scheduler events immediately on a read-only channel. Slow consumers are detected by checking the number of elements in the channel buffer. If the channel buffer on the local side is full, we remove the consumer.

Explicit closure of the event stream is possible and consumers are expected to do it.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1709

### Questions:

### DIFF
--- a/pkg/events/event_ringbuffer.go
+++ b/pkg/events/event_ringbuffer.go
@@ -70,12 +70,36 @@ func (e *eventRingBuffer) Add(event *si.EventRecord) {
 	e.id++
 }
 
-// GetEventsFromID returns "count" number of event records from "id" if possible. The id can be determined from
+func (e *eventRingBuffer) GetRecentEvents(count uint64) []*si.EventRecord {
+	e.RLock()
+	defer e.RUnlock()
+
+	lastID := e.getLastEventID()
+	var startID uint64
+	if lastID < count {
+		startID = 0
+	} else {
+		startID = lastID - count + 1
+	}
+
+	history, _, _ := e.getEventsFromID(startID, count)
+	return history
+}
+
+// GetEventsFromID returns "count" number of event records from id if possible. The id can be determined from
 // the first call of the method - if it returns nothing because the id is not in the buffer, the lowest valid
 // identifier is returned which can be used to get the first batch.
 // If the caller does not want to pose limit on the number of events returned, "count" must be set to a high
 // value, e.g. math.MaxUint64.
 func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64, uint64) {
+	e.RLock()
+	defer e.RUnlock()
+
+	return e.getEventsFromID(id, count)
+}
+
+// getEventsFromID unlocked version of GetEventsFromID
+func (e *eventRingBuffer) getEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64, uint64) {
 	e.RLock()
 	defer e.RUnlock()
 	lowest := e.getLowestID()
--- a/pkg/events/event_streaming.go
+++ b/pkg/events/event_streaming.go
@@ -172,12 +172,9 @@ func (e *EventStreaming) Close() {
 
 // NewEventStreaming creates a new event streaming infrastructure.
 func NewEventStreaming(eventBuffer *eventRingBuffer) *EventStreaming {
-	stopCh := make(chan struct{})
-	e := &EventStreaming{
+	return &EventStreaming{
 		buffer:       eventBuffer,
-		stopCh:       stopCh,
+		stopCh:       make(chan struct{}),
 		eventStreams: make(map[*EventStream]eventConsumerDetails),
 	}
-
-	return e
 }
--- a/pkg/events/event_streaming.go
+++ b/pkg/events/event_streaming.go
@@ -1,0 +1,166 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package events
+
+import (
+	"sync"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/apache/yunikorn-core/pkg/log"
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
+const defaultChannelBufSize = 100
+
+type EventStreaming struct {
+	buffer       *eventRingBuffer
+	stopCh       chan struct{}
+	eventStreams map[*EventStream]eventConsumerDetails
+	sync.Mutex
+}
+
+type eventConsumerDetails struct {
+	local     chan *si.EventRecord
+	consumer  chan<- *si.EventRecord
+	stopCh    chan struct{}
+	name      string
+	createdAt time.Time
+}
+
+// EventStream handle type returned to the client who wants to capture the stream of events.
+type EventStream struct {
+	Events <-chan *si.EventRecord
+}
+
+// PublishEvent publishes an event to all event stream consumers.
+func (e *EventStreaming) PublishEvent(event *si.EventRecord) {
+	e.Lock()
+	defer e.Unlock()
+
+	for consumer, details := range e.eventStreams {
+		if len(details.local) == defaultChannelBufSize {
+			log.Log(log.Events).Warn("Listener buffer full due to potentially slow consumer, removing it")
+			e.removeEventStream(consumer)
+			continue
+		}
+
+		details.local <- event
+	}
+}
+
+// CreateEventStream sets up event streaming for a consumer. The returned EventStream object
+// contains a channel that can be used for reading.
+// When a consumer is finished, it is supposed to call RemoveEventStream to free up resources.
+// Consumers have an arbitrary name for logging purposes. The "count" parameter defines the number
+// of maximum historical events from the ring buffer.
+func (e *EventStreaming) CreateEventStream(name string, count uint64) *EventStream {
+	consumer := make(chan *si.EventRecord, defaultChannelBufSize)
+	stream := &EventStream{
+		Events: consumer,
+	}
+	local := make(chan *si.EventRecord, defaultChannelBufSize)
+	stop := make(chan struct{})
+	history := e.createEventStreamInternal(stream, local, consumer, stop, name, count)
+
+	go func(consumer chan<- *si.EventRecord, local <-chan *si.EventRecord, stop <-chan struct{}) {
+		// store the refs of historical events; it's possible that some events are added to the
+		// ring buffer and also to "local" channel
+		seen := make(map[*si.EventRecord]bool)
+		for _, event := range history {
+			consumer <- event
+			seen[event] = true
+		}
+		for {
+			select {
+			case <-e.stopCh:
+				close(consumer)
+				return
+			case <-stop:
+				close(consumer)
+				return
+			case event := <-local:
+				if seen[event] {
+					continue
+				}
+				// since events are processed in a single goroutine, doubling is no longer
+				// possible at this point
+				seen = make(map[*si.EventRecord]bool)
+				consumer <- event
+			}
+		}
+	}(consumer, local, stop)
+
+	log.Log(log.Events).Info("Created event stream", zap.String("consumer name", name))
+	return stream
+}
+
+func (e *EventStreaming) createEventStreamInternal(stream *EventStream,
+	local chan *si.EventRecord,
+	consumer chan *si.EventRecord,
+	stop chan struct{},
+	name string,
+	count uint64) []*si.EventRecord {
+	// stuff that needs locking
+	e.Lock()
+	defer e.Unlock()
+
+	e.eventStreams[stream] = eventConsumerDetails{
+		local:     local,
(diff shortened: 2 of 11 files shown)


pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,175 @@
...
+//
+// Retrieving the records can be achieved with GetLatestEntriesCount and GetLatestEntries. Since these do not
+// remove the elements, they are not regular dequeue operations either.
+//
+// Entries have a maximum lifespan defined in nanoseconds. Cleanup of expired objects occurs when a call to
+// RemoveExpiredEntries is made.
+//
+// The buffer does not use locking, proper synchronization is delegated to the client.
+type EventRingBuffer struct {
+	events        []*si.EventRecord
```
pbacsko, line 44: IMO we need to switch to `[]si.EventRecord` if we want to store a lot of elements. We want to avoid GC interactions, so using a value type here is very much reasonable.

In fact, I'm considering a specia, internal type here instead of `EventRecord` with interned strings. There will be a lot of string duplications in the events (especially IDs). 
This library looks promising: https://github.com/philpearl/intern

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// PublishEvent publishes an event to event stream consumers. Events are first stored in a ring buffer,
+// then all clients are notified. Clients whose buffer is full are considered slow and removed.
+func (e *EventStreaming) PublishEvent(event *si.EventRecord) {
+	e.Lock()
+	defer e.Unlock()
+
+	e.buffer.Add(event)
+	for consumer, details := range e.eventStreams {
```
pbacsko, line 65: We can add another level of channel communication here, so updating clients will occur on a separate goroutine. So this method can look like:

```
func (e *EventStreaming) PublishEvent(event *si.EventRecord) {
	e.Lock()
	defer e.Unlock()
	e.buffer.Add(event)
        e.events <- event
}
```

So the event flow will be:
incoming event --> `e.events` --> `consumerDetails.local` --> `consumerDetails.consumer`.

It's a matter of taste, this method is already fast and not affected by slow clients.

pbacsko:
Note: we can hold off merging until 1.4.0 branching is ready. This is intended for Yunikorn 1.5.0.

pbacsko:
TODO:
- http status codes on return & error object
- idle timeout
- coverage

pbacsko:
Note: any timeout for the HTTP server can break the streaming (WriteTimeout for 1.1, ReadTimeout and IdleTimeout for 2.0). We either start a new server instance for the streaming or explicitly comment (or even test)  that timeouts must not be used.

pkg/webservice/handlers.go
```diff
@@ -918,3 +918,61 @@ func getEvents(w http.ResponseWriter, r *http.Request) {
 		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
 	}
 }
+
+func getStream(w http.ResponseWriter, r *http.Request) {
```
wilfred-s, line 922: We should limit the number of streams that we allow to be open at the same time here. Even if it is some "high" number but we need to be able to turn down new requests as a safety mechanism.

pkg/webservice/handlers.go
```diff
@@ -918,3 +918,61 @@ func getEvents(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+	}
+
+	enc := json.NewEncoder(w)
+	stream := eventSystem.CreateEventStream(r.Host, count)
+
+	// Reading events in an infinite loop until either the client disconnects or Yunikorn closes the channel.
+	// This results in a persistent HTTP connection where the message body is never closed.
+	// We don't use timeouts and since HTTP 1.1 clients are expected to handle persistent connections by default.
```
wilfred-s, line 951: See comment earlier, we should use the ResponseController with a sliding  write deadline.

pkg/scheduler/objects/common_test.go
wilfred-s, line 1: not sure why these changes are needed.

pbacsko: Interfaces changes in EventSystem (two new methods).

pkg/webservice/handlers.go
pbacsko: Yes, there is a follow up ticket for that: https://issues.apache.org/jira/browse/YUNIKORN-2147

pkg/webservice/webservice.go
```diff
@@ -63,12 +64,13 @@ func loggingHandler(inner http.Handler, name string) http.HandlerFunc {
 // TODO we need the port to be configurable
 func (m *WebService) StartWebApp() {
 	router := newRouter()
+	// Important: do not use ReadTimeout, WriteTimeout or IdleTimeout because those can break the event streaming
```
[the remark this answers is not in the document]
pbacsko: `ReadHeaderTimeout` is fine, it's the rest what can cause issues. Will check out ResponseController, looks like it's relevant to us.

pkg/webservice/handlers.go
pbacsko: Change added. New code uses ResponseController.setWriteDeadline() before each send.

pkg/webservice/webservice.go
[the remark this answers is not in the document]
pbacsko: Write timeout added

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,180 @@
...
+// Events are sent to the "local" channel from where it is forwarded to the "consumer" channel.
+//
+// If "local" is full, it means that the consumer side has not processed the events at an appropriate pace.
+// Such a consumer is removed and the related channels are closed.
+func (e *EventStreaming) PublishEvent(event *si.EventRecord) {
+	e.Lock()
+	defer e.Unlock()
+
+	for consumer, details := range e.eventStreams {
+		if len(details.local) == defaultChannelBufSize {
```
zhuqi-lucas, line 67: Minor question, if we can set custom defaultChannelBufSize or dynamic range for the size, here 100 seems small for some cases.

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,180 @@
...
+// The streaming logic uses bridging to ensure proper ordering of existing and new events.
+// Events are sent to the "local" channel from where it is forwarded to the "consumer" channel.
+//
+// If "local" is full, it means that the consumer side has not processed the events at an appropriate pace.
+// Such a consumer is removed and the related channels are closed.
+func (e *EventStreaming) PublishEvent(event *si.EventRecord) {
+	e.Lock()
+	defer e.Unlock()
+
+	for consumer, details := range e.eventStreams {
```
zhuqi-lucas, line 66: @wilfred-s @pbacsko I think the customer here is a good place for us to set a internal customer, the new trackingService, we can at least use it to streaming resourcce and aggregated tracking events.

pkg/events/event_streaming.go
pbacsko: Yes 100 might be a bit too low. 1000 sounds like a reasonable value.

pkg/events/event_streaming.go
pbacsko: If we decided to this internally, we might as well just run the aggregation on a separate goroutine. 

Wilfred's comments:
"I think we can simply use the existing events and push the whole aggregation out of YuniKorn. The only thing that seems to be missing is the node instance type. If that is available just listening to the app and node events would allow creating the summary outside of the scheduler.
[...]
hat is the best option. A scheduler should schedule. It is not for providing statistics."

To me it sounds like a separate application.

zhuqi-lucas: Yeah, we can have a separate service, but we can let the service as a cusomer here, so we don't need to add another publisher, just one publisher for the streaming, and if we have several  publishers, it seems will cost more overhead.

pbacsko: Yes, a separate service will use the REST API (either batch or streaming, whatever is more convenient), just like everyone else. If this is what you mean, I agree with it.

pkg/events/event_streaming.go
zhuqi-lucas: Agree , 1000 seems more reasonable

pbacsko: Changed

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+
+// CreateEventStream sets up event streaming for a consumer. The returned EventStream object
+// contains a channel that can be used for reading.
+//
+// When a consumer is finished, it must call RemoveEventStream to free up resources.
+//
+// Consumers have an arbitrary name for logging purposes. The "count" parameter defines the number
+// of maximum historical events from the ring buffer.
```
zhuqi-lucas, line 83: Nit:

Add more comments about the count, when we want to get all the historical events from the ring buffer, we can set count to 0

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,180 @@
...
+				consumer <- event
+			}
+		}
+	}(consumer, local, stop)
+
+	log.Log(log.Events).Info("Created event stream", zap.String("consumer name", name))
+	return stream
+}
+
+func (e *EventStreaming) createEventStreamInternal(stream *EventStream,
```
zhuqi-lucas, line 129: Nit
The function name is 
createEventStreamInternal 

But we return the history events for this function, maybe we can get history events after we call this method.

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,180 @@
...
+func (e *EventStreaming) RemoveEventStream(consumer *EventStream) {
+	e.Lock()
+	defer e.Unlock()
+
+	e.removeEventStream(consumer)
+}
+
+func (e *EventStreaming) removeEventStream(consumer *EventStream) {
+	if details, ok := e.eventStreams[consumer]; ok {
+		log.Log(log.Events).Info("Removing event stream consumer", zap.String("name", details.name),
```
zhuqi-lucas, line 160: Nit 
Is it possible, we can logging the remove reason, such as:

"Connection closed for event stream client"
"Cannot set write deadline"
"Marshalling error"

zhuqi-lucas (approved):
LGTM Left final minor comments!

pkg/events/event_streaming.go
pbacsko: The external reason why it was closed is logged at the appropriate place in the handler. Here, the code doesn't have to know about it.

pkg/events/event_streaming.go
pbacsko: Actually "0" means no events. But extended the comment.

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// CreateEventStream sets up event streaming for a consumer. The returned EventStream object
+// contains a channel that can be used for reading.
+//
+// When a consumer is finished, it must call RemoveEventStream to free up resources.
+//
+// Consumers have an arbitrary name for logging purposes. The "count" parameter defines the number
+// of maximum historical events from the ring buffer. "0" is a valid value and means no past events.
+func (e *EventStreaming) CreateEventStream(name string, count uint64) *EventStream {
```
zhuqi-lucas, line 84: Hi @pbacsko , i am thinking if we can have an option can include all history events without setting any count, thanks!

zhuqi-lucas: Sure, it's a solution, agree!

pkg/events/event_system.go
```diff
@@ -94,21 +141,25 @@ func (ec *EventSystemImpl) GetRingBufferCapacity() uint64 {
...
+		streaming:     NewEventStreaming(buffer),
 	}
 }
 
+// StartService starts the event processing in the background. See the interface for details.
 func (ec *EventSystemImpl) StartService() {
 	ec.StartServiceWithPublisher(true)
 }
 
+// VisibleForTesting
```
wilfred-s, line 162: fix comment to use standard go layout as it is an.exported function now.

pkg/events/event_system.go
```diff
@@ -191,11 +245,21 @@ func (ec *EventSystemImpl) isRestartNeeded() bool {
 	return ec.readIsTrackingEnabled() != ec.trackingEnabled
 }
 
+// Restart restarts the event system, used during config update.
 func (ec *EventSystemImpl) Restart() {
 	ec.Stop()
 	ec.StartServiceWithPublisher(true)
 }
 
+// VisibleForTesting
```
wilfred-s, line 254: fix comment to use standard go layout as it is an.exported function now.

pkg/events/event_ringbuffer.go
```diff
@@ -70,14 +70,38 @@ func (e *eventRingBuffer) Add(event *si.EventRecord) {
 	e.id++
 }
 
-// GetEventsFromID returns "count" number of event records from "id" if possible. The id can be determined from
+func (e *eventRingBuffer) GetRecentEvents(count uint64) []*si.EventRecord {
```
wilfred-s, line 73: add comment use standard go layout as it is an.exported function

wilfred-s (approved):
minor changes to comments on exported functions can be fixed during the commit also

pkg/events/event_streaming.go
```diff
@@ -0,0 +1,179 @@
...
+	log.Log(log.Events).Info("Created event stream", zap.String("consumer name", name))
+	return stream
+}
+
+func (e *EventStreaming) createEventStreamInternal(stream *EventStream,
+	local chan *si.EventRecord,
+	consumer chan *si.EventRecord,
+	stop chan struct{},
+	name string,
+	count uint64) {
```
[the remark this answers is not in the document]
wilfred-s: Need a follow up jira to clean that up @pbacsko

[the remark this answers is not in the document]
pbacsko: Correct. I actually have a list of necessary small code changes which are unrelated to each other, I'm adding this to the list.

PR #535, opened 2023-05-03
[YUNIKORN-1714] Fatal error: concurrent write/read when calling Queue.RemoveApplication()

pbacsko:
### What is this PR for?
Race condition in `Queue.RemoveApplication()`: we need to lock when accessing `sq.applications[]`. Currently, we check if the application exists without locking, but the map can be modified concurrently.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1714

### Questions:


wilfred-s (approved):
LGTM

PR #536, opened 2023-05-03
[YUNIKORN-1720] Improve the performance of the node.preAllocateCheck() function

elihschiff:
### What is this PR for?

The node.preAllocateCheck() function gets run a lot during the scheduling cycle. The function is effectively a boolean function, either the node can be allocated on or it cannot. Instead of returning a bool it returns nil for true and a full error message for false. The thing is, that error message is never read anywhere. From what I can tell, the only thing the code does is check if there is an error message but not the contents.

I ran a profiler on the code and noticed that about 25% of the scheduling cycle was spend on generating these error messages that are not used. By switching to a bool for the return value it should improve the performance of this one function and therefore the entire scheduling cycle overall.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
N/A

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1720

I have it running on my end and it seems to work. The unit tests also pass after I updated them to look at the new boolean value over the old error message.

Before this change (note the large `(*Resource).String (resources.go:L#110)` section under the highlighted function):
image

After this change:
image

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1145,7 +1145,7 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator, 
 					zap.String("node", node.NodeID))
 				continue
 			}
-			if err := node.preAllocateCheck(reqFit.GetAllocatedResource(), reservationKey(nil, sa, reqFit)); err != nil {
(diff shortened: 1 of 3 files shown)


wilfred-s (approved):
LGTM

pbacsko (approved):
+1 LGTM

PR #538, opened 2023-05-04
[YUNIKORN-1607] Configuration processing should support limit hierarc…

manirajv06:
### What is this PR for?
Configuration processing should support limit hierarchical quota check.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1607

### Questions:

### DIFF
--- a/pkg/common/configs/config_test.go
+++ b/pkg/common/configs/config_test.go
@@ -22,6 +22,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -1639,7 +1640,7 @@ partitions:
                 - user1
                 groups:
                 - prod
-                maxapplications: 5
+                maxapplications: 2
                 maxresources: {memory: 10000, vcore: 10}
 `
 	// validate the config and check after the update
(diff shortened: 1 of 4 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -139,6 +139,72 @@ func checkQueueResource(cur QueueConfig, parentM *resources.Resource) (*resource
...
+
+	// compare user & group limit setting between the current queue and parent queue
+	for _, limit := range cur.Limits {
+		for _, user := range limit.Users {
+			curM, err := resources.NewResourceFromConf(limit.MaxResources)
+			if err != nil {
+				return err
+			}
+			// Is user limit setting exists?
+			if _, ok := users[user]; ok {
```
0yukali0, line 171: Use a variable to reduce the access count of map repeatedly.
```
maxApplicationsBound, ok := users[user]
```

pkg/common/configs/configvalidator.go
```diff
@@ -139,6 +139,72 @@ func checkQueueResource(cur QueueConfig, parentM *resources.Resource) (*resource
...
+				if !users[user].FitInMaxUndef(curM) {
+					return fmt.Errorf("user %s max resource %s of queue %s is greater than immediate or ancestor parent maximum resource %s", user, curM.String(), cur.Name, users[user].String())
+				}
+				// Override with min resource
+				users[user] = resources.ComponentWiseMinPermissive(curM, users[user])
+			} else {
+				users[user] = curM
+			}
+		}
+		for _, group := range limit.Groups {
```
0yukali0, line 181: The reason is above.

pkg/common/configs/configvalidator.go
```diff
@@ -153,6 +219,56 @@ func checkQueueMaxApplications(cur QueueConfig) error {
...
+			}
+		}
+	}
+
+	// compare user & group limit setting between the current queue and parent queue
+	for _, limit := range cur.Limits {
+		for _, user := range limit.Users {
+			curM := limit.MaxApplications
+			// Is user limit setting exists?
+			if _, ok := users[user]; ok {
```
0yukali0, line 240: The reason is above.

pkg/common/configs/configvalidator.go
```diff
@@ -153,6 +219,56 @@ func checkQueueMaxApplications(cur QueueConfig) error {
...
+				}
+				users[user] = uint64(math.Min(float64(curM), float64(users[user])))
+			} else {
+				users[user] = curM
+			}
+		}
+		for _, group := range limit.Groups {
+			curM := limit.MaxApplications
+			// Is group limit setting exists?
+			if _, ok := groups[group]; ok {
```
0yukali0, line 252: The reason is above.

0yukali0 (changes requested):
Hi @manirajv06, I noticed that this commit use DFS to check limits and the MaxApplications.
In the future work, we will sum up the limits among children queues to check that the sum should not exceed the parent's limits.
I think using BFS can be more intuitive to deal with the problem that the sum from child queues' limits does not exceed the limit of the parent queue.
Maybe we adopt BFS instead of DFS in the future?
I want to know your opinion. :)

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1722,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+                  guaranteed:
+                    {memory: 1000, vcore: 10}
+                  max:
+                    {memory: 10000, vcore: 1000}
+                limits:
+                  - limit:
+                    users: 
+                    - user1
+                    maxapplications: ` + strconv.Itoa(int(leafQueueMaxApps)) + `
+                    maxresources: ` + leafQueueMaxResource + `
```
pbacsko, line 1773: Could you use placeholders like `%d`, `%s`, etc. then replace those with `Sprintf()`? That would make the embedded YAML more readable.

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1722,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+                  guaranteed:
+                    {memory: 1000, vcore: 10}
+                  max:
+                    ` + leafQueueMaxResource + `
+                limits:
+                  - limit:
+                    users: 
+                    - user1
+                    maxapplications: ` + strconv.Itoa(int(leafQueueMaxApps)) + `
+                    maxresources: ` + userMaxResource + `
```
pbacsko, line 1832: Same here, I'd prefer placeholders.

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1722,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+
+	// (More than one level check)
+	// Make sure hierarchical queue user max apps should not less than the child queue max apps
+	// validate the config and check after the update
+	_, err = CreateConfig(prepareUserLimitsWithTwoLevelsConfig(900, "{memory: 10000, vcore: 1000}", "{memory: 1000, vcore: 100}"))
+	assert.ErrorContains(t, err, "user user1 max applications 900 of queue leaf is greater than immediate or ancestor parent max applications 500")
+
+	// Special case: make sure wildcard user with hierarchical queue check
+	data := `
+partitions:
```
pbacsko, line 1874: I haven't compared the test data, but if the hierarchy is the same and only some numbers/settings change, you should consider extracting this to a method which can be parameterized and returns the YAML with the substituted values.

pkg/common/configs/configvalidator.go
```diff
@@ -139,6 +139,72 @@ func checkQueueResource(cur QueueConfig, parentM *resources.Resource) (*resource
...
+					return err
+				}
+			}
+		}
+	}
+
+	// compare user & group limit setting between the current queue and parent queue
+	for _, limit := range cur.Limits {
+		for _, user := range limit.Users {
+			curM, err := resources.NewResourceFromConf(limit.MaxResources)
```
pbacsko, line 166: Two things:
1. Consider a better name for `curM` and `cur`, it's not clear what it stands for.
2. We keep creating `curM` in an embeddef loop, is this necessary? How about creating it outside and maybe clone it if needed (`users[user] = curM.Clone())` ?

pkg/common/configs/configvalidator.go
```diff
@@ -153,6 +219,56 @@ func checkQueueMaxApplications(cur QueueConfig) error {
...
+				if users[user] != 0 && (users[user] < curM || curM == 0) {
+					return fmt.Errorf("user %s max applications %v of queue %s is greater than immediate or ancestor parent max applications %v", user, curM, cur.Name, users[user])
+				}
+				users[user] = uint64(math.Min(float64(curM), float64(users[user])))
+			} else {
+				users[user] = curM
+			}
+		}
+		for _, group := range limit.Groups {
+			curM := limit.MaxApplications
```
pbacsko, line 250: Variable naming.

pkg/common/configs/configvalidator.go
```diff
@@ -613,6 +729,16 @@ func Validate(newConfig *SchedulerConfig) error {
 		if err != nil {
 			return err
 		}
+
+		err = checkLimitResource(partition.Queues[0], nil, make(map[string]*resources.Resource), make(map[string]*resources.Resource))
```
pbacsko, line 733:
```
if err = method(); err != nil {
  return err
}
```

pkg/common/configs/configvalidator.go
```diff
@@ -613,6 +729,16 @@ func Validate(newConfig *SchedulerConfig) error {
...
+		if err != nil {
+			return err
+		}
+
+		err = checkLimitMaxApplications(partition.Queues[0], nil, make(map[string]uint64), make(map[string]uint64))
```
pbacsko, line 738:
```
if err = method(); err != nil {
  return err
}
```

pbacsko (changes requested):
I added some comments which simplifies the code or make it more readable.

pkg/common/configs/configvalidator.go
```diff
@@ -139,6 +139,72 @@ func checkQueueResource(cur QueueConfig, parentM *resources.Resource) (*resource
 	return curG, nil
 }
 
+func checkLimitResource(cur QueueConfig, parent *QueueConfig, users map[string]*resources.Resource, groups map[string]*resources.Resource) error {
```
zhuqi-lucas, line 142: I am thinking, if we can storage the hierarchy queue limit value to memory when traversing, so that we can use it in UGM, so that we don't need to traverse the value again when we want to storage the limit value?

zhuqi-lucas (changes requested):

pkg/common/configs/configvalidator.go
zhuqi-lucas: After traversing, when there is no err, we can return the user limit and group limit all value to store.

manirajv06:
> Hi @manirajv06, I noticed that this commit use DFS to check limits and the MaxApplications. In the future work, we will sum up the limits among children queues to check that the sum should not exceed the parent's limits. I think using BFS can be more intuitive to deal with the problem that the sum from child queues' limits does not exceed the limit of the parent queue. Maybe we adopt BFS instead of DFS in the future? I want to know your opinion. :)

In general, we do the sum only for Guaranteed resources at Queue level, not for max resources. Limit max resources and max applications are being considered as max resources not as guaranteed one to increase the overall efficiency of the overall cluster resources i.e elasticity - user 1 can stretch as much as possible when other users are not using its resources.
Hence, BFS has been used for now similar to other Queue config processing.

manirajv06:
@pbacsko @0yukali0 @zhuqi-lucas Thanks. @wilfred-s can you also take a final look and merge?

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1722,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+                resources:
+                  guaranteed:
+                    {memory: 1000, vcore: 10}
+                  max:
+                    {memory: 10000, vcore: 1000}
+                limits:
+                  - limit:
+                    users: 
+                    - user1
+                    maxapplications: ` + fmt.Sprintf("%d", int(leafQueueMaxApps)) + `
```
pbacsko, line 1772: What I meant is that you use `%d` inside the string, which then gets replaced by `fmt.Sprintf()`.
That is:
```
return fmt.Sprintf(partitions, leafQueueMaxApps, leafQueueMaxResource)
```

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1722,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+                resources:
+                  guaranteed:
+                    {memory: 1000, vcore: 10}
+                  max:
+                    ` + leafQueueMaxResource + `
+                limits:
+                  - limit:
+                    users: 
+                    - user1
+                    maxapplications: ` + fmt.Sprintf("%d", int(leafQueueMaxApps)) + `
```
pbacsko, line 1831: Same as above

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1722,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+                resources:
+                  guaranteed:
+                    {memory: 1000, vcore: 10}
+                  max:
+                    {memory: 10000, vcore: 1000}
+                limits:
+                  - limit:
+                    groups: 
+                    - group1
+                    maxapplications: ` + fmt.Sprintf("%d", int(leafQueueMaxApps)) + `
```
pbacsko, line 1972: Same as above

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1722,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+                resources:
+                  guaranteed:
+                    {memory: 1000, vcore: 10}
+                  max:
+                    ` + leafQueueMaxResources + `
+                limits:
+                  - limit:
+                    groups: 
+                    - group1
+                    maxapplications: ` + strconv.Itoa(int(leafQueueMaxApps)) + `
```
pbacsko, line 2033: Same as above

pkg/common/configs/config_test.go
```diff
@@ -1721,3 +1721,418 @@ func TestGetConfigurationString(t *testing.T) {
...
+                resources:
+                  guaranteed:
+                    {memory: 1000, vcore: 10}
+                  max:
+                    {memory: 10000, vcore: 1000}
+                limits:
+                  - limit:
+                    users: 
+                    - user1
+                    maxapplications: ` + "%d" + `
```
pbacsko, line 1771: Ok, I swear it's the last one... You don't need to concat strings, `%d` will be replaced by `Sprintf()`. Just write `maxapplications: %d`.

pkg/common/configs/configvalidator.go
```diff
@@ -153,6 +211,56 @@ func checkQueueMaxApplications(cur QueueConfig) error {
...
+	// compare user & group limit setting between the current queue and parent queue
+	for _, limit := range cur.Limits {
+		limitMaxApplications := limit.MaxApplications
+		for _, user := range limit.Users {
+			// Is user limit setting exists?
+			if userMaxApplications, ok := users[user]; ok {
+				if userMaxApplications != 0 && (userMaxApplications < limitMaxApplications || limitMaxApplications == 0) {
+					return fmt.Errorf("user %s max applications %v of queue %s is greater than immediate or ancestor parent max applications %v", user, limitMaxApplications, cur.Name, userMaxApplications)
+				}
+				users[user] = uint64(math.Min(float64(limitMaxApplications), float64(userMaxApplications)))
```
wilfred-s, line 237: Introduce a simple util function in our code: `Min(x, y uint64) uint64` that returns one or the other

pkg/common/configs/configvalidator.go
```diff
@@ -153,6 +211,56 @@ func checkQueueMaxApplications(cur QueueConfig) error {
...
+	}
+
+	// compare user & group limit setting between the current queue and parent queue
+	for _, limit := range cur.Limits {
+		limitMaxApplications := limit.MaxApplications
+		for _, user := range limit.Users {
+			// Is user limit setting exists?
+			if userMaxApplications, ok := users[user]; ok {
+				if userMaxApplications != 0 && (userMaxApplications < limitMaxApplications || limitMaxApplications == 0) {
+					return fmt.Errorf("user %s max applications %v of queue %s is greater than immediate or ancestor parent max applications %v", user, limitMaxApplications, cur.Name, userMaxApplications)
```
wilfred-s, line 235: use %d for integers not %v

wilfred-s (changes requested):
minor points to fix

pkg/common/configs/configvalidator.go
```diff
@@ -615,6 +723,13 @@ func Validate(newConfig *SchedulerConfig) error {
 		if err != nil {
 			return err
 		}
+
+		if err = checkLimitResource(partition.Queues[0], nil, make(map[string]*resources.Resource), make(map[string]*resources.Resource)); err != nil {
```
zhuqi-lucas, line 727: If we can pass the init map, and when return err we clean the map.
When we don't return error, we can reuse the map to limit storage, we don't need to load from conifg to storage again after this. cc @wilfred-s @pbacsko

zhuqi-lucas (changes requested):

pkg/common/configs/configvalidator.go
manirajv06: We store limits containing max apps and max resources for users and groups in separate maps. But here, we handle only max resource and max applications in another method. I am not sure how we can club together and also not sure about the net gain. If really needed, can we file a follow up jira?

zhuqi-lucas: Sure @manirajv06 , i think we may file a follow-up jira for this optimize

zhuqi-lucas (approved):
LGTM

pkg/common/utils.go
```diff
@@ -214,3 +214,10 @@ func WaitForCondition(eval func() bool, interval time.Duration, timeout time.Dur
 		time.Sleep(interval)
 	}
 }
+
+func Min(x, y uint64) uint64 {
```
pbacsko, line 218: I might be overly pedantic, but pls add a tiny unit test for this :)

pbacsko (approved):
+1

PR #539, opened 2023-05-05
[YUNIKORN-1722] Remove excessive resource clone calls

wilfred-s:
### What is this PR for?
The node calculateNodesResourceUsage clones the  resource object inside a loop. The clone is correct as the allocated resource could be changed while the calculation is happening in the scheduling cycle. However, it should not happen inside the loop.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1722

unit test cover the code
performance test should improve


pbacsko (approved):
+1

PR #541, opened 2023-05-09
[YUNIKORN-1719] Improve the performance of Application.sortRequests()

pbacsko:
### What is this PR for?
Use `btree.BTree` to store allocation requests, so we don't have to call `Application.sortRequests()` constantly.
Iteration occurs through `BTree.Ascend()`, allocation logic is passed as a function.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1719

### Questions:
* [ ] - The licenses files need update.
* [ ] - There is breaking changes for older versions.
* [ ] - It needs documentation.

PR #547, opened 2023-05-15
[YUNIKORN-1719] Improve the performance of Application.sortRequests()

pbacsko:
### What is this PR for?
Avoid calling `Application.sortRequests()` constantly. We only need to sort if a new request is added. If `pendingAskRepeat` becomes 0 for a request, we simply skip it.
After doing various benchmarks, this approach is simpler and even faster than maintaining requests in a sorted BTree (see #541).

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1719

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -514,6 +514,7 @@ func (sa *Application) removeAsksInternal(allocKey string) int {
 		deltaPendingResource = sa.pending
 		sa.pending = resources.NewResource()
 		sa.requests = make(map[string]*AllocationAsk)
+		sa.sortedRequests = nil
 		sa.askMaxPriority = configs.MinPriority
 		sa.queue.UpdateApplicationPriority(sa.ApplicationID, sa.askMaxPriority)
 	} else {
@@ -536,6 +537,7 @@ func (sa *Application) removeAsksInternal(allocKey string) int {
 			deltaPendingResource = resources.MultiplyBy(ask.GetAllocatedResource(), float64(ask.GetPendingAskRepeat()))
 			sa.pending = resources.Sub(sa.pending, deltaPendingResource)
 			delete(sa.requests, allocKey)
+			sa.sortRequests()
 			if priority := ask.GetPriority(); priority >= sa.askMaxPriority {
 				sa.updateAskMaxPriority()
 			}
@@ -619,7 +621,7 @@ func (sa *Application) AddAllocationAsk(ask *AllocationAsk) error {
 		zap.Bool("placeholder", ask.IsPlaceholder()),
 		zap.Stringer("pendingDelta", delta))
 
-	sa.sortRequests(false)
+	sa.sortRequests()
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


craigcondit (changes requested):
This isn't safe, as there are numerous places where the original requests map is updated, not just on AddAllocationAsk(). Therefore, the sorted list will get out of sync with the map. 

Also, if we go this way, we should refactor sorting to not take a parameter but always sort descending.

wilfred-s (changes requested):
What about the points that we manipulate the request list content? One we can ignore the other two we need to handle for sorting:
* (sa *Application) RecoverAllocationAsk -->  ignore as all asks have a repeat of 0 
* (sa *Application) removeAsksInternal
* (sa *Application) UpdateAskRepeat

The UpdateAskRepeat is special as that is the only one that can increase the ask repeat. It is called if a node is removed while a placeholder replacement is in progress on that node. 

There are also two other locations that we sort the requests:
* (sa *Application) getOutstandingRequests
* (sa *Application) tryPlaceholderAllocate
Both of these need to be updated also, neither should sort and both should skip if `repeat == 0`

We also only sort one way in the production code and that is with `ascending == false`, like @craigcondit said we should remove the parameter.

wilfred-s:
I see that we also have a small leak for the case that we call removeAsksInternal with an empty string. We remove all requests but we do not reset the sortedRequests to nil (line 516) we should add `sa.sortedRequests = nil` It is a leak that eventually will fix itself but we need to be consistent.

pbacsko:
> I see that we also have a small leak for the case that we call removeAsksInternal with an empty string. We remove all requests but we do not reset the sortedRequests to nil (line 516) we should add `sa.sortedRequests = nil` It is a leak that eventually will fix itself but we need to be consistent.

Thinking about creating some new unit tests. Now we maintain more state and we probably should focus on proper coverage.

wilfred-s:
> > (sa *Application) RecoverAllocationAsk --> ignore as all asks have a repeat of 0
> 
> Interesting, but it means that the following branch in the same method is never taken:
> 
> ```
> 	// update app priority
> 	repeat := ask.GetPendingAskRepeat()
> 	priority := ask.GetPriority()
> 	if repeat > 0 && priority > sa.askMaxPriority {
> 		sa.askMaxPriority = priority
> 		sa.queue.UpdateApplicationPriority(sa.ApplicationID, sa.askMaxPriority)
> 	}
> ```

Correct that can be removed. The request is purely there for consistency. It has been allocated already and does not influence priority anymore.
We should file a follow up jira to remove that bit of code.

wilfred-s (approved):
LGTM
Unless Craig has any further review points we can commit this.

craigcondit (approved):
Looks fine.

PR #548, opened 2023-05-15
[YUNIKORN-1716] Improve the performance of baseNodeCollection.getNodeIteratorInternal()

pbacsko:
### What is this PR for?
Improve the performance of node iterators by not creating a sorted slice all the time.
The iterator is reset at each call: we always get a fresh "clone" from `BTree.Clone()`, so tree walking is not affected by concurrent mutations.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1716

### Questions:

### DIFF
--- a/pkg/scheduler/objects/node_collection.go
+++ b/pkg/scheduler/objects/node_collection.go
@@ -31,6 +31,14 @@ import (
 	"github.com/apache/yunikorn-core/pkg/scheduler/policies"
 )
 
+var acceptUnreserved = func(node *Node) bool {
+	return !node.IsReserved()
+}
+
+var acceptAll = func(node *Node) bool {
+	return true
+}
+
 type NodeCollection interface {
 	AddNode(node *Node) error
 	RemoveNode(nodeID string) *Node
@@ -69,6 +77,9 @@ type baseNodeCollection struct {
 	nsp         NodeSortingPolicy   // node sorting policy
 	nodes       map[string]*nodeRef // nodes assigned to this collection
 	sortedNodes *btree.BTree        // nodes sorted by score
+
+	unreservedIterator *treeIterator
+	fullIterator       *treeIterator
 
 	sync.RWMutex
 }
@@ -154,17 +165,15 @@ func (nc *baseNodeCollection) GetNodes() []*Node {
 // Create an ordered node iterator for unreserved nodes based on the sort policy set for this collection.
 // The iterator is nil if there are no unreserved nodes available.
 func (nc *baseNodeCollection) GetNodeIterator() NodeIterator {
-	return nc.getNodeIteratorInternal(func(node *Node) bool {
-		return !node.IsReserved()
-	})
+	nc.unreservedIterator.Reset()
+	return nc.unreservedIterator
 }
 
 // Create an ordered node iterator for all nodes based on the sort policy set for this collection.
 // The iterator is nil if there are no nodes available.
 func (nc *baseNodeCollection) GetFullNodeIterator() NodeIterator {
-	return nc.getNodeIteratorInternal(func(node *Node) bool {
-		return true
-	})
+	nc.fullIterator.Reset()
+	return nc.fullIterator
 }
 
 func (nc *baseNodeCollection) getNodeIteratorInternal(allow func(*Node) bool) NodeIterator {
@@ -242,10 +251,18 @@ func (nc *baseNodeCollection) NodeUpdated(node *Node) {
 
 // Create a new collection for the given partition.
 func NewNodeCollection(partition string) NodeCollection {
-	return &baseNodeCollection{
+	bsc := &baseNodeCollection{
 		Partition:   partition,
 		nsp:         NewNodeSortingPolicy(policies.FairSortPolicy.String(), nil),
 		nodes:       make(map[string]*nodeRef),
 		sortedNodes: btree.New(7), // Degree=7 here is experimentally the most efficient for up to around 5k nodes
 	}
+
+	unreservedIterator := NewTreeIterator(acceptUnreserved, bsc.cloneSortedNodes)
+	fullIterator := NewTreeIterator(acceptAll, bsc.cloneSortedNodes)
+
+	bsc.fullIterator = fullIterator
+	bsc.unreservedIterator = unreservedIterator
+
+	return bsc
 }
--- a/pkg/scheduler/objects/node_collection.go
+++ b/pkg/scheduler/objects/node_collection.go
@@ -163,14 +163,12 @@ func (nc *baseNodeCollection) GetNodes() []*Node {
 // Create an ordered node iterator for unreserved nodes based on the sort policy set for this collection.
 // The iterator is nil if there are no unreserved nodes available.
 func (nc *baseNodeCollection) GetNodeIterator() NodeIterator {
-	nc.unreservedIterator.Reset()
 	return nc.unreservedIterator
 }
 
 // Create an ordered node iterator for all nodes based on the sort policy set for this collection.
 // The iterator is nil if there are no nodes available.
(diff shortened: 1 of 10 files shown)


pbacsko:
Changes to speed up `NodeCollection.GetFullNodeIterator()` and `NodeCollection.GetNodeIterator()`.

pkg/scheduler/objects/node_iterator.go
```diff
@@ -30,6 +32,8 @@ type NodeIterator interface {
 	Next() *Node
 	// reset the iterator to a clean state
 	Reset()
+
+	ForEachNode(func(*Node) bool)
```
pbacsko, line 36: This makes it not really an iterator anymore.

pkg/scheduler/objects/node_iterator.go
```diff
@@ -77,6 +81,42 @@ func NewDefaultNodeIterator(schedulerNodes []*Node) NodeIterator {
 	return it
 }
 
+type treeIterator struct {
```
pbacsko, line 84: An alternative approach with GODS Btree/Redblacktree (don't want to code this): those impls have an {{Iterator}} interface (see https://github.com/emirpasic/gods/blob/master/trees/redblacktree/iterator.go) but don't have {{Clone()}}, so we need to lock a mutex while iterating. The only drawback is that we can receive node updates from different code paths, ie. you might or might not need to obtain a mutex. If you already called `Lock()`, you must not call it again, otherwise it'll be a deadlock. This can be solved in three ways at least (maybe there are some more which involves trickery, but #3 is already somewhat involved):

1. run `baseNodeCollection.NodeUpdated()` calls on a separate goroutine where it receives `Node` objects on a channel
2. propagate a `bool` or a boolean embedded in a `Context` to signal whether we already have a lock or not
3. use CAS + Wait-Notify combo to mark if an iteration is in progress and grab locks accordingly

#2 is too much change and method signatures will be inconsistent (w/ and w/o this extra data) and #3 is too complicated. #1 is nice, we detach the update logic from the scheduling cycle (it does not need to be synchronous).

The upside of using GODS collections is that the nature of iterator will remain, we still use the `Next()` and `HastNext()` methods, it needs less change in other files.

pbacsko:
@craigcondit @wilfred-s could you guys take a look at this and share your opinion?

As other parts of YK is optimized, this path (getting node iterator) is becoming more and more expensive at least in relative terms, so it would be nice if we fixed it. On larger clusters with >1000 nodes, it is becomes costly.

craigcondit:
Needs a rebase, but I think the overall approach seems sane. If we can resolve the conflicts, I can do a deeper review.

craigcondit (changes requested):
Overall, this looks good. However, we now have a lot of dead code. The only iterator implementation used (other than tests) is the tree iterator. Let's remove the other implementations, including the default / baseIterator objects. The existing tests for defaultIterator don't really work for the treeIterator, so we should also create new ones for treeIterator.

pkg/scheduler/objects/node_iterator.go
```diff
@@ -19,103 +19,44 @@
...
-	idx := (ri.countIdx + ri.startIdx) % ri.size
-	value := ri.nodes[idx]
-	ri.countIdx++
-	return value
-}
-
-func (ri *roundRobinNodeIterator) Reset() {
-	ri.countIdx = 0
-	ri.startIdx = -1
+	ti.tree = getTree()
```
craigcondit, line 60: We can do an optimization here by not calling getTree() and storing the result. Instead, just do it in ForEachNode():

Instead of ti.tree.Ascend(...) use ti.getTree().Ascend(...).

This means we only compute the tree if ForEachNode is actually called.

pkg/scheduler/objects/node_iterator.go
```diff
@@ -19,103 +19,44 @@
...
+	tree    *btree.BTree
 }
 
-// Create a new default iterator
-func NewDefaultNodeIterator(schedulerNodes []*Node) NodeIterator {
-	it := &defaultNodeIterator{}
-	it.nodes = schedulerNodes
-	it.size = len(schedulerNodes)
-	return it
+func (ti *treeIterator) Reset() {
```
craigcondit, line 40: The Reset() method and all callers can be removed as this is a No-Op on the tree.

pkg/scheduler/objects/node_iterator.go
```diff
@@ -19,103 +19,44 @@
...
 )
 
 // NodeIterator iterates over a list of nodes based on the defined policy
 type NodeIterator interface {
-	// returns true if there are more values to iterate over
-	HasNext() bool
-	// returns the next node from the iterator
-	Next() *Node
-	// reset the iterator to a clean state
+	// Reset the iterator to a clean state
```
craigcondit, line 27: Remove as this is now a no-op for the only implementation. Calling it has no effect on the result of ForEachNode().

pkg/scheduler/objects/node_iterator.go
```diff
@@ -19,103 +19,44 @@
...
-// Returns nil at the end of iteration.
-func (bi *baseIterator) Next() *Node {
-	if (bi.countIdx + 1) > bi.size {
-		return nil
-	}
 
-	value := bi.nodes[bi.countIdx]
-	bi.countIdx++
-	return value
+	// ForEachNode Calls the provided function on each Node object unil it returns false
```
craigcondit, line 30: Nit: typo (unil -> until)

pkg/scheduler/objects/node_iterator.go
```diff
@@ -19,103 +19,44 @@
...
 }
 
-// Random iterator, wraps the base iterator
-// Iterates over the list from a random starting position in the list.
-// The iterator automatically wraps at the end of the list.
-type roundRobinNodeIterator struct {
-	baseIterator
-	startIdx int
-}
+func (ti *treeIterator) ForEachNode(f func(*Node) bool) {
```
craigcondit, line 44: This function could use documentation, especially wrt the passed-in function.

pkg/scheduler/objects/node_collection.go
```diff
@@ -154,43 +163,15 @@ func (nc *baseNodeCollection) GetNodes() []*Node {
 // Create an ordered node iterator for unreserved nodes based on the sort policy set for this collection.
 // The iterator is nil if there are no unreserved nodes available.
 func (nc *baseNodeCollection) GetNodeIterator() NodeIterator {
-	return nc.getNodeIteratorInternal(func(node *Node) bool {
-		return !node.IsReserved()
-	})
+	nc.unreservedIterator.Reset()
```
craigcondit, line 166: Remove Reset() call as it doesn't do anything.

pkg/scheduler/objects/node_collection.go
```diff
@@ -154,43 +163,15 @@ func (nc *baseNodeCollection) GetNodes() []*Node {
...
-		return true
-	})
-	metrics.GetSchedulerMetrics().ObserveNodeSortingLatency(sortingStart)
-
-	if len(nodes) == 0 {
-		return nil
-	}
-
-	return NewDefaultNodeIterator(nodes)
+	nc.fullIterator.Reset()
```
craigcondit, line 173: Remove Reset() call as it doesn't do anything.

pkg/scheduler/objects/node_iterator.go
craigcondit: Alternatively, we could compute the tree here and simply iterate the same copy of it each time. We should not do both, however.

craigcondit (changes requested):
Almost there, some minor points and a couple of simplifications / optimizations.

PR #559, opened 2023-05-31
[YUNIKORN-1774] Event cache: misc cleanup

pbacsko:
### What is this PR for?
Misc cleanup in the current event implementation:
- "event system" instead of "event cache"
- remove interfaces
- add extra event sending layer in `Application`
- `atomic.Bool` instead of `atomic.Value`
- initialization: start both the event system and publisher in `entrypoint.go`

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1774

### Questions:


mitdesai (changes requested):
I think we should add as many tests as possible to `events.go` which will be helpful when new code gets introduced as part of YUNIKORN-1628.

Looks good to me otherwise.

pkg/events/event_system.go
```diff
@@ -29,39 +29,45 @@ import (
 // need to change for testing
 var defaultEventChannelSize = 100000
 
-var cache *EventCache
+var ev EventSystem
 
-type EventCache struct {
-	Store EventStore // storing eventChannel
+type EventSystem interface {
```
pbacsko, line 34: Changed this to an interface so that we can mock easier when needed

PR #562, opened 2023-06-09
[YUNIKORN-1799] Create basic ringbuffer implementation

pbacsko:
### What is this PR for?
Add a ring buffer implementation which stores `si.EventRecord` types.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1799

### Questions:


pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,173 @@
...
+
+func reverse(r []*si.EventRecord) {
+	for i, j := 0, len(r)-1; i < j; i, j = i+1, j-1 {
+		r[i], r[j] = r[j], r[i]
+	}
+}
+
+func NewEventRingBuffer(capacity int, eventLifeTime time.Duration) *EventRingBuffer {
+	return &EventRingBuffer{
+		capacity:      capacity,
```
FrankYang0529, line 168: Do we want to check whether `capacity` is a positive integer?

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,173 @@
...
+		events:        make([]*si.EventRecord, capacity),
+		lifetimeNanos: int64(eventLifeTime),
```
FrankYang0529, line 170: We can use `eventLifeTime.Nanoseconds()`.

Ref: https://pkg.go.dev/time#Duration.Nanoseconds

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,173 @@
...
+
+package events
+
+import (
+	"time"
+
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
+const latestUnset int64 = -1 << 63
```
FrankYang0529, line 27: Not sure whether using `const latestUnset int64 = math.MinInt64` is more clear?

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+		i = next
+	}
+
+	return records
+}
+
+// RemoveExpiredEntries discards expired entries from the buffer. Current age of an object depends on the
+// TimestampNano field.
+// This method is expected to be called in regular intervals, which usually runs on a separate goroutine.
```
FrankYang0529, line 116: From this comment, it looks like we will use `EventRingBuffer` in more than one gorutines. Do we want to add a lock when we update elements?

FrankYang0529 (commented):
Added some comments. If I misunderstand anything, feel free to ignore my comments. Thanks!

pkg/events/event_ringbuffer.go
pbacsko: This meant to be an internal implementation (perhaps I should unexport this method for this reason). This number is supposed to be validated earlier when the configuration is read.

pkg/events/event_ringbuffer.go
pbacsko: Will add a comment.

pkg/events/event_ringbuffer.go
pbacsko: We'll see how this type will be integrated with the existing code. Right now it's only a basic implementation. If it's accessed through an existing type which has locks, then it's not necessary to add it here.

pkg/events/event_ringbuffer.go
pbacsko: Done

pkg/events/event_ringbuffer.go
pbacsko: Done

pkg/events/event_ringbuffer.go
pbacsko: Unexported type + functions/methods.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+}
+
+// GetLatestEntries returns the most recent items whose age is younger than the current time minus interval.
+func (e *eventRingBuffer) GetLatestEntries(interval time.Duration) []*si.EventRecord {
+	unixNow := now().UnixNano()
+	startTime := unixNow - interval.Nanoseconds()
+
+	if e.noElements == 0 || startTime > e.latest {
+		return nil
```
zhuqi-lucas, line 96: if we need to have some error handling for these operations? Or we will not handle this as an error?

zhuqi-lucas (approved):
LGTM, left minor qustion about error handling, thanks!

pkg/events/event_ringbuffer.go
pbacsko: No, there is no need for error handling. In a classic ringbuffer, `Queue()` can return an error if the buffer is full, or `Dequeue()` can fail if the buffer is empty. In our case, adding an element never fails and returning nil is perfectly fine.

zhuqi-lucas: It makes sense, thanks! LGTM

FrankYang0529 (approved):
Add a minor comment. Feel free to skip it. LGTM. Thanks!

pbacsko:
Just talked to Craig, it makes no sense to remove expired entries. It's better if we reach steady state in memory usage, so just keep the added elements, there's no harm if we don't remove those.

In fact, I think storing `[]si.EventRecord` is preferable to `[]*si.EventRecord`. The only small issue is the map inside `si.Resource`, that must be copied manually.

craigcondit:
> Plus: an event record needs an unique ID so that clients request events that happened after a certain message. This means that `si.EventRecord` needs to be updated.

I don't think we need to store the ID per record. All we need is the ID of the head of the list. So on startup, headID = 0, and after 1k events, headID = 1000. Then you can infer the ID of any message by it's position within the buffer.

craigcondit:
One other question. I don't see any locking here, is that going to be handled externally?

craigcondit:
> I think storing `[]si.EventRecord` is preferable to `[]*si.EventRecord`. The only small issue is the map inside `si.Resource`, that must be copied manually.

I think using pointers is fine. Since we need to return sublists for client paging, that's a lot of copies. We also aren't really going to get the benefit of CPU cache locality when we have to turn around and convert the lists to JSON for transmission anyway. Thinks like reversing and subslicing a list of pointers is going to be more efficient (as the actual data doesn't need to move) the smaller those lists are (and pointers are as small as they can get).

pbacsko:
> One other question. I don't see any locking here, is that going to be handled externally?

I'm thinking about this one, we probably need it. We can insert this logic to various places, but there's no locking where it makes the most sense. I'll add it in the next iteration.

> I don't think we need to store the ID per record. All we need is the ID of the head of the list. So on startup, headID = 0, and after 1k events, headID = 1000. Then you can infer the ID of any message by it's position within the buffer.

Yep, that makes thing easier. But `EventRecord` still needs to be updated so it can be captured on the client side, right?

craigcondit:
> Yep, that makes thing easier. But `EventRecord` still needs to be updated so it can be captured on the client side, right?

Not necessarily. You could make the client response be a tuple of { startingID, eventRecord[] }, along with anything else useful.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,167 @@
...
+			reverse(records)
+			return records
+		}
+
+		i = e.prev(i)
+	}
+}
+
+// GetLatestEntries returns the most recent items whose age is younger than the current time minus interval.
+func (e *eventRingBuffer) GetLatestEntries(interval time.Duration) []*si.EventRecord {
```
craigcondit, line 92: I think we should remove this method entirely.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,174 @@
...
+			return removed
+		}
+	}
+}
+
+func (e *eventRingBuffer) prev(i int) int {
+	i--
+	if i == -1 {
+		i = e.capacity - 1
+	}
```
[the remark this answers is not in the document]
pbacsko: Modified TestRingBuffer_GetLatestEntriesCount_WhenFull which will cause `i == -1`.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		next := e.next(i)
+		i = next
+	}
+
+	return records
+}
+
+func (e *eventRingBuffer) GetEventsFromPosition(pos int) []*si.EventRecord {
```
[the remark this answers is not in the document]
craigcondit: Also, we should probably just combine this with GetLatestEntriesCount() (i.e. GetLatestEntries(pos, count)). Pos == 0 could be used if all values are requested. Also, if we return ([]*si.EventRecord, bool ok) we can mark if the requested data overflowed (i.e. pos was not within the result set).

craigcondit (changes requested):
See comments.

pbacsko:
Latest changes:
1. Two methods to retrieve elements: one for most recent records and one based on IDs
2. Avoiding for loops, built-in functions (`append()` and `copy()`) are used to construct the new slice
3. We no longer care about timestamps, time, etc.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,213 @@
...
+	total := uint64(0)
+	for _, r := range ranges {
+		total += r.end - r.start
+	}
+
+	src := make([]*si.EventRecord, 0)
+	for _, r := range ranges {
+		events := e.events[r.start:r.end]
+		src = append(src, events...)
+	}
```
wilfred-s, line 164: Can we use copy to speed up this process?
```
	dest := make([]*si.EventRecord, total)
	copy(dest, src[r1.start:])
	copyFInish := r1.end - r1.start
	copy(dest[copyFInish:], src[r2.start:])
```
The second copy is only needed if there are two ranges otherwise the first copy will do it all

wilfred-s (changes requested):
performance optimisations in the event copy

pkg/events/event_ringbuffer.go
pbacsko: Please check, now there's no unnecessary copy.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,213 @@
...
+	}), startIdx, resultOK
+}
+
+func (e *eventRingBuffer) getEntriesFromRanges(ranges ...eventRange) []*si.EventRecord {
+	total := uint64(0)
+	for _, r := range ranges {
+		total += r.end - r.start
+	}
+
+	src := make([]*si.EventRecord, 0)
```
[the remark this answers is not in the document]
pbacsko: Done.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,213 @@
...
+type eventRingBuffer struct {
+	events   []*si.EventRecord
+	capacity uint64
+	idx      uint64
+	full     bool
+
+	id       uint64 // increasing unique id
+	startId  uint64 // id of the message at index 0
+	lowestId uint64 // the lowest available id in the buffer (pointed at by e.idx when full)
+
```
craigcondit, line 55: This seems overly complex. We only really need to know:

- capacity of the buffer (i.e. len(events[])
- current count of events
- head position of buffer
- ID of element at head

That's 3 variables. I count 5 plus a full pointer in the implementation. All others can be calculated trivially from these.

craigcondit: On Add, current count would be incremented, and reset to capacity if it exceeds it.
Head pointer would be incremented and on overflow reset to zero.
ID of element at head is ever-increasing.

craigcondit (changes requested):
I think the code could be simplified quite a bit by eliminating full and computing rather than storing some of the internal state.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,213 @@
...
+		end:   e.idx,
+	}, nil), startIdx, resultOK
+}
+
+func (e *eventRingBuffer) getEntriesFromRanges(r1, r2 *eventRange) []*si.EventRecord {
+	var total uint64
+	var src []*si.EventRecord
+	if r2 == nil {
+		total = r1.end - r1.start
+		src = e.events[r1.start:r1.end]
```
wilfred-s, line 159: No need for a re-slice, this can be a simple one step:
```
if r2 == nil {
	dest := make([]*si.EventRecord, r1.end - r1.start)
	// dest limits what can be copied from src
	copy(dest, src[r1.start:])
	return dest
}
```

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,213 @@
...
+	} else {
+		total = (r1.end - r1.start) + (r2.end - r2.start)
+		src = e.events[r1.start:r1.end]
+		src = append(src, e.events[r2.start:r2.end]...)
```
wilfred-s, line 163: The re-slice points to the same underlying storage. When append is called a new array must be allocated so this does the same thing as before. The sharing of the underlying storage is used in the 2nd copy with the re-slice.

When we have two ranges the fist will be the end of the buffer, the second the start, this is true always.

So this will work, with least allocation overhead. No check needed for r2 being nil as that is already handled:
```
total := (r1.end - r1.start) + (r2.end - r2.start)
dest := make([]*si.EventRecord, total)
// all the way to the end of the src, r1.end should be same as len(src)
copy(dest, src[r1.start:])
copyNext := r1.end - r1.start
// dest limits what can be copied from src 
copy(dest[copyNext:], src[r2.start:])
return dest
```

wilfred-s (changes requested):

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,213 @@
...
+	e.Lock()
+	defer e.Unlock()
+
+	e.events[e.idx] = event
+	if e.idx == 0 {
+		e.startId = e.id
+	}
+	if e.idx == e.capacity-1 {
+		e.full = true
+	}
```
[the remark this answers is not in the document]
pbacsko: Done

pkg/events/event_ringbuffer.go
pbacsko: Done

pkg/events/event_ringbuffer.go
pbacsko: Done

pkg/events/event_ringbuffer.go
pbacsko: Removed `lowestId` and `startId`. They are now calculated on-demand in `id2pos()`.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
+type resultState int
+
+const (
+	resultOK resultState = iota
+	bufferEmpty
+	idNotFound
+)
```
craigcondit, line 33: I think `resultState` makes things complex for clients. Why?

resultOK : Normal return status.
bufferEmpty: This can be represented as an empty result (which is a lot more useful).
idNotFound: Ok, now what? Client has to request an ID that is within the buffer, but it doesn't know which one to use.

I would change the structure of GetEventsFromID to the following:

`func (e *eventRingBuffer) GetEventsFromID(id uint64, count unit64) ([]*si.EventRecord, startID uint64)`

Let `startID` be the ID of the first result returned. If the result is empty, let `startID` be the first ID that should be requested on subsequent calls. So in the empty case, this returns `{ [], 0 }`. For the existing `idNotFound` case, if the ID presented is too low, simply start from the real lowest ID and return the first `count` results (or less if there aren't that many). The fact that `startID` in the result didn't equal the ID requested by the client is proof enough that an overflow occurred. If the ID is too high, return an empty list along with the lowest valid starting ID. Again, this is enough to tell the client that the range it requested was not valid, and also gives the startID that should be requested next. In all cases, the returned `startID` + length of the returned list gives the ID that should be requested for the next page.

We should also add `count` as otherwise, the entire buffer may be returned, so we can't paginate properly.

With these changes, I also think we can remove `GetRecentEntries` as this is largely redundant. Newly started clients will likely want to retrieve all history they are not aware of (in which case calling `GetEventsFromID(0, pageSize)` will return the first set of results) or all history beyond a known state (in which case calling `GetEventsFromID(lastKnownEventID+1, pageSize)` will do the job. For completeness we could add `GetLatestEventID() unit64` which would simply return `id`.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,202 @@
...
+// eventRingBuffer A specialized circular buffer to store event objects.
+//
+// Unlike to regular circular buffers, existing entries are never directly removed and new entries can be added if the buffer is full.
+// In this case, the oldest entry is overwritten and can be collected by the GC.
+// Each event has an ID, however, this mapping is not stored directly. If needed, we calculate the id
+// of the event based on slice positions.
+//
+// Retrieving the records can be achieved with GetEventsFromID and GetRecentEntries.
+type eventRingBuffer struct {
+	events   []*si.EventRecord
```
craigcondit, line 49: Can we add comments for each of these so the meaning of them is clear?

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,202 @@
...
+
+	e.events[e.head] = event
+	if !e.full {
+		e.full = e.head == e.capacity-1
+	}
+	e.head = (e.head + 1) % e.capacity
+	e.id++
+}
+
+func (e *eventRingBuffer) GetEventsFromID(id uint64) ([]*si.EventRecord, resultState) {
```
craigcondit, line 72: See comments above regarding `resultState` vs. `startID` and adding `count` parameter.

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,202 @@
...
+		// never points to e.capacity.
+		end = e.capacity
+	}
+	return e.getEntriesFromRanges(&eventRange{
+		start: pos,
+		end:   end,
+	}, nil), resultOK
+}
+
+func (e *eventRingBuffer) GetRecentEntries(count uint64) ([]*si.EventRecord, uint64, resultState) {
```
craigcondit, line 106: See above; I think we can remove this in favor of changing `GetEventsFromID()` slightly.

craigcondit (changes requested):
Internal implementation looks good. I think the external view needs some tweaking; see comments inline.

pkg/events/event_ringbuffer.go
pbacsko: Done.

I kept `GetRecentEntries()` for the time being, so that Wilfred can share his opinion too. This affects the design described in the document.

pkg/events/event_ringbuffer.go
pbacsko: Done.

pkg/events/event_ringbuffer.go
pbacsko: I'll remove it if Wilfred acks the change.

craigcondit (approved):
+1 looks good to me, pending Wilfred's review as well.

pkg/events/event_ringbuffer.go
wilfred-s: I think we can remove it. If we really need the functionality of `GetRecentEntries` then that would be a simple change in `GetEventsFromID` with `id == 0` as a special case here:
```
	if !idFound && id > 0 {
		return nil, lowest
	}
	if id == 0 {
		pos = e.head
	}
```

pkg/events/event_ringbuffer.go
```diff
@@ -0,0 +1,220 @@
...
+	if e.full && e.head == 0 {
+		// Special case: buffer is full and head is pointing at the beginning.
+		// Need to set explicitly to get [0..capacity] range because e.head
+		// never points to e.capacity.
+		end = e.capacity
+	}
+	return e.getEntriesFromRanges(&eventRange{
+		start: pos,
+		end:   end,
+	}, nil, count), lowest
```
wilfred-s, line 97: We know the count here why are we not enforcing the count here instead of passing this as a limitation to be applied on top of the ranges we have created?
```
func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64) {
	e.RLock()
	defer e.RUnlock()
	lowest := e.getLowestId()

	pos, idFound := e.id2pos(id)
	if !idFound {
		return nil, lowest
	}
	// limit count to the capacity
	count = min(count, e.capacity)
	// possible wrap case full buffer and pos after or on the current head
	if e.full && pos >= e.head {
		// first range never passes the end of the slice
		end := min(pos+count, e.capacity)
		r1 := &eventRange{
			start: pos,
			end:   end,
		}
		// second range only if still events left to fetch
		var r2 *eventRange
		end = pos + count - e.capacity
		if end > 0 {
			// never fetch pass the current head
			end = min(end, e.head)
			r2 = &eventRange{
				start: 0,
				end:   end,
			}
		}
		return e.getEntriesFromR(r1, r2), lowest
	}

	// no wrapping: either limited by head position or by the count
	end := min(pos+count, e.head)
	return e.getEntriesFromR(&eventRange{
		start: pos,
		end:   end,
	}, nil), lowest
}

// min a utility function to return the smallest value of two unsigned int
func min(a, b uint64) uint64 {
	m := a
	if b < a {
		m = b
	}
	return m
}
```

wilfred-s (changes requested):

pkg/events/event_ringbuffer.go
wilfred-s: `getEntriesFromRange()` can then be as simple as it was a couple of days ago without all the count handling and ignoring the second range if limited by count

wilfred-s (approved):
LGTM +1
Thank you for working through this big review

PR #567, opened 2023-06-14
[YUNIKORN-1717] cache sortedQueues

FrankYang0529:
### What is this PR for?

Only need sorting queues if the following occurred:
- Allocated resource changed in one of the child queues (most common)
- Pending resource changed from 0 to "n", or from "n" to 0 (affects filtering)
- Child queue got stopped (affects filtering)
- Child queue structure changed on config update

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1717

### Questions:

### DIFF
--- a/pkg/scheduler/objects/object_state.go
+++ b/pkg/scheduler/objects/object_state.go
@@ -78,11 +78,24 @@ func NewObjectState() *fsm.FSM {
 		},
 		fsm.Callbacks{
 			"enter_state": func(_ context.Context, event *fsm.Event) {
+				var object string
+				switch v := event.Args[0].(type) {
+				case *Queue:
+					object = v.QueuePath
+				case string:
+					object = v
+				}
+
 				log.Logger().Info("object transition",
-					zap.Any("object", event.Args[0]),
+					zap.String("object", object),
 					zap.String("source", event.Src),
 					zap.String("destination", event.Dst),
 					zap.String("event", event.Event))
+			},
+			Stopped.String(): func(_ context.Context, event *fsm.Event) {
+				if queue, ok := event.Args[0].(*Queue); ok {
+					queue.parent.clearSortedQueues()
+				}
 			},
 		},
 	)
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -66,6 +66,7 @@ type Queue struct {
 	preemptionPolicy    policies.PreemptionPolicy // preemption policy
 	preemptionDelay     time.Duration             // time before preemption is considered
 	currentPriority     int32                     // the current scheduling priority of this queue
+	sortedQueues        []*Queue                  // sorted list of queues
 
 	// The queue properties should be treated as immutable the value is a merge of the
 	// parent properties with the config for this queue only manipulated during creation
@@ -105,6 +106,7 @@ func newBlankQueue() *Queue {
 		prioritySortEnabled:    true,
 		preemptionDelay:        configs.DefaultPreemptionDelay,
 		preemptionPolicy:       policies.DefaultPreemptionPolicy,
+		sortedQueues:           nil,
 	}
 }
 
@@ -316,6 +318,7 @@ func (sq *Queue) applyConf(conf configs.QueueConfig) error {
 	}
 
 	sq.properties = conf.Properties
+	sq.sortedQueues = nil
 	return nil
 }
 
@@ -464,7 +467,7 @@ func (sq *Queue) CurrentState() string {
 // handleQueueEvent processes the state event for the queue.
 // The state machine handles the locking.
 func (sq *Queue) handleQueueEvent(event ObjectEvent) error {
-	err := sq.stateMachine.Event(context.Background(), event.String(), sq.QueuePath)
+	err := sq.stateMachine.Event(context.Background(), event.String(), sq)
 	// err is nil the state transition was done
 	if err == nil {
 		sq.stateTime = time.Now()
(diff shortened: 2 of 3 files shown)


pkg/scheduler/objects/object_state.go
```diff
@@ -78,12 +78,25 @@ func NewObjectState() *fsm.FSM {
 		},
 		fsm.Callbacks{
 			"enter_state": func(_ context.Context, event *fsm.Event) {
+				var object string
+				switch v := event.Args[0].(type) {
+				case *Queue:
+					object = v.QueuePath
+				case string:
+					object = v
```
pbacsko, line 86: When is it "Queue" and when is it "string"? It seems to me that we only handle events from a single place so `event.Args[0]` will always be a "Queue" from this point on.

pkg/scheduler/objects/queue.go
```diff
@@ -1823,3 +1853,10 @@ func (sq *Queue) recalculatePriority() int32 {
 	sq.currentPriority = curr
 	return priorityValueByPolicy(sq.priorityPolicy, sq.priorityOffset, curr)
 }
+
+// clearSortedQueues remove sortedQueue property. This is used for child queue get into stopped state.
+func (sq *Queue) clearSortedQueues() {
```
pbacsko, line 1858: I believe that this should be called all the way up to root. If a leaf is stopped, then it affects the entire hierarchy which is above, not just the immediate parent.

pkg/scheduler/objects/queue_test.go
```diff
@@ -616,29 +616,83 @@ func TestSortQueue(t *testing.T) {
...
 	res, err = resources.NewResourceFromConf(map[string]string{"first": "1"})
 	assert.NilError(t, err, "failed to create basic resource")
 	leaf.incPendingResource(res)
-	if queues := parent.sortQueues(); len(queues) != 1 {
+	if queues := parent.getSortedQueues(); len(queues) != 1 {
 		t.Errorf("parent queue did not return expected queues: %v", queues)
 	}
+	if !reflect.DeepEqual(parent.sortedQueues, parent.getSortedQueues()) {
+		t.Errorf("parent queue did not use cached sorted queues: %v", parent)
+	}
```
pbacsko, line 643: I know it's based on existing code, but let's use `assert.Equal()` or `assert.Assert()`, there's no need for `t.Errorf()`. It's much cleaner.

pkg/scheduler/objects/queue.go
```diff
@@ -1050,14 +1069,22 @@ func (sq *Queue) sortApplications(filterApps bool) []*Application {
...
 // Only queues with a pending resource request are considered. The queues are sorted using the
 // sorting type for the parent queue.
 // Lock free call all locks are taken when needed in called functions
-func (sq *Queue) sortQueues() []*Queue {
+func (sq *Queue) getSortedQueues() []*Queue {
 	if sq.IsLeafQueue() {
 		return nil
 	}
+
+	sq.RLock()
```
pbacsko, line 1081: Extract this to a separate internal method, that `defer` right before the return doesn't make any sense.
Like:

```
func (sq *Queue) getCachedSortedQueues() []*Queue {
  sq.RLock()
  defer sq.RUnlock()
  
  return sq.sortedQueues
}
```

pbacsko (changes requested):
Thanks for the PR.

There are some things that need to be addressed.
Also, I'd be happy to see some sort of profiling with a slightly more complicated queue hierarchy (not just "'root.default", at least some parents and 15-20 leafs) to see how this performs.

pkg/scheduler/objects/object_state.go
FrankYang0529: The `PartitionContext` also uses the same state machine. It gives a partition name.

https://github.com/apache/yunikorn-core/blob/68a82820f0d7ff82f4d5d9b44629ae3660a184e1/pkg/scheduler/partition.go#L88

https://github.com/apache/yunikorn-core/blob/68a82820f0d7ff82f4d5d9b44629ae3660a184e1/pkg/scheduler/partition.go#L268

FrankYang0529:
Thanks for the review. I will do some profiling and update my code tomorrow.

pkg/scheduler/objects/object_state.go
pbacsko: Then simply introduce a new state machine for Queue.

craigcondit:
I have some concerns about this. Queue sorting has never really showed up as a hot spot for several reasons. For one thing, there are generally far fewer queues than outstanding requests (if not, the cluster is relatively idle and performance isn't an issue). Sorting of queues also happens only once per scheduling cycle, and if we are able to schedule a task, the sorting needs to be re-done anyway, so the only thing caching the sorted queues gets us is a slight speedup when there is nothing available to schedule anyway.

Some of the changes needed here to make this work (including potentially a new FSM for Queue?) strike me as a bridge too far, given the likely negligible impact on performance. Additionally, changes this invasive are going to make implementing additional features more difficult -- things like multiple partition support come to mind. 

In short, I think we should table this JIRA. If at some point in the future this becomes a bottleneck, we can readdress it.

FrankYang0529:
Hi @pbacsko and @craigcondit, I agree with you. Most of the time, we don't have many queues, and cached results may be stale after allocated/pending resources are updated. Also, including cached result makes the code more complex. I will close the PR and see what we can do in the future. Thanks!

PR #569, opened 2023-06-15
[YUNIKORN-1575] [core] Ability to specify guaranteedResources via namespace annotations

zhuqi-lucas:
### What is this PR for?
The core side change to support ability to specify guaranteedResources via namespace annotations.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1575

### Questions:


wilfred-s (changes requested):
We need to add a test for the guaranteed tag to return an empty string (693) and if all resources are 0 (703) we have those for the max tag should be doable for the guaranteed tag too.

pkg/scheduler/objects/queue.go
```diff
@@ -663,29 +663,61 @@ func (sq *Queue) AddApplication(app *Application) {
...
-		return
+	if quota != "" {
+		quotaRes, quotaErr = resources.NewResourceFromString(quota)
+		if quotaErr != nil {
+			log.Log(log.SchedQueue).Warn("application resource quota conversion failure",
+				zap.String("json quota string", quota),
+				zap.Error(quotaErr))
+		} else if !resources.StrictlyGreaterThanZero(quotaRes) {
+			log.Log(log.SchedQueue).Warn("application resource quota has at least one 0 value: cannot set queue limit",
+				zap.Stringer("maxResource", quotaRes))
```
manirajv06, line 684: Error message bit misleading. Many be "Max resource quantities should be greater than zero" ?

pkg/scheduler/objects/queue.go
```diff
@@ -663,29 +663,61 @@ func (sq *Queue) AddApplication(app *Application) {
 	// YUNIKORN-199: update the quota from the namespace
 	// get the tag with the quota
 	quota := app.GetTag(common.AppTagNamespaceResourceQuota)
-	if quota == "" {
+	// get the tag with the guaranteed resource
+	guaranteed := app.GetTag(common.AppTagNamespaceResourceGuaranteed)
+	if quota == "" && guaranteed == "" {
```
manirajv06, line 668: Is guaranteed mandatory? or Use only when annotation has it?

manirajv06 (changes requested):
Overall looks good. Given some minor comments

pkg/scheduler/objects/queue.go
```diff
@@ -663,29 +663,61 @@ func (sq *Queue) AddApplication(app *Application) {
...
+				zap.String("json guaranteed string", guaranteed),
+				zap.Error(guaranteedErr))
+		} else if !resources.StrictlyGreaterThanZero(guaranteedRes) {
+			log.Log(log.SchedQueue).Warn("application guaranteed resource has at least one 0 value: cannot set queue guaranteed resource",
+				zap.Stringer("guaranteedResource", guaranteedRes))
+			guaranteedRes = nil // Skip setting guaranteed resource if it has a value <= 0
+		}
+	}
+
+	if quotaErr != nil && guaranteedErr != nil {
```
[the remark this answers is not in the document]
zhuqi-lucas: I think we will return when both quotaErr and guaranteedErr, for example:
When quotaErr happen but guaranteedErr not happen, we can set for the guranteed resource seperately, every annotaion can set independently

So we only return here when both quotaErr and guaranteedErr happend.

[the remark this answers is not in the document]
zhuqi-lucas: Sure, will address the return when logging.

pkg/scheduler/objects/queue.go
zhuqi-lucas: Good suggestion, will address it.

pkg/scheduler/objects/queue.go
zhuqi-lucas: No @manirajv06 , it only uses when user annotation it just like quota, every annotation can be used independently.

manirajv06: If we can treat each annotation independently, then we should not put together in a single if checks. We can have separate if checks for each one.

pkg/scheduler/objects/queue.go
```diff
@@ -663,29 +663,60 @@ func (sq *Queue) AddApplication(app *Application) {
...
-		return
+
+	// need to set guaranteed resource: convert json string to resource
+	if guaranteed != "" {
+		guaranteedRes, guaranteedErr = resources.NewResourceFromString(guaranteed)
+		if guaranteedErr != nil {
+			log.Log(log.SchedQueue).Warn("application guaranteed resource conversion failure",
+				zap.String("json guaranteed string", guaranteed),
+				zap.Error(guaranteedErr))
+			if quotaErr != nil {
```
manirajv06, line 696: guaranteedErr, not quotaErr :)

manirajv06 (changes requested):

pkg/scheduler/objects/queue.go
zhuqi-lucas: I think we should use quotaErr here, because we only need to return for both quotaErr != nil and guaranteedErr != nil

manirajv06 (approved):

manirajv06:
looks good. See if you can increase the test coverage.

zhuqi-lucas:
Thank you @manirajv06 for review, i wanted to increase the test coverage here, but it is the log error lines, how can we easily cover the log in unit test?

PR #571, opened 2023-06-16
[YUNIKORN-1801] add allocation events

FrankYang0529:
### What is this PR for?

Add allocation events.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1801

### Questions:


FrankYang0529:
Hi @pbacsko, I am not 100% sure that I added events to correct places. May you help me review it first? If all allocation events are fine , I will add test cases. Thanks!

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,87 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+func (evt *applicationEvents) sendTerminateAllocationAskEvent(request *AllocationAsk, terminationType si.TerminationType) {
+	if !evt.enabled {
+		return
+	}
+
+	var message string
+	var event *si.EventRecord
+	switch terminationType {
+	case si.TerminationType_TIMEOUT:
+		message = fmt.Sprintf("Allocation ask '%s' in application '%s' is removed. Reason: timeout", request.GetAllocationKey(), request.GetApplicationID())
```
pbacsko, line 113: Message is not needed.

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,87 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, message, request.GetAllocationKey(), si.EventRecord_REMOVE, si.EventRecord_REQUEST_TIMEOUT, request.GetAllocatedResource())
+	case si.TerminationType_STOPPED_BY_RM:
+		message = fmt.Sprintf("Allocation ask '%s' in application '%s' is removed. Reason: cancel", request.GetAllocationKey(), request.GetApplicationID())
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, message, request.GetAllocationKey(), si.EventRecord_REMOVE, si.EventRecord_APP_REQUEST, request.GetAllocatedResource())
+	}
+
+	if event != nil {
```
pbacsko, line 120: Make sure we always have a valid event so we don't  need this nil check here.

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,87 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
 	evt.eventSystem.AddEvent(event)
 }
 
+func (evt *applicationEvents) sendSuccessfulAllocationEvent(alloc *Allocation) {
```
pbacsko, line 54: Just `sendAllocationEvent()

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,87 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, message, alloc.GetUUID(), si.EventRecord_REMOVE, si.EventRecord_ALLOC_TIMEOUT, alloc.GetAllocatedResource())
+	case si.TerminationType_PREEMPTED_BY_SCHEDULER:
+		message = fmt.Sprintf("Allocation '%s' in application '%s' is removed. Reason: preempted", alloc.GetUUID(), alloc.GetApplicationID())
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, message, alloc.GetUUID(), si.EventRecord_REMOVE, si.EventRecord_ALLOC_PREEMPT, alloc.GetAllocatedResource())
+	case si.TerminationType_PLACEHOLDER_REPLACED:
+		message := fmt.Sprintf("Placeholder '%s' in application '%s' is removed. Reason: replaced", alloc.GetUUID(), alloc.GetApplicationID())
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, message, alloc.GetUUID(), si.EventRecord_REMOVE, si.EventRecord_ALLOC_REPLACED, alloc.GetAllocatedResource())
+	}
+
+	if event != nil {
```
pbacsko, line 99: Make sure we always have a valid event so we don't need this nil check here.

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,87 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+		evt.eventSystem.AddEvent(event)
+	}
+}
+
+func (evt *applicationEvents) sendTerminateAllocationAskEvent(request *AllocationAsk, terminationType si.TerminationType) {
```
pbacsko, line 104: `sendRemoveAskEvent()`

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,87 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+		message = fmt.Sprintf("Allocation ask '%s' in application '%s' is removed. Reason: cancel", request.GetAllocationKey(), request.GetApplicationID())
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, message, request.GetAllocationKey(), si.EventRecord_REMOVE, si.EventRecord_APP_REQUEST, request.GetAllocatedResource())
+	}
+
+	if event != nil {
+		evt.eventSystem.AddEvent(event)
+	}
+}
+
+func (evt *applicationEvents) sendRemoveApplicationTerminateAllocationAskEvent(request *AllocationAsk) {
```
pbacsko, line 125: `sendCancelAskEvent()`

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,87 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+	if !evt.enabled {
+		return
+	}
+
+	message := fmt.Sprintf("Allocation '%s' in application '%s' is successfully allocated to node '%s'", alloc.GetUUID(), alloc.GetApplicationID(), alloc.GetNodeID())
+	event := events.CreateAppEventRecord(evt.app.ApplicationID, message, alloc.GetUUID(), si.EventRecord_ADD, si.EventRecord_APP_ALLOC, alloc.GetAllocatedResource())
+	evt.eventSystem.AddEvent(event)
+}
+
+func (evt *applicationEvents) sendIncomingAllocationAskEvent(request *AllocationAsk) {
```
pbacsko, line 64: `sendNewAskEvent()`

pkg/scheduler/objects/application.go
```diff
@@ -2009,3 +2017,20 @@ func (sa *Application) SetTimedOutPlaceholder(taskGroupName string, timedOut int
 		sa.placeholderData[taskGroupName].TimedOut = timedOut
 	}
 }
+
+func (sa *Application) SendTerminateAllocationAskEvent(allocKey string) {
```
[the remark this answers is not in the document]
FrankYang0529: Canceling AllocationAsk cause of removing an application is in `PartitionContext.removeApplication`. It calls `app. RemoveAllocationAsk("")`.
https://github.com/apache/yunikorn-core/blob/a6a90bb0408a3004592f06d8a79335ac326dcfe4/pkg/scheduler/partition.go#L375-L382

Removing AllocationAsk cause of releasing AllocationAsk request is in `PartitionContext.removeAllocation`. It calls `app.RemoveAllocationAsk(allocKey)`.
https://github.com/apache/yunikorn-core/blob/a6a90bb0408a3004592f06d8a79335ac326dcfe4/pkg/scheduler/partition.go#L1347-L1369

It's not a good idea to check whether `allocKey` is empty in `Application.RemoveAllocationAsk` and sent related events, so I export the function for `PartitionContext` to use.  I thought about giving another argument like [TerminationType](https://github.com/apache/yunikorn-scheduler-interface/blob/12c630245d861e48c66841a05117bfc55c00506b/lib/go/si/si.pb.go#L30-L34), but we can't tell if the application was removed or the release AllocationAsk was sent by the shim.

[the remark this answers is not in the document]
pbacsko: > It's not a good idea to check whether allocKey is empty in Application.RemoveAllocationAsk and sent related events

Why? An empty string means that we cancel all asks. All we have to do is iterate through `sa.requests` and send the related events. If we have 100 asks, then so be it.

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,71 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+
+func (evt *applicationEvents) sendRemoveAllocationEvent(alloc *Allocation, terminationType si.TerminationType) {
+	if !evt.enabled {
+		return
+	}
+
+	var event *si.EventRecord
+	switch terminationType {
+	case si.TerminationType_UNKNOWN_TERMINATION_TYPE:
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, "", alloc.GetUUID(), si.EventRecord_REMOVE, si.EventRecord_ALLOC_NODEREMOVED, alloc.GetAllocatedResource())
```
pbacsko, line 80: This can be greatly simplified: everything is the same except the change detail. Just set a single variable based on `terminationType`.

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,71 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+}
+
+func (evt *applicationEvents) sendRemoveAskEvent(request *AllocationAsk, terminationType si.TerminationType) {
+	if !evt.enabled {
+		return
+	}
+
+	var event *si.EventRecord
+	switch terminationType {
+	case si.TerminationType_TIMEOUT:
```
pbacsko, line 101: Same here, just set a variable.

pkg/scheduler/objects/events.go
```diff
@@ -50,6 +51,71 @@ func (evt *applicationEvents) sendPlaceholderLargerEvent(ph *Allocation, request
...
+	case si.TerminationType_PREEMPTED_BY_SCHEDULER:
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, "", alloc.GetUUID(), si.EventRecord_REMOVE, si.EventRecord_ALLOC_PREEMPT, alloc.GetAllocatedResource())
+	case si.TerminationType_PLACEHOLDER_REPLACED:
+		event = events.CreateAppEventRecord(evt.app.ApplicationID, "", alloc.GetUUID(), si.EventRecord_REMOVE, si.EventRecord_ALLOC_REPLACED, alloc.GetAllocatedResource())
+	}
+
+	evt.eventSystem.AddEvent(event)
+}
+
+func (evt *applicationEvents) sendRemoveAskEvent(request *AllocationAsk, terminationType si.TerminationType) {
```
pbacsko, line 94: It's ok to have a single method for remove/cancel, just merge the two.

pbacsko (changes requested):
Some simplifications.

pbacsko:
@FrankYang0529 another important thing is to update the tests:
1. Update `events_test.go` based on existing test cases.
2. Update `application_test.go` to see that new events are indeed properly generated.

pkg/scheduler/objects/application.go
```diff
@@ -2009,3 +2018,20 @@ func (sa *Application) SetTimedOutPlaceholder(taskGroupName string, timedOut int
 		sa.placeholderData[taskGroupName].TimedOut = timedOut
 	}
 }
+
+func (sa *Application) sendRemoveAskEvent(allocKey string) {
+	sa.RLock()
+	defer sa.RUnlock()
```
pbacsko, line 2024: This lock must be eliminated because the method is called from a locked call site. If it's needed then you have to create another one without locks.

pkg/scheduler/objects/events.go
FrankYang0529: It looks like we don't have a termination type distinguish between "application removal" and "request removal". Currently, I use `TerminationType_STOPPED_BY_RM` for "request removal". Can I use `TerminationType_UNKNOWN_TERMINATION_TYPE` for "application removal"?

pbacsko: You shouldn't be concerned with application removal, that's a different JIRA ticket. This task is only about allocations/asks.

pkg/scheduler/objects/application.go
FrankYang0529: Removed it. Thanks!

pkg/scheduler/objects/events.go
FrankYang0529: Updated it. Thanks!

pkg/scheduler/objects/events.go
FrankYang0529: Updated it. Thanks.

pkg/scheduler/objects/events.go
FrankYang0529: Updated it. Thanks.

pbacsko:
@FrankYang0529 thanks for the update, the patch now is in very good shape. There are some minor complaints from codecov, please check those, new lines should be covered.
I'll let Wilfred/Mani take a look as well.

FrankYang0529:
> There are some minor complaints from codecov, please check those, new lines should be covered.

* Line `pkg/scheduler/objects/application.go#L1150` is covered by `TestPlaceholderMatch`.
* Line `pkg/scheduler/objects/application.go#L1206` is covered by `TestFailReplacePlaceholder`.

Not sure why CodeCov shows it's not covered.

FrankYang0529:
We can get the correct coverage result if we add `-coverpkg` to the test command like:
```
go test ./... -run TestFailReplacePlaceholder -tags deadlock -coverprofile=coverage.txt -covermode=atomic -coverpkg=`go list ./pkg/... | paste -sd "," -`
```

@pbacsko, Do we want to update it in `Makefile`? Thanks

pbacsko:
> We can get the correct coverage result if we add `-coverpkg` to the test command like:
> 
> ```
> go test ./... -run TestFailReplacePlaceholder -tags deadlock -coverprofile=coverage.txt -covermode=atomic -coverpkg=`go list ./pkg/... | paste -sd "," -`
> ```
> 
> @pbacsko, Do we want to update it in `Makefile`? Thanks

I'll take a note. If everything passes and is OK besides this, we're good to go.

wilfred-s:
> Not sure why CodeCov shows it's not covered.

The tests are not in the objects package that skews the details in the codecov sometimes as the existing test coverage is not counted.
Change looks good.

wilfred-s (approved):
LGTM

PR #573, opened 2023-06-19
[YUNIKORN-1802] Add node events

pbacsko:
### What is this PR for?
Create wrapper type for node events and integrate it with the existing codebase.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1802

### Questions:


pbacsko:
cc @FrankYang0529 you might need some stuff from this PR, like `MockEventSystem`. If this gets merged, you can rebase your PR on this.

pkg/scheduler/objects/node.go
```diff
@@ -27,6 +27,7 @@ import (
 	"go.uber.org/zap"
```
pbacsko, line 27: General comment: we don't have event change details for placeholder replacement and reservation/unreservation. I don't know if we need those or not, I assume the replace would be useful.

pkg/events/events.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
+const Empty = ""
+
```
wilfred-s, line 29: Move to the common package

pkg/scheduler/objects/node.go
wilfred-s: Placeholders: no that is just another allocation for the node. Nothing special from that side.
Reservation: file a follow up jira to look at that. We use a reservation in different ways specially in preemption. We need to think where we report it (app or node) and how we report it.

wilfred-s (changes requested):

wilfred-s:
PR needs a rebase after other event changes have been committed.

pkg/scheduler/partition_test.go
```diff
@@ -355,15 +355,15 @@ func TestRemoveNodeWithAllocations(t *testing.T) {
 	// wait for events to be processed
 	err = common.WaitFor(10*time.Millisecond, time.Second, func() bool {
 		fmt.Printf("checking event length: %d\n", eventSystem.Store.CountStoredEvents())
-		return eventSystem.Store.CountStoredEvents() == 1
+		return eventSystem.Store.CountStoredEvents() == 4 // 3 node alloc + 1 app event
```
pbacsko, line 358: Created a follow-up to address this: https://issues.apache.org/jira/browse/YUNIKORN-1833.
The tests are becoming unmaintainable, this cannot be tested here.

pkg/scheduler/objects/node.go
pbacsko: Created the follow-up.

pkg/scheduler/objects/node.go
```diff
@@ -584,4 +592,13 @@ func (sn *Node) SetReady(ready bool) {
 	sn.Lock()
 	defer sn.Unlock()
 	sn.ready = ready
+	sn.nodeEvents.sendNodeReadyChangedEvent(sn.ready)
+}
+
+func (sn *Node) SendNodeAddedEvent() {
+	sn.nodeEvents.sendNodeAddedEvent()
```
wilfred-s, line 599: Should this not happen under a read lock? (same for remove)

wilfred-s (changes requested):
one minor point on locking

PR #574, opened 2023-06-20
[YUNIKORN-1803] Add queue events

pbacsko:
### What is this PR for?
Add queue events to the core scheduler.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1803

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -32,6 +32,7 @@ import (
 	"github.com/apache/yunikorn-core/pkg/common/configs"
 	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/apache/yunikorn-core/pkg/common/security"
+	"github.com/apache/yunikorn-core/pkg/events"
 	"github.com/apache/yunikorn-core/pkg/log"
 	"github.com/apache/yunikorn-core/pkg/metrics"
 	"github.com/apache/yunikorn-core/pkg/scheduler/objects/template"
@@ -66,6 +67,7 @@ type Queue struct {
 	preemptionPolicy    policies.PreemptionPolicy // preemption policy
 	preemptionDelay     time.Duration             // time before preemption is considered
 	currentPriority     int32                     // the current scheduling priority of this queue
+	initialized         bool                      // whether the queue is fully initialized or not
 
 	// The queue properties should be treated as immutable the value is a merge of the
 	// parent properties with the config for this queue only manipulated during creation
@@ -83,6 +85,7 @@ type Queue struct {
 	runningApps            uint64
 	allocatingAcceptedApps map[string]bool
 	template               *template.Template
+	queueEvents            *queueEvents
 
 	sync.RWMutex
 }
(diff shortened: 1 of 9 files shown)


pkg/scheduler/partition.go
```diff
@@ -196,6 +196,7 @@ func (pc *PartitionContext) addQueue(conf []configs.QueueConfig, parent *objects
 		if err != nil {
 			return err
 		}
+		thisQueue.SendNewQueueEvent()
```
manirajv06, line 199: I am not really sure about triggering new queue event from this place as there are some other places where this needs to be triggered. Instead of doing this in 'n' number of places, I think it is better to trigger this even after queue creation in corresponding constructors in `Queue` itself. For example, NewConfiguredQueue and NewDynamicQueue

pkg/scheduler/objects/queue_events.go
```diff
@@ -0,0 +1,127 @@
...
+	if !q.enabled {
+		return
+	}
+
+	var detail si.EventRecord_ChangeDetail
+	if q.queue.IsManaged() {
+		detail = si.EventRecord_DETAILS_NONE
+	} else {
+		detail = si.EventRecord_QUEUE_DYNAMIC
+	}
```
manirajv06, line 43: detail could be assigned with most frequently used value by default and change it only when necessary.

pkg/scheduler/objects/queue_events.go
```diff
@@ -0,0 +1,127 @@
...
+	if !q.enabled {
+		return
+	}
+
+	var message string
+	if q.queue.isLeaf {
+		message = "leaf queue: true"
+	} else {
+		message = "leaf queue: false"
+	}
```
manirajv06, line 114: same like before

manirajv06 (changes requested):
Overall looks good. Given some minor comments.

pkg/scheduler/objects/queue_events.go
```diff
@@ -0,0 +1,127 @@
...
+type queueEvents struct {
+	enabled     bool
+	eventSystem events.EventSystem
+	queue       *Queue
+}
+
+func (q *queueEvents) sendNewQueueEvent() {
+	if !q.enabled {
+		return
+	}
```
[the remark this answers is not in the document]
pbacsko: This will change due to #579. Not sure if this or that PR goes in first, but let's keep this as it is for now.

pkg/scheduler/partition.go
pbacsko: I was thinking about this before, it seemed to be a better idea because it makes testing slightly easier (no extra event when creating queues). I can check it and see how much changes is needed.

pkg/scheduler/objects/queue.go
```diff
@@ -296,12 +302,18 @@ func (sq *Queue) applyConf(conf configs.QueueConfig) error {
...
 	}
 
+	prevLeaf := sq.isLeaf
 	sq.isLeaf = !conf.Parent
 	// Make sure the parent flag is set correctly: config might expect auto parent type creation
 	if len(conf.Queues) > 0 {
 		sq.isLeaf = false
 	}
 
+	if prevLeaf != sq.isLeaf && sq.initialized {
```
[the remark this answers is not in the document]
pbacsko: Let me check, it looks reasonable...

pkg/scheduler/partition.go
manirajv06: Ok, see if you can address now or file a new jira

pkg/scheduler/objects/queue_events.go
manirajv06: I think you missed this place.

manirajv06 (changes requested):

manirajv06 (approved):

PR #581, opened 2023-06-26
[YUNIKORN-1800] Extend Application event wrapper with new events

lixmgl:
### What is this PR for?
Extend Application event wrapper with new events.
Generated event based on "Generic events" from [Historical Usage Tracking](https://docs.google.com/document/d/1g4tzGNyiSzIecPlQUI0R2rcFs6bBvHhf0xSjZoNJBX4/edit?pli=1#heading=h.idw8khlh4ag1)

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1800

Tested locally
make test passed

### Questions:


pbacsko (changes requested):
Two things right now:
1. Fix the unit tests, it's very likely that you're calling locked methods while already holding a lock
2. Add new test cases to `application_test.go`

lixmgl:
> Two things right now:
> 
> 1. Fix the unit tests: you're calling a locked method `Application.GetAllocatedResource()` while already having the lock
> 2. Add new test cases to `application_test.go`

Yes @pbacsko will update the unit tests in next round.
Question for State change: New: Since NEW is the initial state
Do we need to send event when state change: new?

pbacsko:
> 
> Yes @pbacsko will update the unit tests in next round. Question for State change: New: Since NEW is the initial state Do we need to send event when state change: new?

It's mentioned in the design document, but I feel it's safe to skip it. New application means that its state is New, so this doesn't give us new information.

lixmgl:
> > Yes @pbacsko will update the unit tests in next round. Question for State change: New: Since NEW is the initial state Do we need to send event when state change: new?
> 
> It's mentioned in the design document, but I feel it's safe to skip it. New application means that its state is New, so this doesn't give us useful information.

sounds good. thanks!

pbacsko:
Now the tests are failing in `partition_test.go` because we have new events... it's painful.
You either fix it yourself or wait for YUNIKORN-1833.

Another thing to consider is to comment out or remove the event verification in those tests temporarily in this PR and the whole thing will be resolved in YUNIKORN-1833.

pbacsko:
@lixmgl you're now a committer, you can review & approve changes here:
https://github.com/apache/yunikorn-core/pull/583

It's straightforward and I already discussed this with Wilfred. You don't even need to rebase this PR. Just give me a +1.

lixmgl:
> Now the tests are failing in `partition_test.go` because we have new events... it's painful. You either fix it yourself or wait for YUNIKORN-1833.
> 
> Another thing to consider is to comment out or remove the event verification in those tests temporarily in this PR and the whole thing will be resolved in YUNIKORN-1833.
> 
> I'll think about what's best and add a comment.

Yes, I commented locally and it works

lixmgl:
> @lixmgl you're now a committer, you can review & approve changes here: #583
> 
> It's straightforward and I already discussed this with Wilfred. You don't even need to rebase this PR. Just give me a +1.
> 
> I think I'll add a follow-up to ensure converage for `REQUEST` events (and coverage in general), but for now, it's enough to unblock you.

Sounds good, R+ on 583

pbacsko:
Build keeps failing because it checks out an earlier commit id. Let's rebuild later, hopefully it'll pick up the changes.

craigcondit:
If you rebase and force push to this PR, it should work.

craigcondit:
I just rebased on master and force-pushed. Hopefully this will check properly.

craigcondit:
Still failing (as it does locally too). Master branch succeeds, so it seems the failures are related to the patch.

pbacsko:
> Still failing (as it does locally too). Master branch succeeds, so it seems the failures are related to the patch.

Correct, this time it's different test cases. I created [YUNIKORN-1846](https://issues.apache.org/jira/browse/YUNIKORN-1846) to address this. @lixmgl please wait until YUNIKORN-1846.

lixmgl:
> > Still failing (as it does locally too). Master branch succeeds, so it seems the failures are related to the patch.
> 
> Correct, this time it's different test cases. I created [YUNIKORN-1846](https://issues.apache.org/jira/browse/YUNIKORN-1846) to address this. @lixmgl please wait until YUNIKORN-1846 is done. I'll ping you once there is a PR.

sounds good, thanks!

lixmgl:
hi @pbacsko @craigcondit @wilfred-s Please review the pr, thanks.

pkg/scheduler/objects/application_state_test.go
```diff
@@ -19,15 +19,22 @@
...
 
 	"gotest.tools/v3/assert"
 
 	"github.com/apache/yunikorn-core/pkg/common"
+	"github.com/apache/yunikorn-core/pkg/events"
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
 func TestAcceptStateTransition(t *testing.T) {
+	events.CreateAndSetEventSystem()
```
pbacsko, line 34: It's too much change of the existing tests. 
I suggest that you create a new test case which triggers all state transitions then you verify the events. That results in a more compact and likely smaller change.

pkg/scheduler/objects/application_test.go
```diff
@@ -269,26 +269,19 @@ func TestAppAllocReservation(t *testing.T) {
...
-	assert.Equal(t, ask.applicationID, allocAskCancelRecord.ObjectID, "incorrect object ID, expected application ID")
-	assert.Equal(t, ask.allocationKey, allocAskCancelRecord.ReferenceID, "incorrect reference ID, expected alloc ask ID")
-	assert.Equal(t, si.EventRecord_REMOVE, allocAskCancelRecord.EventChangeType, "incorrect change type, expected remove")
-	assert.Equal(t, si.EventRecord_REQUEST_CANCEL, allocAskCancelRecord.EventChangeDetail, "incorrect change detail, expected new alloc ask")
+	assert.Equal(t, 5, len(records), "expecting 5 events")
+	isNewApplicationEvent(t, app, records[0])
+	isStateChangeEvent(t, app, si.EventRecord_APP_ACCEPTED, records[1])
+	isNewAllocAskEvent(t, ask, records[2])
+	isAllocCancelEvent(t, ask, records[3])
+	isStateChangeEvent(t, app, si.EventRecord_APP_COMPLETING, records[4])
```
pbacsko, line 284: Remove this completely.
We have the relevant checks in `TestAllocationEvents` and `TestAskEvents`.

pkg/scheduler/objects/application_test.go
```diff
@@ -385,20 +378,17 @@ func TestAddAllocAsk(t *testing.T) {
...
-	record := records[0]
-	assert.Equal(t, si.EventRecord_APP, record.Type, "incorrect event type, expect app")
-	assert.Equal(t, appID1, record.ObjectID, "incorrect object ID, expected application ID")
-	assert.Equal(t, aKey, record.ReferenceID, "incorrect reference ID, expected placeholder alloc ID")
-	assert.Equal(t, si.EventRecord_ADD, record.EventChangeType, "incorrect change type, expected add")
-	assert.Equal(t, si.EventRecord_APP_REQUEST, record.EventChangeDetail, "incorrect change detail, expected app request")
+	assert.Equal(t, 3, len(records), "expecting 3 events")
+	isNewApplicationEvent(t, app, records[0])
+	isStateChangeEvent(t, app, si.EventRecord_APP_ACCEPTED, records[1])
+	isNewAllocAskEvent(t, ask, records[2])
```
pbacsko, line 391: You can remove this completely.
We have the relevant checks in `TestAllocationEvents` and `TestAskEvents`.

pkg/scheduler/objects/application_test.go
```diff
@@ -1990,17 +1980,16 @@ func TestAskEvents(t *testing.T) {
...
-	assert.Equal(t, si.EventRecord_APP_REQUEST, records[0].EventChangeDetail)
-	assert.Equal(t, si.EventRecord_APP, records[1].Type)
-	assert.Equal(t, si.EventRecord_REMOVE, records[1].EventChangeType)
-	assert.Equal(t, si.EventRecord_REQUEST_CANCEL, records[1].EventChangeDetail)
+	assert.Equal(t, 5, len(records), "number of events")
+	isNewApplicationEvent(t, app, records[0])
+	isStateChangeEvent(t, app, si.EventRecord_APP_ACCEPTED, records[1])
+	isNewAllocAskEvent(t, ask, records[2])
+	isAllocCancelEvent(t, ask, records[3])
+	isStateChangeEvent(t, app, si.EventRecord_APP_COMPLETING, records[4])
```
pbacsko, line 1992: All these new changes should not be needed... Too much noise. These are tested in a new test (which I suggested above).

There are multiple things we can do, but I believe the simplest is to use a function as variables so that can be replaced with nop:
```
func (sa *Application) HandleApplicationEvent(event applicationEvent) error {
	sa.handleAppEventFn(event)
}

// HandleApplicationEventWithInfo handles the state event for the application with associated info object.
// The application lock is expected to be held.
func (sa *Application) HandleApplicationEventWithInfo(event applicationEvent, eventInfo string) error {
	sa.handleAppEventWithInfoFn(event)
}
```

In both cases, the functions are variables inside `Application`. By default, they point to the current implementation:
```
type Application struct {
    ...
	rmEventHandler              handler.EventHandler
	rmID                        string
	terminatedCallback          func(appID string)
	appEvents                   *applicationEvents
	handleAppEventFn            func(applicationEvent)
    handleAppEventWithInfoFn    func(applicationEvent,string)
	sync.RWMutex
}

func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eventHandler handler.EventHandler, rmID string) *Application {
...
   app.handleAppEventFn = app.handleAppEvent
   app.handleAppEventWithInfoFn = app.handleAppEventWithInfoFn
...
}
```

The new methods `handleAppEvent()` and `handleAppEventWithInfoFn()` are the same as their exported counterpart, so you move the current implementation to these new methods which start with a lowercase. 

In the tests, you simply replace these with dummy nop functions so no transition events will be generated.

pbacsko (changes requested):
There is enough room for simplification. See comments.

pkg/scheduler/objects/application_state_test.go
lixmgl: Done. Thanks.

pkg/scheduler/objects/application_test.go
lixmgl: Done.

pkg/scheduler/objects/application_test.go
lixmgl: Done.

pkg/scheduler/objects/application_test.go
lixmgl: That's a good idea. But it seems disabling the state transits in test will break some of the existing tests. Instead, I added a bool to enable/disable sending app state event which we can use to disabling state change event in test. Thanks.

pkg/scheduler/objects/queue_test.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	"gotest.tools/v3/assert"
 
-	"github.com/apache/yunikorn-core/pkg/common"
+	pkg_common "github.com/apache/yunikorn-core/pkg/common"
```
pbacsko, line 31: We have another common from SI, but that's referenced as `siCommon`, is this necessary?

pkg/scheduler/objects/application.go
```diff
@@ -198,9 +200,18 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
...
 	return app
 }
 
+func (sa *Application) SetEnableAppStateEvent(enable bool) {
+	sa.enableAppStateEvent = enable
+}
+
+func (sa *Application) ResetAppEvents() {
+	sa.appEvents = newApplicationEvents(sa, events.GetEventSystem())
+}
```
pbacsko, line 213: Do not export these two methods. In fact, these are only called from unit tests, so I suggest that you move them, to `application_test.go` (see examples like `addPlaceholderDataWithLocking()`, `handleApplicationEventWithLocking()`). Obviously `enableAppStateEvent` must stay as a member variable.

pbacsko (changes requested):
OK, this is much better now.
There are a lot of complaints from the linters, plus I have some comments. Please address those.

pbacsko:
IMO you should update the PR with master changes otherwise depguard will keep complaining.

pkg/scheduler/objects/application.go
lixmgl: Makes sense. Thanks.

pkg/scheduler/objects/application.go
```diff
@@ -156,7 +157,7 @@ func (sa *Application) GetApplicationSummary(rmID string) *ApplicationSummary {
 	return appSummary
 }
 
-func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eventHandler handler.EventHandler, rmID string) *Application {
+func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eventHandler handler.EventHandler, rmID string, disableNewAppEvent ...bool) *Application {
```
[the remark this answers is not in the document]
lixmgl: Done. Thanks.

pkg/scheduler/objects/queue_test.go
lixmgl: Yes, this is not needed. Thanks.

pbacsko:
+1 LGTM
will merge soon

PR #587, opened 2023-07-03
[YUNIKORN-1835] User belong to zero or multiple groups

manirajv06:
### What is this PR for?

User may belong to zero or multiple groups. On the other hand, Limits are configured for different groups at different queue hierarchy. Among these multiple groups stored in security.UserGroup, matching against group for which limit has been configured happens from leaf to root and first matching group would be picked and used as user's group for all activities in UGM module.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1835

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -41,6 +41,7 @@ type Manager struct {
 	groupTrackers             map[string]*GroupTracker
 	userWildCardLimitsConfig  map[string]*LimitConfig // Hold limits settings of user '*'
 	groupWildCardLimitsConfig map[string]*LimitConfig // Hold limits settings of group '*'
+	configuredGroups          []string                // Hold groups only for which limit has been configured in bottom up queue hierarchical order starting from leaf to root.
 	sync.RWMutex
 }
 
@@ -98,7 +99,7 @@ func (m *Manager) IncreaseTrackedResource(queuePath, applicationID string, usage
 	if err := m.ensureGroupTrackerForApp(queuePath, applicationID, user); err != nil {
 		return false
 	}
-	group, err := m.getGroup(user)
+	group, err := m.ensureGroup(user)
 	if err != nil {
 		return false
 	}
@@ -160,7 +161,7 @@ func (m *Manager) DecreaseTrackedResource(queuePath, applicationID string, usage
 		delete(m.userTrackers, user.User)
 	}
 
-	group, err := m.getGroup(user)
+	group, err := m.ensureGroup(user)
 	if err != nil {
 		return false
 	}
@@ -255,7 +256,7 @@ func (m *Manager) ensureGroupTrackerForApp(queuePath string, applicationID strin
 	userTracker := m.userTrackers[user.User]
 	if !userTracker.hasGroupForApp(applicationID) {
 		var groupTracker *GroupTracker
-		group, err := m.getGroup(user)
+		group, err := m.internalEnsureGroup(user)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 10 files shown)


wilfred-s (changes requested):
How does this take into account the hierarchy?
I would expect a similar setup as for the wildcard: a map key of `string`, queue path, containing a `[]string` which represents the groups in order of the config.
This cannot be a single slice as the hierarchy can split and not have any groups at any level.
When adding a new application the group check is performed starting at the leaf. The slice, if defined, is checked start to finish and the first group matched is returned. If nothing matches the level above is checked, until the root is checked.

manirajv06:
user and group linkage has been stored in appGroupTrackers map only for the first activity. for successive activities with the  same application, already linked group would be used. 

When new queue config doesn't contain old group limit settings, all user linked to that removed group through appGroupTrackers map earlier needs to be set with "nil" value. But that removed group resource usage cannot be decreased accordingly because there is no accountability of app wise resource usage with in group. Have explained this limitations as "to do" in https://github.com/apache/yunikorn-core/pull/587/commits/c944be6b4f71215b4f1779e239628c7fa5a8d570#diff-54ba4ed0ff6a89621a3c9bccc59177673c12f38b6bbce5acdf5df8e702e0b5b2R483.

manirajv06:
Filed https://issues.apache.org/jira/browse/YUNIKORN-1858 to handle the group resource usage during config changes.

pkg/scheduler/ugm/manager.go
```diff
@@ -438,58 +475,100 @@ func (m *Manager) clearEarlierSetLimits(userLimits map[string]bool, groupLimits
...
+							zap.String("application id", appID),
+							zap.String("queue path", queuePath))
+						// removing the linkage only happens here by setting it to nil and deleting app id
+						// but group resource usage so far remains as it is because we don't have app id wise resource usage with in group as of now.
+						// YUNIKORN-1858 handles the group resource usage properly
+						// In case of only one (last) application, group tracker would be removed from the manager.
+						ut.setGroupForApp(appID, nil)
+						gt.removeApp(appID)
+						if len(gt.getTrackedApplications()) == 0 {
+							log.Log(log.SchedUGM).Debug("Is this app the only running application in group?",
```
pbacsko, line 496: Nit: to me this is a weird sounding log message. The answer is a yes/no, but that's actually not logged. I suggest printing a statement here.

pkg/scheduler/ugm/manager.go
```diff
@@ -363,12 +388,24 @@ func (m *Manager) internalProcessConfig(cur configs.QueueConfig, queuePath strin
...
 			}
+			m.configuredGroups[queuePath] = append(m.configuredGroups[queuePath], group)
 		}
 	}
 	if err := m.clearEarlierSetLimits(userLimits, groupLimits, queuePath); err != nil {
 		return err
 	}
 
+	// clear earlier set matched group of user for which limits have been configured earlier and even now.
+	// Matched Group would be set as part of increase calls when user carriers out any activity next time
```
pbacsko, line 399: Nit: "carriers out"

pkg/scheduler/ugm/utilities.go
```diff
@@ -36,3 +36,16 @@ func getChildQueuePath(queuePath string) (string, string) {
 	}
 	return childQueuePath, childQueuePath[:idx]
 }
+
+func getParentQueuePath(queuePath string) (string, string) {
```
pbacsko, line 40: Please add a short unit test to `utilities_test.go`.

wilfred-s (changes requested):
There is a basic issue with the tracking, enforcement and group resolution.
On calling IncreaseTrackedResource we create a new user if it does not exist. We also know based on the user object if the application is new or not. Only for a new application we resolve a group. That can be a named group, wildcard or none. That will create a new group tracker if needed. It will always add an entry in the appGroupTrackers for the user. It either points to an object or is a nil but the key (application ID) exists in the map. Next time only the key existence is checked. That is how we know if the app is new or not.
Even the new group tracker creation should be triggered there if it does not exist at that point time. All that should happen before increasing the usage of the user or group.

pkg/scheduler/ugm/user_tracker.go
```diff
@@ -51,7 +54,28 @@ func newUserTracker(user string) *UserTracker {
...
+		log.Log(log.SchedUGM).Debug("Increasing resource usage for group",
+			zap.String("group", gt.getName()),
+			zap.String("queue path", queuePath),
+			zap.String("application", applicationID),
+			zap.Stringer("resource", usage))
+		increasedGroupUsage := gt.increaseTrackedResource(queuePath, applicationID, usage)
+		if !increasedGroupUsage {
+			_, decreased := ut.queueTracker.decreaseTrackedResource(queuePath, applicationID, usage, false)
+			if !decreased {
+				log.Log(log.SchedUGM).Error("Problem in increasing group resource usage. Hence tried to rollback user tracked usage, but failed to do.",
```
wilfred-s, line 69: NIT: fix to make the message more relevant to log

wilfred-s (approved):
LGTM
improvements to the messages for the two entries that are tagged should be added in a next PR that touches the code

pkg/scheduler/ugm/manager.go
wilfred-s: fixing in the next step

PR #588, opened 2023-07-11
[YUNIKORN-1859] add omitempty

FrankYang0529:
### What is this PR for?

Use "omitempty" in json REST object definition.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1859

### Questions:


pkg/webservice/state_dump.go
```diff
@@ -37,16 +37,16 @@ const (
...
+	Timestamp        int64                            `json:",omitempty"`
+	Partitions       []*dao.PartitionInfo             `json:",omitempty"`
+	Applications     []*dao.ApplicationDAOInfo        `json:",omitempty"`
+	AppHistory       []*dao.ApplicationHistoryDAOInfo `json:",omitempty"`
+	Nodes            []*dao.NodesDAOInfo              `json:",omitempty"`
+	ClusterInfo      []*dao.ClusterDAOInfo            `json:",omitempty"`
+	ContainerHistory []*dao.ContainerHistoryDAOInfo   `json:",omitempty"`
+	Queues           []dao.PartitionQueueDAOInfo      `json:",omitempty"`
+	RMDiagnostics    map[string]interface{}           `json:",omitempty"`
+	LogLevel         string                           `json:",omitempty"`
```
wilfred-s, line 49: we should set the json name for these fields inline with all our other objects that we create

wilfred-s (changes requested):
minor comment on the state dump

pkg/webservice/state_dump.go
FrankYang0529: Hi @wilfred-s, thanks for the review. We didn't set json name, so the output key is like `Timestamp`, not `timestamp`. Do we want to add `timestamp` as json name, or `Timestamp`? Thank you.

FrankYang0529: Updated it. Thank you.

wilfred-s (approved):
LGTM

PR #589, opened 2023-07-12
[YUNIKORN-1804] Add REST endpoint for batch event retrieval

pbacsko:
### What is this PR for?
Add new REST endpoints so that users can retrieve events in a batch.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1804

### Questions:

### DIFF
--- a/pkg/events/event_ringbuffer.go
+++ b/pkg/events/event_ringbuffer.go
@@ -69,7 +69,7 @@ func (e *eventRingBuffer) Add(event *si.EventRecord) {
 func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64, uint64) {
 	e.RLock()
 	defer e.RUnlock()
-	lowest := e.getLowestId()
+	lowest := e.getLowestID()
 
 	pos, idFound := e.id2pos(id)
 	if !idFound {
@@ -159,7 +159,7 @@ func (e *eventRingBuffer) id2pos(id uint64) (uint64, bool) {
 	var calculatedID uint64 // calculated ID based on index values
 	if pos > e.head {
 		diff := pos - e.head
-		calculatedID = e.getLowestId() + diff
+		calculatedID = e.getLowestID() + diff
 	} else {
 		pId := e.id - 1
 		idAtZero := pId - (pId % e.capacity) // unique id at slice position 0
@@ -184,8 +184,8 @@ func (e *eventRingBuffer) id2pos(id uint64) (uint64, bool) {
 	return pos, true
 }
 
-// getLowestId returns the current lowest available id in the buffer.
-func (e *eventRingBuffer) getLowestId() uint64 {
+// getLowestID returns the current lowest available id in the buffer.
+func (e *eventRingBuffer) getLowestID() uint64 {
 	if !e.full {
 		return 0
 	}
--- a/pkg/events/event_ringbuffer.go
+++ b/pkg/events/event_ringbuffer.go
@@ -66,14 +66,14 @@ func (e *eventRingBuffer) Add(event *si.EventRecord) {
 // identifier is returned which can be used to get the first batch.
 // If the caller does not want to pose limit on the number of events returned, "count" must be set to a high
 // value, e.g. math.MaxUint64.
-func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64) {
+func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64, uint64) {
 	e.RLock()
 	defer e.RUnlock()
 	lowest := e.getLowestId()
 
 	pos, idFound := e.id2pos(id)
 	if !idFound {
-		return nil, lowest
+		return nil, lowest, e.getLastEventID()
 	}
 	// limit count to the capacity
 	count = min(count, e.capacity)
@@ -96,7 +96,7 @@ func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventR
 				end:   end,
 			}
 		}
-		return e.getEntriesFromRanges(r1, r2), lowest
+		return e.getEntriesFromRanges(r1, r2), lowest, e.getLastEventID()
 	}
 
 	// no wrapping: either limited by head position or by the count
@@ -104,7 +104,7 @@ func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventR
 	return e.getEntriesFromRanges(&eventRange{
 		start: pos,
 		end:   end,
-	}, nil), lowest
+	}, nil), lowest, e.getLastEventID()
 }
 
 // min a utility function to return the smallest value of two unsigned int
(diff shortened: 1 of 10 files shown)


pkg/webservice/handlers.go
```diff
@@ -869,3 +891,90 @@ func getGroupResourceUsage(w http.ResponseWriter, r *http.Request) {
...
+	getEvents(w, r, nil)
+}
+
+func getEvents(w http.ResponseWriter, r *http.Request, filter func(record *si.EventRecord) bool) {
+	writeHeaders(w)
+	eventSystem := events.GetEventSystem()
+	if eventSystem == nil {
+		buildJSONErrorResponse(w, "Event system is disabled", http.StatusBadRequest)
+		return
+	}
```
pbacsko, line 928: This will be affected by #579

pkg/webservice/handlers.go
```diff
@@ -869,3 +891,90 @@ func getGroupResourceUsage(w http.ResponseWriter, r *http.Request) {
 		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
 	}
 }
+
+func getEventsByType(w http.ResponseWriter, r *http.Request) {
```
craigcondit, line 895: I think this may be problematic. Since we filter after retrieving a page of results, there's no way for the client to know what ID to request next (a filtered page will look the same as the last page with < {pageSize} entries). It may be better to have the clients do their own filtering. Alternatively, we need to provide metadata in the response of each ID along with its associated message. Otherwise it's impossible for a client to reliably interpret the results.

craigcondit: Also, this could catastrophically fail if all results were filtered out. Imagine if ID 1-1000 are all of the wrong type; the client would request a page of 1000 entries starting with entry 1, and interpret the empty result as end-of-stream. On the next request, it would re-request the same ID and get the same results, leading to deadlock.

craigcondit (changes requested):
See comments regarding filtering. I think this needs some design work.

pkg/webservice/handlers.go
pbacsko: What about providing both startId and endId? Or perhaps the number of filtered entries.

craigcondit: That's better than nothing but doesn't give any ability to correlate messages with the ID they were generated with. Also, if a client is interested in more than one filtered data stream, they now need to query for all of them individually resulting in more work server-side. Why not just filter on the client? Data transfer is cheaper than CPU time.

pbacsko:
I removed the filtering from the handlers. We can put it back later if it's needed.

pkg/events/event_ringbuffer.go
```diff
@@ -116,9 +116,20 @@ func min(a, b uint64) uint64 {
...
-	return e.id
+	e.RLock()
+	defer e.RUnlock()
+	return e.getLastEventID()
+}
+
+// unlocked version of GetLastEventID()
+func (e *eventRingBuffer) getLastEventID() uint64 {
+	if e.id == 0 {
+		return 0
```
pbacsko, line 130: Special case: no queues, no apps, no nodes, no nothing. It will never happen in real life (we can't return -1 due to "id" being uint64).

pkg/webservice/handlers.go
pbacsko: I removed the filtering.

pkg/webservice/handlers.go
```diff
@@ -869,3 +870,53 @@ func getGroupResourceUsage(w http.ResponseWriter, r *http.Request) {
...
+			}
+			if i < 0 {
+				buildJSONErrorResponse(w, fmt.Sprintf("Illegal id: %d", i), http.StatusBadRequest)
+				return
+			}
+			fromId = uint64(i)
+		}
+	}
+
+	records, lowestId, highestID := eventSystem.GetEventsFromId(fromId, count)
```
craigcondit, line 913: Nit: use lowestID to match highestID.

craigcondit (approved):
+1 (tiny nit regarding variable name, will fix on commit).

pkg/events/event_system.go
```diff
@@ -28,38 +28,46 @@ import (
...
 var defaultEventChannelSize = 100000
+var defaultRingBufferSize uint64 = 100000
 
 var ev EventSystem
 
 type EventSystem interface {
 	AddEvent(event *si.EventRecord)
 	StartService()
 	Stop()
+	GetEventsFromId(uint64, uint64) ([]*si.EventRecord, uint64, uint64)
```
craigcondit, line 39: ID should be capitalized here. (GetEventsFromId -> GetEventsFromID).

pkg/events/event_system.go
```diff
@@ -28,38 +28,46 @@ import (
...
 
 	channel chan *si.EventRecord // channelling input eventChannel
-	stop    chan bool            // whether the service is stop
+	stop    chan bool            // whether the service is stopped
 	stopped bool
 
 	sync.Mutex
 }
 
+func (ec *EventSystemImpl) GetEventsFromId(id, count uint64) ([]*si.EventRecord, uint64, uint64) {
```
craigcondit, line 54: Let's add names for the returned values for clarity.

pkg/scheduler/objects/common_test.go
```diff
@@ -39,6 +39,10 @@ func (m *EventSystemMock) Reset() {
 	m.events = make([]*si.EventRecord, 0)
 }
 
+func (m *EventSystemMock) GetEventsFromId(uint64, uint64) ([]*si.EventRecord, uint64, uint64) {
```
craigcondit, line 42: As above, GetEventsFromId -> GetEventsFromID. Also, name the output variables.

pkg/events/event_ringbuffer.go
```diff
@@ -66,14 +66,14 @@ func (e *eventRingBuffer) Add(event *si.EventRecord) {
 // identifier is returned which can be used to get the first batch.
 // If the caller does not want to pose limit on the number of events returned, "count" must be set to a high
 // value, e.g. math.MaxUint64.
-func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64) {
+func (e *eventRingBuffer) GetEventsFromID(id uint64, count uint64) ([]*si.EventRecord, uint64, uint64) {
```
craigcondit, line 69: Let's name the output variables for clarity.

pkg/webservice/handlers.go
```diff
@@ -869,3 +870,53 @@ func getGroupResourceUsage(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+			if c <= 0 {
+				buildJSONErrorResponse(w, fmt.Sprintf("Illegal number of events: %d", c), http.StatusBadRequest)
+				return
+			}
+			count = uint64(c)
+		}
+
+		if fromIdStr := vars.ByName("fromId"); fromIdStr != "" {
```
craigcondit, line 899: I think we should name this parameter "start" instead of fromID. It's simpler to type and no casing issues to address. As such, startStr should be the variable name used.

craigcondit (changes requested):
Mostly minor changes, this looks good for the most part.

wilfred-s (approved):
LGTM

PR #591, opened 2023-07-17
[YUNIKORN-1864] Apply wild card group limit settings for users without any matching group

manirajv06:
### What is this PR for?

At times, config may have limits only for users but not for its group. Additionally, if there are any limit settings explicitly configured for wild card group, then wild card group limit settings should be applied for those users. In this case, group tracker object would be created for "*" and mapped to the user in appGroupTrackers map like any other named group.

Also, this PR ensures user tracking changes would be reverted in case of any issues(group quota exceeded etc) in group tracking flow.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1864

### Questions:

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -51,6 +51,6 @@ func CreateNodeEventRecord(objectID, message, referenceID string, changeType si.
 	return createEventRecord(si.EventRecord_NODE, objectID, referenceID, message, changeType, changeDetail, resource)
 }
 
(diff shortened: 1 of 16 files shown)


manirajv06:
This PR contains changes on top of [https://github.com/apache/yunikorn-core/pull/587]587 pr. Once [https://github.com/apache/yunikorn-core/pull/587]587 pr is merged, will need to do the cleanup

pbacsko (changes requested):
Please do a proper rebase, event related changes such as YUNIKORN-1803 should not be in this PR.

manirajv06:
Created a new pr https://github.com/apache/yunikorn-core/pull/599. Please refer the same.

PR #596, opened 2023-07-25
[YUNIKORN-1858] Handle group resource usage properly during config changes

manirajv06:
### What is this PR for?

1. App wise usage accounting with in queue tracker object.
2. With help of above changes#1, group resource usage can be decremented for the specific app during config changes. For example, Whenever earlier used group removed from config, all users mapped to that removed group through some application has to handled properly. So far, we just remove the linkage between the user and group in appGroupTrackers map by setting group tracker object to nil. With this pr, even group resource usage also has been handled properly by decrementing that app resource usage.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1858

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application_test.go
+++ b/pkg/scheduler/objects/application_test.go
@@ -1602,34 +1602,34 @@ func TestIncAndDecUserResourceUsageInSameGroup(t *testing.T) {
 
 	// Increase testuser
 	app.incUserResourceUsage(res)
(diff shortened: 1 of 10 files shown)


manirajv06:
Opened new pr https://github.com/apache/yunikorn-core/pull/600. hence closing this.

PR #597, opened 2023-07-26
[YUNIKORN-1881] Improve the performance of tryPlaceholderAllocate

pbacsko:
### What is this PR for?
`tryPlaceholderAllocate()` is too expensive. It copies and sorts apps and tries to find schedulable tasks even if we do not have placeholders at all. Applications should track if they have an active placeholder allocation to avoid unnecessary copy/sorting.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1881

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -24,6 +24,7 @@ import (
 	"math"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/looplab/fsm"
@@ -103,6 +104,7 @@ type Application struct {
 	stateLog             []*StateLogEntry            // state log for this application
 	placeholderData      map[string]*PlaceholderData // track placeholder and gang related info
 	askMaxPriority       int32                       // highest priority value of outstanding asks
+	hasPlaceholderAlloc  atomic.Bool                 // Whether there is at least one allocated placeholder (has to be atomic.Bool due to queue/app lock ordering)
 
 	rmEventHandler        handler.EventHandler
 	rmID                  string
@@ -1638,6 +1640,7 @@ func (sa *Application) addAllocationInternal(info *Allocation) {
 					zap.Error(err))
 			}
 		}
+		sa.hasPlaceholderAlloc.Store(true)
 	} else {
 		// skip the state change if this is the first replacement allocation as we have done that change
 		// already when the last placeholder was allocated
@@ -1712,6 +1715,17 @@ func (sa *Application) ReplaceAllocation(uuid string) *Allocation {
 	if sa.placeholderData != nil {
 		sa.placeholderData[ph.GetTaskGroup()].Replaced++
 	}
+
+	// check if we have active placeholder allocation left
+	hasReplaceablePlaceholder := false
+	for _, phData := range sa.placeholderData {
+		if phData.Count > (phData.Replaced + phData.TimedOut) {
+			hasReplaceablePlaceholder = true
+			break
+		}
+	}
+	sa.hasPlaceholderAlloc.Store(hasReplaceablePlaceholder)
+
 	return ph
 }
 
@@ -2011,6 +2025,10 @@ func (sa *Application) LogAppSummary(rmID string) {
 	appSummary := sa.GetApplicationSummary(rmID)
 	appSummary.DoLogging()
 	appSummary.ResourceUsage = nil
+}
+
+func (sa *Application) HasPlaceholderAllocation() bool {
+	return sa.hasPlaceholderAlloc.Load()
 }
 
 // test only
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1624,6 +1624,7 @@ func (sa *Application) addAllocationInternal(info *Allocation) {
 		// If we would start it when we just try to allocate, that is something very unstable, and we don't really have any
 		// impact on what is happening until this point
 		if resources.IsZero(sa.allocatedPlaceholder) {
+			sa.hasPlaceholderAlloc = true
 			sa.initPlaceholderTimer()
 		}
 		// User resource usage needs to be updated even during resource allocation happen for ph's itself even though state change would happen only after all ph allocation completes.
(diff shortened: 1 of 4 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -1712,6 +1715,17 @@ func (sa *Application) ReplaceAllocation(uuid string) *Allocation {
...
+
+	// check if we have active placeholder allocation left
+	hasReplaceablePlaceholder := false
+	for _, phData := range sa.placeholderData {
+		if phData.Count > (phData.Replaced + phData.TimedOut) {
+			hasReplaceablePlaceholder = true
+			break
+		}
+	}
+	sa.hasPlaceholderAlloc.Store(hasReplaceablePlaceholder)
```
pbacsko, line 1727: This part is crucial and needs to be checked carefully. If we replaced all PHs, we set it back to false so we won't bother with further PH allocations.

craigcondit (approved):
Looks ok I think. Would like @wilfred-s to take a look as well. +1.

pkg/scheduler/objects/application.go
```diff
@@ -1638,6 +1640,7 @@ func (sa *Application) addAllocationInternal(info *Allocation) {
 					zap.Error(err))
 			}
 		}
+		sa.hasPlaceholderAlloc.Store(true)
```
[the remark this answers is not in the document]
wilfred-s: -1 that cannot work.

Applications at the partition level are a map of all applications in the partition. The queue that we are checking needs to make the decision. At the partition level we do not know if the specific queue the app is running in has pending requests. We also do not want to walk over all applications at the partition level as that would not take into account any queue details.

Filtering out that early does not work.

pkg/scheduler/objects/application.go
```diff
@@ -103,6 +104,7 @@ type Application struct {
 	stateLog             []*StateLogEntry            // state log for this application
 	placeholderData      map[string]*PlaceholderData // track placeholder and gang related info
 	askMaxPriority       int32                       // highest priority value of outstanding asks
+	hasPlaceholderAlloc  atomic.Bool                 // Whether there is at least one allocated placeholder (has to be atomic.Bool due to queue/app lock ordering)
```
wilfred-s, line 107: Does not need to be atomic, application sorting uses the calls to get other data from the application which require locking the application. Can use the same mechanism. Sticking with that makes sure that if a release is in progress it works.

pkg/scheduler/objects/application.go
wilfred-s: See earlier comment, placeholder allocation based is more stable and does not need any new looping as we already have the checks for other things

pkg/scheduler/objects/application.go
```diff
@@ -2013,6 +2027,10 @@ func (sa *Application) LogAppSummary(rmID string) {
 	appSummary.ResourceUsage = nil
 }
 
+func (sa *Application) HasPlaceholderAllocation() bool {
+	return sa.hasPlaceholderAlloc.Load()
```
wilfred-s, line 2031: See above: a mutex can be used. Setting the boolean happens inside a mutex locked application. This is the only read point.

wilfred-s (changes requested):

pkg/scheduler/objects/application.go
```diff
@@ -1712,6 +1715,17 @@ func (sa *Application) ReplaceAllocation(uuid string) *Allocation {
 	if sa.placeholderData != nil {
 		sa.placeholderData[ph.GetTaskGroup()].Replaced++
 	}
+
+	// check if we have active placeholder allocation left
+	hasReplaceablePlaceholder := false
+	for _, phData := range sa.placeholderData {
```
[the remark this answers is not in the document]
manirajv06: Yes, it is good to set/unset in existing places (appropriate as described). do we need to unset even when app transition to work as non-gang because of "soft" style"?

manirajv06 (commented):
Few more observations, but not exactly specific to this pr.

1. Some checks are repeated in mutiple try...() methods in `queue`. For example, `app.IsAccepted()` and `sq.canRunApp()`
2. Some checks can be carried out much more earlier in `queue` itself instead of doing in `application`. For example, queue and user headroom checks.

I think these checks can be kept in one single place.

pkg/scheduler/objects/application.go
pbacsko: This is only possible if we call `GetCopyOfApps()` then go through it and filter out the ones without a PH allocation. I wanted to avoid this because we copy the apps unnecessarily. Otherwise, we have a locked mutex order queue-> application and the inverse application->queue can happen in a separate goroutine (eg. `AddAllocationAsk()`, `removeAsksInternal()`) at the same time.

pkg/scheduler/objects/application.go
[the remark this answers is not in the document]
pbacsko: You're right, we have `hasPlaceHolderAllocations := len(sa.getPlaceholderAllocations()) > 0` which can be used for this purpose.

pkg/scheduler/objects/application.go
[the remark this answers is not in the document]
pbacsko: If there's no application with an active placeholder allocation, then we don't have to call tryPlaceholderAllocate()` because there's no placeholder to replace. Obviously if this counter is >1 then we have to descend the hierarchy and see where can we can do a replacement.

pkg/scheduler/objects/application.go
pbacsko: Ok, I modified it regardless, let's see if this is any good.

pbacsko: Update: I used the new unit test to check performance, we really want to avoid `GetCopyOfApps()`. That slows things down significantly. I believe we still have a speedup because we avoid the actual sorting, but the speedup is much smaller this way.

pkg/scheduler/objects/application.go
```diff
@@ -1638,6 +1639,7 @@ func (sa *Application) addAllocationInternal(info *Allocation) {
 					zap.Error(err))
 			}
 		}
+		sa.hasPlaceholderAlloc = true
```
wilfred-s, line 1642: NIT: this could move inside the if checking for zero allocated placeholders  at line 1626

wilfred-s (approved):
LGTM

pkg/scheduler/objects/application.go

PR #600, opened 2023-07-31
[YUNIKORN-1858] Handle group resource usage properly during config ch…

manirajv06:
…anges

### What is this PR for?
1. App wise usage accounting with in queue tracker object.
2. With help of above changes#1, group resource usage can be decremented for the specific app during config changes. For example, Whenever earlier used group removed from config, all users mapped to that removed group through some application has to handled properly. So far, we just remove the linkage between the user and group in appGroupTrackers map by setting group tracker object to nil. With this pr, even group resource usage also has been handled properly by decrementing that app resource usage. 
3. Used zap.stringer instead of zap.string for resources, removed unnecessary logs etc

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1858

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -408,6 +408,10 @@ func (m *Manager) processGroupConfig(group string, limitConfig *LimitConfig, que
 
 // clearEarlierSetLimits Clear already configured limits of users and groups for which limits have been configured before but not now
 func (m *Manager) clearEarlierSetLimits(userLimits map[string]bool, groupLimits map[string]bool, queuePath string) error {
+	// Clear already configured limits of group for which limits have been configured before but not now
+	for _, gt := range m.groupTrackers {
+		m.clearEarlierSetGroupLimits(gt, queuePath, groupLimits)
+	}
 	// Clear already configured limits of user for which limits have been configured before but not now
 	for u, ut := range m.userTrackers {
 		// Is this user already tracked for the queue path?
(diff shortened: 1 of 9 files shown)


pkg/scheduler/ugm/manager.go
```diff
@@ -421,32 +425,25 @@ func (m *Manager) clearEarlierSetLimits(userLimits map[string]bool, groupLimits
...
-							delete(m.groupTrackers, g)
+						removeQT, decreased := gt.decreaseTrackedResourceUsage(appID, true)
+						if decreased {
+							if removeQT {
+								log.Log(log.SchedUGM).Debug("Removing group from manager",
+									zap.String("group", gt.groupName),
+									zap.String("queue path", queuePath),
+									zap.String("application", appID))
+								delete(m.groupTrackers, gt.groupName)
+							}
```
FrankYang0529, line 439: I tried to apply the change to #595, and the following test case couldn't pass. I think we may also check `m.userWildCardLimitsConfig` or `m.groupWildCardLimitsConfig` before removing trackers.

```go
func TestMultipleGroupLimitChange(t *testing.T) {
	setupUGM()

	manager := GetUserManager()
	conf := createConfigWithLimits([]configs.Limit{
		createLimit(nil, []string{"group1", "group2"}, largeResource, 2),
		createLimit(nil, []string{"*"}, mediumResource, 1),
	})
	assert.NilError(t, manager.UpdateConfig(conf.Queues[0], "root"))

	user1 := security.UserGroup{User: "user1", Groups: []string{"group1"}}
	user2 := security.UserGroup{User: "user2", Groups: []string{"group2"}}
	user3 := security.UserGroup{User: "user3", Groups: []string{"group3"}}

	usage, err := resources.NewResourceFromConf(mediumResource)
	if err != nil {
		t.Errorf("new resource create returned error or wrong resource: error %t, res %v", err, usage)
	}

	// all users can increate usage within the quota
	increased := manager.IncreaseTrackedResource(queuePathParent, "test-app-1-1", usage, user1)
	assert.Equal(t, increased, true, "unable to increase tracked resource: queuepath "+queuePathParent+", app test-app-1-1, res "+usage.String())

	increased = manager.IncreaseTrackedResource(queuePathParent, "test-app-2-1", usage, user2)
	assert.Equal(t, increased, true, "unable to increase tracked resource: queuepath "+queuePathParent+", app test-app-2-1, res "+usage.String())

	increased = manager.IncreaseTrackedResource(queuePathParent, "test-app-3-1", usage, user3)
	assert.Equal(t, increased, true, "unable to increase tracked resource: queuepath "+queuePathParent+", app test-app-3-1, res "+usage.String())

	// remove group2 from the specific group
	conf.Queues[0].Queues[0].Limits[0].Groups = []string{"group1"}
	assert.NilError(t, manager.UpdateConfig(conf.Queues[0], "root"))

	// user1 still can increase usage within the quota
	increased = manager.IncreaseTrackedResource(queuePathParent, "test-app-1-2", usage, user1)
	assert.Equal(t, increased, true, "unable to increase tracked resource: queuepath "+queuePathParent+", app test-app-1-2, res "+usage.String())

	// user2 can't increase usage more than wildcard limit
	increased = manager.IncreaseTrackedResource(queuePathParent, "test-app-2-2", usage, user2)
	assert.Equal(t, increased, false, "should not increase tracked resource: queuepath "+queuePathParent+", app test-app-2-2, res "+usage.String())

	// user3 can't increase usage more than wildcard limit
	increased = manager.IncreaseTrackedResource(queuePathParent, "test-app-3-2", usage, user3)
	assert.Equal(t, increased, false, "should not increase tracked resource: queuepath "+queuePathParent+", app test-app-3-2, res "+usage.String())
}
```

wilfred-s (changes requested):
Why do we need to track usage per app in the group? If the group is removed all usage is removed independent of the individual application usage.
I dropped a large comment in the jira to assess this.

wilfred-s (approved):
LGTM

PR #604, opened 2023-08-01
[YUNIKORN-1886] Optimize scheduling cycle if there's no placeholder allocation or reservation

pbacsko:
### What is this PR for?
Count reservations and active placeholder allocations so that we can avoid calling allocation code paths unnecessarily.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1886

### Questions:

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -569,9 +569,6 @@ func (pc *PartitionContext) AddNode(node *objects.Node, existingAllocations []*o
 				metrics.GetSchedulerMetrics().IncFailedNodes()
 				return err
 			}
-			if alloc.IsPlaceholder() {
-				pc.incPhAllocationCount()
-			}
 		}
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/scheduler/partition_test.go
```diff
@@ -2513,6 +2513,7 @@ func TestUpdateNode(t *testing.T) {
 }
 
 func TestAddTGApplication(t *testing.T) {
+
```
wilfred-s, line 2516: unexpected newline

pkg/scheduler/partition.go
```diff
@@ -567,6 +569,9 @@ func (pc *PartitionContext) AddNode(node *objects.Node, existingAllocations []*o
 				metrics.GetSchedulerMetrics().IncFailedNodes()
 				return err
 			}
+			if alloc.IsPlaceholder() {
+				pc.incPhAllocationCount()
+			}
```
wilfred-s, line 574: This one needs to be reversed also if the node is removed.
The node removal path is not currently covered: see line 698 (after change) or 693 (before change) need a decrement placeholder count in that check

pbacsko: Now we add and remove in two methods:
* `removeNodeAllocations()` / `removeAllocation()` (one/ all when uuid = "")
* `allocate()` (normal scheduling) / `addAllocation()` (startup)

manirajv06 (approved):
looks good

wilfred-s (approved):
LGTM

PR #607, opened 2023-08-03
[YUNIKORN-1896] CalculateAbsUsedCapacity must be consistent

wilfred-s:
### What is this PR for?
The overflow detection is still showing different results on ARM vs Intel. Simplifying the calculation in CalculateAbsUsedCapacity to always return the same result on all architectures.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1896

new unit tests, must all pass on ARM and intel
mixed set of reviewers: some with Intel and some with ARM local systems

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -936,6 +936,12 @@ func IsZero(zero *Resource) bool {
 	return true
 }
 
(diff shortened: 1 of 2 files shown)


manirajv06 (approved):
looks good.

PR #608, opened 2023-08-03
[YUNIKORN-1885] Update Go version filter

wilfred-s:
### What is this PR for?
The Go version filter pulls in patch release details which shows errors when running tests with pre-release (release candidate) go version to check for compatibility.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1885

Install a rc for Go and set a Go version before running a test, example:
GO="go1.21rc3" make test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ ifeq ($(GO),)
 GO := go
 endif
 
-GO_VERSION := $(shell "$(GO)" version | awk '{print substr($$3, 3, 10)}')
+GO_VERSION := $(shell "$(GO)" version | awk '{print substr($$3, 3, 4)}')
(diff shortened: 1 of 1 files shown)


mitdesai:
LGTM
+1 (non-binding)

manirajv06:
By any chance, If version is 2.1rc1 ?

pbacsko (approved):
+1

wilfred-s:
> By any chance, If version is 2.1rc1 ?

Go version 2 will have breaking incompatible changes. That would be the reason to change to Go 2. We will most likely not support Go 1 and Go 2 at the same time...

PR #618, opened 2023-08-11
[YUNIKORN-1916] Align headroom and resource tracking resolution logic

wilfred-s:
### What is this PR for?
Headroom and increase need to run the exact same logic for user and group tracker create and checks. The wildcard group is a separate group tracker that gets returned based. Wildcard groups are only tracked in a leaf and all use the same group tracker object.

Only calculate headroom once for each application in the scheduling cycle not for each ask.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1916

existing unit tests should all still pass

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -950,6 +950,8 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, preemptionDelay
 	if sa.sortedRequests == nil {
 		return nil
 	}
+	// calculate the users' headroom, includes group check which requires the applicationID
+	userHeadroom := ugm.GetUserManager().Headroom(sa.queuePath, sa.ApplicationID, sa.user)
 	// get all the requests from the app sorted in order
 	for _, request := range sa.sortedRequests {
 		if request.GetPendingAskRepeat() == 0 {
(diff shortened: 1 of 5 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -959,13 +961,11 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, preemptionDelay
 		if sa.canReplace(request) {
 			continue
 		}
-
-		userHeadroom := ugm.GetUserManager().Headroom(sa.queuePath, sa.user)
+		// check if this fits in the users' headroom first, if that fits check the queues' headroom
+		// NOTE: preemption most likely will not help in this case. The chance that preemption helps is mall
```
FrankYang0529, line 965: typo: mall -> small.

FrankYang0529:
In resource tracking resolution logic, we also check whether `runningApplications` is less than `maxRunningApps`. However, we don't do it in headroom logic. Do we also want to align it? If yes, I can help to create another PR for it. Thanks.

https://github.com/apache/yunikorn-core/blob/5f812c1d34e212ebc3126143592233587f12cf51/pkg/scheduler/ugm/queue_tracker.go#L92

wilfred-s:
> In resource tracking resolution logic, we also check whether `runningApplications` is less than `maxRunningApps`. However, we don't do it in headroom logic. Do we also want to align it? If yes, I can help to create another PR for it. Thanks.

Max running apps from a scheduling cycle is checked outside the headroom. It is checked from the queue before we even get to the application. We need a similar check for the user at the same point. See canRunApp in the queue. That would need a different jira.

manirajv06 (approved):
looks good.

PR #619, opened 2023-08-15
[YUNIKORN-1919] decrease runningApps when app changes from starting to completing

FrankYang0529:
### What is this PR for?

We didn't decrease `runningApps` if the single-allocation app changes from starting to completing.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1919

Add a unit test for it.


wilfred-s (approved):
LGTM
Would like @lixmgl and @pbacsko to have a look too.

PR #620, opened 2023-08-15
[YUNIKORN-1920] headroom with parent queue usage

wilfred-s:
### What is this PR for?
Headroom needs to take into account the usage in each queue and not assume that the full max resource is available if the queue is not the leaf queue.
Add a queue tracker object for any queue that is not found while traversing the hierarchy. The user/group has an application running in that queue and will add usage at some point.

Add tests to cover usage and max with different resource types set.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1920

New unit tests added to cover the scenario

### DIFF
--- a/pkg/scheduler/ugm/queue_tracker.go
+++ b/pkg/scheduler/ugm/queue_tracker.go
@@ -245,34 +245,32 @@ func (qt *QueueTracker) setLimit(queuePath string, maxResource *resources.Resour
 	childQueueTracker.maxResources = maxResource
 }
 
-func (qt *QueueTracker) headroom(queuePath string) *resources.Resource {
+func (qt *QueueTracker) headroom(hierarchy []string) *resources.Resource {
 	log.Log(log.SchedUGM).Debug("Calculating headroom",
-		zap.String("queue path", queuePath))
-	childQueuePath, immediateChildQueueName := getChildQueuePath(queuePath)
-	if childQueuePath != common.Empty {
-		if qt.childQueueTrackers[immediateChildQueueName] != nil {
-			headroom := qt.childQueueTrackers[immediateChildQueueName].headroom(childQueuePath)
-			if headroom != nil {
-				return resources.ComponentWiseMinPermissive(headroom, qt.maxResources)
-			}
-		} else {
-			log.Log(log.SchedUGM).Error("Child queueTracker tracker must be available in child queues map",
-				zap.String("child queueTracker name", immediateChildQueueName))
-			return nil
+		zap.Strings("queue path", hierarchy))
+	// depth first: all the way to the leaf, create if not exists
+	// more than 1 in the slice means we need to recurse down
+	var headroom, childHeadroom *resources.Resource
+	if len(hierarchy) > 1 {
+		childName := hierarchy[1]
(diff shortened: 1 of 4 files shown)


pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -245,34 +245,32 @@ func (qt *QueueTracker) setLimit(queuePath string, maxResource *resources.Resour
...
+			zap.String("queue path", qt.queuePath),
 			zap.String("max resource", qt.maxResources.String()),
 			zap.String("headroom", headroom.String()))
-		return headroom
 	}
-	return nil
+	if headroom == nil {
+		return childHeadroom
+	}
+	return resources.ComponentWiseMinPermissive(headroom, childHeadroom)
```
manirajv06, line 273: I tried to reproduce the issue by using the modified test method and thought of simplifying the changes and also to retain the same signature like given below:

mani@mani-MBP16 incubator-yunikorn-core % git diff pkg/scheduler/ugm/queue_tracker.go
diff --git a/pkg/scheduler/ugm/queue_tracker.go b/pkg/scheduler/ugm/queue_tracker.go
index c793ffc..1b089de 100644
--- a/pkg/scheduler/ugm/queue_tracker.go
+++ b/pkg/scheduler/ugm/queue_tracker.go
@@ -253,7 +253,11 @@ func (qt *QueueTracker) headroom(queuePath string) *resources.Resource {
                if qt.childQueueTrackers[immediateChildQueueName] != nil {
                        headroom := qt.childQueueTrackers[immediateChildQueueName].headroom(childQueuePath)
                        if headroom != nil {
-                               return resources.ComponentWiseMinPermissive(headroom, qt.maxResources)
+                               pHeadroom := qt.maxResources.Clone()
+                               if !resources.Equals(resources.NewResource(), pHeadroom) {
+                                       pHeadroom.SubOnlyExisting(qt.resourceUsage)
+                               }
+                               return resources.ComponentWiseMinPermissive(headroom, pHeadroom)
                        }
                } else {
                        log.Log(log.SchedUGM).Error("Child queueTracker tracker must be available in child queues map",

Please assess this alternative approach.

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -245,34 +245,32 @@ func (qt *QueueTracker) setLimit(queuePath string, maxResource *resources.Resour
 	childQueueTracker.maxResources = maxResource
 }
 
-func (qt *QueueTracker) headroom(queuePath string) *resources.Resource {
+func (qt *QueueTracker) headroom(hierarchy []string) *resources.Resource {
```
pbacsko, line 248: I suggest receiving the full path and using `getChildQueuePath()` from `utilities.go` to keep the code consistent (also the UGM API remains the same and simpler).

wilfred-s: That utility function will be removed as part of a next change

manirajv06 (approved):
looks good.

Filed https://issues.apache.org/jira/browse/YUNIKORN-1940 to do the similar changes.

PR #621, opened 2023-08-16
[YUNIKORN-1917] enforce ugm maxapplications rule

FrankYang0529:
### What is this PR for?

In increaseTrackedResource, we check `maxRunningApps` before we increase the resource. However, we don't check it before we create a new allocation. In this case, the real applications will be more than the limit.

https://github.com/apache/yunikorn-core/blob/5f812c1d34e212ebc3126143592233587f12cf51/pkg/scheduler/ugm/queue_tracker.go#L92

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1917

Add unit test cases for it.

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -584,6 +584,35 @@ func (m *Manager) Headroom(queuePath, applicationID string, user security.UserGr
 	return resources.ComponentWiseMinPermissive(userHeadroom, groupHeadroom)
 }
 
+// CanRunApp checks the maxApplications for this specific application that runs as the user and group.
+func (m *Manager) CanRunApp(queuePath, applicationID string, user security.UserGroup) bool {
+	userTracker := m.getUserTracker(user.User)
+	userCanRunApp := userTracker.canRunApp(queuePath, applicationID)
+	log.Log(log.SchedUGM).Debug("Check whether user can run app",
+		zap.String("user", user.User),
+		zap.String("queue path", queuePath),
+		zap.Bool("can run app", userCanRunApp))
+	// make sure the user has a groupTracker for this application, if not yet there add it
+	if !userTracker.hasGroupForApp(applicationID) {
+		m.ensureGroupTrackerForApp(queuePath, applicationID, user)
+	}
+	// check if this application now has group tracking, if not we're done
+	appGroup := userTracker.getGroupForApp(applicationID)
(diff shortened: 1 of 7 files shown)


pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -364,3 +364,53 @@ func (qt *QueueTracker) UnlinkQT(queuePath string) bool {
...
+			allowed := qt.childQueueTrackers[immediateChildQueueName].canRunApp(childQueuePath, applicationID, trackType)
+			if !allowed {
+				return false
+			}
+		}
+	}
+
+	running := len(qt.runningApplications)
+	if existingApp := qt.runningApplications[applicationID]; !existingApp {
+		running += 1
```
manirajv06, line 385: Can we return when app is already running? so that rest of the following code would execute only for new application.

pkg/scheduler/ugm/manager.go
```diff
@@ -607,6 +607,35 @@ func (m *Manager) Headroom(queuePath, applicationID string, user security.UserGr
 	return resources.ComponentWiseMinPermissive(userHeadroom, groupHeadroom)
 }
 
+// CanRunApp checks the maxApplications for this specific application that runs as the user and group.
+func (m *Manager) CanRunApp(queuePath, applicationID string, user security.UserGroup) bool {
+	userTracker := m.getUserTracker(user.User)
```
manirajv06, line 612: Does userTracker always exists? I think this is going to be the first entry point into ugm module as this is being called from the very early steps.

manirajv06 (changes requested):

pkg/scheduler/ugm/manager.go
manirajv06: yes, correct

manirajv06 (commented):
looks good.

Can you add a case to cover this behaviour? when both user & group limits are present? may be user max app is lower and group max app is higher ? and vice-versa

FrankYang0529:
> Can you add a case to cover this behaviour? when both user & group limits are present? may be user max app is lower and group max app is higher ? and vice-versa

Yeah, added it. Thanks for the review!

manirajv06 (approved):
looks good.

FrankYang0529:
Resolved conflicts. Thanks.

PR #622, opened 2023-08-22
[YUNIKORN-1918] Remove UpdateSchedulerConfig from core

Paul-Lyu:
### What is this PR for?
In the ClusterContext there is a locked version to update the scheduler config. The locked version was introduced to allow setting the config via the web service. This has been removed in a previous version and now the function UpdateSchedulerConfig() is no longer useful and as dead code should be removed.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1918

### Questions:

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -324,26 +324,6 @@ func (cc *ClusterContext) removePartitionsByRMID(event *rmevent.RMPartitionsRemo
 	}
 }
 
-// Locked version of the configuration update called from the webservice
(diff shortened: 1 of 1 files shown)


pbacsko (approved):
+1

PR #623, opened 2023-08-22
[YUNIKORN-1928] Exceed queue quota requests should be skipped when cl…

zhuqi-lucas:
…uster auto scaling.

### What is this PR for?
In shim side we have skip autoscaling logic both for yunikorn mode and plugin mode, we need to make that exceed queue quota requests should be skipped when cluster auto scaling.

```
switch request.State {
case si.UpdateContainerSchedulingStateRequest_SKIPPED:
    // auto-scaler scans pods whose pod condition is PodScheduled=false && reason=Unschedulable
    // if the pod is skipped because the queue quota has been exceed, we do not trigger the auto-scaling
    task.SetTaskSchedulingState(interfaces.TaskSchedSkipped)
    if ctx.updatePodCondition(task,
       &v1.PodCondition{
          Type:    v1.PodScheduled,
          Status:  v1.ConditionFalse,
          Reason:  "SchedulingSkipped",
          Message: request.Reason,
       }) {
       events.GetRecorder().Eventf(task.pod.DeepCopy(), nil,
          v1.EventTypeNormal, "PodUnschedulable", "PodUnschedulable",
          "Task %s is skipped from scheduling because the queue quota has been exceed", task.alias)
    }
```
It will be over auto scaling.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation_ask.go
+++ b/pkg/scheduler/objects/allocation_ask.go
@@ -54,6 +54,7 @@ type AllocationAsk struct {
 	allocLog            map[string]*AllocationLogEntry
 	preemptionTriggered bool
 	preemptCheckTime    time.Time
+	skipAutoScaling     bool // skip auto-scaling for this ask when it exceeds the queue quota
 
 	sync.RWMutex
 }
@@ -93,6 +94,7 @@ func NewAllocationAskFromSI(ask *si.AllocationAsk) *AllocationAsk {
 		allowPreemptOther: common.IsAllowPreemptOther(ask.PreemptionPolicy),
 		originator:        ask.Originator,
 		allocLog:          make(map[string]*AllocationLogEntry),
+		skipAutoScaling:   false,
 	}
 	// this is a safety check placeholder and task group name must be set as a combo
 	// order is important as task group can be set without placeholder but not the other way around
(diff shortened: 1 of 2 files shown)


zhuqi-lucas:
A draft code without unit testing, will submit unit test.

zhuqi-lucas:
My PR want to improve, when a request is rejected due to queue limit, we don't hurry to auto-scaling for this request, we can make other requests which are not rejected by queue limit to auto scale, i am not sure if it ls a improvement @wilfred-s

craigcondit:
See my comment on JIRA -- I don't think this is necessary at all:

I don't think there's any change required here as this logic is already handled properly in the core. We don't reject requests when queues are full, we simply keep the pod in a pending state (but not unschedulable). This prevents the autoscaler from scaling up. However, if the pod would fit in the queue but the cluster itself is full, then we mark it unschedulable to allow the autoscaler to work.

As for making other requests, this also happens - YK is a greedy scheduler and will attempt to fit smaller pods that are further back in the queue if there is room for them.

zhuqi-lucas:
Thank you @craigcondit and @wilfred-s for confirming this, i will remove the skip logic in another PR.

PR #625, opened 2023-08-23
[YUNIKORN-1929] fix update group configs

FrankYang0529:
### What is this PR for?

Keep a group tracker if it's from a specific limit to wildcard limit.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1929

Covered by unit tests.


FrankYang0529:
Close the PR since we only track a group which is defined in queue limit config.

PR #626, opened 2023-08-23
[YUNIKORN-1930] Add message to node added/removed event

pbacsko:
### What is this PR for?
Messages that are sent to the shim should have a Message field set properly. This will be visible when the events for a given object (pod/node) are listed.
The node added/removed event will be sent out as an event to K8s, so we add an informative message.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1930

### Questions:


craigcondit (approved):
+1 LGTM.

PR #628, opened 2023-08-25
[YUNIKORN-1937] Extract URL query properly in getEvents()

pbacsko:
### What is this PR for?
Extract URL query parameters properly. We need to use `URL.GetQuery()`.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1937

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -874,40 +874,38 @@ func getGroupResourceUsage(w http.ResponseWriter, r *http.Request) {
 func getEvents(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	eventSystem := events.GetEventSystem()
-	if eventSystem == nil {
-		buildJSONErrorResponse(w, "Event system is disabled", http.StatusBadRequest)
+	if !eventSystem.IsEventTrackingEnabled() {
+		buildJSONErrorResponse(w, "Event tracking is disabled", http.StatusBadRequest)
 		return
 	}
 
 	count := uint64(10000)
 	var start uint64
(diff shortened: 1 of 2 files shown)


manirajv06 (commented):
looks good.

https://issues.apache.org/jira/browse/YUNIKORN-1472 migrated the handler from https://github.com/gorilla/mux to [httprouter](https://github.com/julienschmidt/httprouter) recently. So, if you can explain the problem highlighted in jira, it would be great. Based on the severity, most likely, we need to file a new jira to replace in all places.

pbacsko:
> looks good.
> 
> https://issues.apache.org/jira/browse/YUNIKORN-1472 migrated the handler from https://github.com/gorilla/mux to [httprouter](https://github.com/julienschmidt/httprouter) recently. So, if you can explain the problem highlighted in jira, it would be great. Based on the severity, most likely, we need to file a new jira to replace in all places.

I think this is the only instance where this is needed. We don't use URL query parameters (`/ws/v1/xxxx?param1=val1&param2=val2` etc) in other paths.

zhuqi-lucas (approved):
LGTM And i agree this is the speical case.
We don't use URL query parameters (/ws/v1/xxxx?param1=val1&param2=val2 etc) in other paths.
In other place, one level paras, we can use vars.ByName, it's enough.

PR #629, opened 2023-08-30
[YUNIKORN-1945] Add missing application state transition event

pbacsko:
### What is this PR for?
We dont send event when the app transitions to "Running" state.
Also, a small cleanup in the event sending code to make it nicer and centralized.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1945

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application_state.go
+++ b/pkg/scheduler/objects/application_state.go
@@ -74,6 +74,17 @@ const (
 	Resuming
 )
 
+var stateEvents = map[string]si.EventRecord_ChangeDetail{
+	Accepted.String():   si.EventRecord_APP_ACCEPTED,
+	Starting.String():   si.EventRecord_APP_STARTING,
+	Running.String():    si.EventRecord_APP_RUNNING,
+	Rejected.String():   si.EventRecord_APP_REJECT,
+	Completing.String(): si.EventRecord_APP_COMPLETING,
+	Completed.String():  si.EventRecord_APP_COMPLETED,
+	Failing.String():    si.EventRecord_APP_FAILING,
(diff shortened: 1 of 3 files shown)


pkg/scheduler/objects/application_state.go
```diff
@@ -142,6 +153,13 @@ func NewAppState() *fsm.FSM {
 				} else {
 					app.OnStateChange(event, "")
 				}
+				eventDetails, ok := stateEvents[event.Dst]
+				if !ok {
+					log.Log(log.SchedFSM).Error("event details not found",
+						zap.String("state", event.Dst))
+					return
+				}
```
manirajv06, line 161: We are sending events based on certain cases. Which means, there is no need to send event while entering certain state. So, do we really need to log this as error? May be debug in case if there is no need to send event

manirajv06 (approved):
looks good.

Minor nit.

pkg/scheduler/objects/application_state.go
pbacsko: Ideally we want to send out an event for every state except New (because that's the initial state and it carries no extra info). So right now if it's missing it means a bug, that's why I dediced to log as an error. Let's wait for Wilfred's/Craig's input.

manirajv06: Ok. Expired, Resuming ?

pbacsko: Added.

PR #635, opened 2023-08-31
[YUNIKORN-1942] Null Batch API Response when we reset ring buffer size

pbacsko:
### What is this PR for?
`ringBuffer.id2pos()` does not work properly after resizing the buffer. We need to track the ID of position 0 because the id-to-position calculation becomes invalid. By doing so, `id2pos()` also becomes simpler.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1942

### Questions:

### DIFF
--- a/pkg/events/event_ringbuffer.go
+++ b/pkg/events/event_ringbuffer.go
@@ -45,9 +45,10 @@ type eventRingBuffer struct {
 	events   []*si.EventRecord
 	capacity uint64 // capacity of the buffer
 	head     uint64 // position of the next element (no tail since we don't remove elements)
-	full     bool   // indicates whether the buffer if full - once it is, it stays full unless buffer is resized
+	full     bool   // indicates whether the buffer if full - once it is, it stays full unless the buffer is resized
 	id       uint64 // unique id of an event record
 	lowestId uint64 // lowest id of an event record available in the buffer at any given time
+	idAtZero uint64 // id at index 0, used to calculate position based on id
 
 	sync.RWMutex
 }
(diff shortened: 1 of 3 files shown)


pbacsko:
Note that the following changes are also present in the PR:
* removed comments which don't add any meaningful information
* changed the path of the batch query API (removed the slash from the end because it's not possible to use the optional query parameters like `/ws/v1/events/batch?start=x&count=n`)

wilfred-s (changes requested):
If head == 0 the next event that will be written at position 0 is never the lowestID it is the current id which is incremented later. You are tracking the previous id at position 0.

The only point at which we need to update the "id at pos 0" is when we resize. That is the only point it can change from the base 0 (or the multiple of length). The value will be the lowest ID as currently calculated in the resize. After that change in the resize it does not need updating until the next resize. The % calculation will handle it.

The name is also incorrect and this is what really confused me to start with. It is more like the id offset due to resize: `resizeOffset` covers the value better.

NIT: keeping comments in the code so we know what is being done is not a bad thing. The code is complex enough that comments help. We also need further code comments to explain what is being done not less.

pbacsko:
> If head == 0 the next event that will be written at position 0 is never the lowestID it is the current id which is incremented later. You are tracking the previous id at position 0.
> 
> The only point at which we need to update the "id at pos 0" is when we resize. That is the only point it can change from the base 0 (or the multiple of length). The value will be the lowest ID as currently calculated in the resize. After that change in the resize it does not need updating until the next resize. The % calculation will handle it.
> 
> The name is also incorrect and this is what really confused me to start with. It is more like the id offset due to resize: `resizeOffset` covers the value better.

I agree the name is misleading. I was thinking about something very similar but somehow I conviced myself that this is better, hah. 

> 
> NIT: keeping comments in the code so we know what is being done is not a bad thing. The code is complex enough that comments help. We also need further code comments to explain what is being done not less.

I examined comments in detail and only removed those which do not convey extra information and it's obvious what the code is doing (eg "update buffer state"). I kept those which explains something which is not straightforward.

pbacsko:
I modified the variable name & made changes to the comments (formatting, minor wording).

wilfred-s (approved):

PR #636, opened 2023-09-01
[YUNIKORN-1951] The queues only set max resource shouldn't be eligibl…

wusamzong:
…e victim

### What is this PR for?
The function findEligiblePreemptionVictims() is used to find preemption victims.
The code that determines the relationship between guaranteed and preemption victim is as follows:

```
guaranteed := resources.ComponentWiseMinPermissive(sq.GetActualGuaranteedResource(), sq.GetMaxResource())
if guaranteed.FitInMaxUndef(sq.GetAllocatedResource()) {
    return
}
```

When guaranteed is not set, the maximum configuration is used. In normal circumstances, this allows the `FitInMaxUndef()` condition to return true, preventing the queue from becoming a victim. However, the problem arises when the queue's maximum resources are set to `{memory:200, pods:0}`, and yet applications with`{memory:100, pods:1}`can still be allocated to the queue.
Which causing `FitInMaxUndef()` to always return `false`, thus considering the queue as a victim.

```
max = &Resource{Resources: map[string]Quantity{"memory": 200, "pods": 0}}
allocated = &Resource{Resources: map[string]Quantity{"memory": 100, "pods": 1}}
max.FitInMaxUndef(allocated) // false
```

For the solution, just take the guaranteed resources when selecting preemption victims.
Furthermore, ensure that the unit test takes the pods into consideration.
Additionally, create a new unit test to verify that queues without guaranteed resources set are not considered as potential victims.

### What type of PR is it?
* [x] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1951/

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -1718,7 +1718,7 @@ func (sq *Queue) findEligiblePreemptionVictims(results map[string]*QueuePreempti
 		}
 
 		// skip this queue if we are within guaranteed limits
-		guaranteed := resources.ComponentWiseMinPermissive(sq.GetActualGuaranteedResource(), sq.GetMaxResource())
+		guaranteed := sq.GetActualGuaranteedResource()
 		if guaranteed.FitInMaxUndef(sq.GetAllocatedResource()) {
 			return
 		}
--- a/pkg/scheduler/objects/queue_test.go
+++ b/pkg/scheduler/objects/queue_test.go
@@ -1919,8 +1919,8 @@ func TestFindQueueByAppID(t *testing.T) {
 
 // nolint: funlen
 func TestFindEligiblePreemptionVictims(t *testing.T) {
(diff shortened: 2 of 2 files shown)


craigcondit (commented):
This looks good to me. @wilfred-s can you also take a look? Want to be sure we're not introducing any edge cases.

craigcondit (changes requested):
I don't think this approach is correct... I'm working on where I think the real issue is and once I've confirmed this will update further.

craigcondit:
@wusamzong can you review https://github.com/apache/yunikorn-core/pull/639 as an alternative and see if you still experience this issue? I think that PR is covering the real problem.

wilfred-s:
I agree when I look at this PR and walk through the issue, I do not think this will fix it. The calculation should include the max resources of the queue but only if there is a guaranteed resource. If guaranteed in the whole root to leaf hierarchy is empty we should not even attempt preemption.

PR #639, opened 2023-09-07
[YUNIKORN-1951] Ensure that queues with no guaranteed resources do not trigger preemption

craigcondit:
### What is this PR for?
When a queue has no guaranteed resources, yunikorn always sees that queue as being within guaranteed limits. For unspecified resources, this is correct, but in the case where no guarantees are given at all, this should not be the case.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1951

New unit tests added to cover this case.

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -907,6 +907,11 @@ func (r *Resource) HasNegativeValue() bool {
 	return false
 }
 
+// IsEmpty returns true if the resource is nil or has no component resources specified.
+func (r *Resource) IsEmpty() bool {
+	return r == nil || len(r.Resources) == 0
+}
+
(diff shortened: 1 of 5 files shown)


wilfred-s (approved):
LGTM +1

wusamzong (approved):
LGTM+1. I have written an E2E test on my local side to test this, and it's working well.

zhuqi-lucas (approved):
LGTM

PR #642, opened 2023-09-10
[YUNIKORN-1944] calculate headroom for wildcard cases

FrankYang0529:
### What is this PR for?

Can't get the correct headroom for the user wildcard limit.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1944

Covered by unit tests.

### DIFF
--- a/pkg/scheduler/ugm/queue_tracker.go
+++ b/pkg/scheduler/ugm/queue_tracker.go
@@ -237,9 +237,12 @@ func (qt *QueueTracker) setLimit(hierarchy []string, maxResource *resources.Reso
 	}
 }
 
-func (qt *QueueTracker) headroom(hierarchy []string) *resources.Resource {
+func (qt *QueueTracker) headroom(hierarchy []string, trackType trackingType) *resources.Resource {
 	log.Log(log.SchedUGM).Debug("Calculating headroom",
-		zap.Strings("queue path", hierarchy))
+		zap.String("queue path", qt.queuePath),
+		zap.Strings("hierarchy", hierarchy),
+		zap.Int("tracking type", int(trackType)),
+	)
 	// depth first: all the way to the leaf, create if not exists
 	// more than 1 in the slice means we need to recurse down
 	var headroom, childHeadroom *resources.Resource
@@ -248,7 +251,7 @@ func (qt *QueueTracker) headroom(hierarchy []string) *resources.Resource {
 		if qt.childQueueTrackers[childName] == nil {
 			qt.childQueueTrackers[childName] = newQueueTracker(qt.queuePath, childName)
 		}
-		childHeadroom = qt.childQueueTrackers[childName].headroom(hierarchy[1:])
+		childHeadroom = qt.childQueueTrackers[childName].headroom(hierarchy[1:], trackType)
 	}
 
 	// arrived at the leaf or on the way out: check against current max if set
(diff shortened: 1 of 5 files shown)


manirajv06 (commented):
Just for my understanding: Why https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/ugm/queue_tracker.go#L124-L135 block of code is not working correctly for max resources when it works for max applications?

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -256,16 +259,34 @@ func (qt *QueueTracker) headroom(hierarchy []string) *resources.Resource {
...
+			zap.Int("tracking type", int(trackType)),
 			zap.Stringer("max resource", qt.maxResources),
 			zap.Stringer("headroom", headroom))
+	} else if resources.Equals(nil, childHeadroom) {
+		// Only user wildcard case gets to this block, so we only check user tracking type here.
+		// All non-specific groups get into a same wildcard cumulative group limit.
+		// The maxResources in wildcard group limit is not empty. Its qt.maxResources is not empty,
+		// so it gets calculated headroom in the upper block.
+		// ref: https://docs.google.com/document/d/1yCZSoHTrsistHktmPwgreY05u26eI35XwOYdHWP0LrQ/edit#heading=h.2zonl4qy889f
+		if trackType == user {
```
manirajv06, line 279: Please use the below comment:

fall back on wild card user limit settings to calculate headroom only for unnamed users. For unnamed groups, "*" group tracker object would be used using the above if block to calculate headroom because resource usage added together for all unnamed groups under "*" group tracker object.

manirajv06 (changes requested):
Sorry for the delay.

PR depends on wild card user limit settings to calculate headroom for unnamed users. So, we need to add config validations to ensure named user limit settings should not exceed wild card user limit settings irrespective of queue levels. By catching this, we can set the correct expectations for the end users. 

For example, 

root.p1.c1 -> p1: wild user “*” config is 10 cores, c1 -> user “user1" config is 11 cores. Is this allowed? Since we are comparing c1:user1:cores with p1:*:cores for headroom calculation, we should this validation in config parsing.

Please file a separate jira to handle this.

pkg/scheduler/ugm/queue_tracker.go
FrankYang0529: Updated comment. I change few words for better understanding. Thanks for the review.

FrankYang0529:
> For example,
> 
> root.p1.c1 -> p1: wild user “_” config is 10 cores, c1 -> user “user1" config is 11 cores. Is this allowed? Since we are comparing c1:user1:cores with p1:_:cores for headroom calculation, we should this validation in config parsing.
> 
> Please file a separate jira to handle this.

I think the parent level wildcard limit lower than child level user limit is not a problem, because we check whether there is `childHeadrrom` first. If there is `childHeadroom`, we don't check wildcard limit in parent level.

However, the reverse case is a problem. We need to check whether parent level user limit is lower than child level wildcard limit. Thanks for catching this. 👍 I will file a jira after this PR is merged.

pkg/scheduler/ugm/queue_tracker.go
manirajv06: I don't think comment has been updated. Providing links to the design doc and all is unnecessary. At times, we keep changing the implementation based on the situations.

Just replace it with the below comment:

// Fall back on wild card user limit settings to calculate headroom only for unnamed users. 
// For unnamed groups, "\*" group tracker object would be used using the above if block to calculate headroom 
// because resource usage added together for all unnamed groups under "\*" group tracker object.

manirajv06:
Since this pr is bit old, Can you rebase and run the tests again?

manirajv06 (changes requested):
Please check the comments

pkg/scheduler/ugm/queue_tracker.go
FrankYang0529: Okay. I will update it. Just curious, is "if" a redundant work in the sentence?

// For unnamed groups, "*" group tracker object would be used using the above "if" block to calculate headroom

FrankYang0529:
> Since this pr is bit old, Can you rebase and run the tests again?

Updated the comment and rebase it. Thanks

manirajv06 (approved):

FrankYang0529:
> > For example,
> > root.p1.c1 -> p1: wild user “_” config is 10 cores, c1 -> user “user1" config is 11 cores. Is this allowed? Since we are comparing c1:user1:cores with p1:_:cores for headroom calculation, we should this validation in config parsing.
> > Please file a separate jira to handle this.
> 
> I think the parent level wildcard limit lower than child level user limit is not a problem, because we check whether there is `childHeadrrom` first. If there is `childHeadroom`, we don't check wildcard limit in parent level.
> 
> However, the reverse case is a problem. We need to check whether parent level user limit is lower than child level wildcard limit. Thanks for catching this. 👍 I will file a jira after this PR is merged.

Create [YUNIKORN-2119](https://issues.apache.org/jira/browse/YUNIKORN-2119) to follow up.

PR #643, opened 2023-09-10
[YUNIKORN-1967] Update the help message in the application_total metric

brandboat:
### What is this PR for?
The help message of the application_total metric  lack of `failed` state.
https://github.com/apache/yunikorn-core/blob/master/pkg/metrics/scheduler.go#L89
application_total metric record applications under failed state
https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/objects/application_state.go#L233

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1967

this pr improve help message in prometheus GaugeOpts, no need to add test

### Questions:

### DIFF
--- a/pkg/metrics/scheduler.go
+++ b/pkg/metrics/scheduler.go
@@ -86,7 +86,7 @@ func InitSchedulerMetrics() *SchedulerMetrics {
 			Namespace: Namespace,
 			Subsystem: SchedulerSubsystem,
 			Name:      "application_total",
-			Help:      "Total number of applications. State of the application includes `running` and `completed`.",
+			Help:      "Total number of applications. State of the application includes `running`, `completed` and `failed`.",
 		}, []string{"state"})
 
 	s.node = prometheus.NewGaugeVec(


pkg/metrics/scheduler.go
```diff
@@ -86,7 +86,7 @@ func InitSchedulerMetrics() *SchedulerMetrics {
 			Namespace: Namespace,
 			Subsystem: SchedulerSubsystem,
 			Name:      "application_total",
-			Help:      "Total number of applications. State of the application includes `running` and `completed`.",
+			Help:      "Total number of applications. State of the application includes `running`, `completed` and `failed`.",
```
wusamzong, line 89: I think the state can also be either 'accepted' or 'rejected'[1], and we can also include these states in the message!

[1] [metrics/init.go#78](https://github.com/apache/yunikorn-core/blob/master/pkg/metrics/init.go#L78C38-L78C38)

brandboat: Correct me if I'm wrong, but I didn't see those labels (i.e. accepted/rejected) in https://github.com/apache/yunikorn-core/blob/1540a5508db1816f11645ab86781a4685de08ebc/pkg/metrics/scheduler.go#L267-L317. Not sure where are these two labels implemented ?

wusamzong: These two labels are implemented in [here](https://github.com/apache/yunikorn-core/blob/master/pkg/metrics/scheduler.go#L251-L265)

wusamzong: Sorry I missed the part about application_submission_total in metrics/scheduler.go.

FrankYang0529 (approved):
LGTM.

manirajv06 (approved):

PR #645, opened 2023-09-11
[YUNIKORN-1971] Update linter configuration

brandboat:
### What is this PR for?
update linter configuration

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [x] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1971

### Questions:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,15 +15,13 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-#
 
 # options for analysis running
 run:
-  go: '1.21'
   issues-exit-code: 1
   skip-dirs-use-default: true
   modules-download-mode: readonly
-  deadline: 5m
+  timeout: 5m
 
 # settings of specific linters
 linters-settings:
@@ -98,4 +96,3 @@ issues:
 
   # Show only new issues created in git patch with set file path.
   # new-from-patch: path/to/patch/file
-


.golangci.yml
```diff
@@ -15,15 +15,13 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-#
 
 # options for analysis running
 run:
-  go: '1.21'
```
doupache, line 22: If we remove this line , which golang version will be used ?

craigcondit: I believe it will use the default version executed with 'go'.

brandboat: > If we remove this line , which golang version will be used ?

Default: use Go version from the go.mod file

Quoted from golangci-lint configuration
```
  # Define the Go version limit.
  # Mainly related to generics support since go1.18.
  # Default: use Go version from the go.mod file, fallback on the env var `GOVERSION`, fallback on 1.18
  go: '1.19'
```

brandboat: While go.mod still use version 1.20. Should we upgrade it to 1.21 ? gentle ping @craigcondit, @FrankYang0529 . If so, I'm willing to do this.

craigcondit: No, we should leave it at 1.20 as that impacts downstream consumers. We don't (yet) require any go 1.21 features to compile.

craigcondit (approved):
+1 LGTM.

PR #646, opened 2023-09-12
[YUNIKORN-1973] Queue Limits configuration at same level should not c…

manirajv06:
…ompare each other

### What is this PR for?
Limits configured for queues at same level should not be compared against each other. As long as it lesser than its immediate or ancestor parent limit settings, it should be allowed.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1973

### Questions:


pkg/common/configs/configvalidator.go
```diff
@@ -58,6 +58,7 @@ const (
 	errNonExistingQueue
 	errQueueNotLeaf
 	errLastQueueLeaf
+	Separator = "~"
```
wilfred-s, line 61: Why is that here? Seems unused.

wilfred-s (changes requested):
The tests for the configs belong in the config validation not in the web handler.
Besides that the code looks much better now.

manirajv06:
> The tests for the configs belong in the config validation not in the web handler.

Yes, moved related test code to the correct place. Retained few cases in handler test itself to ensure "max application as zero" goes through without throwing up any errors.

wilfred-s (approved):
LGTM

PR #650, opened 2023-09-14
[YUNIKORN-1986] add ugm CanRunApp to TryReservedAllocate

FrankYang0529:
### What is this PR for?

Need to check whether an user can run a reserved allocation.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1986

Covered by unit tests.

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -1435,7 +1435,7 @@ func (sq *Queue) TryReservedAllocate(iterator func() NodeIterator) *Allocation {
 						zap.String("appID", appID))
 					return nil
 				}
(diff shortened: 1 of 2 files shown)


wilfred-s:
We should never have created a reservation for an application unless that application was allowed to run already. The only time that we can return an allocation that is a reservation on a node is from `application.tryAllocate()` or the preemptor.

Not sure how often we will hit the case that we have a reservation for an app that we then cannot start but the overhead should be minimal.

wilfred-s (approved):
LGTM
@manirajv06 if you have nothing to add we can commit this.

manirajv06 (approved):

PR #652, opened 2023-09-17
[YUNIKORN-1984] Expose other config value in the REST response

FrankYang0529:
### What is this PR for?

Expose other config values in the REST response.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1984

Covered by unit tests.

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -424,9 +424,17 @@ func getContainerHistory(w http.ResponseWriter, r *http.Request) {
 func getClusterConfig(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 
-	conf := configs.ConfigContext.Get(schedulerContext.GetPolicyGroup())
 	var marshalledConf []byte
 	var err error
+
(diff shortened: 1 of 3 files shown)


pkg/webservice/handlers.go
```diff
@@ -58,6 +58,11 @@ const GroupNameMissing = "Group name is missing"
 const ApplicationDoesNotExists = "Application not found"
 const NodeDoesNotExists = "Node not found"
 
+type outputConfig struct {
+	*configs.SchedulerConfig
+	Extra map[string]string `yaml:",omitempty" json:",omitempty"`
+}
```
manirajv06, line 64: Can we move this struct to `webservice/dao/config_info`? so that https://github.com/apache/yunikorn-core/pull/664 can also use this. Also lets keep it as just `config`.

manirajv06 (changes requested):
looks good. Minor nit

manirajv06 (approved):

PR #653, opened 2023-09-18
[YUNIKORN-1980] Remove deprecated log REST calls

brandboat:
### What is this PR for?
~~return 410 in deprecated REST APIs `PUT /ws/v1/loglevel/{level}` and `GET /ws/v1/loglevel`~~
remove deprecated REST APIs `PUT /ws/v1/loglevel/{level}` and `GET /ws/v1/loglevel`

### What type of PR is it?
* [x] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1980

~~Covered by unit tests.~~

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -622,20 +622,6 @@ func getApplication(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func setLogLevel(w http.ResponseWriter, r *http.Request) {
-	writeHeaders(w)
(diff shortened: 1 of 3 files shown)


brandboat:
gentle ping @wilfred-s , could you take a look ?

brandboat:
Thanks for the comment
> If we're going to break compatibility in 1.4, let's go all the way and get rid of the dead code. A 404 (the default) is just as good as a 410.
This sounds reasonable, I'll remove these two deprecated apis in the next commit, many thanks !

manirajv06 (approved):

PR #654, opened 2023-09-18
[YUNIKORN-1991] Fix wrong log key in Application

yui2010:
### What is this PR for?
we show task user info log in AddAllocationAsk method of Application as
zap.String("appID",sa.user.User)
it should be zap.String("user",sa.user.User)

### What type of PR is it?

 - [x] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1991
### Questions:


craigcondit (approved):
+1 LGTM.

PR #658, opened 2023-09-20
[YUNIKORN-1995] Preemption CheckPreemptionQueueGuarantee need to compare aggregated Victims with ask resource

zhuqi-lucas:
### What is this PR for?
Preemption CheckPreemptionQueueGuarantee need to compare aggregated Victims with ask resource
PR try to fix following problems:
1. Compare the aggregated PotentialVictims, to see if it is enough to fit the ask resource.
2. And about the aggregated process, make sure preempted queue after preempted, it should be larger than the guaranteed resource.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1995

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -189,13 +189,21 @@ func (p *Preemptor) checkPreemptionQueueGuarantees() bool {
 		return false
 	}
 
+	// If current queue after adding the ask is already at or above guaranteed, we can't preempt
 	currentQueue.AddAllocation(p.ask.GetAllocatedResource())
+	if currentQueue.IsAtOrAboveGuaranteedResource() {
+		return false
+	}
 
+	aggregatedVictims := resources.NewResource()
 	// remove each allocation in turn, validating that at some point we free enough resources to allow this ask to fit
 	for _, snapshot := range queues {
 		for _, alloc := range snapshot.PotentialVictims {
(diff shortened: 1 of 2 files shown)


zhuqi-lucas:
Hi @craigcondit, @wilfred-s 
I am not sure if we need to fix about the queue check for preemption, i push a draft code about this.

And one question about the unit tests errors, 
For example, when the preemption ask in the queue root.A, it preemption other queue's asks such as queue root.b, if we need to make sure after preempting other queues, the preemption ask's queue root.A not equal or exceed the guaranteed resources?

And if we can let root.A just equal to the guaranteed resource? I see unit test errors, the unit test case is after root.a preempt other asks, it will just equal to the guaranteed resources, but not execeed.

craigcondit (changes requested):
I don't think this is the correct approach. We simply need to check that when iterating the allocations here, we skip those that are part of the current queue. This method is only used to determine if it's even possible to free enough resources for the ask to be satisfied by preemption. This is why we keep reducing the snapshot usage until we either fit the ask or run out of allocations.

zhuqi-lucas:
Thank you, @craigcondit, for explanation, close this PR, duplicated with @lixmgl  jira.

PR #659, opened 2023-09-20
[YUNIKORN-1990] Fixing intra-queue preemption: updating checkPreemption…

lixmgl:
…QueueGuarantees to check both snapshot and currentQueue

### What is this PR for?
Found intra-queue preemption happened, which caused high tier spark application got shut down unexpectedly in production.

Previous logic seems not working for snapshot is a different queue as currentQueue.
More details can be found in first comment on https://issues.apache.org/jira/browse/YUNIKORN-1990

This pr including changes:
1. A new variable availableResource is defined as a clone of p.headRoom
2. Within each snapshot of the queue, it will check if it's above or at its guaranteed resource before removing any allocations (alloc). 
3. If the snapshot is not above or at its guaranteed resource, it will add back the alloc that was removed

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1990

Will update test in next iteration. 

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -190,13 +190,22 @@ func (p *Preemptor) checkPreemptionQueueGuarantees() bool {
 	}
 
 	currentQueue.AddAllocation(p.ask.GetAllocatedResource())
+	if !currentQueue.IsWithinMaxResource() {
+		return false
+	}
+	availableResource := p.headRoom.Clone()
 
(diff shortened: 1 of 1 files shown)


wilfred-s:
JIra as been closed as won't fix, closing the PR also

PR #660, opened 2023-09-20
[YUNIKORN-1988] Check additional victims and the satisfaction of guaranteed resource at a parent queue level

lixmgl:
…n of guaranteed resource at a parent queue level

### What is this PR for?
We noticed Preemption happened when a queue lower than its guaranteed capacity, caused issue in production.
This pr including changes:

1. The calculation for extraVictims and appending it to victims is now within a condition that checks if needsAdditionalVictims is true. extraVictims are only calculated and appended when there's a need for additional victims based on the tryNodes function result.
2. p.calculateAdditionalVictims(victims, nodeID) now takes in nodeID as a parameter, showing that this function may now consider the nodeID in its calculation for extra victims.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1988

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -399,7 +399,8 @@ func (p *Preemptor) checkPreemptionPredicates(predicateChecks []*si.PreemptionPr
 }
 
 // calculateAdditionalVictims finds additional preemption victims necessary to ensure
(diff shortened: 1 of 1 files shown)


lixmgl:
Closing this pr, I will create a new one soon.

PR #661, opened 2023-09-20
[YUNIKORN-1996] Change a log about queue update failure due to max capacity reached from Warn to Debug

yzhangal:
### What is this PR for?
Tons of logs  (62k of them in 3 seconds for the same request) because the max capacity of a queue has reached. 
This jira changes the Warn to Debug to avoid the log spam for the same reason.

### What type of PR is it?
* [ ] - Bug Fix
* [ x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-1996

Observe the log level of the particular changed log is now Debug rather than Warn.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1490,7 +1490,7 @@ func (sa *Application) tryNode(node *Node, ask *AllocationAsk) *Allocation {
 	alloc := NewAllocation(common.GetNewUUID(), node.NodeID, ask)
 	if node.AddAllocation(alloc) {
 		if err := sa.queue.IncAllocatedResource(alloc.GetAllocatedResource(), false); err != nil {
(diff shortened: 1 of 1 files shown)


craigcondit (changes requested):
We should not change this. It is a warning because if this condition occurs, scheduling has broken as we are no longer accounting for resources properly. The "log spew" is a symptom of a much larger problem, and hiding this away by making it a debug message is the wrong approach.

yzhangal:
> We should not change this. It is a warning because if this condition occurs, scheduling has broken as we are no longer accounting for resources properly. The "log spew" is a symptom of a much larger problem, and hiding this away by making it a debug message is the wrong approach.

Thanks Craig, I closed the jira and I'm closing this PR too.

PR #662, opened 2023-09-21
[YUNIKORN-1993] race condition in allocation removal

wilfred-s:
### What is this PR for?
Race between allocation removal and the timed transition to Completed state change in the partition code.
If the removal of an allocation is in progress when the state transition causes the queue removal (entry of the Completed state) the queue will not be updated and the allocation will leak.

No test case as the timing required is not reproducible in unit or e2e tests.

### What type of PR is it?
* [X ] - Bug Fix
* [X] - Hot Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1993

NA, race condition not testable

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -685,6 +685,9 @@ func (pc *PartitionContext) removeNodeAllocations(node *objects.Node) ([]*object
 				zap.String("nodeID", node.NodeID))
 			continue
 		}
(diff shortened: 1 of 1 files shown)


zhuqi-lucas (approved):
LGTM And it seems hard to mock the test cases for this.

craigcondit (approved):
+1 LGTM.

PR #667, opened 2023-10-02
[YUNIKORN-2010] Yunikorn web shows negative number of applications in application history

chenyulin0719:
### What is this PR for?
This PR fixed the issue of nagative running application count in core.
Add below actions in application fsm:
- Increase running application count when Completing -> Running
- Decrease running application count when Starting -> Failing

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2010

I put the reproduce steps in YUNIKORN-2010.

NA

### Questions:
NA

### DIFF
--- a/pkg/scheduler/objects/application_state.go
+++ b/pkg/scheduler/objects/application_state.go
@@ -169,24 +169,29 @@ func NewAppState() *fsm.FSM {
 			fmt.Sprintf("enter_%s", Starting.String()): func(_ context.Context, event *fsm.Event) {
 				app := event.Args[0].(*Application) //nolint:errcheck
 				app.startTime = time.Now()
-				app.queue.incRunningApps(app.ApplicationID)
 				app.setStateTimer(app.startTimeout, app.stateMachine.Current(), RunApplication)
+				app.queue.incRunningApps(app.ApplicationID)
 				metrics.GetQueueMetrics(app.queuePath).IncQueueApplicationsRunning()
 				metrics.GetSchedulerMetrics().IncTotalApplicationsRunning()
 			},
-			fmt.Sprintf("enter_%s", Completing.String()): func(_ context.Context, event *fsm.Event) {
-				app := event.Args[0].(*Application) //nolint:errcheck
-				if event.Src == Starting.String() {
+			fmt.Sprintf("leave_%s", Starting.String()): func(_ context.Context, event *fsm.Event) {
+				if event.Dst != Running.String() {
+					app := event.Args[0].(*Application) //nolint:errcheck
 					app.queue.decRunningApps()
 					metrics.GetQueueMetrics(app.queuePath).DecQueueApplicationsRunning()
 					metrics.GetSchedulerMetrics().DecTotalApplicationsRunning()
 				}
+			},
+			fmt.Sprintf("enter_%s", Completing.String()): func(_ context.Context, event *fsm.Event) {
+				app := event.Args[0].(*Application) //nolint:errcheck
 				app.setStateTimer(completingTimeout, app.stateMachine.Current(), CompleteApplication)
 			},
 			fmt.Sprintf("leave_%s", New.String()): func(_ context.Context, event *fsm.Event) {
-				app := event.Args[0].(*Application) //nolint:errcheck
-				metrics.GetQueueMetrics(app.queuePath).IncQueueApplicationsAccepted()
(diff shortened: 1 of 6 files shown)


pkg/scheduler/objects/application_state.go
```diff
@@ -205,6 +205,14 @@ func NewAppState() *fsm.FSM {
 					app.rejectedMessage = event.Args[1].(string) //nolint:errcheck
 				}
 			},
+			fmt.Sprintf("enter_%s", Running.String()): func(_ context.Context, event *fsm.Event) {
```
wilfred-s, line 208: This counts apps twice: Starting increased the value and now again in Running

pkg/scheduler/objects/application_state.go
```diff
@@ -231,6 +239,11 @@ func NewAppState() *fsm.FSM {
 				app := event.Args[0].(*Application) //nolint:errcheck
 				metrics.GetQueueMetrics(app.queuePath).IncQueueApplicationsFailed()
 				metrics.GetSchedulerMetrics().IncTotalApplicationsFailed()
+				if event.Src == Starting.String() {
+					app.queue.decRunningApps()
+					metrics.GetQueueMetrics(app.queuePath).DecQueueApplicationsRunning()
+					metrics.GetSchedulerMetrics().DecTotalApplicationsRunning()
+				}
```
wilfred-s, line 246: We should track this from the Starting state. We are not tracking the metrics in the same way for different state entries or exits. Makes it difficult to comprehend what we do

wilfred-s (changes requested):
Also see the comment in the jira.
We need a more consistent solution and fix one more broken tracking point in Resuming

pkg/scheduler/objects/application_state.go
chenyulin0719: Actually it didn't count twice here, 'enter_Running' only increase app count if it's from Completing state. 

Accepting-> Starting (+1)
Starting -> Completing (-1)
Completing ->Running (+1)

But I agree that it's difficult to track. To simplify it, I will modify this PR to follow what you suggest in Jira.

pkg/scheduler/objects/application_state.go
chenyulin0719: Thanks @wilfred-s,

I just updated the PR,
1. Cleanup resuming
2. Below is all ‘Running Application’ metric changes in application_state. (queue, queue metric, scheduler metric)
   - enter Starting → increment
   - leave Completing → increment if destination is Running
   - leave Starting → decrement, unless destination is Running
   - leave Running → decrement
3. And I found that ‘Accepted Applications’ was incorrect, so I fixed it in the same PR:
   - Was: leave New → increment
   - Is:     leave New → ncrement if destination is not Reject
4. Add test to walk through all nodes.

pbacsko (commented):
Looks good, I just have one question.

pkg/scheduler/objects/application_state.go
```diff
@@ -216,6 +216,8 @@ func NewAppState() *fsm.FSM {
 				// account for going back into running state
 				if event.Dst == Running.String() {
 					app.queue.incRunningApps(app.ApplicationID)
+					metrics.GetQueueMetrics(app.queuePath).IncQueueApplicationsRunning()
+					metrics.GetSchedulerMetrics().IncTotalApplicationsRunning()
```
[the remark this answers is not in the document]
chenyulin0719: Hi @pbacsko, 
Your suggestion make sense to me. In the case we can bound all RunningApps metrics changes to Starting/Running state.
- enter/leave Starting
- enter/leave Running

Let me change this now.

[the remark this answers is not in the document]
chenyulin0719: @pbacsko PR updated.

Below is all the changes for ‘Running Application’ metric in application_state. (queue, queue metric, scheduler metric)
   - enter Starting → increment
   - leave Starting → decrement, unless destination is Running
   - enter Running → increment if source is Completing
   - leave Running → decrement

wilfred-s (approved):
LGTM
thank you for that last optimisation around the enter running call.

PR #668, opened 2023-10-03
[YUNIKORN-2012] group quota is not considered during pod scheduling

manirajv06:
### What is this PR for?

Problem: When config has limits set only for group, its quota was not getting honoured. User belonging to the group are allowed to use more resources than the allowed limit.

Besides fixing the issue, this pr also simplifies the "cleaning up old user/group limits resource usage when config changes" codepath by removing the unnecessary clean up processing for the queue path. For example, say group "test" limit is set for queue path a.b.c. A config change had happened for the removal of "test" from a.b.c. So, clean up is required only when a.b.c queue path comes. But, clean up happens even when we are processing a & b from queue configs just because those queues has been tracked for the group "test". So far, as long as the queue has been tracked for a group, then, even that queue enters the clean up loop. This unnecessary processing reset the values already set not only for the a & b, even for c. which is not needed and correct.

Now simplified the whole codepath by comparing the old config with the new config. If new config doesn't have the limit set for group "test", then clean up triggers.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2012

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -415,21 +415,17 @@ func (m *Manager) clearEarlierSetLimits(newUserLimits map[string]map[string]*Lim
 	m.clearEarlierSetUserLimits(newUserLimits)
 }
 
+// clearEarlierSetUserLimits Traverse new user config and decide whether earlier usage needs to be cleared or not
+// by comparing with the existing config. Reset earlier usage only config set earlier but not now
 func (m *Manager) clearEarlierSetUserLimits(newUserLimits map[string]map[string]*LimitConfig) {
-	if len(m.userLimits) == 0 {
-		return
-	}
 	m.RLock()
 	defer m.RUnlock()
-	// Need to clear user config only when
-	// 1. config set earlier but not now
-	// 2. user already tracked for the queue path
 	for queuePath, limitConfig := range m.userLimits {
 		// Is queue path exists?
 		if newUserLimit, ok := newUserLimits[queuePath]; !ok {
 			for u := range limitConfig {
 				if ut, utExists := m.userTrackers[u]; utExists {
-					m.clearEarlierSetUserLimit(ut, queuePath)
+					m.resetUserEarlierUsage(ut, queuePath)
 				}
 			}
 		} else {
@@ -437,7 +433,7 @@ func (m *Manager) clearEarlierSetUserLimits(newUserLimits map[string]map[string]
 			for u := range limitConfig {
 				if _, ulExists := newUserLimit[u]; !ulExists {
 					if ut, utExists := m.userTrackers[u]; utExists {
-						m.clearEarlierSetUserLimit(ut, queuePath)
+						m.resetUserEarlierUsage(ut, queuePath)
 					}
 				}
 			}
@@ -445,7 +441,10 @@ func (m *Manager) clearEarlierSetUserLimits(newUserLimits map[string]map[string]
 	}
 }
 
-func (m *Manager) clearEarlierSetUserLimit(ut *UserTracker, queuePath string) {
+// resetUserEarlierUsage Clear or reset earlier usage only when user already tracked for the queue path.
+// Reset the max apps and max resources to default, unlink the end leaf queue of queue path from its immediate parent and
+// eventually remove user tracker object itself from ugm if it can be removed.
+func (m *Manager) resetUserEarlierUsage(ut *UserTracker, queuePath string) {
 	// Is this user already tracked for the queue path?
 	if ut.IsQueuePathTrackedCompletely(queuePath) {
 		log.Log(log.SchedUGM).Debug("Need to clear earlier set configs for user",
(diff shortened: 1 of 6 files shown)


manirajv06:
Notes:

Instead of locking objects for longer duration, moved the locks to the nearest place so that scheduler performance doesn't degrade during config changes. Changed the whole approach of clearing old configs by storing the new configs in separate maps and do the clearance only comparing the existing configs with new configs. For the newer limit configs, things happen while during the traversal and storing in new configs itself. Once all the work is over, replace the existing configs with new configs by keeping the write lock. This way, mostly read lock happens on the existing configs and write lock only during the final switch over.

In addition, some more hidden issues has been identified and fixed them too.

1. When group limits removed in new config (root.parent.leaf) and contains the same group config in another level (root.parent),  during the clearance step, we reduce/reset the resource usage for the entire group queue hierarchy (root.parent.leaf). i.e for root.parent though it has same group limit configs.
2. Changes in unlinking queue trackers.

pkg/scheduler/ugm/manager.go
```diff
@@ -365,127 +381,159 @@ func (m *Manager) internalProcessConfig(cur configs.QueueConfig, queuePath strin
...
-		if len(appUsersMap) > 0 {
-			for app, user := range appUsersMap {
-				ut := m.userTrackers[user]
-				ut.setGroupForApp(app, nil)
+func (m *Manager) clearEarlierSetUserLimits(newUserLimits map[string]map[string]*LimitConfig) {
+	if len(m.userLimits) == 0 {
+		return
+	}
+	m.RLock()
+	defer m.RUnlock()
```
wilfred-s, line 423: needs to be before the check of the userlimits

pkg/scheduler/ugm/manager.go
```diff
@@ -365,127 +381,159 @@ func (m *Manager) internalProcessConfig(cur configs.QueueConfig, queuePath strin
...
-// clearEarlierSetLimits Clear already configured limits of users and groups for which limits have been configured before but not now
-func (m *Manager) clearEarlierSetLimits(userLimits map[string]bool, groupLimits map[string]bool, queuePath string) error {
-	// Clear already configured limits of group for which limits have been configured before but not now
-	for _, gt := range m.groupTrackers {
-		appUsersMap := m.clearEarlierSetGroupLimits(gt, queuePath, groupLimits)
-		if len(appUsersMap) > 0 {
-			for app, user := range appUsersMap {
-				ut := m.userTrackers[user]
-				ut.setGroupForApp(app, nil)
+func (m *Manager) clearEarlierSetUserLimits(newUserLimits map[string]map[string]*LimitConfig) {
```
wilfred-s, line 418: keep the comment

pkg/scheduler/ugm/manager.go
```diff
@@ -365,127 +381,159 @@ func (m *Manager) internalProcessConfig(cur configs.QueueConfig, queuePath strin
...
-
-	// Clear already configured limits of user for which limits have been configured before but not now
-	for _, ut := range m.userTrackers {
-		m.clearEarlierSetUserLimits(ut, queuePath, userLimits)
-	}
-	return nil
 }
 
-func (m *Manager) clearEarlierSetUserLimits(ut *UserTracker, queuePath string, userLimits map[string]bool) {
+func (m *Manager) clearEarlierSetUserLimit(ut *UserTracker, queuePath string) {
```
wilfred-s, line 448: This really removes, while the previous function does the looping:
the names look too much alike difficult to understand the difference between the two.

pkg/scheduler/ugm/manager.go
```diff
@@ -365,127 +381,159 @@ func (m *Manager) internalProcessConfig(cur configs.QueueConfig, queuePath strin
...
+		log.Log(log.SchedUGM).Debug("Cleared earlier set limit configs for user",
+			zap.String("user", ut.userName),
+			zap.String("queue path", queuePath))
+		if ut.canBeRemoved() {
+			delete(m.userTrackers, ut.userName)
+		}
+	}
+}
+
+func (m *Manager) clearEarlierSetGroupLimits(newGroupLimits map[string]map[string]*LimitConfig) {
```
wilfred-s, line 468: Similar comment as for the user: comments for the functions to explain what they do and the names look too much alike

pkg/scheduler/ugm/manager.go
```diff
@@ -365,127 +381,159 @@ func (m *Manager) internalProcessConfig(cur configs.QueueConfig, queuePath strin
...
+	if len(m.groupLimits) == 0 {
+		return
+	}
+	m.RLock()
+	defer m.RUnlock()
```
wilfred-s, line 473: move before the check

wilfred-s (changes requested):

manirajv06:
@wilfred-s Addressed review comments. Also, tried running #669 pr test changes with this pr and can see tests passing through without any issues. Can you take a final look and approve?

wilfred-s (approved):
LGTM

PR #669, opened 2023-10-06
[YUNIKORN-2024] clear limits only upon the right queue

doupache:
### What is this PR for?
Currently, we update the limits by pre-order traversing the queue configuration, so we call 'clearEarlierSetLimits' before the child queue gets visited. (https://github.com/apache/yunikorn-core/blob/bc7c00124c4cfb77c1f4cffe434227ac33d3e875/pkg/scheduler/ugm/manager.go#L324C19-L324C40)

For example. 

If the 'root' queue doesn't have any limits, and we update the 'parent' queue's limits from 'g0' to 'g1', we will lose any resource usage the parent is using because 'clearEarlierSetLimits' will first be called on the root queue.

 root        **limits : None**
   |
parent    **limits: groups:[ 'g1']**

We need to change to post-order traversal and keep every limit that has been set in 'userLimits' and 'groupLimits'. Then we can correctly determine which limits should be cleared."

 
This is a bug  when i working on [YuniKorn-1950](https://issues.apache.org/jira/browse/YUNIKORN-1950)

### What type of PR is it?
* [x] - Bug Fix

### What is the Jira issue?
[YuniKorn-2024](https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-2024)

new test case in manager_test.go

cc @manirajv06  please take a look.


doupache:
https://github.com/apache/yunikorn-core/pull/600#discussion_r1286898707
also pass on this pr ! 

cc @FrankYang0529

doupache:
CI Pass on my fork  https://github.com/doupache/yunikorn-core/actions/runs/6637146374
So I think is ready for review! 

the git diff may look a little bit confused. I basically just 

1. move 

```
userLimits := make(map[string]bool)
groupLimits := make(map[string]bool)
```
those initialization from `internalProcessConfig` to `UpdateConfig`
because we need to keep userLimits  and groupLimits in the whole queue traversing .
And pass them to `internalProcessConfig`.


2. preorder to postorder.

3. add test from https://github.com/apache/yunikorn-core/pull/600#discussion_r1286898707

@FrankYang0529 @manirajv06 please take a look.

wilfred-s:
There is an overlap with YUNIKORN-2012 in #668.
That PR fixes some further issues we saw. Before we proceed here we need to have that one committed.

doupache:
I will update this pr when https://github.com/apache/yunikorn-core/pull/668 get merge

wilfred-s:
@doupache please update the PR: in #668 the code fix went in for the issue. We still want and need the test you created to be added.

doupache:
cc @wilfred-s @FrankYang0529 @manirajv06 Please take a look

manirajv06 (approved):

PR #670, opened 2023-10-06
[YUNIKORN-2015] separate container metric from application in queue

FrankYang0529:
### What is this PR for?

Queue metric mixes application and container in one metric. It's better to separate them like scheduler metric.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2015

### DIFF
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -27,8 +27,9 @@ import (
 
 // QueueMetrics to declare queue metrics
 type QueueMetrics struct {
-	appMetrics      *prometheus.GaugeVec
-	ResourceMetrics *prometheus.GaugeVec
+	appMetrics       *prometheus.GaugeVec
+	containerMetrics *prometheus.CounterVec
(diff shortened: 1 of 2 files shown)


pkg/metrics/queue.go
```diff
@@ -42,7 +43,15 @@ func InitQueueMetrics(name string) CoreQueueMetrics {
 			Namespace: Namespace,
 			Subsystem: replaceStr,
 			Name:      "queue_app",
-			Help:      "Queue application metrics. State of the application includes `running`, `accepted`, `rejected`, `failed`, `completed`, `allocated`, `released`.",
+			Help:      "Queue application metrics. State of the application includes `running`, `accepted`, `rejected`, `running`, `failed`, `completed`.",
```
manirajv06, line 46: "running" state mentioned twice

manirajv06 (changes requested):

manirajv06 (approved):

PR #671, opened 2023-10-11
[YUNIKORN-1987] Identity config fields not set explicitly

Haser0305:
### What is this PR for?
Identuty config fields not set explicitly. 

I have added a struct called `DefaultValues` to keep track of the fields that the user has not entered. 
The function `CheckAndSetDefault` is used to determine if the value inside this field matches the predefined default value. 
If it does, the field name will be added to `DefaultValues.defaultFields`, and the value of this field will be overwritten with the corresponding default value of its data type. 

In the default settings, `uint64` is set to the maximum value of `uint64`, and `string` is set to `"defaultStringValue"`. 
I did not handle operations for map, slice, etc. because if the user does not provide any input, it will be nil.

To enable the above functionality, an `UnmarshalYAML` method has been implemented for each object that includes the `DefaultValues` struct. 

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-1987

### Questions:
* [ ] - Not sure if setting uint64 default value to the maximum is appropriate.

### DIFF
--- a/pkg/common/configs/config.go
+++ b/pkg/common/configs/config.go
@@ -32,11 +32,20 @@ import (
 	"github.com/apache/yunikorn-core/pkg/log"
 )
 
+const (
+	stringDefaultValue = "defaultStringValue"
+	uint64DefaultValue = uint64(18446744073709551615) // max in uint64
+)
+
 // The configuration can contain multiple partitions. Each partition contains the queue definition for a logical
 // set of scheduler resources.
 type SchedulerConfig struct {
 	Partitions []PartitionConfig
 	Checksum   string `yaml:",omitempty" json:",omitempty"`
+}
+
+type DefaultValues struct {
+	defaultFields []string
 }
 
 // The partition object for each partition:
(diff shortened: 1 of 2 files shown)


pbacsko (changes requested):
This approach is excessively complicated and invasive. We really need to think about whether this is truly needed. At many points, I can't just figure out right away what happens. Eg. what's this `UnmarshalYAML(func(interface{}) error)` stuff? How is this called? Does it implement an interface so it becomes a callback?

After re-reading a couple of times, the solution becomes clearer, but still, it's complicated enough to stop and think.

If differentation is necessary (as described in the JIRA), I suggest we handle individual fields when needed.

craigcondit (changes requested):
-1 on this. This adds significant complexity and doesn't really get us much benefit. Recommend scrapping this idea entirely.

Haser0305:
> Eg. what's this `UnmarshalYAML(func(interface{}) error)` stuff? How is this called? Does it implement an interface so it becomes a callback?

Yes, the Unmarshaler interface is used in the package yaml, which we used for decoding the configuration. It allows customization of the behavior during unmarshaling from a YAML document. In this case, it detects the default value, then records the field name into an additional field (DefaultValues) and updates the value.

If we don't need to distinguish between the not set and the null value, I think there is indeed no need to add this code.

Thank you for taking the time to review my pull request.

pbacsko:
Thanks for the effort @Haser0305. For the time being I'm closing this, let's continue the discussion with Mani under YUNIKORN-1987.

PR #673, opened 2023-10-11
[YUNIKORN-2031] Fix a bug on metric of pending pods/resources

Huang-Wei:
### What is this PR for?

Fix a bug in metric of pending pods/resources. The bug seems to present in all YK versions.

### What type of PR is it?

* [x] - Bug Fix

### What is the Jira issue?

- https://issues.apache.org/jira/browse/YUNIKORN-2031

I cannot find any existing test suite except for pkg/metrics/queue_test.go, but that's more simply checking the metric's existence. Pplease advice if you want tests in this PR, or in follow-up to add tests to cover metric accuracy generally.

For now, I did a local ad-hoc test to verify this PR works:

image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,7 @@ require (
 	github.com/HdrHistogram/hdrhistogram-go v1.0.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.1 // indirect
 	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
--- a/pkg/scheduler/objects/queue_test.go
+++ b/pkg/scheduler/objects/queue_test.go
@@ -23,15 +23,19 @@ import (
 	"reflect"
 	"sort"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	promtu "github.com/prometheus/client_golang/prometheus/testutil"
 	"gotest.tools/v3/assert"
 
 	"github.com/apache/yunikorn-core/pkg/common"
 	"github.com/apache/yunikorn-core/pkg/common/configs"
 	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/apache/yunikorn-core/pkg/events"
+	"github.com/apache/yunikorn-core/pkg/metrics"
 	"github.com/apache/yunikorn-core/pkg/scheduler/objects/template"
 	"github.com/apache/yunikorn-core/pkg/scheduler/policies"
 	siCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
@@ -266,6 +270,11 @@ func TestPriorityCalc(t *testing.T) {
 }
 
 func TestPendingCalc(t *testing.T) {
+	// Reset existing metric storage; otherwise this unit test would get metrics populated by other UTs.
+	// In long run, to make the metrics code more testable, we should pass instantiable Metrics obj to Queue
+	// instead of using a global Metrics obj at pkg/metrics/init.go.
+	metrics.Reset()
+
 	// create the root
 	root, err := createRootQueue(nil)
 	assert.NilError(t, err, "queue create failed")
@@ -284,6 +293,16 @@ func TestPendingCalc(t *testing.T) {
 	if !resources.Equals(leaf.pending, allocRes) {
 		t.Errorf("leaf queue pending allocation failed to increment expected %v, got %v", allocRes, leaf.pending)
 	}
+	metrics := []string{"yunikorn_root_queue_resource", "yunikorn_root_leaf_queue_resource"}
+	want := concatQueueResourceMetric(metrics, []string{`
+yunikorn_root_queue_resource{resource="memory",state="pending"} 100
+yunikorn_root_queue_resource{resource="vcores",state="pending"} 10
+`, `
+yunikorn_root_leaf_queue_resource{resource="memory",state="pending"} 100
+yunikorn_root_leaf_queue_resource{resource="vcores",state="pending"} 10
+`},
+	)
+	assert.NilError(t, promtu.GatherAndCompare(prometheus.DefaultGatherer, strings.NewReader(want), metrics...), "unexpected metrics")
 	leaf.decPendingResource(allocRes)
 	if !resources.IsZero(root.pending) {
 		t.Errorf("root queue pending allocation failed to decrement expected 0, got %v", root.pending)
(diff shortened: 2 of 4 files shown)


yzhangal (commented):
Thanks for the patch Wei. It seems

func (sq *Queue) decPendingResource(delta *resources.Resource) {

need a similar change. Right?

manirajv06:
> Thanks for the patch Wei. It seems
> 
> func (sq *Queue) decPendingResource(delta *resources.Resource) {
> 
> need a similar change. Right?

yes

manirajv06:
> Pplease advice if you want tests in this PR, or in follow-up to add tests to cover metric accuracy generally.

Yes, we should be doing metrics assertions in all possible places, at least for important state change events in objects like Queues, Application etc. Please file a separate jira. Though no. of changes is going to be high, it is good to do especially given that we are seeing increase in no. of bugs with respect to metrics recently.

Can we start with this? It is better to start by making changes in queue_test#TestPendingCalc only.

manirajv06 (changes requested):
Please see comments.

Huang-Wei:
Thanks @yzhangal @manirajv06 for the review. I added a minimum UT to repro the bug:

```console
=== RUN   TestPendingCalc
2023-10-12T15:57:52.289-0700	INFO	core.scheduler.queue	objects/queue.go:144	configured queue added to scheduler	{"queueName": "root"}
2023-10-12T15:57:52.290-0700	INFO	core.scheduler.queue	objects/queue.go:144	configured queue added to scheduler	{"queueName": "root.leaf"}
    queue_test.go:300: assertion failed: error is not nil: 
        
        Diff:
        --- metric output does not match expectation; want
        +++ got:
        @@ -1,10 +1,2 @@
        -(bytes.Buffer) # HELP yunikorn_root_leaf_queue_resource Queue resource metrics. State of the resource includes `guaranteed`, `max`, `allocated`, `pending`, `preempting`.
        -# TYPE yunikorn_root_leaf_queue_resource gauge
        -yunikorn_root_leaf_queue_resource{resource="memory",state="pending"} 100
        -yunikorn_root_leaf_queue_resource{resource="vcores",state="pending"} 10
        -# HELP yunikorn_root_queue_resource Queue resource metrics. State of the resource includes `guaranteed`, `max`, `allocated`, `pending`, `preempting`.
        -# TYPE yunikorn_root_queue_resource gauge
        -yunikorn_root_queue_resource{resource="memory",state="pending"} 100
        -yunikorn_root_queue_resource{resource="vcores",state="pending"} 10
        +(bytes.Buffer) 
         
        -
        : unexpected metrics
--- FAIL: TestPendingCalc (0.00s)
```

To ensure we have enough UT coverage on metrics, a follow-up PR would be a better bet.

go.mod
```diff
@@ -45,6 +45,7 @@ require (
 	github.com/HdrHistogram/hdrhistogram-go v1.0.1 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
```
Huang-Wei, line 48: This is generated by `go mod tidy`.

pkg/scheduler/objects/queue_test.go
```diff
@@ -285,13 +288,32 @@ func TestPendingCalc(t *testing.T) {
...
+	metrics := []string{"yunikorn_root_queue_resource", "yunikorn_root_leaf_queue_resource"}
+	want := concatQueueResourceMetric(metrics, []string{`
+yunikorn_root_queue_resource{resource="memory",state="pending"} 100
+yunikorn_root_queue_resource{resource="vcores",state="pending"} 10
+`, `
+yunikorn_root_leaf_queue_resource{resource="memory",state="pending"} 100
+yunikorn_root_leaf_queue_resource{resource="vcores",state="pending"} 10
+`},
+	)
+	assert.NilError(t, promtu.GatherAndCompare(prometheus.DefaultGatherer, strings.NewReader(want), metrics...), "unexpected metrics")
```
Huang-Wei, line 300: This style followed how Kubernetes' codebase test metrics.

go.mod
wilfred-s: Checked locally and it does not get added:
```
yunikorn-core % go mod tidy
yunikorn-core % git status
On branch master
Your branch is up to date with 'origin/master'.
yunikorn-core % go version
go version go1.21.3 darwin/arm64
yunikorn-core % go mod why github.com/davecgh/go-spew
# github.com/davecgh/go-spew
(main module does not need package github.com/davecgh/go-spew)
```

wilfred-s (changes requested):
This adds a lot of overhead. The call to `updateAllocatedAndPendingResourceMetrics()` updates three metrics when it only should do one. Three loops are run setting metrics for values that we know have not changed. We call the specific increment or decrement for each type separately. We need to update the metrics separately also.

The function should be split into three calls one for each type we track: allocated, pending and preempting.

wilfred-s:
> To ensure we have enough UT coverage on metrics, a follow-up PR would be a better bet.
Completely agree, a follow up jira is needed. Probably linked to the changes for using labels in the queue metrics instead of the names like we do now.

Huang-Wei:
> This adds a lot of overhead. The call to updateAllocatedAndPendingResourceMetrics() updates three metrics when it only should do one.

Agree. However, the function `updateAllocatedAndPendingResourceMetrics()` and how this pattern is used has been across several releases. In this PR, I followed existing pattern and focus on the **_accuracy_** issue, and try to make it as neat as possible so it's cherrypick-able.

For radical performance optimization on how metrics get updated, we should track in a separate JIRA.

go.mod
Huang-Wei: once switch to this PR's branch, `go mod tidy` shows the diff.

wilfred-s:
> Agree. However, the function updateAllocatedAndPendingResourceMetrics() and how this pattern is used has been across several releases. In this PR, I followed existing pattern and focus on the accuracy issue, and try to make it as neat as possible so it's cherrypick-able.

I am also looking at keeping accurate track. We generate metrics entries at a point in time without changes to the underlying values. This could hide the fact that we do not call the metrics at the right time or don't call it for certain changes. They seem accurate in the long run as they are updated as a side effect of another change. This might not show if churn and update rate is high but could show for lower change rates.

Huang-Wei:
@wilfred-s how should I proceed, like creating a separate PR to address the inefficiency issue and then rebase here?

Huang-Wei:
> I doubt that the extra change will affect the cherry pick.

That would cause the rebased PR not cherrypickable. That's why earlier I proposed to have this PR a pure fix (leave perf bits apart), and that would cherrypickable.

Anyway, let me try to include perf bits in this PR.

Huang-Wei:
> let me try to include perf bits in this PR.

This is done in the first commit, and followed by the pending metric fix.

(I mark updateAllocatedAndPendingResourceMetrics() as DEPRECATED for now. LMK if you want to remove it entirely)

PTAL @wilfred-s @yzhangal @manirajv06

wilfred-s (approved):
LGTM

PR #687, opened 2023-10-29
[YUNIKORN-2042] REST API for specific queue

steinsgateted:
### What is this PR for?
Expose a REST API for specific queue:
`/ws/v1/partition/%s/queue/%s/
`
Other modifications:
1
Modify the test items, which were not actually tested originally.
`pkg/scheduler/objects/queue_test.go` `TestGetPartitionQueueDAOInfo` function

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-2042

example:http://localhost:9080/ws/v1/partition/default/queue/root.namespaces.level1.level2

### Questions:


pkg/webservice/handlers.go
```diff
@@ -519,6 +519,31 @@ func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
...
+func getPartitionQueue(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	vars := httprouter.ParamsFromContext(r.Context())
+	if vars == nil {
+		buildJSONErrorResponse(w, MissingParamsName, http.StatusBadRequest)
+		return
+	}
+	partition := vars.ByName("partition")
+	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
+	if partitionContext != nil {
```
wilfred-s, line 531: Turn the check around: 	if partitionContext == nil return error

pkg/webservice/handlers.go
```diff
@@ -519,6 +519,31 @@ func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
...
+		queueName := vars.ByName("queue")
```
wilfred-s, line 532: Leverage `validateQueue(queuePath)` to do a syntax check before calling the queue retrieval

pkg/webservice/handlers.go
```diff
@@ -519,6 +519,31 @@ func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
...
+		queue := partitionContext.GetQueue(queueName)
+		if queue == nil {
+			buildJSONErrorResponse(w, QueueDoesNotExists, http.StatusBadRequest)
+			return
+		}
+		queueDao := getPartitionQueueDAO(partitionContext, queue)
```
wilfred-s, line 538: If we have the queue we can directly call the queue without using the partition.

wilfred-s (changes requested):
I was expecting an API that would return a specific queue with its details. In this case we return a subtree rooted at a certain point. I would have expected just the one queue to be returned as a lightweight object without the whole subtree attached. There is no difference for a leaf queue but further up the hierarchy there is.

pkg/scheduler/objects/queue.go
```diff
@@ -669,6 +669,52 @@ func (sq *Queue) GetPartitionQueueDAOInfo() dao.PartitionQueueDAOInfo {
...
 
+// GetSingleQueueDAOInfo returns the single queue as an object for a REST call.
+func (sq *Queue) GetSingleQueueDAOInfo() dao.PartitionQueueDAOInfo {
+	queueInfo := dao.PartitionQueueDAOInfo{}
+	childes := sq.GetCopyOfChildren()
+	queueInfo.ChildrenNames = make([]string, 0, len(childes))
+	for _, child := range childes {
+		queueInfo.ChildrenNames = append(queueInfo.ChildrenNames, child.Name)
+	}
+	// we have held the read lock so following method should not take lock again.
```
chia7712, line 680: those code is almost equal to `GetPartitionQueueDAOInfo`. Could you do a bit refactor for them?

pkg/scheduler/objects/queue.go
```diff
@@ -669,6 +669,52 @@ func (sq *Queue) GetPartitionQueueDAOInfo() dao.PartitionQueueDAOInfo {
 	return queueInfo
 }
 
+// GetSingleQueueDAOInfo returns the single queue as an object for a REST call.
+func (sq *Queue) GetSingleQueueDAOInfo() dao.PartitionQueueDAOInfo {
+	queueInfo := dao.PartitionQueueDAOInfo{}
+	childes := sq.GetCopyOfChildren()
+	queueInfo.ChildrenNames = make([]string, 0, len(childes))
```
chia7712, line 676: maybe `GetPartitionQueueDAOInfo` could fill up this field also?

pkg/webservice/routes.go
```diff
@@ -98,6 +98,12 @@ var webRoutes = routes{
 		"/ws/v1/partition/:partition/queues",
 		getPartitionQueues,
 	},
+	route{
+		"Scheduler",
+		"GET",
+		"/ws/v1/partition/:partition/queue/:queue/single/:single",
```
chia7712, line 104: It seems to me the behavior of "single" is to include/exclude some fields, right? if so, could we replace the "single" by "exclude/include" query? for example, user can receive thin response by sending `/ws/v1/partition/:partition/queue/:queue?exclude=children`. Also, the "exclude/include" can be implemented to other APIs to have consistent behavior and flexible response.

chia7712 (commented):
@steinsgateted nice patch. there are some minor comments, and please take a look. thanks!

steinsgateted:
@wilfred-s @chia7712
Sorry for the late reply, thanks for the review.
A few changes:

- In `pkg/scheduler/objects/queue.go`, change `GetPartitionQueueDAOInfo()` to `GetPartitionQueueDAOInfo(exclude bool)`

    exclude is true, which means there is no children Queue Object, and only the children's queue path will be recorded.

- In `pkg/webservice/handlers_test.go`, `assertAppStateAllow` function rename to `assertAppStateNotAllow`

- In `pkg/webservice/handlers_test.go`, `assertActiveStateAllow` function rename to `assertActiveStateNotAllow`

- In `pkg/webservice/handlers_test.go`, add new function `assertQueueInvalid`

Also, I would like some clarifications.
Do we recommend renaming the QueueName of `PartitionQueueDAOInfo` to QueuePath?
Because it is actually QueuePath
`queueInfo.QueueName = sq.QueuePath`

pkg/scheduler/objects/queue.go
```diff
@@ -624,12 +624,19 @@ func (sq *Queue) CheckAdminAccess(user security.UserGroup) bool {
...
+	children := sq.GetCopyOfChildren()
+	queueInfo.Children = make([]dao.PartitionQueueDAOInfo, 0, len(children))
+	if exclude {
+		for _, child := range children {
+			queueInfo.ChildrenNames = append(queueInfo.ChildrenNames, child.QueuePath)
+		}
+	} else {
+		for _, child := range children {
+			queueInfo.Children = append(queueInfo.Children, child.GetPartitionQueueDAOInfo(false))
+		}
```
wilfred-s, line 639: refactor: only need to get a copy of children if we need to recurse into them:
```
if !exclude {
	children := sq.GetCopyOfChildren()
	queueInfo.Children = make([]dao.PartitionQueueDAOInfo, 0, len(children))
	for _, child := range children {
		queueInfo.Children = append(queueInfo.Children, child.GetPartitionQueueDAOInfo(false))
	}
}
```
Leave adding the child path names to the locked code below

wilfred-s: Add this:
```
for _, child := range children {
	queueInfo.ChildrenNames = append(queueInfo.ChildrenNames, child.QueuePath)
}
```
around properties for loop line 660

pkg/webservice/handlers.go
```diff
@@ -642,6 +642,53 @@ func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
...
+	}
+	// include children
+	if include := strings.ToLower(r.URL.Query().Get("include")); include != "" {
+		if include == "children" {
+			queueDao = queue.GetPartitionQueueDAOInfo(false)
+		} else {
+			buildJSONErrorResponse(w, "Only following include is allowed: children", http.StatusBadRequest)
+			return
+		}
+	}
```
wilfred-s, line 686: A simple `subtree` query parameter is a much more elegant solution.
If the parameter is there we call `GetPartitionQueueDAOInfo(true)` otherwise `GetPartitionQueueDAOInfo(false)` The default, and not present, is `false` so code wise it would look a bit like this:
```
queueDao := queue.GetPartitionQueueDAOInfo(r.URL.Query().Has("subtree"))
```
It checks for the presence of the key `subtree`

wilfred-s (changes requested):
almost there, some small simplifications and cleanup requested

pkg/webservice/routes.go
wilfred-s: `subtree` is a clearer option than exclude or include. I cannot tell what to exclude or include. Could point to applications or maybe resources. The `subtree` refers back to the same object and we get the subtree rooted at the requested point.

steinsgateted:
@wilfred-s Thank you for your review. I modified.

PR #688, opened 2023-10-29
[YUNIKORN-2073] add rate limiter to event system

FrankYang0529:
### What is this PR for?

The `sendAppDoesNotFitEvent()` causes a spew in the event system. We need to limit the number of those events with some throttle.

The event will be send every time it does not fit in the queue headroom. That could occur really often if it is a high priority request on an application that does not fit but other requests get scheduled (eager scheduling).

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2073

Covered by unit tests.

### DIFF
--- a/pkg/scheduler/objects/application_events.go
+++ b/pkg/scheduler/objects/application_events.go
@@ -20,6 +20,9 @@ package objects
 
 import (
 	"fmt"
+	"time"
+
+	"golang.org/x/time/rate"
 
 	"github.com/apache/yunikorn-core/pkg/common"
 	"github.com/apache/yunikorn-core/pkg/events"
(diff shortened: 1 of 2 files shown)


FrankYang0529:
Hi @wilfred-s, I introduced rate limiter for the event system, not only for `sendAppDoesNotFitEvent`. Not sure whether it's good to do that? If yes, do you think it's good to add config for rate limiter in the event system? Like enable/disable, maximum keys, etc. Thank you.

pbacsko:
@FrankYang0529 this solutions seems to be unnecessarily complicated. 
We need to timit `applicationEvents.sendAppDoesNotFitEvent()` and nothing else.

Can't we just use `golang.org/x/time/rate` like in YUNIKORN-1985?

pbacsko (changes requested):
Please check how the current approach can be simplified. It's unnecessarily complex.

FrankYang0529:
> Please check how the current approach can be simplified. It's unnecessarily complex.

Hi @pbacsko, I updated it. May you review it again? Thank you.

pbacsko:
Way better! I like this.
+1

cc @wilfred-s for second opinion.

pbacsko (approved):
Way better! I like this.
+1

cc @wilfred-s for second opinion.

wilfred-s (approved):
LGTM, nice and simple

PR #690, opened 2023-11-02
[YUNIKORN-2107] Allow preemption to be disabled globally

craigcondit:
### What is this PR for?
Added configuration `service.disablePreemption` which can be used to disable the preemption subsystem entirely when set to `true`.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2107

Added unit tests.

### Questions:

### DIFF
--- a/pkg/common/configs/config.go
+++ b/pkg/common/configs/config.go
@@ -50,14 +50,14 @@ type PartitionConfig struct {
 	Queues            []QueueConfig
 	PlacementRules    []PlacementRule           `yaml:",omitempty" json:",omitempty"`
 	Limits            []Limit                   `yaml:",omitempty" json:",omitempty"`
-	Preemption        PartitionPreemptionConfig `yaml:",omitempty" json:",omitempty"` // deprecated
+	Preemption        PartitionPreemptionConfig `yaml:",omitempty" json:",omitempty"`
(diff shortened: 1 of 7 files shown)


wilfred-s:
Why not use the existing parameter in the partition config for this? (See jira for details also)

craigcondit:
> Why not use the existing parameter in the partition config for this? (See jira for details also)

Because then we end up with opt-in preemption rather than opt out as there's no distinction in the existing configuration between Enabled: false and not specified. That's backwards-incompatible.

craigcondit:
> > Why not use the existing parameter in the partition config for this? (See jira for details also)
> 
> Because then we end up with opt-in preemption rather than opt out as there's no distinction in the existing configuration between Enabled: false and not specified. That's backwards-incompatible.

Managed to get this working by changing the type of the Enabled field in the config to *bool, so we can detect unset values and default to true.

manirajv06 (approved):

pbacsko (approved):
+1

PR #693, opened 2023-11-03
[YUNIKORN-2120] Fix incorrect error message when check maxApplications in configValidator

brandboat:
### What is this PR for?
The error message use terms maxRunningApps when setup incorrect maxapplications in ConfigMap. Would be more explicit if we replace `maxRunningApps` with `maxApplications`.
```
error: configmaps "yunikorn-configs" could not be patched: admission webhook "admission-webhook.yunikorn.validate-conf" denied the request: parent maxRunningApps must be larger than child maxRunningApps
```

### What type of PR is it?
* [x] - Improvement

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2120

N/A

N/A

### Questions:
N/A


FrankYang0529 (approved):
+1. Thanks!

PR #696, opened 2023-11-06
[YUNIKORN-2124] Fix incorrect test in queue_test.go#TestNewConfiguredQueue

brandboat:
### What is this PR for?
in unit test queue_test.go#TestNewConfiguredQueue

`// case 0: leaf can use template`
https://github.com/apache/yunikorn-core/blob/284cc92974a7330c95ad3c6ecdf05ea68b9eee9d/pkg/scheduler/objects/queue_test.go#L2225-L2226

This test is flawed as it erroneously allows managed leaf queues to utilize child template and it consistently passes because the `getResourceConf` and `getProperties` lack the required randomness.

### What type of PR is it?
* [x] - Bug Fix

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2124

covered by unit test

N/A

### Questions:
N/A

### DIFF
--- a/pkg/scheduler/objects/queue_test.go
+++ b/pkg/scheduler/objects/queue_test.go
@@ -2222,23 +2222,18 @@ func TestNewConfiguredQueue(t *testing.T) {
 	assert.Assert(t, resources.Equals(resourceStruct, parent.template.GetMaxResource()))
 	assert.Assert(t, resources.Equals(resourceStruct, parent.template.GetGuaranteedResource()))
 
-	// case 0: leaf can use template
+	// case 0: managed leaf queue can't use template
 	leafConfig := configs.QueueConfig{
-		Name:       "leaf_queue",
-		Parent:     false,
-		Properties: getProperties(),
-		Resources: configs.Resources{
-			Max:        getResourceConf(),
-			Guaranteed: getResourceConf(),
-		},
+		Name:   "leaf_queue",
+		Parent: false,
 	}
 	childLeaf, err := NewConfiguredQueue(leafConfig, parent)
 	assert.NilError(t, err, "failed to create queue: %v", err)
 	assert.Equal(t, childLeaf.QueuePath, "parent_queue.leaf_queue")
 	assert.Assert(t, childLeaf.template == nil)
-	assert.DeepEqual(t, childLeaf.properties, parent.template.GetProperties())
-	assert.Assert(t, resources.Equals(childLeaf.maxResource, parent.template.GetMaxResource()))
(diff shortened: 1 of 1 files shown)


wilfred-s:
@brandboat Please file a follow up jira to generate more random data in `getResourceConf()`, something more in the direction of `rand.Intn(10000)` instead of the seconds based setup.

brandboat:
> @brandboat Please file a follow up jira to generate more random data in getResourceConf(), something more in the direction of rand.Intn(10000) instead of the seconds based setup.

Roger that ! Already filed follow up jira https://issues.apache.org/jira/browse/YUNIKORN-2168

pkg/scheduler/objects/queue_test.go
```diff
@@ -2214,23 +2214,18 @@ func TestNewConfiguredQueue(t *testing.T) {
...
 	childLeaf, err := NewConfiguredQueue(leafConfig, parent)
 	assert.NilError(t, err, "failed to create queue: %v", err)
 	assert.Equal(t, childLeaf.QueuePath, "parent_queue.leaf_queue")
 	assert.Assert(t, childLeaf.template == nil)
-	assert.DeepEqual(t, childLeaf.properties, parent.template.GetProperties())
-	assert.Assert(t, resources.Equals(childLeaf.maxResource, parent.template.GetMaxResource()))
-	assert.Assert(t, resources.Equals(childLeaf.guaranteedResource, parent.template.GetGuaranteedResource()))
+	assert.Assert(t, !reflect.DeepEqual(childLeaf.properties, parent.template.GetProperties()))
+	assert.Assert(t, !resources.Equals(childLeaf.maxResource, parent.template.GetMaxResource()))
+	assert.Assert(t, !resources.Equals(childLeaf.guaranteedResource, parent.template.GetGuaranteedResource()))
```
wilfred-s, line 2228: Since you have removed setting of the properties and the resources the check is now irrelevant and the two test cases really could use the same checks (compare to nil)
You need to leave the child as is, resources defined, and make sure there is no override.

compare `resources.NewResourceFromConf(leafConfig.Resources.Max)` to `childLeaf.maxResource` etc

wilfred-s (changes requested):

pkg/scheduler/objects/queue_test.go
brandboat: Already addressed comments in the latest commit, many thanks !

pbacsko (approved):
+1

PR #697, opened 2023-11-06
[YUNIKORN-2033] Add MaxApplications in Template

brandboat:
### What is this PR for?
This is a follow up in YUNIKORN-193.
Currently we leave MaxApplications in ChildTemplate while Template didn't use it.

### What type of PR is it?
* [x] - Improvement

### Todos
* [ ] - Add e2e tests for child template max applications in k8shim (will fire another jira after this pr merged)

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2033

covered by unit tests

N/A

### Questions:
N/A

### DIFF
--- a/pkg/scheduler/objects/template/template.go
+++ b/pkg/scheduler/objects/template/template.go
@@ -25,6 +25,7 @@ import (
 )
 
 type Template struct {
+	maxApplications    uint64
 	properties         map[string]string
 	maxResource        *resources.Resource
 	guaranteedResource *resources.Resource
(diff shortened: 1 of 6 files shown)


craigcondit (approved):
+1, @wilfred-s can you take a look as well?

pkg/scheduler/objects/template/template.go
```diff
@@ -44,7 +45,7 @@ func FromConf(template *configs.ChildTemplate) (*Template, error) {
 		return true
 	}
 
-	if template == nil || (isMapEmpty(template.Properties) && isMapEmpty(template.Resources.Guaranteed) && isMapEmpty(template.Resources.Max)) {
+	if template == nil || (template.MaxApplications == 0 && isMapEmpty(template.Properties) && isMapEmpty(template.Resources.Guaranteed) && isMapEmpty(template.Resources.Max)) {
```
pbacsko, line 48: Could you refactor this part and `isMapEmpty` to a separate method for readability?
```
if template == nil || template.IsEmpty() {
  return nil, nil
}

...

func (t *Template) IsEmpty() bool {
...
}

func isMapEmpty(m map[string]string) bool {
...
}
```

pbacsko (changes requested):
Nit for refactor

pkg/scheduler/objects/template/template.go
brandboat: Already address comments in the latest comment, many thanks !

pbacsko (approved):
+1

PR #698, opened 2023-11-06
[YUNIKORN-2123] Modify json.Unmarshal error message

makinyemi:
### What is this PR for?
Modifying the json.Unmarshal error message to better explain the failure of the specific function.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-2123

n/a
### Questions:
n/a

### DIFF
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -49,6 +49,8 @@ import (
 )
 
 const unmarshalError = "Failed to unmarshal error response from response body"
+const statusCodeError = "Incorrect Status code"
+const jsonMessageError = "JSON error message is incorrect"
 
 const partitionNameWithoutClusterID = "default"
 const normalizedPartitionName = "[rm-123]default"
@@ -323,7 +325,7 @@ func TestValidateConf(t *testing.T) {
 		validateConf(resp, req)
 		var vcr dao.ValidateConfResponse
 		err := json.Unmarshal(resp.outputBytes, &vcr)
-		assert.NilError(t, err, "%s (ValidateConfResponse)", unmarshalError)
+		assert.NilError(t, err, unmarshalError)
 		assert.Equal(t, vcr.Allowed, test.expectedResponse.Allowed, "allowed flag incorrect")
 		assert.Equal(t, vcr.Reason, test.expectedResponse.Reason, "response text not as expected")
 	}
@@ -364,7 +366,7 @@ func TestUserGroupLimits(t *testing.T) {
 		validateConf(resp, req)
 		var vcr dao.ValidateConfResponse
 		err := json.Unmarshal(resp.outputBytes, &vcr)
-		assert.NilError(t, err, "%s (ValidateConfResponse)", unmarshalError)
+		assert.NilError(t, err, unmarshalError)
 		assert.Equal(t, vcr.Allowed, test.expectedResponse.Allowed, "allowed flag incorrect")
 		assert.Equal(t, vcr.Reason, test.expectedResponse.Reason, "response text not as expected")
 	}
@@ -382,9 +384,9 @@ func TestApplicationHistory(t *testing.T) {
 
 	var errInfo dao.YAPIError
 	err := json.Unmarshal(resp.outputBytes, &errInfo)
-	assert.NilError(t, err, "%s (App History): %s", unmarshalError, string(resp.outputBytes))
+	assert.NilError(t, err, unmarshalError)
 	assert.Equal(t, http.StatusNotImplemented, resp.statusCode, "app history handler returned wrong status")
-	assert.Equal(t, errInfo.Message, "Internal metrics collection is not enabled.", "JSON error message is incorrect")
+	assert.Equal(t, errInfo.Message, "Internal metrics collection is not enabled.", jsonMessageError)
 	assert.Equal(t, errInfo.StatusCode, http.StatusNotImplemented)
 
 	// init should return null and thus no records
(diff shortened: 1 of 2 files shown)


steinsgateted:
@makinyemi Thank you for the PR.
I found that there are a lot of "failed to unmarshal..."
Maybe we could make unified changes.

makinyemi:
> @makinyemi Thank you for the PR. I found that there are a lot of "failed to unmarshal..." Maybe we could make unified changes.

@steinsgateted No problem! Ok, were you thinking unified in the sense of adding some sort of reusable function or constant string that we could format as needed to substitute for all of the occurrences of "failed to unmarshal..."?

manirajv06:
> > I found that there are a lot of "failed to unmarshal..." Maybe we could make unified changes.
> 
> @steinsgateted No problem! Ok, were you thinking unified in the sense of adding some sort of reusable function or constant string that we could format as needed to substitute for all of the occurrences of "failed to unmarshal..."?

Introduce a const something like this.
STRING1 = "failed to unmarshal DAONAME response from response body"
replace DAONAME with actual name.

Also you can introduce constants for 

"Incorrect Status code"
"JSON error message is incorrect"

craigcondit (changes requested):
-1. Every one of of the new messages is incorrect. It's the YAPIError structure that we're attempting to parse (Contains StatusCode, Message, Description). All of these functions are used to test error handling, and we assert in each of them that the *error object* is parsed correctly. Just change every message to "failed to unmarshal error response from response body".

craigcondit (changes requested):
Now you've got additional changes that are *not* error handling. You need to pay attention to the context for each one.

makinyemi:
> Now you've got additional changes that are _not_ error handling. You need to pay attention to the context for each one.

Thanks for the feedback. Will reset and go back through

craigcondit (approved):
+1, pending precommit checks.

steinsgateted:
@makinyemi @manirajv06 @craigcondit 
Thank you for your discussion.
nit: I think other "failed to unmarshal" information can also be changed.

makinyemi:
> nit: I think other "failed to unmarshal" information can also be changed.

Similar to my last commit? Make a constant for specific test functions?

pkg/webservice/handlers_test.go
```diff
@@ -48,6 +48,8 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
+const UnmarshalError = "Failed to unmarshal error response from response body"
```
steinsgateted, line 51: UnmarshalError -> unmarshalError

pkg/webservice/handlers_test.go
```diff
@@ -321,7 +323,7 @@ func TestValidateConf(t *testing.T) {
 		validateConf(resp, req)
 		var vcr dao.ValidateConfResponse
 		err := json.Unmarshal(resp.outputBytes, &vcr)
-		assert.NilError(t, err, "failed to unmarshal ValidateConfResponse from response body")
+		assert.NilError(t, err, "%s (ValidateConfResponse)", UnmarshalError)
```
steinsgateted, line 326: dao can be removed
`assert.NilError(t, err, unmarshalError)`

steinsgateted (commented):
> Just change every message to "failed to unmarshal error response from response body".

> Similar to my last commit? Make a constant for specific test functions?

@makinyemi @craigcondit 
Thank you.
I think it's like @craigcondit said.
Replace all messages with unmarshalError.
No need to retain the original DAONAME part.

makinyemi:
> Replace all messages with unmarshalError. No need to retain the original DAONAME part.

@steinsgateted I did something like that in a previous commit that i reverted where @craigcondit mentioned in the above comment

> Now you've got additional changes that are _not_ error handling. You need to pay attention to the context for each one.

If i understand correctly, as @craigcondit explained it certain test were testing for specific failure and the messages reflected the error they were handling.

If i am mistaken i can definitely make the corrections

makinyemi:
@steinsgateted @craigcondit Actually, I remember now. I added DAOInfo to every message, which was wrong. 
So if i understand correctly. We are good to remove all DAOInfo?

steinsgateted:
@makinyemi Thank you
I think we can use two variables to deal with different situations.
Create a new variable:
`const unmarshalDAOError = "Failed to unmarshal dao response from response body"`
In test error handling situation:
`assert.NilError(t, err, unmarshalError)`
in dao:
`assert.NilError(t, err, "%s (ValidateConfResponse)", unmarshalDAOError)`

> Also you can introduce constants for
> 
> "Incorrect Status code"
> "JSON error message is incorrect"

we could add two new consts:
```
const statusCodeError = "Incorrect Status code"
const JsonMessageError = "JSON error message is incorrect"
```

steinsgateted:
@makinyemi @manirajv06 @craigcondit Thank you.
I would like some clarifications.
I think there are three expressions of unmarshal dao messages.
Which expression method do you think makes more sense?

- Using the original method is like:

`assert.NilError(t, err, "failed to unmarshal ValidateConfResponse from response body")`

- `assert.NilError(t, err, "%s (ValidateConfResponse)", unmarshalDAOError)`

- `assert.NilError(t, err, unmarshalDAOError)`

craigcondit:
> @makinyemi @manirajv06 @craigcondit Thank you. I would like some clarifications. I think there are three expressions of unmarshal dao messages. Which expression method do you think makes more sense?
> 
> * Using the original method is like:
> 
> `assert.NilError(t, err, "failed to unmarshal ValidateConfResponse from response body")`
> 
> * `assert.NilError(t, err, "%s (ValidateConfResponse)", unmarshalDAOError)`
> * `assert.NilError(t, err, unmarshalDAOError)`

What I think is that *far* too much time and effort has been spent on this. We're talking about the wording of a diagnostic message echoed by test cases when they fail (which they basically never do). Pick something reasonable and go with it.

steinsgateted:
@makinyemi @craigcondit Thank you. +1

steinsgateted (approved):

pbacsko (approved):
+1

PR #700, opened 2023-11-07
[YUNIKORN-2085] CurrentPriority field missing from Queues API respons…

Haser0305:
### What is this PR for?
The `CurrentPriority` field is currently tagged with "omitempty", which causes it to be omitted from json response when the value is 0. 

Omitting the field in this case could mislead consumers or cause potential errors when consuming the API, such as throwing exceptions upon finding the expected field missing from the response.
Removing omitempty ensures the field will always be included with its actual value, even if 0. This avoids situations where the field might be missing.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-2085


craigcondit (approved):
+1 LGTM.

PR #701, opened 2023-11-07
[YUNIKORN-2127] fix stateAwareFilter

FrankYang0529:
### What is this PR for?

In stateAwareFilter description, it only allows one (1) application with a state that is not running in the list of candidates. The non-running state can be Starting or Accepted. However, in the function, it can allow more than one starting application. We should either update the description or fix the function.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2127

Covered by unit tests.

### DIFF
--- a/pkg/scheduler/objects/sorters_test.go
+++ b/pkg/scheduler/objects/sorters_test.go
@@ -385,6 +385,18 @@ func TestSortAppsStateAware(t *testing.T) {
 	assert.NilError(t, err, "state change failed for app-3")
 	list = sortApplications(input, policies.StateAwarePolicy, false, nil)
 	assertAppListLength(t, list, []string{appID0, appID1, appID3}, "state not app-2")
+
+	// move 3rd to starting: 1st, 3rd and 4th in that order
+	err = input[appID2].HandleApplicationEvent(RunApplication)
+	assert.NilError(t, err, "state change failed for app-2")
+	list = sortApplications(input, policies.StateAwarePolicy, false, nil)
+	assertAppListLength(t, list, []string{appID0, appID2, appID3}, "state not app-1")
+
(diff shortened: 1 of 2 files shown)


craigcondit:
My understanding is if anything, the description is incorrect, not the implementation. @wilfred-s ?

pkg/scheduler/objects/sorters_test.go
```diff
@@ -385,6 +385,18 @@ func TestSortAppsStateAware(t *testing.T) {
...
+	err = input[appID2].HandleApplicationEvent(RunApplication)
+	assert.NilError(t, err, "state change failed for app-2")
+	list = sortApplications(input, policies.StateAwarePolicy, false, nil)
+	assertAppListLength(t, list, []string{appID0, appID2, appID3}, "state not app-1")
+
+	// move 2nd to starting: 1st, 2nd and 4th in that order
+	err = input[appID1].HandleApplicationEvent(RunApplication)
+	assert.NilError(t, err, "state change failed for app-1")
+	list = sortApplications(input, policies.StateAwarePolicy, false, nil)
+	assertAppListLength(t, list, []string{appID0, appID1, appID3}, "state not app-2")
```
wilfred-s, line 399: This is an incorrect test case. You have two applications in a Starting state. In the Stateaware setup that cannot and must not happen as it breaks the guarantee that is given.

Both "app-1" and "app-2" have been progressed to Starting. You can only do this by manipulating the application states outside of the normal flow.
Since you need to move from Accepted to Starting it will not happen.

wilfred-s (changes requested):
This cannot happen in a normal flow.

FrankYang0529:
@wilfred-s, thanks for the information. Close the PR because the function works fine.

PR #706, opened 2023-11-09
[YUNIKORN-2134] Use nil resource instead of NewResource()

manirajv06:
### What is this PR for?
Initialize resourceUsage and maxResources variables to nil and does empty check by comparing it to nil to avoid expensive checks against resources.NewResource()

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2134

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/group_tracker_test.go
+++ b/pkg/scheduler/ugm/group_tracker_test.go
@@ -86,6 +86,7 @@ func TestGTDecreaseTrackedResource(t *testing.T) {
 	// Queue setup:
 	// root->parent->child1
 	// root->parent->child2
+	// Initialize ugm
 	GetUserManager()
 	user := &security.UserGroup{User: "test", Groups: []string{"test"}}
 	groupTracker := newGroupTracker(user.User)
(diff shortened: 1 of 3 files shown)


pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -89,13 +89,16 @@ func (qt *QueueTracker) increaseTrackedResource(hierarchy []string, applicationI
 		}
 	}
 
+	if resources.IsZero(qt.resourceUsage) {
```
FrankYang0529, line 92: We can just use `qt.resourceUsage == nil`. If `qt.resourceUsage` is empty or all resources is 0, it will pass `resources.IsZero` and we will reassign a similar value to it.

https://github.com/apache/yunikorn-core/blob/0fd743614b973650fc3fa24488be240e2e0b3121/pkg/common/resources/resources.go#L930-L942

FrankYang0529 (approved):
+1. Leave a minor suggestion. Thank you.

pkg/scheduler/ugm/group_tracker_test.go
```diff
@@ -86,6 +86,7 @@ func TestGTDecreaseTrackedResource(t *testing.T) {
 	// Queue setup:
 	// root->parent->child1
 	// root->parent->child2
+	GetUserManager()
```
pbacsko, line 89: What's the purpose of this? Initialization? Pls add a short comment.

pbacsko (approved):
+1 LGTM I just have a minor question.

PR #709, opened 2023-11-10
[YUNIKORN-2139] Move queuepath split from user and group trackers to …

manirajv06:
…manager.go

### What is this PR for?
Instead of doing strings.split on queue path all over the places, keep it in one place (manager.go) for ease of maintanence. Also some code clean up.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2139

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/group_tracker.go
+++ b/pkg/scheduler/ugm/group_tracker.go
@@ -19,11 +19,9 @@
 package ugm
 
 import (
-	"strings"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


wilfred-s (approved):
LGTM
Fixing two new `zap.Any` calls on commit to use `zap.Strings`.
Filing follow up jira to remove all older `zap.Any` calls for resources.

PR #710, opened 2023-11-11
YUNIKORN-2141 Should not preempt placeholders which has been released

zhuqi-lucas:
### What is this PR for?
We should not preempt placeholders which has been released, it will cause the real pod pending due to not get the response:
```
The real pod will always pending, because the core side doesn't receive the release response for PLACEHOLDER_REPLACED
// if we have an uuid the termination type is important
if release.TerminationType == si.TerminationType_PLACEHOLDER_REPLACED {
    log.Logger().Info("replacing placeholder allocation",
       zap.String("appID", appID),
       zap.String("allocationId", uuid))
    if alloc := app.ReplaceAllocation(uuid); alloc != nil {
       released = append(released, alloc)
    }
} else {
    log.Logger().Info("removing allocation from application",
       zap.String("appID", appID),
       zap.String("allocationId", uuid),
       zap.String("terminationType", release.TerminationType.String()))
    if alloc := app.RemoveAllocation(uuid); alloc != nil {
       released = append(released, alloc)
    }
} 
```

Details are in the jira description:
https://issues.apache.org/jira/browse/YUNIKORN-2141

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-2141

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -1752,6 +1752,11 @@ func (sq *Queue) findEligiblePreemptionVictims(results map[string]*QueuePreempti
 					continue
 				}
 
+				// skip placeholder tasks which are marked released
+				if alloc.IsReleased() {
+					continue
+				}
+
 				// if we have encountered a fence then all tasks are eligible for preemption
 				// otherwise the task is a candidate if its priority is less than or equal to the ask priority
 				if fenced || int64(alloc.GetPriority()) <= askPriority {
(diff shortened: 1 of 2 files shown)


pbacsko (approved):
+1 LGTM 
I'll let Wilfred/Craig take a look to be sure.

wilfred-s (approved):
LGTM

PR #711, opened 2023-11-13
[YUNIKORN-2146] skip preempted placeholder in allocate

wilfred-s:
### What is this PR for?
Skip preempted placeholders in tryPlaceholderAllocate() as the replacement will most likely never happen on time. Not skipping could also cause other allocations to keep waiting for the confirmation that preemption has worked.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2146

unit test included


zhuqi-lucas (approved):
LGTM Agree, we need to skip preempted placeholder in allocate, thanks!

PR #712, opened 2023-11-13
[YUNIKORN-2136] change IsZero to StrictlyGreaterThanZero

FrankYang0529:
### What is this PR for?

Limit max resource should be validated using resources.StrictlyGreaterThanZero() check to ensure that we are not allowing resources with zero values.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2136

Covered by unit test.

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -539,8 +539,8 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
 				zap.Error(err))
 			return err
 		}
-		if resources.IsZero(limitResource) {
-			return fmt.Errorf("MaxResources is zero in '%s' limit, all resource types are zero", limit.Limit)
+		if !resources.StrictlyGreaterThanZero(limitResource) {
+			return fmt.Errorf("MaxResources should be greater than zero in '%s' limit", limit.Limit)
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -539,8 +539,8 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
 				zap.Error(err))
 			return err
 		}
-		if resources.IsZero(limitResource) {
-			return fmt.Errorf("MaxResources is zero in '%s' limit, all resource types are zero", limit.Limit)
+		if !resources.StrictlyGreaterThanZero(limitResource) {
+			return fmt.Errorf("MaxResources is not greater than zero in '%s' limit", limit.Limit)
```
manirajv06, line 543: MaxResources is not greater than zero in '%s' limit -> MaxResources should be greater than zero in '%s' limit ?

manirajv06 (changes requested):

pkg/common/configs/configvalidator.go
FrankYang0529: Updated it. Thank you.

manirajv06 (approved):

PR #720, opened 2023-11-20
[YUNIKORN-2119] check wildcard limit in different queue level

FrankYang0529:
### What is this PR for?

If a non-wildcard limit is not defined in parent queue, but it's defined in child queue, we need to check whether it's large than wildcard limit in parent queue.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2119

Covered by unit test.

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -172,6 +172,11 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 				}
 				// Override with min resource
 				users[curQueuePath][user] = resources.ComponentWiseMinPermissive(limitMaxResources, userMaxResource)
+			} else if wildcardMaxResource, ok := users[queuePath][common.Wildcard]; user != common.Wildcard && ok {
+				if !wildcardMaxResource.FitInMaxUndef(limitMaxResources) {
+					return fmt.Errorf("user %s max resource %s of queue %s is greater than wildcard maximum resource %s of immediate or ancestor parent queue", user, limitMaxResources.String(), cur.Name, wildcardMaxResource.String())
+				}
+				users[curQueuePath][user] = limitMaxResources
 			} else {
 				users[curQueuePath][user] = limitMaxResources
 			}
@@ -184,6 +189,11 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 				}
 				// Override with min resource
 				groups[curQueuePath][group] = resources.ComponentWiseMinPermissive(limitMaxResources, groupMaxResource)
+			} else if wildcardMaxResource, ok := groups[queuePath][common.Wildcard]; group != common.Wildcard && ok {
+				if !wildcardMaxResource.FitInMaxUndef(limitMaxResources) {
+					return fmt.Errorf("group %s max resource %s of queue %s is greater than wildcard maximum resource %s of immediate or ancestor parent queue", group, limitMaxResources.String(), cur.Name, wildcardMaxResource.String())
+				}
+				groups[curQueuePath][group] = limitMaxResources
 			} else {
 				groups[curQueuePath][group] = limitMaxResources
 			}
--- a/pkg/common/configs/configvalidator_test.go
+++ b/pkg/common/configs/configvalidator_test.go
@@ -740,32 +740,34 @@ func TestServiceAccountGroupName(t *testing.T) {
 
 func TestCheckLimitResource(t *testing.T) { //nolint:funlen
 	testCases := []struct {
-		name     string
-		config   QueueConfig
-		hasError bool
+		name   string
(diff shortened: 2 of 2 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -172,6 +172,11 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 				}
 				// Override with min resource
 				users[curQueuePath][user] = resources.ComponentWiseMinPermissive(limitMaxResources, userMaxResource)
+			} else if wildcardMaxResource, ok := users[queuePath][common.Wildcard]; user != common.Wildcard && ok {
+				if !wildcardMaxResource.FitInMaxUndef(limitMaxResources) {
+					return fmt.Errorf("user %s max resource %s of queue %s is greater than wildcard maximum resource %s of immediate or ancestor parent queue", user, limitMaxResources.String(), cur.Name, wildcardMaxResource.String())
+				}
+				users[curQueuePath][user] = limitMaxResources
```
manirajv06, line 179: I think we would need to call `ComponentWiseMinPermissive` similar to earlier block before writing it into map. Otherwise, following case does not throw error. 

root: * => mem: 100
root.parent: user1=> mem: 50
root.parent.leaf: user2=> mem: 150

manirajv06 (changes requested):

pkg/common/configs/configvalidator.go
FrankYang0529: The case is covered by test cases `leaf queue user maxresources exceed grandparent queue wildcard maxresources` and `leaf queue group maxresources exceed grandparent queue wildcard maxresources` in this PR. Yeah, we throw error for this config, because `user2` in `root.parent.leaf` is larger than `*` in `root`.

manirajv06: Ok. I tried to recollect but not able to capture completely. May be because of error message mismatch in the above case, I would have that it has failed. Please try the below test code and you can see error message mismatch.

{
			name: "leaf queue user maxresources exceed grandparent queue wildcard maxresources",
			config: QueueConfig{
				Name: "parent",
				Limits: createLimitMaxResources(
					map[string]map[string]string{
						"*": {"memory": "50"},
					},
					nil),
				Queues: []QueueConfig{
					{
						Name: "child1",
						Limits: createLimitMaxResources(
							map[string]map[string]string{
								"user1": {"memory": "25"},
							},
							nil),
						Queues: []QueueConfig{
							{
								Name: "child2",
								Limits: createLimitMaxResources(
									map[string]map[string]string{
										"user2": {"memory": "75"},
									},
									nil),
							},
						},
					},
				},
			},
			hasError: true,
		},

Also, Are doing hierarchical comparison checks for wildcard user and group similar to named users and groups? I doubt this.

manirajv06:
@FrankYang0529 Any updates?

pkg/common/configs/configvalidator.go
FrankYang0529: Hi @manirajv06, thanks for the review. My bad, I changed `hasError` check to `errMsg` and forgot to remove the `hasError` attribute. That is why your test case is failed. May you change `hasError: true` to `errMsg: "is greater than wildcard maximum resource"`? I can pass with that. Thank you.

manirajv06: Ok, Please go ahead.

 BTW, can you check the 2nd comment?

manirajv06 (approved):

pkg/common/configs/configvalidator.go
FrankYang0529: > Are doing hierarchical comparison checks for wildcard user and group similar to named users and groups? I doubt this.

IMO, the validation is reasonable. If an admin gives limits like `user1: 100 vcores` and `*: 50 vcores` in the `root` queue, they may think user1 can use 100 vcores, but each of others can use only 50 vcores. In this situation, it's weird to set `user2: 100 vcores` in `root.child` queue.

manirajv06: My question is,

root.parent => * => mem:10
root.parent.child1 => * => mem:15

Are we doing this wild card limit check between queues at different hierarchy? Ideally, we should not allow this.

FrankYang0529: Agree. Add more test cases to cover it.

* leaf queue user wildcard maxresources are within immediate parent queue wildcard maxresources
* leaf queue user wildcard maxresources are within grandparent queue wildcard maxresources
* leaf queue user wildcard maxresources exceed immediate parent queue wildcard maxresources
* leaf queue user wildcard maxresources exceed grandparent queue wildcard maxresources

manirajv06 (approved):

PR #724, opened 2023-11-22
[YUNIKORN-2171] race between node removal and scheduling

wilfred-s:
### What is this PR for?
When a node gets removed the partition resources and thus the root max resources are decreased. Cleanup of the allocations happens after that. This means that for a short period of time the root queue max resources are already decreased while the usage is not.

If the queue headroom is limited by the root queue then we could have a race between the removal of the node allocations and scheduling adding a new allocation. Scheduling could fail with the root queue usage being over the max resources.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2171

not been able to find a way to test this

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -611,9 +611,8 @@ func (pc *PartitionContext) addNodeToList(node *objects.Node) error {
 	return nil
 }
 
-// Update the partition details when removing a node.
-// This locks the partition. The partition may not be locked when we process the allocation
-// removal from the node as that takes further app, queue or node locks
+// removeNodeFromList removes the node from the list of partition nodes.
+// This locks the partition.
 func (pc *PartitionContext) removeNodeFromList(nodeID string) *objects.Node {
 	pc.Lock()
 	defer pc.Unlock()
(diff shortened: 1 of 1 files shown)


pbacsko:
I think this is related to #505. I proposed a solution with an extra lock but Craig recommended holding the partition read lock during the scheduling cycle. If we do that, it would solve this and also #505 at the same time.

manirajv06:
As a first step in decommissioning, Node has been marked `schedulable` as false. This happens well before all the allocations cleanup, root queue resource adjustments etc. Since we are doing the check `is node schedulable` in all try*() calls, just wondering how do we encounter this problem between root queue adjustments & allocations cleanup?

wilfred-s:
I think I need to answer the questions in the reverse order they were made:

> As a first step in decommissioning, Node has been marked schedulable as false. This happens well before all the allocations cleanup, root queue resource adjustments etc. Since we are doing the check is node schedulable in all try*() calls, just wondering how do we encounter this problem between root queue adjustments & allocations cleanup?

The problem is not the fact that the node is marked unschedulable or not. The node resources are removed from the root queue max resources. That happens before we remove the allocations from the application(s) and queue(s).
So assume the following: 
- cluster with 2 nodes A and B: each with 32GB
- the root queue max is set to 64GB.
- both nodes have a usage of 30GB, root queue usage is thus 60GB
- root queue has a headroom of 4GB (64GB - 60GB).
- queues below the root do not have a max set (simplify the example)

Now we're allocating a 2GB allocation. Fits well inside the headroom of the root queue.
Allocation is assigned to node A, all fits and looks OK. When the allocation is added to the node and queue is when the issue starts.
During the time the scheduler is trying to find a node for the allocation the removal of node B is requested, and partially processed: the root queue max is set to 32GB. The allocations are not yet removed, which means the root queue usage is still 60GB.
We now try to update the usage of the queues as part of the node A confirmation. At the root queue the usage _would_ become 60GB + 2GB == 62GB. This fails.

The max of the root queue was set to 32GB due to the node removal. _Any_ increment would thus put the root queue over its max. Root queue usage will still be somewhere between 30GB and 60GB. Any allocation from the removed node that is left would cause the root queue to go over its max value as it already is over...
After this scheduling will not happen until all allocations from the removed node are removed. The headroom of the root queue will be negative until all is done. It thus only happens once and it fixes itself. We should prevent this from happening anyway.

> I think this is related to https://github.com/apache/yunikorn-core/pull/505. I proposed a solution with an extra lock

This update is not linked to configuration. Node additions or removals are not config and are not part of the scheduling cycle. The change proposed there does not cover this behaviour.

pbacsko (approved):
+1 LGTM

PR #727, opened 2023-11-22
YUNIKORN-2175 Add queue headRoom for Rest API querying and improve logs

zhuqi-lucas:
### What is this PR for?
1. We should support query headroom using rest API.

2. And from the logs, i can't find any info about the headroom value, when application request can't meet the headroom.

```
Application **** does not fit into *** queue\" timestampNano:1700625301297929026 referenceID:\"******\" resource:{resources:{key:\"ephemeral-storage\" value:{}} resources:{key:\"memory\" value:{value:24000000000}} resources:{key:\"pods\" value:{value:1}} resources:{key:\"vcore\" value:{value:13000}}}"} 
```
We need to add more logs about this case!

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-2175

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application_events.go
+++ b/pkg/scheduler/objects/application_events.go
@@ -20,12 +20,12 @@ package objects
 
 import (
 	"fmt"
-	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"time"
 
 	"golang.org/x/time/rate"
 
 	"github.com/apache/yunikorn-core/pkg/common"
+	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/apache/yunikorn-core/pkg/events"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
--- a/pkg/scheduler/objects/application_events.go
+++ b/pkg/scheduler/objects/application_events.go
@@ -40,7 +40,7 @@ func (evt *applicationEvents) sendAppDoesNotFitEvent(request *AllocationAsk, hea
 	if !evt.eventSystem.IsEventTrackingEnabled() || !evt.limiter.Allow() {
 		return
 	}
-	message := fmt.Sprintf("Application %s does not fit into %s queue, request resoure %s can't fit headroom %s", request.GetApplicationID(), evt.app.queuePath, request.GetAllocatedResource().String(), headroom.String())
+	message := fmt.Sprintf("Application %s does not fit into %s queue (request resoure %s, headroom %s)", request.GetApplicationID(), evt.app.queuePath, request.GetAllocatedResource(), headroom)
 	event := events.CreateRequestEventRecord(request.GetAllocationKey(), request.GetApplicationID(), message, request.GetAllocatedResource())
 	evt.eventSystem.AddEvent(event)
 }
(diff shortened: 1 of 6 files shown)


manirajv06 (changes requested):
Can you add tests to cover this change?

zhuqi-lucas:
Thank you @manirajv06 for review, add more logs improvement, and also added test cases!

pkg/scheduler/objects/application_events.go
```diff
@@ -20,6 +20,7 @@ package objects
 
 import (
 	"fmt"
+	"github.com/apache/yunikorn-core/pkg/common/resources"
```
pbacsko, line 23: linter

pkg/scheduler/objects/application_events_test.go
```diff
@@ -19,6 +19,7 @@
 package objects
 
 import (
+	"github.com/apache/yunikorn-core/pkg/common/resources"
```
pbacsko, line 22: linter

pkg/scheduler/objects/application_events.go
zhuqi-lucas: Thank you @pbacsko , fixed in latest PR.

pkg/webservice/dao/queue_info.go
```diff
@@ -33,6 +33,7 @@ type PartitionQueueDAOInfo struct {
 	GuaranteedResource     map[string]int64        `json:"guaranteedResource,omitempty"`
 	AllocatedResource      map[string]int64        `json:"allocatedResource,omitempty"`
 	PreemptingResource     map[string]int64        `json:"preemptingResource,omitempty"`
+	HeadRoom               map[string]int64        `json:"headroom,omitempty"`
```
pbacsko, line 36: It's not tested whether this value appears properly after a web service call. It's not just this field, but quite a few inside `PartitionQueueDAOInfo`. There's a test case `TestGetPartitionQueuesHandler` which should be improved to see if values from the schedulerContext are mapped back properly. This can be done in a follow-up JIRA.

pbacsko: Created https://issues.apache.org/jira/browse/YUNIKORN-2179 so that we can proceed with this one.

pbacsko (approved):
+1

pkg/scheduler/objects/application_events.go
```diff
@@ -35,11 +36,11 @@ type applicationEvents struct {
 	limiter     *rate.Limiter
 }
 
-func (evt *applicationEvents) sendAppDoesNotFitEvent(request *AllocationAsk) {
+func (evt *applicationEvents) sendAppDoesNotFitEvent(request *AllocationAsk, headroom *resources.Resource) {
 	if !evt.eventSystem.IsEventTrackingEnabled() || !evt.limiter.Allow() {
 		return
 	}
-	message := fmt.Sprintf("Application %s does not fit into %s queue", request.GetApplicationID(), evt.app.queuePath)
+	message := fmt.Sprintf("Application %s does not fit into %s queue, request resoure %s can't fit headroom %s", request.GetApplicationID(), evt.app.queuePath, request.GetAllocatedResource().String(), headroom.String())
```
pbacsko, line 43: Minor nit no.1: calling String() is not needed, fmt.Sprintf() can recognize the Stringer interface.
Minor nit no.2: slight rewording: "Application %s does not fit into %s queue (request resoure %s, headroom %s)". The proposed message sounds like two separate problems.

pbacsko (changes requested):
Two minor comment

pkg/scheduler/objects/application_events.go
zhuqi-lucas: Thanks, addressed in latest PR.

pkg/webservice/dao/queue_info.go
zhuqi-lucas: Good catch @pbacsko !

pbacsko (approved):
+1

PR #713, opened 2023-11-13
[YUNIKORN-2089] Move usedResource type and tests to their own files

vinayakphegde:
### What is this PR for?
* Moved usedResource type and the related functions to their own files
* Created unit tests for usedResource functions

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2089

PR #734, opened 2023-11-27
[YUNIKORN-2145] refactor: ApplicationSummary into its own file

brandboat:
### What is this PR for?
ApplicationSummary is a type on its own which should not be included in the generic application file. Move the type and related functions into its own application_summary.go file. Move the related tests out too.

### What type of PR is it?
* [x] - Improvement

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2145

covered by unit tests

N/A

### Questions:
N/A


pkg/scheduler/objects/application_summary.go
```diff
@@ -0,0 +1,65 @@
...
+		as.SubmissionTime.UnixMilli(),
+		as.StartTime.UnixMilli(),
+		as.FinishTime.UnixMilli(),
+		as.User,
+		as.Queue,
+		as.State,
+		as.RmID,
+		as.ResourceUsage.TrackedResourceMap,
+		as.PreemptedResource.TrackedResourceMap,
+		as.PlaceholderResource.TrackedResourceMap)
```
brandboat, line 60: The `TrackedResource` to String method has been implemented in https://github.com/apache/yunikorn-core/pull/713. Currently, it has not been merged into the master branch. Therefore, we will continue to use TrackedResourceMap here.

pbacsko: I'll ping the owner of #713.

pkg/scheduler/objects/application_summary_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+package objects
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestNilApplicationSummaryToString(t *testing.T) {
```
pbacsko, line 27: Is this necessary? I haven't seen any unit test which tests sth like that... I don't think we need this.

pkg/scheduler/objects/application_summary.go
```diff
@@ -0,0 +1,65 @@
...
+	RmID                string
+	ResourceUsage       *resources.TrackedResource
+	PreemptedResource   *resources.TrackedResource
+	PlaceholderResource *resources.TrackedResource
+}
+
+func (as *ApplicationSummary) String() string {
+	if as == nil {
+		return "nil ApplicationSummary"
+	}
```
pbacsko, line 47: Do we need this "nil" check?

pkg/scheduler/objects/application_summary.go
```diff
@@ -0,0 +1,65 @@
...
+
+package objects
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+	"github.com/apache/yunikorn-core/pkg/log"
+	"go.uber.org/zap"
```
pbacsko, line 27: Fix the imports.

pbacsko (changes requested):
See comments. I don't think we need the nil check & test.

pkg/scheduler/objects/application_summary.go
brandboat: For now, we don't need this nil check. I'll remove the nil check and test code in the next commit, many thanks.

PR #736, opened 2023-11-28
[YUNIKORN-2203] Possible log spew in UGM code

pbacsko:
Don't log from methods which can be repeatedly called from the scheduling cycle as it floods the log.

### What is this PR for?
Remove logging from hot code path. Yunikorn logs will be flooded if we switch to DEBUG level or an enforcement kicks in and Yunikorn constantly tries to schedule the app.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2203

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -133,7 +133,7 @@ func (m *Manager) DecreaseTrackedResource(queuePath, applicationID string, usage
 		return decreased
 	}
 	if removeQT {
-		log.Log(log.SchedUGM).Debug("Removing user from manager",
+		log.Log(log.SchedUGM).Info("Removing user from manager",
 			zap.String("user", user.User))
 		delete(m.userTrackers, user.User)
 	}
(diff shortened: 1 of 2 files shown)


FrankYang0529:
@pbacsko, Wilfred gave another suggestion in the thread. Do we want to do it in another Jira? If yes, I can help it. Thank you.

```
Other locations that do log in the UGM at DEBUG and I think should be logged at an INFO level:

L136: "Removing user from manager"
L162: "Removing group from manager"
L222: "Group tracker doesn't exists. Creating appGroup tracker"
L231: "Group tracker set for user application"
L565: "User tracker doesn't exists. Creating user tracker."
These are all major changes and low volume which should be tracked in the logs.
```

FrankYang0529 (approved):
+1

pbacsko:
I'll change those.

FrankYang0529 (approved):

PR #750, opened 2023-12-07
[YUNIKORN-2235] Add new RESTful API for retrieving application

laysfire:
### What is this PR for?
This PR add 2 RESTful API. 
One for retrieving one application object directly via /ws/v1/partition/{partitionName}/application/{applicationID}.
The other for listing application IDs via /ws/v1/partition/{partitionName}/applications/{state}.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2235

It's been locally tested using go test
N/A

### Questions:
N/A

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -425,6 +425,10 @@ func (pc *PartitionContext) removeAppInternal(appID string) *objects.Application
 	return app
 }
 
+func (pc *PartitionContext) GetApplication(appID string) *objects.Application {
+	return pc.getApplication(appID)
+}
+
 func (pc *PartitionContext) getApplication(appID string) *objects.Application {
 	pc.RLock()
 	defer pc.RUnlock()
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -58,35 +58,21 @@ const (
 	GroupNameMissing         = "Group name is missing"
 	ApplicationDoesNotExists = "Application not found"
 	NodeDoesNotExists        = "Node not found"
-	ActiveState              = "active"
 )
 
-var allowedStatesMsg string
 var allowedActiveStatesMsg string
-var allowedAppStates map[string]bool
 var allowedAppActiveStates map[string]bool
 
 func init() {
-	allowedAppStates = make(map[string]bool)
 	allowedAppActiveStates = make(map[string]bool)
 
-	allowedAppStates[ActiveState] = true
-	allowedAppStates[strings.ToLower(objects.Rejected.String())] = true
-	allowedAppStates[strings.ToLower(objects.Completed.String())] = true
-
-	var states []string
-	for k := range allowedAppStates {
-		states = append(states, k)
-	}
-	allowedStatesMsg = fmt.Sprintf("Only following application states are allowed: %s", strings.Join(states, ","))
-
-	allowedAppActiveStates[strings.ToLower(objects.New.String())] = true
-	allowedAppActiveStates[strings.ToLower(objects.Accepted.String())] = true
-	allowedAppActiveStates[strings.ToLower(objects.Starting.String())] = true
-	allowedAppActiveStates[strings.ToLower(objects.Running.String())] = true
-	allowedAppActiveStates[strings.ToLower(objects.Completing.String())] = true
-	allowedAppActiveStates[strings.ToLower(objects.Failing.String())] = true
-	allowedAppActiveStates[strings.ToLower(objects.Resuming.String())] = true
+	allowedAppActiveStates["new"] = true
+	allowedAppActiveStates["accepted"] = true
+	allowedAppActiveStates["starting"] = true
+	allowedAppActiveStates["running"] = true
+	allowedAppActiveStates["completing"] = true
+	allowedAppActiveStates["failing"] = true
+	allowedAppActiveStates["resuming"] = true
 
 	var activeStates []string
 	for k := range allowedAppActiveStates {
@@ -669,13 +655,13 @@ func getPartitionApplicationsByState(w http.ResponseWriter, r *http.Request) {
 		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusNotFound)
 		return
 	}
-	if !allowedAppStates[appState] {
-		buildJSONErrorResponse(w, allowedStatesMsg, http.StatusBadRequest)
+	if appState != "active" && appState != "rejected" && appState != "completed" {
+		buildJSONErrorResponse(w, "Only following application states are allowed: active, rejected, completed", http.StatusBadRequest)
 		return
 	}
 	var appList []*objects.Application
 	switch appState {
-	case ActiveState:
+	case "active":
 		if status := strings.ToLower(r.URL.Query().Get("status")); status != "" {
 			if !allowedAppActiveStates[status] {
 				buildJSONErrorResponse(w, allowedActiveStatesMsg, http.StatusBadRequest)
@@ -689,15 +675,15 @@ func getPartitionApplicationsByState(w http.ResponseWriter, r *http.Request) {
 		} else {
 			appList = partitionContext.GetApplications()
 		}
-	case strings.ToLower(objects.Rejected.String()):
+	case "rejected":
 		appList = partitionContext.GetRejectedApplications()
-	case strings.ToLower(objects.Completed.String()):
+	case "completed":
 		appList = partitionContext.GetCompletedApplications()
 	default:
 		buildJSONErrorResponse(w, "BUG: unknown state", http.StatusInternalServerError)
 		return
 	}
-	appsDao := make([]*dao.ApplicationDAOInfo, 0)
+	appsDao := make([]*dao.ApplicationDAOInfo, 0, len(appList))
 	for _, app := range appList {
 		appsDao = append(appsDao, getApplicationDAO(app))
 	}
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -58,7 +58,42 @@ const (
 	GroupNameMissing         = "Group name is missing"
 	ApplicationDoesNotExists = "Application not found"
 	NodeDoesNotExists        = "Node not found"
+	ActiveState              = "Active"
 )
+
+var allowedStatesMsg string
+var allowedActiveStatesMsg string
+var allowedAppStates map[string]bool
+var allowedAppActiveStates map[string]bool
+
+func init() {
+	allowedAppStates = make(map[string]bool)
+	allowedAppActiveStates = make(map[string]bool)
+
+	allowedAppStates[ActiveState] = true
+	allowedAppStates[objects.Rejected.String()] = true
+	allowedAppStates[objects.Completed.String()] = true
+
+	var states []string
+	for k := range allowedAppStates {
+		states = append(states, k)
+	}
+	allowedStatesMsg = fmt.Sprintf("Only following application states are allowed: %s", strings.Join(states, ","))
+
+	allowedAppActiveStates[objects.New.String()] = true
+	allowedAppActiveStates[objects.Accepted.String()] = true
+	allowedAppActiveStates[objects.Starting.String()] = true
+	allowedAppActiveStates[objects.Running.String()] = true
+	allowedAppActiveStates[objects.Completing.String()] = true
+	allowedAppActiveStates[objects.Failing.String()] = true
+	allowedAppActiveStates[objects.Resuming.String()] = true
+
+	var activeStates []string
+	for k := range allowedAppActiveStates {
+		activeStates = append(activeStates, k)
+	}
+	allowedActiveStatesMsg = fmt.Sprintf("Only following active states are allowed: %s", strings.Join(activeStates, ","))
+}
 
 func getStackInfo(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
@@ -619,6 +654,55 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getPartitionApplicationsByState(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	vars := httprouter.ParamsFromContext(r.Context())
+	if vars == nil {
+		buildJSONErrorResponse(w, MissingParamsName, http.StatusBadRequest)
+		return
+	}
+	partition := vars.ByName("partition")
+	appState := vars.ByName("state")
+
+	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
+	if partitionContext == nil {
+		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusNotFound)
+		return
+	}
+	if !allowedAppStates[appState] {
+		buildJSONErrorResponse(w, allowedStatesMsg, http.StatusBadRequest)
+		return
+	}
+	var appList []*objects.Application
+	switch appState {
+	case ActiveState:
+		if status := r.URL.Query().Get("status"); status != "" {
+			if !allowedAppActiveStates[status] {
+				buildJSONErrorResponse(w, allowedActiveStatesMsg, http.StatusBadRequest)
+				return
+			}
+			for _, app := range partitionContext.GetApplications() {
+				if app.CurrentState() == status {
+					appList = append(appList, app)
+				}
+			}
+		} else {
+			appList = partitionContext.GetApplications()
+		}
+	case objects.Rejected.String():
+		appList = partitionContext.GetRejectedApplications()
+	case objects.Completed.String():
+		appList = partitionContext.GetCompletedApplications()
+	}
+	appsDao := make([]*dao.ApplicationDAOInfo, 0)
+	for _, app := range appList {
+		appsDao = append(appsDao, getApplicationDAO(app))
+	}
+	if err := json.NewEncoder(w).Encode(appsDao); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+	}
+}
+
 func getApplication(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	vars := httprouter.ParamsFromContext(r.Context())
@@ -629,22 +713,27 @@ func getApplication(w http.ResponseWriter, r *http.Request) {
 	partition := vars.ByName("partition")
 	queueName := vars.ByName("queue")
 	application := vars.ByName("application")
-	queueErr := validateQueue(queueName)
-	if queueErr != nil {
-		buildJSONErrorResponse(w, queueErr.Error(), http.StatusBadRequest)
-		return
-	}
 	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
 	if partitionContext == nil {
 		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusNotFound)
 		return
 	}
-	queue := partitionContext.GetQueue(queueName)
-	if queue == nil {
-		buildJSONErrorResponse(w, QueueDoesNotExists, http.StatusNotFound)
-		return
+	var app *objects.Application
+	if len(queueName) == 0 {
+		app = partitionContext.GetApplication(application)
+	} else {
+		queueErr := validateQueue(queueName)
+		if queueErr != nil {
+			buildJSONErrorResponse(w, queueErr.Error(), http.StatusBadRequest)
+			return
+		}
+		queue := partitionContext.GetQueue(queueName)
+		if queue == nil {
+			buildJSONErrorResponse(w, QueueDoesNotExists, http.StatusNotFound)
+			return
+		}
+		app = queue.GetApplication(application)
 	}
-	app := queue.GetApplication(application)
 	if app == nil {
 		buildJSONErrorResponse(w, ApplicationDoesNotExists, http.StatusNotFound)
 		return
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -61,24 +61,24 @@ const (
 )
 
 var allowedActiveStatusMsg string
-var allowedAppActiveStates map[string]bool
+var allowedAppActiveStatuses map[string]bool
 
 func init() {
-	allowedAppActiveStates = make(map[string]bool)
+	allowedAppActiveStatuses = make(map[string]bool)
 
-	allowedAppActiveStates["new"] = true
-	allowedAppActiveStates["accepted"] = true
-	allowedAppActiveStates["starting"] = true
-	allowedAppActiveStates["running"] = true
-	allowedAppActiveStates["completing"] = true
-	allowedAppActiveStates["failing"] = true
-	allowedAppActiveStates["resuming"] = true
+	allowedAppActiveStatuses["new"] = true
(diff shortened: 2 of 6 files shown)


pkg/webservice/handlers.go
```diff
@@ -619,6 +628,58 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+
+func getPartitionApplication(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	vars := httprouter.ParamsFromContext(r.Context())
+	if vars == nil {
+		buildJSONErrorResponse(w, MissingParamsName, http.StatusBadRequest)
+		return
+	}
+	partition := vars.ByName("partition")
+	application := vars.ByName("application")
```
pbacsko, line 666: Check valid partition and application (`partition != ""` and `application != ""`)

pkg/webservice/handlers.go
```diff
@@ -619,6 +628,58 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getPartitionApplicationsByState(w http.ResponseWriter, r *http.Request) {
```
pbacsko, line 631: Function name: you're not returning applications but application IDs. A better name is `getPartitionApplicationIDsByState()`.

pkg/webservice/handlers_test.go
```diff
@@ -1088,8 +1088,13 @@ func addAppWithUserGroup(t *testing.T, id string, part *scheduler.PartitionConte
 	assert.Equal(t, 1+initSize, len(part.GetApplications()))
 	if isCompleted {
 		app.SetState(objects.Completing.String())
+		currentCount := len(part.GetCompletedApplications())
 		err = app.HandleApplicationEvent(objects.CompleteApplication)
 		assert.NilError(t, err, "The app should have completed")
+		err = common.WaitFor(10*time.Millisecond, time.Duration(1000)*time.Millisecond, func() bool {
```
pbacsko, line 1094: `time.Duration(1000)*time.Millisecond` --> `time.Second()`

pbacsko (changes requested):
See comments.

pkg/scheduler/partition.go
```diff
@@ -425,6 +425,10 @@ func (pc *PartitionContext) removeAppInternal(appID string) *objects.Application
 	return app
 }
 
+func (pc *PartitionContext) GetApplication(appID string) *objects.Application {
+	return pc.getApplication(appID)
+}
```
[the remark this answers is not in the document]
laysfire: Hi, pbacsko, thanks for you reivew. There are RLock/RUnlock call in getApplication function.
It seems that RWMutex is not reentrant. Adding RLock/RUnlock in GetApplication may lead deadlock.

pkg/webservice/handlers.go
```diff
@@ -88,6 +88,15 @@ func getClusterInfo(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func validateApplicationState(appType string) error {
```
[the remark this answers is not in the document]
laysfire: Remove this helper function.

pkg/webservice/handlers.go
laysfire: partition = "" & application = "" will be handle by the following call schedulerContext.GetPartitionWithoutClusterID(partition) and partitionContext.GetApplication(application)

pkg/scheduler/partition.go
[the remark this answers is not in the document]
pbacsko: True, I missed that.

pkg/webservice/handlers.go
```diff
@@ -60,6 +60,12 @@ const (
 	NodeDoesNotExists        = "Node not found"
 )
 
+var allowedAppStates = map[string]bool{
```
chia7712, line 63: Pardon me, why only partial states are allowed?

pkg/webservice/dao/application_info.go
```diff
@@ -56,3 +56,7 @@ type PlaceholderDAOInfo struct {
 	Replaced      int64            `json:"replaced,omitempty"`
 	TimedOut      int64            `json:"timedout,omitempty"`
 }
+
+type ApplicationIDsDAOInfo struct {
+	AppIDs []string
```
chia7712, line 61: Does it need json tag?

chia7712 (commented):
nice patch! a couple of comments left.

pkg/scheduler/partition.go
```diff
@@ -425,6 +425,10 @@ func (pc *PartitionContext) removeAppInternal(appID string) *objects.Application
 	return app
 }
 
+func (pc *PartitionContext) GetApplication(appID string) *objects.Application {
```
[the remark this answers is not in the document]
laysfire: hi @chia7712 thanks for review.
The are so many places call the internal method.
Which way do you think is better?
First,
- Add RLock/RUnlock in exposed method & Remove RLock/RUnlock in internal method
- Add explicit RLock/RUnlock snippet in every place that call the internal method
Second,
Just change the exposed name

pkg/webservice/dao/application_info.go
laysfire: yes, i will add json tag & keep consistency

pkg/webservice/handlers.go
```diff
@@ -619,6 +625,57 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+	partition := vars.ByName("partition")
+	appState := vars.ByName("state")
+
+	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
+	if partitionContext == nil {
+		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusNotFound)
+		return
+	}
+	if !allowedAppStates[appState] {
+		buildJSONErrorResponse(w, "Application state is not allowed: "+appState, http.StatusBadRequest)
```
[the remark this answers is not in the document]
laysfire: yes, make sense.

pkg/scheduler/partition.go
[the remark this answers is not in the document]
chia7712: > Add RLock/RUnlock in exposed method & Remove RLock/RUnlock in internal method

yep

> Add explicit RLock/RUnlock snippet in every place that call the internal method

How about just replacing internal method `getApplication` by public version `GetApplication`? Small change is good at avoiding new bugs :)

Also, you can file new Jira as follow-up to improve lock usage if you observe any weird use case.

pkg/scheduler/partition.go
```diff
@@ -1066,6 +1070,18 @@ func (pc *PartitionContext) GetRejectedApplications() []*objects.Application {
 	return appList
 }
 
+func (pc *PartitionContext) GetAppIDsByState(state string) []string {
```
pbacsko, line 1073: See below, I don't think this method needed as is.

pkg/webservice/handlers.go
```diff
@@ -619,6 +625,57 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	appIDs := partitionContext.GetAppIDsByState(appState)
```
pbacsko, line 647: I've just realized this... A big question here: if we already do state-based filtering here, do we need to do that again in `GetAppIDsByState()`? We shouldn't have the same logic in two places, it's not needed. What we expose on REST is an interface decision, so I'd rather have it here only. `getAppsByState()` should have an exposed pair `GetAppsByState()` which we call here.

pbacsko (changes requested):
Some new comments.

pkg/webservice/handlers.go
laysfire: For the Rejected applications, they were stored in rejectedApplications map. And the get function is getRejectedAppsByState(). Same for completed application.
Should we place the judge logic in handlers or partition?

laysfire: hi @pbacsko , If i understand correctly, you mean create exposed pair method for getAppsByState(), getRejectedAppsByState() & getCompletedAppsByState(). And let restful handler decide which function to call.
plz correct me if i'm wrong.

pbacsko: Yes that's the idea, something like that. Just avoid the double filtering which is error prone. Make a decision at one place and the right place is here, the handler method.

manirajv06:
@laysfire I have shared my views in jira. Please check.

laysfire:
@manirajv06 @chia7712 @pbacsko Address some comment, please help review this again.

pkg/webservice/handlers.go
chia7712: @laysfire could you share the discussion to me? Also, adding the comment here can helper readers in the future :)

pkg/webservice/routes.go
```diff
@@ -122,6 +122,18 @@ var webRoutes = routes{
...
+	route{
+		"Scheduler",
+		"GET",
+		"/ws/v1/partition/:partition/application/:application",
+		getApplication,
+	},
+	route{
+		"Scheduler",
+		"GET",
+		"/ws/v1/partition/:partition/applications/:state",
```
chia7712, line 134: this endpoint is under discussion. https://issues.apache.org/jira/browse/YUNIKORN-2235

pkg/webservice/handlers.go
```diff
@@ -619,6 +625,41 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
+	if partitionContext == nil {
+		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusNotFound)
+		return
+	}
+	if !allowedAppStates[appState] {
+		buildJSONErrorResponse(w, fmt.Sprintf("Only %s,%s,%s state is allowed",
+			objects.Running.String(), objects.Rejected.String(), objects.Completed.String()), http.StatusBadRequest)
+		return
+	}
```
pbacsko, line 647: I don't want to overengineer things, but it's safer to auto-construct this:
```
var allowedStatesMsg string

var allowedAppStates map[string]bool

func init() {
	allowedAppStates[objects.Running.String()] = true
	allowedAppStates[objects.Rejected.String()] = true
	allowedAppStates[objects.Completed.String()] = true

	var states []string
	for k := range allowedAppStates {
		states = append(states, k)
	}
	allowedStatesMsg = fmt.Sprintf("Only following application states are allowed: %s", strings.Join(states, ","))
}

...

	if !allowedAppStates[appState] {
		buildJSONErrorResponse(w, allowedStatesMsg, http.StatusBadRequest)
		return
	}
```

pkg/webservice/routes.go
manirajv06: Updated jira with more details. Please check.

pkg/scheduler/partition.go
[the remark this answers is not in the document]
wilfred-s: It might be a package internal method until this point. Lots of places calling it but they are all in the package. The partition is the place where the events from the shim get processed **and** the scheduling cycle runs from. That means we have multiple go routines that could be changing values stored in the partition. The locking makes sure we have no data races and a consistent view.
The internal method was locked for that reason and needs to stay like that.

wilfred-s:
Dropped in a large comment in the [jira](https://issues.apache.org/jira/browse/YUNIKORN-2235?focusedCommentId=17798448&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17798448):

laysfire:
@manirajv06 @wilfred-s @chia7712 @pbacsko please help review again.
Summary are as follows:
- Expose GetApplication on partition with nolock version
- Make getApplication handler handle `/ws/v1/partition/{partitionName}/application/{applicationID} `
- Add getPartitionApplicationByState to retrieve  applications by types
- We only allow 3 state: Active(fake), Rejected and Completed
- We only allow New, Accepted, Starting, Running, Completing, Failing and Resuming status for Active state
- I do not modify the return DAOs, like clean allocations for completed application. I think we can done this in follow up jira.

pkg/webservice/handlers.go
```diff
@@ -619,6 +654,55 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
 
+func getPartitionApplicationsByState(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	vars := httprouter.ParamsFromContext(r.Context())
+	if vars == nil {
+		buildJSONErrorResponse(w, MissingParamsName, http.StatusBadRequest)
+		return
+	}
+	partition := vars.ByName("partition")
+	appState := vars.ByName("state")
```
pbacsko, line 665: not sure if it's common practice or not, but doing a `toLower()` makes it more resilient

pkg/webservice/handlers.go
```diff
@@ -619,6 +654,55 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusNotFound)
+		return
+	}
+	if !allowedAppStates[appState] {
+		buildJSONErrorResponse(w, allowedStatesMsg, http.StatusBadRequest)
+		return
+	}
+	var appList []*objects.Application
+	switch appState {
+	case ActiveState:
```
pbacsko, line 678: We don't have "default" branch. Shouldn't we have one and return with an error message (ie. "unsupported state")?

pbacsko (changes requested):
Two questions.

pkg/webservice/handlers.go
```diff
@@ -629,22 +715,27 @@ func getApplication(w http.ResponseWriter, r *http.Request) {
...
-		return
+	var app *objects.Application
+	if len(queueName) == 0 {
+		app = partitionContext.GetApplication(application)
+	} else {
+		queueErr := validateQueue(queueName)
+		if queueErr != nil {
+			buildJSONErrorResponse(w, queueErr.Error(), http.StatusBadRequest)
+			return
+		}
```
pbacsko, line 731: Please add coverage to this branch.

pkg/webservice/handlers.go
```diff
@@ -619,6 +654,57 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
 
+func getPartitionApplicationsByState(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	vars := httprouter.ParamsFromContext(r.Context())
+	if vars == nil {
+		buildJSONErrorResponse(w, MissingParamsName, http.StatusBadRequest)
+		return
+	}
+	partition := vars.ByName(strings.ToLower("partition"))
+	appState := vars.ByName(strings.ToLower("state"))
```
pbacsko, line 665: Partitions are case-sensitive, so drop that.
For `appState`, the proper approach is:
```
appState := strings.ToLower(vars.ByName("state"))
```

pbacsko (changes requested):
Almost. One more round.

pkg/webservice/handlers.go
```diff
@@ -619,6 +654,57 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+			}
+		} else {
+			appList = partitionContext.GetApplications()
+		}
+	case objects.Rejected.String():
+		appList = partitionContext.GetRejectedApplications()
+	case objects.Completed.String():
+		appList = partitionContext.GetCompletedApplications()
+	default:
+		buildJSONErrorResponse(w, "BUG: unknown state", http.StatusInternalServerError)
```
pbacsko, line 697: `return` statement missing here

laysfire:
@pbacsko Thanks for review. Have support case-insensitive for appState & activeState. But i add many strings.ToLower in code. I'm not sure if this is a good practice?

pkg/webservice/handlers.go
```diff
@@ -58,8 +58,43 @@ const (
...
+	}
+	allowedStatesMsg = fmt.Sprintf("Only following application states are allowed: %s", strings.Join(states, ","))
+
+	allowedAppActiveStates[strings.ToLower(objects.New.String())] = true
+	allowedAppActiveStates[strings.ToLower(objects.Accepted.String())] = true
+	allowedAppActiveStates[strings.ToLower(objects.Starting.String())] = true
+	allowedAppActiveStates[strings.ToLower(objects.Running.String())] = true
+	allowedAppActiveStates[strings.ToLower(objects.Completing.String())] = true
+	allowedAppActiveStates[strings.ToLower(objects.Failing.String())] = true
+	allowedAppActiveStates[strings.ToLower(objects.Resuming.String())] = true
```
pbacsko, line 89: We might need a better naming here. It's a bit difficult to distingush between `allowedAppStates` and `allowedAppActiveStates`. At least some comments:

```
var allowedStatesMsg string  // returned error message when the requested application state is invalid
var allowedActiveStatesMsg string // returned error message when the actual state of the application is not an active state like Running, Accepted, etc.
var allowedAppStates map[string]bool // ??? (this one is a bit tough to explain, we mix two existing states with a higher level one which)
var allowedAppActiveStates map[string]bool // list of application states that are valid for filtering
```

We can actually just drop `allowedAppStates` because it only contains 3 values. Readability is important, so this is probably better:
```
	if appState != "active" && appState != "rejected" && appState != "completed" {
		buildJSONErrorResponse(w, "Only following application states are allowed: active, rejected, completed", http.StatusBadRequest)
		return
	}
```

You can also just drop the `strings.ToLower()` calls, it makes the code a bit clumsy. Just write:
```
        allowedAppActiveStates["new"] = true
	allowedAppActiveStates["accepted"] = true
	allowedAppActiveStates["starting"] = true
...
```

Obviously this is hard-coded and we no longer reference the code which define the states, but something for something. However, it's pretty unlikely that the state machine will ever change. You can add a comment to `application_state.go` above the `const` section:

`// Application states are used for filtering in the webservice handlers. Please check&update the logic if needed if the state machine is modified`

Sometimes we need to be pragmatic.

pkg/webservice/handlers.go
```diff
@@ -619,6 +654,58 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+		}
+	case strings.ToLower(objects.Rejected.String()):
+		appList = partitionContext.GetRejectedApplications()
+	case strings.ToLower(objects.Completed.String()):
+		appList = partitionContext.GetCompletedApplications()
+	default:
+		buildJSONErrorResponse(w, "BUG: unknown state", http.StatusInternalServerError)
+		return
+	}
+	appsDao := make([]*dao.ApplicationDAOInfo, 0)
```
pbacsko, line 700: Nit: how about `appsDao := make([]*dao.ApplicationDAOInfo, 0, len(appList))` ?

pkg/webservice/handlers.go
```diff
@@ -619,6 +654,58 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+					appList = append(appList, app)
+				}
+			}
+		} else {
+			appList = partitionContext.GetApplications()
+		}
+	case strings.ToLower(objects.Rejected.String()):
+		appList = partitionContext.GetRejectedApplications()
+	case strings.ToLower(objects.Completed.String()):
+		appList = partitionContext.GetCompletedApplications()
```
pbacsko, line 695: Just "rejected" and "completed", drop the `ToLower()` call.

pbacsko (changes requested):
I agree that this change too much distracting stuff. Let's get rid of the `ToLower()` calls and just use plain string literals. See my comments.

pbacsko (approved):
+1 LGTM

I'll let others review it, too.

pkg/webservice/handlers.go
```diff
@@ -619,6 +640,58 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+	partition := vars.ByName("partition")
+	appState := strings.ToLower(vars.ByName("state"))
+
+	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
+	if partitionContext == nil {
+		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusNotFound)
+		return
+	}
+	if appState != "active" && appState != "rejected" && appState != "completed" {
+		buildJSONErrorResponse(w, "Only following application states are allowed: active, rejected, completed", http.StatusBadRequest)
```
chia7712, line 659: Could we move this line to line#683? for example:
```go
	case "active":
		if status := strings.ToLower(r.URL.Query().Get("status")); status != "" {
			if !allowedAppActiveStates[status] {
				buildJSONErrorResponse(w, allowedActiveStatesMsg, http.StatusBadRequest)
				return
			}
			for _, app := range partitionContext.GetApplications() {
				if strings.ToLower(app.CurrentState()) == status {
				}
			}
		} else {
			appList = partitionContext.GetApplications()
		}
	case "rejected":
		appList = partitionContext.GetRejectedApplications()
	case "completed":
		appList = partitionContext.GetCompletedApplications()
	default:
		buildJSONErrorResponse(w, "Only following application states are allowed: active, rejected, completed", http.StatusBadRequest)
		return
	}
```

pkg/webservice/handlers.go
```diff
@@ -619,6 +640,58 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	var appList []*objects.Application
+	switch appState {
+	case "active":
+		if status := strings.ToLower(r.URL.Query().Get("status")); status != "" {
+			if !allowedAppActiveStates[status] {
+				buildJSONErrorResponse(w, allowedActiveStatesMsg, http.StatusBadRequest)
```
chia7712, line 667: Should `allowedActiveStatesMsg` be renamed to `allowedActiveStatusMsg` if the query key is called "statue"?

chia7712 (commented):
@laysfire thanks for this nice feature. two small comments left. please take a look.

pkg/webservice/handlers.go
laysfire: Cool. Move #659 to "default" branch make code clearer and default branch would be possible to hit.

pkg/webservice/handlers.go
```diff
@@ -60,6 +60,27 @@ const (
...
+	allowedAppActiveStates["running"] = true
+	allowedAppActiveStates["completing"] = true
+	allowedAppActiveStates["failing"] = true
+	allowedAppActiveStates["resuming"] = true
+
+	var activeStates []string
+	for k := range allowedAppActiveStates {
+		activeStates = append(activeStates, k)
+	}
+	allowedActiveStatusMsg = fmt.Sprintf("Only following active states are allowed: %s", strings.Join(activeStates, ","))
```
chia7712, line 81: Could you change `active states` to `active statuses`? The query key is `status` so I prefer naming consistency.

chia7712 (commented):
@laysfire sorry that I am overengineering today. one small comment left.

BTW, please file follow-up to doc the new APIs

pkg/webservice/handlers.go
laysfire: @chia7712 Thanks for review. In order to keep consistency, should change allowedAppActiveStates to allowedAppActiveStatuses?

chia7712: yep!

chia7712 (approved):
LGTM

PR #752, opened 2023-12-08
[YUNIKORN-2233] Scheduler core cannot be stopped properly

pbacsko:
### What is this PR for?
After calling `MockScheduler.Stop()`, we still have a bunch of goroutines running. This is not a problem on a real cluster, because a restart means a complete shutdown of the Yunikorn process. During testing, this leaves garbage behind and generates possible unwanted interactions between test cases.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2233

Insert a breakpoint after `ms.Stop()` in a test which uses the Mock scheduler. Check the running goroutines with the debugger.

### Questions:

### DIFF
--- a/pkg/entrypoint/service_context.go
+++ b/pkg/entrypoint/service_context.go
@@ -21,25 +21,36 @@ package entrypoint
 import (
 	"go.uber.org/zap"
 
+	"github.com/apache/yunikorn-core/pkg/events"
 	"github.com/apache/yunikorn-core/pkg/log"
+	"github.com/apache/yunikorn-core/pkg/metrics"
+	"github.com/apache/yunikorn-core/pkg/rmproxy"
 	"github.com/apache/yunikorn-core/pkg/scheduler"
 	"github.com/apache/yunikorn-core/pkg/webservice"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/api"
 )
 
 type ServiceContext struct {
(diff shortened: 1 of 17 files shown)


pkg/entrypoint/service_context.go
```diff
@@ -21,25 +21,36 @@ package entrypoint
...
 		if err := s.WebApp.StopWebApp(); err != nil {
 			log.Log(log.Entrypoint).Error("failed to stop web-app",
 				zap.Error(err))
 		}
 	}
+	if s.MetricsCollector != nil {
+		s.MetricsCollector.Stop()
+	}
+	s.Scheduler.Stop()
+	if proxyImpl, ok := s.RMProxy.(*rmproxy.RMProxy); ok {
```
pbacsko, line 52: NOTE: Currently there's no Stop() method on  `api.SchedulerAPI`.

wilfred-s: change committed in SI

craigcondit (approved):
+1 LGTM.

wilfred-s (approved):
LGTM
nice cleanup for the stops

PR #754, opened 2023-12-11
[YUNIKORN-2257] Add rest API to retrieve node utilization for multiple resource types

chenyulin0719:
### What is this PR for?
Add new rest endpoint to return multiple resource types. ([]*dao.NodesUtilDAOInfo)
- /ws/v1/partition/:partition/node-utilization

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
Add api document.

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2257

1. make e2e test should passed.
2. deploy yunikorn-core and call http://localhost:9889/ws/v1/partition/default/node-utilization

YUNIKORN-2257 API Result

### Questions:
NA

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -378,6 +378,34 @@ func getNodeUtilisation(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getPartitionNodeUtilisation(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	vars := httprouter.ParamsFromContext(r.Context())
+	if vars == nil {
+		buildJSONErrorResponse(w, MissingParamsName, http.StatusBadRequest)
+		return
+	}
+	partition := vars.ByName("partition")
+	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
+
+	if partitionContext == nil {
+		buildJSONErrorResponse(w, PartitionDoesNotExists, http.StatusInternalServerError)
+		return
+	}
+
+	var result []*dao.NodesUtilDAOInfo
+	if partitionContext.GetTotalPartitionResource() != nil { // if have resource in partition
+		for name := range partitionContext.GetTotalPartitionResource().Resources {
+			nodesUtil := getNodesUtilJSON(partitionContext, name)
+			result = append(result, nodesUtil)
+		}
+	}
+
+	if err := json.NewEncoder(w).Encode(result); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+	}
+}
+
 // getNodesUtilJSON loads the nodes utilisation for a partition for a specific resource type.
 func getNodesUtilJSON(partition *scheduler.PartitionContext, name string) *dao.NodesUtilDAOInfo {
 	mapResult := make([]int, 10)
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -727,17 +727,10 @@ func TestGetNodeUtilisation(t *testing.T) {
 	assert.Assert(t, confirmNodeCount(utilisation.NodesUtil, 0), "unexpected number of nodes returned should be 0")
 
 	// create test nodes
-	nodeRes := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 10}).ToProto()
-	nodeRes2 := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 10, "second": 5}).ToProto()
 	node1ID := "node-1"
-	node1 := objects.NewNode(&si.NodeInfo{NodeID: node1ID, SchedulableResource: nodeRes})
 	node2ID := "node-2"
-	node2 := objects.NewNode(&si.NodeInfo{NodeID: node2ID, SchedulableResource: nodeRes2})
-
-	err = partition.AddNode(node1, nil)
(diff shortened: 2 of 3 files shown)


chenyulin0719:
Checked the CI result. I think the missing coverage(3 lines) is accpeted because the logic is straightforward, and it's diffcult to force 'json.NewEncoder(w).Encode(result') to return an error. (That would be unusual conditions.)  I can also see same code snippet was used in serveral APIs.
> 
	if err := json.NewEncoder(w).Encode(result); err != nil {
	}

manirajv06:
@chenyulin0719 I don't think adding new API is necessary here. https://issues.apache.org/jira/browse/YUNIKORN-2087 had already exposed an API to return the node utilisation for dominant resource type only . We should extend/generalise the same API to cover other resource types too. We can add a new field in the response to differentiate the dominant one from others if needed. However, at the same time, we need to ensure backward compatibility as well.

chenyulin0719:
> @chenyulin0719 I don't think adding new API is necessary here. https://issues.apache.org/jira/browse/YUNIKORN-2087 had already exposed an API to return the node utilisation for dominant resource type only . We should extend/generalise the same API to cover other resource types too. We can add a new field in the response to differentiate the dominant one from others if needed. However, at the same time, we need to ensure backward compatibility as well.

Hi @manirajv06, thanks for the feedback.

In YUNIKORN-2087, the new AP(/ws/v1/scheduler/node-utilization) returned domaint resource utilization but only for 'default' partition.  That was a workaround. ([comment in core](https://github.com/apache/yunikorn-core/blob/master/pkg/webservice/handlers.go#L354-L357))  Therefore, we have YUNIKORN-2088 to get more resource types in partition level)

My original idea for this PR is to modify the existing one to be cluster-wide and introduce a new partition-level API.
- /ws/v1/scheduler/node-utilization (change from 'default' to combine all partitions)
- /ws/v1/partition/:partition/node-utilization (new)

But after rethink about your suggestion, I believe you're correct. Those could be done by a generalized one.
We can change the existing api /ws/v1/scheduler/node-utilization to return all the necessary information.

**Current**: Return dao.NodesUtilDAOInfo (Domaint Resource only.)
**To be**: return all partitions, all type's resources utilization.

> [{
>        "partition": "default",
>        "nodeUtilizationList" : []*dao.NodesUtilDAOInfo
>     }, 
>     {
>        "partition": "other-partitions",
>        "nodeUtilizationList" : []*dao.NodesUtilDAOInfo
>     }]

I think we could remove the dominant calculation from yunikorn-core.  Craig has suggested an algorithm to calculate the weight of dao.NodesUtilDAOInfo, so we can get domaint resources indirectly in partitions/cluster level.   (https://yunikornworkspace.slack.com/archives/CL9CRJ1KM/p1701713212026479?thread_ts=1701680922.070739&cid=CL9CRJ1KM)

This change will break backward compatibility in UI. (the current node utilization chart will fail, we need a quick fix) 
For api, I think the change is acceptible because the existing api was not formal released in 1.4.0 (The document is going to be updated in 1.5.0)

Hi @manirajv06 , @wilfred-s , @craigcondit ,
If I can obtain your approval for my proposal, I will withdraw this pull request and modify the Jira.

Thanks for reading.

manirajv06:
+1 on extending/generalise the existing API to return Array of dao.NodesUtilDAOInfo objects for every partition. Though we say, every partition, in reality, today it is going to be "default" only. So, I don't think "partition" field carries any value here.

Though this proposal looks good, I am not sure whether UI changes can be pushed through around same time. To minimize the effort on UI side, it is better to keep the dominant resource check in the core for some more time and always keep the dominant resource as the fist item in the response array.

chenyulin0719:
Hi @manirajv06,

> Though this proposal looks good, I am not sure whether UI changes can be pushed through around same time. 
I'll push these changes together (Core/UI), this might require the assistance of someone from the committer group. (I should push it after slow down?)

> To minimize the effort on UI side, it is better to keep the dominant resource check in the core for some more time and always keep the dominant resource as the fist item in the response array.

Base on the Slack discussion, we'll need to calculate the weight of multiple resources in UI after YUNIKORN-2088.  (ex: to show only top N resoruces) 
So the domain resource calculation in core is redundant, and the UI calculation is quite simple, which should not impact the UI performance. So I still prefer to remove dominant resource check in core.

chenyulin0719:
I think we have 2 choices here.
1. Put all domaint reosurce check in UI
2. Put all domaint reosurce check in Core

My opionion is #1

manirajv06:
@wilfred-s and I had a offline discussion about this. Sharing the summary below.

1. It is good to proceed with the changes on both core and UI together in the same release to overcome the back compatibility problems. Please raise a jira for UI too and work on it once you are done with core.
2. Need to take a fresh look at getNodesUtilDAO. To be revisited from the scratch as lot of unnecessary calls or loops can be avoided because we would be doing it for all resource types.
3. Yes, we can do dominant resource check in UI if needed, but not anymore in core. REST API should expose the data based on the need and it is upto the UI to present the data in whatever way we want. As we agreed, response would have Array of `NodesUtilDAOInfo` objects.

chenyulin0719:
I will proceed to modify the PR/Jira based on the discussion conclusion.

Many thanks to @manirajv06 and @wilfred-s for providing your feedback.

PR #755, opened 2023-12-11
[YUNIKORN-2190] Using empty array to replace null response from /ws/v1/partition/{partitionName}/usage/users

brandboat:
### What is this PR for?
The endpoint `/ws/v1/partition/{partitionName}/usage/users` and `/ws/v1/partition/{partitionName}/usage/groups`
will return nil when user / group usages are nonexistent. The return type is expected to be array, so it seems to me the empty array is more suitable to be returned.

### What type of PR is it?
* [x] - Improvement

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2190

covered by unit tests

N/A

### Questions:
N/A

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -823,7 +823,7 @@ func getUsersResourceUsage(w http.ResponseWriter, _ *http.Request) {
 	writeHeaders(w)
 	userManager := ugm.GetUserManager()
 	usersResources := userManager.GetUsersResources()
-	var result []*dao.UserResourceUsageDAOInfo
+	result := []*dao.UserResourceUsageDAOInfo{}
 	for _, tracker := range usersResources {
 		result = append(result, tracker.GetUserResourceUsageDAOInfo())
 	}
@@ -859,7 +859,7 @@ func getGroupsResourceUsage(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	userManager := ugm.GetUserManager()
 	groupsResources := userManager.GetGroupsResources()
-	var result []*dao.GroupResourceUsageDAOInfo
+	result := []*dao.GroupResourceUsageDAOInfo{}
 	for _, tracker := range groupsResources {
 		result = append(result, tracker.GetGroupResourceUsageDAOInfo())
 	}
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -823,9 +823,9 @@ func getUsersResourceUsage(w http.ResponseWriter, _ *http.Request) {
 	writeHeaders(w)
 	userManager := ugm.GetUserManager()
 	usersResources := userManager.GetUsersResources()
-	result := []*dao.UserResourceUsageDAOInfo{}
-	for _, tracker := range usersResources {
-		result = append(result, tracker.GetUserResourceUsageDAOInfo())
+	result := make([]*dao.UserResourceUsageDAOInfo, len(usersResources))
+	for i, tracker := range usersResources {
+		result[i] = tracker.GetUserResourceUsageDAOInfo()
 	}
 	if err := json.NewEncoder(w).Encode(result); err != nil {
 		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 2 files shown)


pkg/webservice/handlers.go
```diff
@@ -823,7 +823,7 @@ func getUsersResourceUsage(w http.ResponseWriter, _ *http.Request) {
 	writeHeaders(w)
 	userManager := ugm.GetUserManager()
 	usersResources := userManager.GetUsersResources()
-	var result []*dao.UserResourceUsageDAOInfo
+	result := []*dao.UserResourceUsageDAOInfo{}
```
chia7712, line 826: how about using `make` with initial capacity? 

```
result := make([]*dao.UserResourceUsageDAOInfo, len(usersResources))
```

pkg/webservice/handlers.go
```diff
@@ -859,7 +859,7 @@ func getGroupsResourceUsage(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	userManager := ugm.GetUserManager()
 	groupsResources := userManager.GetGroupsResources()
-	var result []*dao.GroupResourceUsageDAOInfo
+	result := []*dao.GroupResourceUsageDAOInfo{}
```
chia7712, line 862: ditto

chia7712 (commented):
@brandboat thanks for this patch. a couple of comments below.

pkg/webservice/handlers.go
pbacsko: +1 for that

pbacsko: Actually I think it has to be `result := make([]*dao.UserResourceUsageDAOInfo, 0, len(usersResources))` so that appending starts at position 0.

wilfred-s: The even faster solution is to use this:
```
	result := make([]*dao.UserResourceUsageDAOInfo, len(usersResources))
	for i, tracker := range usersResources {
		result[i] = tracker.GetUserResourceUsageDAOInfo()
	}
```

brandboat: Already addressed comments, thank you for everyone's valuable suggestions !

pkg/webservice/handlers_test.go
```diff
@@ -1500,7 +1500,7 @@ func TestUsersAndGroupsResourceUsage(t *testing.T) {
 	assert.NilError(t, err, "Get Groups Resource Usage Handler request failed")
 
 	var groupsResourceUsageDao []*dao.GroupResourceUsageDAOInfo
-	var expGroupsResourceUsageDao []*dao.GroupResourceUsageDAOInfo
+	expGroupsResourceUsageDao := []*dao.GroupResourceUsageDAOInfo{}
```
chia7712, line 1503: pardon me, is this change related to PR? or it is a potential issue?

chia7712 (approved):
overall +1, one small question left.

pkg/webservice/handlers_test.go
brandboat: Yes it's related. Since the response from the code you mentioned is now empty array instead of nil as we changed the behavior in this pr.

IMO, the test should be adjusted as follows: set up a scenario where the group resource usage is not empty, ensuring that there is content in groupsResourceUsageDao. Otherwise, it appears to be duplicating the functionality of TestEmptyGroupsResourceUsageHandler.

chia7712: > IMO, the test should be adjusted as follows: set up a scenario where the group resource usage is not empty, ensuring that there is content in groupsResourceUsageDao. Otherwise, it appears to be duplicating the functionality of TestEmptyGroupsResourceUsageHandler.

that makes sense to me. Please make sure the test is NOT duplicate to `TestEmptyGroupsResourceUsageHandler`

manirajv06: @brandboat Empty response is one of the basic test cases while hitting those API's. So, can we add the test related changes in `TestUsersAndGroupsResourceUsage` itself instead of two new & separate test methods?

brandboat: Already addressed comments in the latest commit, many thanks !

pkg/webservice/handlers_test.go
```diff
@@ -1405,7 +1422,7 @@ func TestFullStateDumpPath(t *testing.T) {
 }
 
 func TestSpecificUserAndGroupResourceUsage(t *testing.T) {
-	prepareUserAndGroupContext(t)
+	prepareUserAndGroupContext(t, configDefault)
```
chia7712, line 1425: (this is not related to this PR)
it seems the `configDefault` does not include any group/user, so all asserts in this test case are related to "nonexistent" case, right?

pbacsko (approved):
+1 LGTM

pkg/webservice/handlers_test.go
brandboat: > it seems the configDefault does not include any group/user, so all asserts in this test case are related to "nonexistent" case, right?

You're correct, except for the test case under comment `// Test existed user query`. I believe the test case mentioned in that comment is missing an assertion to verify the existence of the user resource `testuser`.

chia7712: That is unrelated to this PR, so I file ticket (https://issues.apache.org/jira/browse/YUNIKORN-2272) to trace it.

chia7712 (approved):
LGTM

pkg/webservice/handlers_test.go
brandboat: Thanks, already take over the ticket, will come up with another pr ASAP.

PR #757, opened 2023-12-12
[YUNIKORN-2249] Add compression option to getQueueApplicaion API

targetoee:
### What is this PR for?
When the cluster has many pods, the API response size can be quite large (e.g. 50,000 pod creating a 30MB response in getQueueApplications). It causes long response time due to data transmission.
So this PR add a compression option to getQueueApplications API. If user specifies the 'Accept-Encoding' header as gzip, then the scheduler will compress the data before sending back to user.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
The compression option can be applied to other APIs facing similar response size issues.

### What is the Jira issue?
[YUNIKORN-2249](https://issues.apache.org/jira/browse/YUNIKORN-2249)

local build

N/A

### Questions:
N/A

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -19,6 +19,8 @@
 package webservice
 
 import (
+	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -612,6 +614,33 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 	appsDao := make([]*dao.ApplicationDAOInfo, 0)
 	for _, app := range queue.GetCopyOfApps() {
 		appsDao = append(appsDao, getApplicationDAO(app))
+	}
+
+	if strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
+		w.Header().Set("Content-Encoding", "gzip")
+
+		response, err := json.Marshal(appsDao)
+		if err != nil {
+			buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		var comp bytes.Buffer
+		writer := gzip.NewWriter(&comp)
+		_, err = writer.Write(response)
+		if err != nil {
+			buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		err = writer.Close()
+		if err != nil {
+			buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		if _, err = w.Write(comp.Bytes()); err != nil {
+			buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+		}
+		return
 	}
 
 	if err := json.NewEncoder(w).Encode(appsDao); err != nil {


pkg/webservice/handlers.go
```diff
@@ -614,6 +616,33 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 		appsDao = append(appsDao, getApplicationDAO(app))
 	}
 
+	if strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
```
chia7712, line 619: Could we have test for gzip header?

manirajv06: Can we try to make it modular? We already thinking of doing it in two more places. We should be able to use the method wherever required.

pbacsko: 1. Test is missing
2. Pls extract all headers to a `const` section, it's better for maintainability. We have string literals all over the place, eg. `writeHeaders()`. I don't think we need to import a separate library for this, but at least have a single place where all of them are defined. Also do the same for MIME types.

pbacsko (changes requested):
Test + small refactor

pkg/webservice/handlers.go
wilfred-s: By modular I mean create two functions. One that tests for the header content see other review point and one that does all the compress work and takes two parameters and returns nothing. We can use that compress function anywhere.
```
func checkHeader(h http.Header, key, value string) bool
func compress(w http.ResponseWriter, v any)
```
Need to file a follow up jira when we add this to add support for all REST endpoints to support compressed data...

wilfred-s (changes requested):

targetoee:
Thanks @chia7712 @pbacsko @wilfred-s for the advice. 
I've adjust the function so that it can be reused now. The test case is added also.
Please take a look.

pkg/webservice/handlers.go
```diff
@@ -746,6 +752,11 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 		appsDao = append(appsDao, getApplicationDAO(app))
 	}
 
+	if checkHeader(r.Header, "Content-Encoding", "gzip") {
```
chia7712, line 755: This PR aims to enhance only `getQueueApplication`. However, I'd like to open the room to discuss whether we should bring this enhancement to all restful APIs. @targetoee WDYT?

targetoee: That's a good idea. It can be an optional choice for user to provide more flexibility.
Some APIs typically don't return large amounts of data in common cases, so it may be necessary to discuss which ones require this functionality.

chia7712: > it may be necessary to discuss which ones require this functionality.

The improvement you propose is a kind of infra to our Restful APIs, so I prefer to bring such benefit to all APIs if there is no obvious side-effect or cost.

wilfred-s: That is exactly why I would make the two functions so we can easily apply it to any REST end points. Anything that sends more than a single IP packet as the response can benefit. 
For the streaming API, which uses really small messages that fit in a single IP packet, compressing might be more overhead than the gains we get so that one might not be a candidate everything else is.

pkg/webservice/handlers.go
```diff
@@ -1216,3 +1227,40 @@ func getStream(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 }
+
+func checkHeader(h http.Header, key string, value string) bool {
```
chia7712, line 1231: Should we return error if users use unsupported compression?

chia7712 (commented):
I leave two comments below. overall LGTM. We should have some follow-up to enhance it.

I also run YK with this patch, and all LGTM

pkg/webservice/handlers.go
```diff
@@ -1216,3 +1227,40 @@ func getStream(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	var compressedData bytes.Buffer
+	writer := gzip.NewWriter(&compressedData)
+	_, err = writer.Write(response)
+	if err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
targetoee: Done

pkg/webservice/handlers.go
targetoee: That may be more clearer to users and I added it.

targetoee:
I closed the writer in the missing case and returned an error if receiving an unsupported compression type request.
Also, the compression step would be ignored in case of the data size is smaller than MTU. This can prevent compression in unnecessary case.

pkg/webservice/handlers.go
targetoee: This may not be necessary. Skipping it is a solution when dealing with unsupported compression types in requests. Do you think it's essential for users to require this one?

targetoee:
Some adjust based on the review.
Additionally, the compression step will be skipped if the data size is smaller than MTU. This can prevent unnecessary compression.
The check of header is corrected also (in previous commit, it use 'Content-Encoding' to determine whether the request is asking for compression, but it's not the typical usage of this header).

pkg/webservice/handlers.go
```diff
@@ -1216,3 +1227,53 @@ func getStream(w http.ResponseWriter, r *http.Request) {
...
+
+func compress(w http.ResponseWriter, data any) {
+	response, err := json.Marshal(data)
+	if err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// don't compress the data if it is smaller than MTU size
+	if len(response) < 1500 {
```
craigcondit, line 1253: This is really arbitrary. Not every interface uses 1500 (and this is really only at the ethernet level anyway). Practically speaking, 64K is at least the max packet size for IPv4. We should really just focus on endpoints which can produce large outputs and not speculatively encode first. Also, we should be compressing the stream, rather than writing the JSON data to a byte array first. This can get very large on big results and use excessive RAM on the server. Instead, we should create a helper function that decides based on Accept-Encoding that takes the result stream and passes back a wrapped, compressed stream instead. This can then easily be called from any endpoint we want to compress. I think we should skip endpoints that only produce a single object as output (single app, summary endpoints, etc.) as these won't be large enough to benefit from compression anyway.

pkg/webservice/handlers.go
chia7712: Please take a look https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

It seems to me following the standard error can avoid the misunderstanding in the future.

craigcondit: > It seems to me following the standard error can avoid the misunderstanding in the future.

*NO*. We should never return an error to the client if it request an encoding we don't understand. `Accept-Encoding: identity` is the default, which means the identity encoding is always allowed. Therefore, if an unacceptable encoding is requested, we simply send the request uncompressed and without a `Content-Encoding: gzip` header.

chia7712: > NO. We should never return an error to the client if it request an encoding we don't understand. Accept-Encoding: identity is the default, which means the identity encoding is always allowed. Therefore, if an unacceptable encoding is requested, we simply send the request uncompressed and without a Content-Encoding: gzip header.

That is an acceptable way to me, but I'd like to have more discussion for my own education :)

Should we support full representation of `Accept-Encoding` ( `weight` and `coding` )? If yes, we need to consider the `Accept-Encoding: gzip;q=1.0, identity; q=0`.

Or we ignore the `weight` and only check the existence of `gzip` from the `Accept-Encoding`. This is the solution adopted by this PR.

craigcondit: There's a common saying in network protocol design: Be liberal in what you accept, strict in what you produce. In other words, we can get away with just checking for the substring `gzip` in `Accept-Encoding`, and produce *exactly* `Content-Encoding: gzip` in that case. If we choose not to compress due to size, or gzip was not requested, then we use the standard identity version. Weights are not really necessary; yes, they are part of the spec, but the client is only giving its preference; we do not have to honor it.

craigcondit: I'd like to see a very simple function like: `GetCompressedWriter(headers, writer) (writer)` that checks for the gzip header and wraps the given writer with a gzip-compressed one, else returns the original writer. Then in any endpoint we want to (potentially compress), we just replace our writer with that one instead.

chia7712:
@targetoee please take a look at https://issues.apache.org/jira/browse/YUNIKORN-2499 for the failed test

pkg/webservice/handlers.go
```diff
@@ -59,6 +60,7 @@ const (
 	GroupNameMissing         = "Group name is missing"
 	ApplicationDoesNotExists = "Application not found"
 	NodeDoesNotExists        = "Node not found"
+	UnsupportedCompType      = "Compression type not support"
```
pbacsko, line 63: nit: "not supported"

pkg/webservice/handlers.go
```diff
@@ -1216,3 +1226,38 @@ func getStream(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func compress(w http.ResponseWriter, data any) {
+	response, err := json.Marshal(data)
+	if err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	writer := gzip.NewWriter(w)
```
pbacsko, line 1247: close the writer with a defer call eg. (`defer writer.Close()` or similar)

pbacsko (changes requested):
Two minor nits

pkg/webservice/handlers.go
craigcondit: This shouldn't be here at all. As I've repeatedly indicated we should never fail with an error, just fall back to no compression.

pkg/webservice/handlers.go
wilfred-s: That will cause a leak as the gzip writer must be closed for it not to leak. Calling close on the http.ResponseWriter is not possible so we need more. Probably the easiest solution is to use the same solution as we have for the loggingHandler(). We wrap the compression choice in a handler function, which then gets wrapped in the logging handler. That means we have it all in one place and expand on it with compressor pooling or other things in the future.

Example code, which is not complete but gives some idea on how we can close the compressor. That can be expanded to use a sync.Pool to not recreate the
```
type gzipResponseWriter struct {
	io.Writer
	http.ResponseWriter
}

func (w gzipResponseWriter) Write(b []byte) (int, error) {
	return w.Writer.Write(b)
}

func makeGzipHandler(fn http.HandlerFunc) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		if !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
			fn(w, r)
			return
		}
		w.Header().Set("Content-Encoding", "gzip")
                w.Header().Del("Content-Length")
		gz := gzip.NewWriter(w)
		defer gz.Close()
		gzr := gzipResponseWriter{Writer: gz, ResponseWriter: w}
		fn(gzr, r)
	}
}
```

pbacsko:
@targetoee is this still a draft?

pkg/webservice/handlers.go
targetoee: Thanks to everyone for all the advice. I have got some questions about what @wilfred-s said. 
In the provided example, the compress handler would be wrapped within the logging handler. So we first see which API is being called, and then decide whether to use the gzip handler, since we haven't decided to compress all APIs yet? Is there any misunderstanding?

targetoee:
> @targetoee is this still a draft?
Yes, there is still something which needs to be modified. I will change the tag after the modification.

pkg/webservice/handlers.go
targetoee: Thanks for pointing out the defect and that's really helpful. I've changed it and still adjust the function.

pbacsko:
> > @targetoee is this still a draft?
> 
> Yes, there is still something which needs to be modified. I will change the tag after the modification.

A draft means that you don't want the changes to be reviewed just yet because modifications are still being made. We're already reviewing this PR.

pkg/webservice/handlers_test.go
```diff
@@ -2602,3 +2607,82 @@ func NewResponseRecorderWithDeadline() *ResponseRecorderWithDeadline {
...
+
+	m := NewWebApp(schedulerContext, nil)
+	m.StartWebApp()
+	defer func() {
+		err = m.StopWebApp()
+		assert.NilError(t, err, "Error when closing webapp service.")
+	}()
+
+	err = common.WaitFor(500*time.Millisecond, 2*time.Second, func() bool {
+		conn, connErr := net.DialTimeout("tcp", net.JoinHostPort("127.0.0.1", "9080"), time.Second)
```
pbacsko, line 2628: It's safer to immediately call `conn.Close()` if possible because the connection is not needed (this is also more readable):
```
conn, connErr := net.DialTimeout("tcp", net.JoinHostPort("127.0.0.1", "9080"), time.Second)
if conn != nil {
  conn.Close()
}
return connErr == nil
```

pkg/webservice/handlers_test.go
```diff
@@ -2602,3 +2607,82 @@ func NewResponseRecorderWithDeadline() *ResponseRecorderWithDeadline {
 		ResponseRecorder: httptest.NewRecorder(),
 	}
 }
+
+func TestCompressGetQueueApplicationAPI(t *testing.T) {
```
chia7712, line 2611: As we apply the compression to all restful APIs by changing the `webservice` directly, it seems we should add test for `webservice` instead of verifying all APIs here. For example: we can modify `newRouter` to make it accept custom routes, and then we write some dumb routes in testing to check the gzip compression.

pkg/webservice/handlers_test.go
```diff
@@ -2602,3 +2607,82 @@ func NewResponseRecorderWithDeadline() *ResponseRecorderWithDeadline {
...
+	var buf io.ReadWriter
+	req, err := http.NewRequest("GET", u.String(), buf)
+	assert.NilError(t, err, "Create new http request failed.")
+	req.Header.Set("Accept", "application/json")
+	// prevent http.DefaultClient from automatically adding gzip header
+	req.Header.Set("Accept-Encoding", "deflate")
+	resp, err := http.DefaultClient.Do(req)
+	assert.NilError(t, err, "Request failed to send.")
+	data, err := io.ReadAll(resp.Body)
+	assert.NilError(t, err, "Failed when reading data.")
```
pbacsko, line 2653: nit: "Failed to read body"

pkg/webservice/handlers_test.go
```diff
@@ -2602,3 +2607,82 @@ func NewResponseRecorderWithDeadline() *ResponseRecorderWithDeadline {
...
+
+	// request without gzip compression
+	var buf io.ReadWriter
+	req, err := http.NewRequest("GET", u.String(), buf)
+	assert.NilError(t, err, "Create new http request failed.")
+	req.Header.Set("Accept", "application/json")
+	// prevent http.DefaultClient from automatically adding gzip header
+	req.Header.Set("Accept-Encoding", "deflate")
+	resp, err := http.DefaultClient.Do(req)
+	assert.NilError(t, err, "Request failed to send.")
```
pbacsko, line 2651: nit: "Http request failed"

pkg/webservice/handlers_test.go
```diff
@@ -2602,3 +2607,82 @@ func NewResponseRecorderWithDeadline() *ResponseRecorderWithDeadline {
...
+	assert.NilError(t, err, "Failed when reading data.")
+	defer resp.Body.Close()
+	var originalResp []*dao.ApplicationDAOInfo
+	err = json.Unmarshal(data, &originalResp)
+	assert.NilError(t, err, unmarshalError)
+
+	// request with gzip compression enabled
+	req.Header.Set("Accept-Encoding", "gzip")
+	resp2, err := http.DefaultClient.Do(req)
+	assert.NilError(t, err, "Request failed to send.")
```
pbacsko, line 2662: nit: "Http request failed"

pkg/webservice/handlers_test.go
```diff
@@ -2602,3 +2607,82 @@ func NewResponseRecorderWithDeadline() *ResponseRecorderWithDeadline {
...
+	gzipReader, err := gzip.NewReader(resp2.Body)
+	assert.NilError(t, err, "Failed to create gzip reader.")
+	data2, err := io.ReadAll(gzipReader)
+	assert.NilError(t, err, "Failed when reading data.")
```
pbacsko, line 2666: nit: "Failed to read body"

pkg/webservice/handlers_test.go
```diff
@@ -2602,3 +2607,82 @@ func NewResponseRecorderWithDeadline() *ResponseRecorderWithDeadline {
...
+	err = json.Unmarshal(data, &originalResp)
+	assert.NilError(t, err, unmarshalError)
+
+	// request with gzip compression enabled
+	req.Header.Set("Accept-Encoding", "gzip")
+	resp2, err := http.DefaultClient.Do(req)
+	assert.NilError(t, err, "Request failed to send.")
+	gzipReader, err := gzip.NewReader(resp2.Body)
+	assert.NilError(t, err, "Failed to create gzip reader.")
+	data2, err := io.ReadAll(gzipReader)
```
pbacsko, line 2665: nit: ˛`compressedData` instead of  `data2`

pkg/webservice/handlers_test.go
pbacsko: +1 for the dummy route

pbacsko (changes requested):
Some minor nits

pkg/webservice/handlers_test.go
targetoee: The test about `webservice` has been added. I don't think we should modify `newRouter` just for the test case, so I kept it as it is. Instead, I simulate the action in `webservice` and then check whether the result is correct.

pkg/webservice/handlers_test.go
targetoee: Thanks for the advice. Done.

pkg/webservice/handlers_test.go
targetoee: Done (in corresponding part)

pkg/webservice/handlers_test.go
targetoee: Done

pkg/webservice/handlers_test.go
targetoee: Done

pkg/webservice/handlers_test.go
targetoee: Done

pkg/webservice/handlers_test.go
targetoee: Done

pkg/webservice/webservice_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/julienschmidt/httprouter"
+	"go.uber.org/zap"
+	"gotest.tools/v3/assert"
+
+	"github.com/apache/yunikorn-core/pkg/common"
+	"github.com/apache/yunikorn-core/pkg/log"
+	"github.com/apache/yunikorn-core/pkg/scheduler"
+)
+
+func TestCompressionWithDummyRoute(t *testing.T) {
```
pbacsko, line 40: Nit: you can just name this `TestCompression`

pkg/webservice/webservice_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}()
+
+	err := common.WaitFor(500*time.Millisecond, 5*time.Second, func() bool {
+		conn, connErr := net.DialTimeout("tcp", net.JoinHostPort("127.0.0.1", "9080"), time.Second)
+		if connErr == nil {
+			conn.Close()
+		}
+		return connErr == nil
+	})
+	assert.NilError(t, err, "Ｗeb app failed to start in 2 seconds.")
```
pbacsko, line 74: nit: "2 seconds" -> "5 seconds"

pbacsko (changes requested):
2 small comments from me.

pkg/webservice/webservice_test.go
targetoee: Done. Thanks for point out that.

pkg/webservice/webservice_test.go
```diff
@@ -0,0 +1,125 @@
...
+		Path:   "/ws/v1/helloWorld",
+	}
+
+	// request without gzip compression
+	var buf io.ReadWriter
+	req, err := http.NewRequest("GET", u.String(), buf)
+	assert.NilError(t, err, "Create new http request failed.")
+	req.Header.Set("Accept", "application/json")
+
+	// prevent http.DefaultClient from automatically adding gzip header
```
chia7712, line 88: We can merge those code into a helper function. For example:
```go
	check := func(compression bool) {
		req, err := http.NewRequest("GET", u.String(), nil)
		assert.NilError(t, err, "Create new http request failed.")
		req.Header.Set("Accept", "application/json")
		if compression {
			req.Header.Set("Accept-Encoding", "gzip")
		} else {
			req.Header.Set("Accept-Encoding", "deflate")
		}

		resp, err := http.DefaultClient.Do(req)
		assert.NilError(t, err, "Http request failed.")
		defer resp.Body.Close()

		var reader io.Reader
		if compression {
			gzipReader, err := gzip.NewReader(resp.Body)
			assert.NilError(t, err, "Failed to create gzip reader.")
			defer gzipReader.Close()
			reader = gzipReader
		} else {
			reader = resp.Body
		}
		byteArr, err := io.ReadAll(reader)
		assert.NilError(t, err, "Failed to read body.")

		var respMsg map[string]string
		err = json.Unmarshal(byteArr, &respMsg)
		assert.Equal(t, len(respMsg), 1)
		assert.Equal(t, respMsg["data"], "hello world")
	}

	check(false)
	check(true)
```

chia7712 (commented):
@targetoee thanks for updated PR. two major comments are left.

pkg/webservice/webservice.go
```diff
@@ -43,7 +46,7 @@ type WebService struct {
 func newRouter() *httprouter.Router {
 	router := httprouter.New()
 	for _, webRoute := range webRoutes {
-		handler := loggingHandler(webRoute.HandlerFunc, webRoute.Name)
+		handler := gzipHandler(loggingHandler(webRoute.HandlerFunc, webRoute.Name))
```
craigcondit, line 49: I don't think we should unconditionally GZIP every response, only those that might be large. I'd suggest adding a "Compress" boolean to the route object in routes.go and use that to determine whether to add the gzip handler.

craigcondit: For example, none of the /debug routes should have it. It should probably not be on anything that typically results in < 10 KB of data.

PR #758, opened 2023-12-13
[YUNIKORN-2209] Remove limit checks in QueueTracker

manirajv06:
### What is this PR for?
If there is no limit configured for user but wild card has some limit, then use that limit for all users permanently till next config change. Since limit has been defined in corresponding user queue tracker object, there is no need to wild card checks during headroom, canrunapp and increase calls etc thereby increases the performance of critical path. While clearing old configs, in case wild card user limit has been removed, ensure all users used that wild card user limit settings has reset to its original value. Introduced "useWildCard" bool flag in queue tracker to handle this.

### What type of PR is it?
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2209

Existing unit tests should continue to run as usual without any failures.
Added new unit test to cover this change.

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/group_tracker_test.go
+++ b/pkg/scheduler/ugm/group_tracker_test.go
@@ -32,7 +32,7 @@ func TestGTIncreaseTrackedResource(t *testing.T) {
 	// root->parent->child1->child12
 	// root->parent->child2
 	// root->parent->child12 (similar name like above leaf queue, but it is being treated differently as similar names are allowed)
-	GetUserManager()
+	manager := GetUserManager()
 	user := &security.UserGroup{User: "test", Groups: []string{"test"}}
 	groupTracker := newGroupTracker(user.User)
 
@@ -40,6 +40,8 @@ func TestGTIncreaseTrackedResource(t *testing.T) {
 	if err != nil {
 		t.Errorf("new resource create returned error or wrong resource: error %t, res %v", err, usage1)
 	}
+
+	manager.Headroom(queuePath1, TestApp1, *user)
 	result := groupTracker.increaseTrackedResource(hierarchy1, TestApp1, usage1, user.User)
 	if !result {
 		t.Fatalf("unable to increase tracked resource: queuepath %+q, app %s, res %v", hierarchy1, TestApp1, usage1)
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -311,6 +311,9 @@ func (m *Manager) UpdateConfig(config configs.QueueConfig, queuePath string) err
 	// compare existing config with new configs stored in above temporary maps
 	m.clearEarlierSetLimits(userLimits, groupLimits)
 
+	// compare existing wild card user config with new configs stored in above temporary maps
+	m.clearEarlierSetUserWildCardLimits(userWildCardLimitsConfig)
+
 	// switch over - replace the existing config with new configs
 	m.replaceLimitConfigs(userLimits, groupLimits, userWildCardLimitsConfig, groupWildCardLimitsConfig, configuredGroups)
 
@@ -399,6 +402,27 @@ func (m *Manager) clearEarlierSetLimits(newUserLimits map[string]map[string]*Lim
 
 	// Clear already configured limits of user for which limits have been configured before but not now
 	m.clearEarlierSetUserLimits(newUserLimits)
+}
+
+// clearEarlierSetUserWildCardLimits Traverse new wild card user config and decide whether earlier usage needs to be cleared or not
+// by comparing with the existing config. If config set earlier not now, then traverse all users, check whether wild card limit has been applied/used or not.
+// Reset earlier settings for the users only when wild card user limit has been applied.
+func (m *Manager) clearEarlierSetUserWildCardLimits(newUserWildCardLimits map[string]*LimitConfig) {
+	for queuePath := range m.userWildCardLimitsConfig {
+		hierarchy := strings.Split(queuePath, configs.DOT)
+		// Is queue path exists?
+		if _, ok := newUserWildCardLimits[queuePath]; !ok {
+			for u, ut := range m.userTrackers {
+				// Has wild card user limit applied for this user?
+				if m.HasWildCardApplied(u, hierarchy) {
+					log.Log(log.SchedUGM).Debug("Need to clear earlier set configs for user because wild card limit has been applied earlier",
+						zap.String("user", ut.userName),
+						zap.String("queue path", queuePath))
+					m.resetUserEarlierUsage(ut, hierarchy)
+				}
+			}
+		}
+	}
 }
 
 // clearEarlierSetUserLimits Traverse new user config and decide whether earlier usage needs to be cleared or not
@@ -629,6 +653,21 @@ func (m *Manager) CanRunApp(queuePath, applicationID string, user security.UserG
 	}
 	groupCanRunApp := groupTracker.canRunApp(hierarchy, applicationID)
 	return userCanRunApp && groupCanRunApp
+}
+
+func (m *Manager) HasWildCardApplied(user string, hierarchy []string) bool {
+	m.RLock()
+	defer m.RUnlock()
+	log.Log(log.SchedUGM).Debug("Has wild card user limit applied?",
+		zap.String("user", user),
+		zap.Strings("queue path", hierarchy))
+	userTracker := m.GetUserTracker(user)
+	if userTracker == nil {
+		log.Log(log.SchedUGM).Error("user tracker must be available in userTrackers map",
+			zap.String("user", user))
+		return false
+	}
+	return userTracker.hasWildCardApplied(hierarchy)
 }
 
 // ClearUserTrackers only for tests
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -281,20 +281,6 @@ func (m *Manager) ensureGroupInternal(userGroups []string, queuePath string) str
 	return m.ensureGroupInternal(userGroups, parentPath)
 }
 
-func (m *Manager) isUserRemovable(ut *UserTracker) bool {
-	if len(ut.getTrackedApplications()) == 0 && resources.IsZero(ut.queueTracker.resourceUsage) {
-		return true
-	}
-	return false
-}
-
-func (m *Manager) isGroupRemovable(gt *GroupTracker) bool {
-	if len(gt.getTrackedApplications()) == 0 && resources.IsZero(gt.queueTracker.resourceUsage) {
-		return true
-	}
-	return false
-}
-
 func (m *Manager) UpdateConfig(config configs.QueueConfig, queuePath string) error {
 	userWildCardLimitsConfig := make(map[string]*LimitConfig)
 	groupWildCardLimitsConfig := make(map[string]*LimitConfig)
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -298,7 +298,10 @@ func (m *Manager) UpdateConfig(config configs.QueueConfig, queuePath string) err
 	m.clearEarlierSetLimits(userLimits, groupLimits)
 
 	// compare existing wild card user config with new configs stored in above temporary maps
-	m.clearEarlierSetUserWildCardLimits(userWildCardLimitsConfig)
+	m.clearEarlierSetUserWildCardLimits(userWildCardLimitsConfig, userLimits)
+
+	// apply wild card user limits to all existing users for which no limits configured explicitly
+	m.applyWildCardUserLimits(userWildCardLimitsConfig, userLimits)
 
 	// switch over - replace the existing config with new configs
 	m.replaceLimitConfigs(userLimits, groupLimits, userWildCardLimitsConfig, groupWildCardLimitsConfig, configuredGroups)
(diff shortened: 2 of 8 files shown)


pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -94,67 +97,33 @@ func (qt *QueueTracker) increaseTrackedResource(hierarchy []string, applicationI
 	}
 	finalResourceUsage := qt.resourceUsage.Clone()
 	finalResourceUsage.AddTo(usage)
-	wildCardQuotaExceeded := false
 	existingApp := qt.runningApplications[applicationID]
 
 	// apply user/group specific limit settings set if configured, otherwise use wild card limit settings
 	if qt.maxRunningApps != 0 && !resources.IsZero(qt.maxResources) {
-		log.Log(log.SchedUGM).Debug("applying enforcement checks using limit settings of specific user/group",
+		log.Log(log.SchedUGM).Debug("applying enforcement checks using limit settings",
```
pbacsko, line 104: I might be wrong, wasn't getting rid of this whole thing the main idea? As far as I can remember, we agreed that this check here is nothing more than an internal state check at this point, because maxrunning apps & resource limits are already verified at this point by `CanRunApp()` & `Headroom()` so technically (if the impl is 100% correct) we should not enter this branch.

manirajv06: We do many things between `CanRunApp()/Headroom()` and `Increase` as part of scheduling cycle. So, things could have changed in the middle, which may cause issues to one of the two different go routines crossed earlier check simultaneously.

pbacsko: "So, things could have changed in the middle" - how exactly? That would mean that invariants can be violated. I was under the impression that certain things only change from the scheduling cycle. So we deny an app to run (from `canRunApp()`) and for some reason the number of running apps still goes above the limit?

pbacsko (approved):
+1 LGTM

pkg/scheduler/ugm/group_tracker_test.go
```diff
@@ -32,14 +32,16 @@ func TestGTIncreaseTrackedResource(t *testing.T) {
...
-	GetUserManager()
+	manager := GetUserManager()
 	user := &security.UserGroup{User: "test", Groups: []string{"test"}}
 	groupTracker := newGroupTracker(user.User)
 
 	usage1, err := resources.NewResourceFromConf(map[string]string{"mem": "10M", "vcore": "10"})
 	if err != nil {
 		t.Errorf("new resource create returned error or wrong resource: error %t, res %v", err, usage1)
 	}
+
```
wilfred-s, line 43: Explain why we need this

pkg/scheduler/ugm/manager.go
```diff
@@ -401,6 +390,27 @@ func (m *Manager) clearEarlierSetLimits(newUserLimits map[string]map[string]*Lim
 	m.clearEarlierSetUserLimits(newUserLimits)
 }
 
+// clearEarlierSetUserWildCardLimits Traverse new wild card user config and decide whether earlier usage needs to be cleared or not
+// by comparing with the existing config. If config set earlier not now, then traverse all users, check whether wild card limit has been applied/used or not.
+// Reset earlier settings for the users only when wild card user limit has been applied.
+func (m *Manager) clearEarlierSetUserWildCardLimits(newUserWildCardLimits map[string]*LimitConfig) {
+	for queuePath := range m.userWildCardLimitsConfig {
+		hierarchy := strings.Split(queuePath, configs.DOT)
```
wilfred-s, line 398: This should be inside the if `newUserWildCardLimits`

pkg/scheduler/ugm/manager.go
```diff
@@ -401,6 +390,27 @@ func (m *Manager) clearEarlierSetLimits(newUserLimits map[string]map[string]*Lim
...
+		// Is queue path exists?
+		if _, ok := newUserWildCardLimits[queuePath]; !ok {
+			for u, ut := range m.userTrackers {
+				// Has wild card user limit applied for this user?
+				if m.HasWildCardApplied(u, hierarchy) {
```
wilfred-s, line 403: Here we find out if it has a wildcard applied and then in line 407 we do the work again to remove the settings. Why are we not just doing the reset directly?

pkg/scheduler/ugm/manager.go
```diff
@@ -401,6 +390,27 @@ func (m *Manager) clearEarlierSetLimits(newUserLimits map[string]map[string]*Lim
...
+					log.Log(log.SchedUGM).Debug("Need to clear earlier set configs for user because wild card limit has been applied earlier",
```
wilfred-s, line 404: This should be printed inside `resetUserEarlierUsage()` when we do the work
btw: the name of the function is not correct we do not reset usage we reset the limit enforced.

pkg/scheduler/ugm/manager.go
```diff
@@ -631,6 +641,21 @@ func (m *Manager) CanRunApp(queuePath, applicationID string, user security.UserG
 	return userCanRunApp && groupCanRunApp
 }
 
+func (m *Manager) HasWildCardApplied(user string, hierarchy []string) bool {
```
wilfred-s, line 644: This is only used in the clear case but I think we can fold this into the clear itself and only traverse the hierarchy once.
I think that would remove `HasWildCardApplied()`,  `ut.hasWildCardApplied()` and `qt.hasWildCardApplied()`

pkg/scheduler/ugm/manager_test.go
```diff
@@ -307,28 +304,19 @@ func TestUpdateConfig(t *testing.T) {
...
 	// configure max resource for root.parent lesser than current resource usage. should be allowed to set but user cannot be allowed to do any activity further
 	conf = createConfig(user.User, user.Groups[0], "memory", "50", 40, 4)
 	err = manager.UpdateConfig(conf.Queues[0], "root")
 	assert.NilError(t, err)
-	increased := manager.IncreaseTrackedResource(queuePath1, TestApp1, usage, user)
-	if increased {
-		t.Fatalf("unable to increase tracked resource: queuepath %s, app %s, res %v", queuePath1, TestApp1, user)
-	}
+	headroom := manager.Headroom(queuePath1, TestApp1, user)
+	assert.Equal(t, headroom.FitInMaxUndef(usage), false)
```
wilfred-s, line 312: use `assert.Assert()` for boolean checks (multiple instances found)

pkg/scheduler/ugm/manager_test.go
```diff
@@ -337,10 +325,93 @@ func TestUpdateConfig(t *testing.T) {
...
+}
+
+func TestUseWildCard(t *testing.T) {
+	setupUGM()
+	manager := GetUserManager()
+	user := security.UserGroup{User: "user1", Groups: []string{"group1"}}
+
+	expectedResource, err := resources.NewResourceFromConf(map[string]string{"memory": "50", "vcores": "50"})
+	if err != nil {
```
wilfred-s, line 338: use `assert.NilError()` to check for nil error returns (multiple instances found)

pkg/scheduler/ugm/manager_test.go
```diff
@@ -1116,9 +1183,10 @@ func TestUserGroupLimit(t *testing.T) { //nolint:funlen
 			assert.NilError(t, err, fmt.Sprintf("can't create resource from %v", tc.finalExpectedHeadroomResource))
 			headroom = manager.Headroom(queuePathParent, TestApp1, tc.user)
 			assert.Equal(t, resources.Equals(headroom, finalExpectedHeadroom), true, "final headroom is not expected")
-
-			increased = manager.IncreaseTrackedResource(queuePathParent, TestApp2, usage, tc.user)
-			assert.Equal(t, increased, false, "should not increase tracked resource: queuepath "+queuePathParent+", app "+TestApp2+", res "+usage.String())
+			if manager.CanRunApp(queuePathParent, TestApp2, tc.user) {
```
wilfred-s, line 1186: use assert to check return of `CanRunApp()` and fail test if not good

pkg/scheduler/ugm/manager_test.go
```diff
@@ -1425,8 +1493,10 @@ func TestUserGroupLimitChange(t *testing.T) { //nolint:funlen
 			conf.Queues[0].Queues[0].Limits = tc.newLimits
 			assert.NilError(t, manager.UpdateConfig(conf.Queues[0], "root"))
 
-			increased = manager.IncreaseTrackedResource(queuePathParent, TestApp2, usage, tc.user)
-			assert.Equal(t, increased, false, "should not increase tracked resource: queuepath "+queuePathParent+", app "+TestApp2+", res "+usage.String())
+			if manager.CanRunApp(queuePathParent, TestApp2, tc.user) {
```
wilfred-s, line 1496: use assert to check return of `CanRunApp()` and fail test if not good

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -56,6 +57,7 @@ func newQueueTracker(queuePath string, queueName string) *QueueTracker {
 		maxResources:        nil,
 		maxRunningApps:      0,
 		childQueueTrackers:  make(map[string]*QueueTracker),
+		useWildCard:         false,
```
wilfred-s, line 60: boolean zero  value is false so no need to set it explicitly

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -266,6 +209,15 @@ func (qt *QueueTracker) headroom(hierarchy []string, trackType trackingType) *re
 		if trackType == user {
 			if config := m.getUserWildCardLimitsConfig(qt.queuePath); config != nil {
 				headroom = config.maxResources.Clone()
+				log.Log(log.SchedUGM).Debug("Use wild card limit settings as there is no limit set explicitly",
```
wilfred-s, line 212: Wildcard use implicitly means no limit set. No need to add that to the message.
We also should only need this check as part of the create of a new queue tracker. After that the config change should be setting or removing the value. We should not need it at all at this point.

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -448,8 +400,19 @@ func (qt *QueueTracker) canRunApp(hierarchy []string, applicationID string, trac
 		} else if trackType == group {
 			config = m.getGroupWildCardLimitsConfig(qt.queuePath)
 		}
-		if config != nil && config.maxApplications != 0 && running > int(config.maxApplications) {
-			return false
+		if config != nil && config.maxApplications != 0 {
```
wilfred-s, line 403: See above comment we should not do this as part of the headroom check. We do it once as part of the queue tracker create and then depend on the config update to maintain it

wilfred-s (changes requested):

pkg/scheduler/ugm/manager.go
```diff
@@ -631,6 +641,21 @@ func (m *Manager) CanRunApp(queuePath, applicationID string, user security.UserG
...
+	defer m.RUnlock()
+	log.Log(log.SchedUGM).Debug("Has wild card user limit applied?",
+		zap.String("user", user),
+		zap.Strings("queue path", hierarchy))
+	userTracker := m.GetUserTracker(user)
+	if userTracker == nil {
+		log.Log(log.SchedUGM).Error("user tracker must be available in userTrackers map",
+			zap.String("user", user))
+		return false
+	}
```
pbacsko, line 655: This branch is easy to test, let's add test for it to enhance coverage.

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -480,3 +443,22 @@ func (qt *QueueTracker) canBeRemovedInternal() bool {
...
+
+func (qt *QueueTracker) hasWildCardApplied(hierarchy []string) bool {
+	// depth first: all the way to the leaf, child queue tracker should exist
+	// more than 1 in the slice means we need to recurse down
+	if len(hierarchy) > 1 {
+		childName := hierarchy[1]
+		if qt.childQueueTrackers[childName] == nil {
+			log.Log(log.SchedUGM).Error("Child queueTracker tracker must be available in child queues map",
+				zap.String("child queueTracker name", childName))
+			return false
```
pbacsko, line 455: Add test for this branch to make sure we don't end up with a panic.

pkg/scheduler/ugm/group_tracker_test.go
manirajv06: As discussed, we can take this up in a follow up jira as there are lot more places to cover

pkg/scheduler/ugm/manager.go
manirajv06: At times, config may not even have a single limit settings. So, we cannot depend on len of newUserWildCardLimits

pkg/scheduler/ugm/manager.go
manirajv06: Taken care

manirajv06:
Earlier commit takes care of Wild card user limit removal case only. Recent commit covered two more cases:

1. Wild card user limit updated, but not removed:

In this case, all existing users using the wild card limits needs to be updated.

2. Wild card user limit added now:

In this case, all existing users (who don't have explicit limit config) limit settings needs to be updated.

pkg/scheduler/ugm/manager.go
manirajv06: Filed [Fix codecoverage warnings in UGM](https://issues.apache.org/jira/browse/YUNIKORN-2307) to address this.

pbacsko (changes requested):
I found two smaller things, please check those.

Also if you addressed Wilfred's comments or created a follow-up JIRA, please close his comments to see that everything is up to date.

pbacsko (approved):
+1

PR #759, opened 2023-12-16
[YUNIKORN-2270] GPU Preemption is not triggered as expected when all available GPUs are used.

yangwwei:
### What is this PR for?
I am testing GPU preemption scenarios, and in the following scenario, the preemption isn't happening as expected. 

queue structure is pretty simple:

```
root
 - a
   min=100
   max=300
 - b
   min=0
   max=300
```
the cluster has a total of 300 GPUs available, no autoscaling. Reproducing steps:

1. Create 600 pods in `root.b queue`, each needs 1 GPU. This will consume all 300 GPUs available in the cluster, and 300 pods pending
2. Create 100 pods in `root.a` queue, each needs 1 GPU. The expectation is queue `a` will preempt 100 GPU from queue `b` reach the guarantee.

observation: a small number of pods preempted resources from queue `b` got started on queue `a`, the result is not stable. it could not reach guaranteed resources.

When I debug, I found in these lines: https://github.com/apache/yunikorn-core/blob/620687afe10638d3e191edffbc81959985abbbb4/pkg/scheduler/objects/preemption.go#L576-L587. When it marks an ask as triggered preemption, when the next time doing `tryPreemption()`, it will be skipped. However, when the queue headroom is enough, in my case, because all GPUs are utilized, the headroom for GPU is 0, it skipped to create the reservation. Then even one pod gets evicted, the pod initiated this preemption cannot get the spot.

Not sure why we need the check here, when I removed it and always return a reservation, it works much better and much stable now.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
N/A

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2270

### DIFF
--- a/pkg/scheduler/objects/application_test.go
+++ b/pkg/scheduler/objects/application_test.go
@@ -1871,7 +1871,7 @@ func TestTryAllocatePreemptQueue(t *testing.T) {
 	// pass the time and try again
 	ask3.createTime = ask3.createTime.Add(-30 * time.Second)
 	alloc3 = app2.tryAllocate(resources.NewResourceFromMap(map[string]resources.Quantity{"first": 0}), true, 30*time.Second, &preemptionAttemptsRemaining, iterator, iterator, getNode)
-	assert.Assert(t, alloc3 == nil, "alloc3 not expected")
+	assert.Assert(t, alloc3 != nil && alloc3.result == Reserved, "alloc3 should be a reservation")
 	assert.Assert(t, alloc2.IsPreempted(), "alloc2 should have been preempted")
 }
 
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -573,20 +573,11 @@ func (p *Preemptor) TryPreemption() (*Allocation, bool) {
 		"preempting allocations to free up resources to run ask: "+p.ask.GetAllocationKey())
 
 	// reserve the selected node for the new allocation if it will fit
-	if p.headRoom.FitInMaxUndef(p.ask.GetAllocatedResource()) {
-		log.Log(log.SchedPreemption).Info("Reserving node for ask after preemption",
-			zap.String("allocationKey", p.ask.GetAllocationKey()),
-			zap.String("nodeID", nodeID),
-			zap.Int("victimCount", len(victims)))
-		return newReservedAllocation(nodeID, p.ask), true
-	}
-
-	// can't reserve as queue is still too full, but scheduling should succeed after preemption occurs
-	log.Log(log.SchedPreemption).Info("Preempting allocations for ask, but not reserving yet as queue is still above capacity",
+	log.Log(log.SchedPreemption).Info("Reserving node for ask after preemption",
 		zap.String("allocationKey", p.ask.GetAllocationKey()),
+		zap.String("nodeID", nodeID),
 		zap.Int("victimCount", len(victims)))
-
-	return nil, true
+	return newReservedAllocation(nodeID, p.ask), true
 }
 
 type predicateCheckResult struct {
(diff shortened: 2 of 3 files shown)


yangwwei:
@craigcondit please help to take a look, thanks!

zhuqi-lucas:
Good catch @yangwwei , i think it's because we allow preemption when request don't fit the headroom, so we should allow the reservation when it don't fit the headroom, thanks. Need @craigcondit @wilfred-s for double check.
```
// resource must fit in headroom otherwise skip the request (unless preemption could help)
			// attempt preemption
			if allowPreemption && *preemptAttemptsRemaining > 0 {
				*preemptAttemptsRemaining--
				fullIterator := fullNodeIterator()
				if fullIterator != nil {
					if alloc, ok := sa.tryPreemption(headRoom, preemptionDelay, request, fullIterator, false); ok {
						// preemption occurred, and possibly reservation
						return alloc
					}
				}
			}
			sa.appEvents.sendAppDoesNotFitEvent(request, headRoom)
			continue
		}
```

zhuqi-lucas (approved):

wilfred-s:
@manirajv06 this is what we have been discussing around preemption and reservations. It came up after the preemption discussion we had with Rainie.

manirajv06:
> a small number of pods preempted resources from queue b got started on queue a

My understanding is, even small number of pods preempted here should have not happened through reserved allocation because p.headroom check would have always failed. How old is p.headroom? It holds the available resource calculated much earlier in the allocation process. Even though, some pods has been removed completely (after confirmation from shim) as part of current preemption loop, p.headroom still carries the old value. So, it is always bound to fail. I think none of the preemption would have happened through reserved allocation so far for the cases like this. By any chance, are you seeing "Reserving node for ask after preemption" in the logs before applying the fix?

manirajv06:
On further analysis, observed that ask (preemptor) acquired resources through normal preemption process may be preempted by requiredNodePreemptor to free up spaces for daemon sets. It may happen once ask (preemptor) starts running on the node. So, whole preemption effort in making up spaces for this ask (preemptor) won't be beneficial as expected. Filing a [jira](https://issues.apache.org/jira/browse/YUNIKORN-2279) for now. Need to investigate more on this.

yangwwei:
Thanks @manirajv06 . 

In my testing, without the patch, it gets me to 9 out of 20 replicas:

```
job-a-preempt-job   9/20    20           9           29m
```
with the patch, it can reach 20. It should get all these allocated because this is under the guaranteed capacity.

I saw some logs like:

```
2023-12-20T10:28:36.542-0800    INFO    core.scheduler.preemption       objects/preemption.go:585       Preempting allocations for ask, but not reserving yet as queue is still above capacity  {"allocationKey": "a3ea8a58-7594-4d11-907c-6b603ecea3ec", "victimCount": 1}
```

I have attached the log for reference.
[yk.log](https://github.com/apache/yunikorn-core/files/13731227/yk.log)

manirajv06:
I don't see any reserved allocation ""Reserving node for ask after preemption"" in the logs. I can see "Preempting allocations for ask, but not reserving yet as queue is still above capacity" for 7 allocations. As expected, none of the preemption has happened through reserved allocation process.

We can go ahead with the changes for now. [jira](https://issues.apache.org/jira/browse/YUNIKORN-2279) would improve this situation much better.

manirajv06 (approved):

Interaze:
Do you know when we'll see a release for this? Thanks!

manirajv06:
We haven't yet planned release 1.5.0 and it may take couple of months from now. If we plan 1.4.1 release in the interim, then we will include this.

PR #760, opened 2023-12-16
[YUNIKORN-2273] checkLimitMaxApplications fails if there is missing limit in a middle queue

FrankYang0529:
### What is this PR for?
The checkLimitMaxApplications function rejects config which child limit is large than parent limit. However, it doesn't cover all cases like child limit is large than grandparent limit.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2273

Covered by unit tests.

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -171,7 +171,6 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 				if !userMaxResource.FitInMaxUndef(limitMaxResources) {
 					return fmt.Errorf("user %s max resource %s of queue %s is greater than immediate or ancestor parent maximum resource %s", user, limitMaxResources.String(), cur.Name, userMaxResource.String())
 				}
-				// Override with min resource
 				users[curQueuePath][user] = resources.ComponentWiseMinPermissive(limitMaxResources, userMaxResource)
 			} else if wildcardMaxResource, ok := users[queuePath][common.Wildcard]; user != common.Wildcard && ok {
 				if !wildcardMaxResource.FitInMaxUndef(limitMaxResources) {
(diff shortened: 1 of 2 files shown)


manirajv06 (commented):
Ah ! Is this the one I had come across and explained in https://github.com/apache/yunikorn-core/pull/720#discussion_r1401827861 ? I think so :)

BTW, https://github.com/apache/yunikorn-core/pull/720 has gone in. Can you rebase?

FrankYang0529:
> Ah ! Is this the one I had come across and explained in [#720 (comment)](https://github.com/apache/yunikorn-core/pull/720#discussion_r1401827861) ? I think so :)

Yes, the case is covered by:
* leaf queue user maxapplications exceed grandparent queue wildcard maxapplications
* leaf queue group maxapplications exceed grandparent queue wildcard maxapplications

> BTW, #720 has gone in. Can you rebase?

Yes, thanks for the review.

pkg/common/configs/configvalidator.go
```diff
@@ -232,54 +232,53 @@ func checkLimitMaxApplications(cur QueueConfig, users map[string]map[string]uint
...
+	users[curQueuePath] = make(map[string]uint64)
+	groups[curQueuePath] = make(map[string]uint64)
+
+	// Carry forward (populate) the parent limit settings to the next level
+	for u, v := range users[queuePath] {
+		users[curQueuePath][u] = v
+	}
+	for g, v := range groups[queuePath] {
+		groups[curQueuePath][g] = v
+	}
```
pbacsko, line 244: Nit: without having a context, it's difficult to interpret these letters. Ok, I guess "u" = user, "g" = group, but what is "v"?

pkg/common/configs/configvalidator.go
```diff
@@ -232,54 +232,53 @@ func checkLimitMaxApplications(cur QueueConfig, users map[string]map[string]uint
...
-						users[curQueuePath] = make(map[string]uint64)
-					}
-					users[curQueuePath][user] = limitMaxApplications
+	users[curQueuePath] = make(map[string]uint64)
+	groups[curQueuePath] = make(map[string]uint64)
+
+	// Carry forward (populate) the parent limit settings to the next level
+	for u, v := range users[queuePath] {
+		users[curQueuePath][u] = v
+	}
```
pbacsko, line 241: Nit: a short comment about the possible values of "curQueuePath" and "queuePath" to help understand the code (eg "root.parent" vs "root.parent.subparent").

pbacsko (changes requested):
Two minor things about the code

pkg/common/configs/configvalidator.go
```diff
@@ -232,54 +232,53 @@ func checkLimitMaxApplications(cur QueueConfig, users map[string]map[string]uint
...
+				if userMaxApplications != 0 && (userMaxApplications < limitMaxApplications || limitMaxApplications == 0) {
+					return fmt.Errorf("user %s max applications %d of queue %s is greater than immediate or ancestor parent max applications %d", user, limitMaxApplications, cur.Name, userMaxApplications)
+				}
+				// Override with min resource
+				users[curQueuePath][user] = common.Min(limitMaxApplications, userMaxApplications)
+			} else if wildcardMaxApplications, ok := users[queuePath][common.Wildcard]; user != common.Wildcard && ok {
+				if wildcardMaxApplications != 0 && (wildcardMaxApplications < limitMaxApplications || limitMaxApplications == 0) {
+					return fmt.Errorf("user %s max applications %d of queue %s is greater than wildcard max applications %d of immediate or ancestor parent queue", user, limitMaxApplications, cur.Name, wildcardMaxApplications)
 				}
+				users[curQueuePath][user] = common.Min(limitMaxApplications, wildcardMaxApplications)
```
manirajv06, line 261: If we apply `min` here, then we would need to apply even for wild card max resources block in https://github.com/apache/yunikorn-core/blob/master/pkg/common/configs/configvalidator.go#L179 as well.

manirajv06 (changes requested):

pkg/common/configs/configvalidator.go
FrankYang0529: I change the `v` to `maxapplications` to give more context. Thanks!

pkg/common/configs/configvalidator.go
FrankYang0529: I add comment and example for `queuePath` and `curQueuePath`. Thank you!

pkg/common/configs/configvalidator.go
FrankYang0529: After double check, I think we can just use `limitMaxApplications`. The `wildcardMaxApplications` is like parent maxapplications and `limitMaxApplications` is like child, so we will have following cases:

* parent == 0 and child == 0, this case is fine to assign child value.
* parent != 0 and child == 0, throw error in line 258.
* parent != 0 and child != 0, and parent > child, it's ok to assign child value.
* parent != 0 and child != 0, and parent > child, throw error in line 258.
* parent == 0 and child != 0, it's ok to assign child value.

pbacsko:
@FrankYang0529 have you changed anything? You performed a force push but I can still see "v" variable which I commented about.
Next time, instead of rebasing, pls add a new commit (eg. "Code review") so we can see what changed exactly. Thanks.

pkg/common/configs/configvalidator_test.go
```diff
@@ -1580,3 +1825,56 @@ func TestCheckLimits(t *testing.T) { //nolint:funlen
...
+					},
+				},
+			},
+		},
+		make(map[string]map[string]uint64),
+		make(map[string]map[string]uint64),
+		common.Empty,
+	)
+	assert.Equal(t, err != nil, true, "user2 cpu maxapplications in root.parent.child should not be large than root") // this will fail
+}
```
manirajv06, line 1880: I think this single test case has been already covered in earlier method. Isn't it? Having new method just for one case can be avoided.

manirajv06 (approved):
looks good.

Please take care of minor nits.

pkg/common/configs/configvalidator_test.go
FrankYang0529: Yeah, good catch. Removed it. Thank you!

FrankYang0529:
> have you changed anything? You performed a force push but I can still see "v" variable which I commented about.

Sorry, I may revert the change locally. I address your comments in the last change. May you check it again? Thank you.

pbacsko (approved):
+1 LGTM

pkg/common/configs/configvalidator.go
```diff
@@ -236,54 +236,55 @@ func checkLimitMaxApplications(cur QueueConfig, users map[string]map[string]uint
...
-						groups[curQueuePath] = make(map[string]uint64)
-					}
-					groups[curQueuePath][group] = limitMaxApplications
+		for _, group := range limit.Groups {
+			// Is user limit setting exists?
+			if groupMaxApplications, ok := groups[queuePath][group]; ok {
+				if groupMaxApplications != 0 && (groupMaxApplications < limitMaxApplications || limitMaxApplications == 0) {
+					return fmt.Errorf("group %s max applications %d of queue %s is greater than immediate or ancestor parent max applications %d", group, limitMaxApplications, cur.Name, groupMaxApplications)
 				}
+				// Override with min resource
```
manirajv06, line 279: Can you remove this comment? In other places too..

manirajv06 (commented):
Minor nit.

Did we covered all these different possible combinations (child against parent, child against grandparent, for specific user, for wild card,  specific user against wild card etc) for max resources as well? If not, can you file a follow up jira?

pkg/common/configs/configvalidator.go
FrankYang0529: Sure, removed it. Thank you.

FrankYang0529:
> Did we covered all these different possible combinations (child against parent, child against grandparent, for specific user, for wild card, specific user against wild card etc) for max resources as well? If not, can you file a follow up jira?

Yeah, I think we already covered it in https://github.com/apache/yunikorn-core/pull/704 and https://github.com/apache/yunikorn-core/pull/720.

PR #761, opened 2023-12-23
[YUNIKORN-2259] add more meaningful error message when the invalid config is caused by undefined maxapplications

brandboat:
### What is this PR for?
add more detailed error message when validating maxApplications

### What type of PR is it?
* [x] - Improvement

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2259

covered by existing unit tests

N/A

### Questions:
N/A

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -213,10 +213,10 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 func checkQueueMaxApplications(cur QueueConfig) error {
 	var err error
 	for _, child := range cur.Queues {
-		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications) {
+		if cur.MaxApplications != 0 && cur.MaxApplications < child.MaxApplications {
 			return fmt.Errorf("parent maxApplications must be larger than child maxApplications")
 		}
-		if cur.MaxApplications != 0 && (child.MaxApplications == 0) {
+		if cur.MaxApplications != 0 && child.MaxApplications == 0 {
 			return fmt.Errorf("maxApplications is either undefined or zero, which is not allowed when parent queue's maxApplications is defined")
 		}
 		err = checkQueueMaxApplications(child)
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -213,8 +213,11 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 func checkQueueMaxApplications(cur QueueConfig) error {
 	var err error
 	for _, child := range cur.Queues {
-		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications || child.MaxApplications == 0) {
+		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications) {
 			return fmt.Errorf("parent maxApplications must be larger than child maxApplications")
+		}
+		if cur.MaxApplications != 0 && (child.MaxApplications == 0) {
+			return fmt.Errorf("maxApplications is either undefined or zero, which is not allowed when parent queue's maxApplications is defined")
 		}
 		err = checkQueueMaxApplications(child)
 		if err != nil {


pkg/common/configs/configvalidator.go
```diff
@@ -213,9 +213,12 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 func checkQueueMaxApplications(cur QueueConfig) error {
 	var err error
 	for _, child := range cur.Queues {
-		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications || child.MaxApplications == 0) {
+		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications) {
 			return fmt.Errorf("parent maxApplications must be larger than child maxApplications")
 		}
+		if cur.MaxApplications != 0 && (child.MaxApplications == 0) {
```
chia7712, line 219: how about `cur.MaxApplications != 0 && child.MaxApplications == 0`?

pkg/common/configs/configvalidator.go
```diff
@@ -213,9 +213,12 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 func checkQueueMaxApplications(cur QueueConfig) error {
 	var err error
 	for _, child := range cur.Queues {
-		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications || child.MaxApplications == 0) {
+		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications) {
```
chia7712, line 216: `cur.MaxApplications != 0 && cur.MaxApplications < child.MaxApplications`

chia7712 (commented):
@brandboat thanks for patch. a couple of small comments left. PTAL

pkg/common/configs/configvalidator.go
brandboat: oops...

pkg/common/configs/configvalidator.go
```diff
@@ -213,9 +213,12 @@ func checkLimitResource(cur QueueConfig, users map[string]map[string]*resources.
 func checkQueueMaxApplications(cur QueueConfig) error {
 	var err error
 	for _, child := range cur.Queues {
-		if cur.MaxApplications != 0 && (cur.MaxApplications < child.MaxApplications || child.MaxApplications == 0) {
+		if cur.MaxApplications != 0 && cur.MaxApplications < child.MaxApplications {
 			return fmt.Errorf("parent maxApplications must be larger than child maxApplications")
 		}
+		if cur.MaxApplications != 0 && child.MaxApplications == 0 {
+			return fmt.Errorf("maxApplications is either undefined or zero, which is not allowed when parent queue's maxApplications is defined")
+		}
```
pbacsko, line 221: Does this mean that whenever a parent has maxApps defined, all the child queues are required to have this setting too?

If that's the case, I'm -1 on this and 100% against this change. Imagine "root.dev" having maxapps with a bunch of parent queues for teams and leaf queues for two dozen developers. Once you set it, you have to define for the entire hierarchy.

pbacsko (changes requested):
-1 if I'm interpreting this correctly. We must not mandate this setting for an entire sub-hiearchy once it's defined somewhere.

pkg/common/configs/configvalidator.go
chia7712: > Does this mean that whenever a parent has maxApps defined, all the child queues are required to have this setting too?

This PR aims to add more meaningful message to inspire users how to fix configs. The behavior you described is already there, and I had filed a ticket to discuss such weird behavior. https://issues.apache.org/jira/browse/YUNIKORN-2258

pbacsko: I see. We must get rid of this then. Don't make this change to the config validator and remove the existing logic which performs the runtime checks.

pbacsko: Added my thoughts to YUNIKORN-2258.

chia7712: @pbacsko thanks for feedback. Do you approve of this PR? I'd like to merge it before YUNIKORN-2258 :)

chia7712 (approved):
LGTM

@pbacsko I will merge it tomorrow

pkg/common/configs/configvalidator.go
pbacsko: @chia7712 my stance is that we don't need this check. If we submit it, then it has to be reverted in YUNIKORN-2258. So I don't see the point of adding it.

pbacsko:
> LGTM
> 
> @pbacsko I will merge it tomorrow

See my comment in the thread. If we add it now, it has to be reverted later per YUNIKORN-2258.

pkg/common/configs/configvalidator.go
chia7712: > my stance is that we don't need this check. 

Sorry that I neglected your comment before.

> If we submit it, then it has to be reverted in YUNIKORN-2258. So I don't see the point of adding it.

The reasons I want to merge it are shown below.

1. YUNIKORN-2258 hasn't yet reached consensus, so it may still open in next release. Hence, the error message added by this PR can give more readable output when users encounter such issue.

2. The changes introduced by this PR is small, so it wouldn't burn out our servers 😀

pbacsko (approved):
+1 OK go ahead with this and we'll see if anything needs to be done later.

PR #762, opened 2023-12-25
[YUNIKORN-2257] Add rest API to retrieve node utilization for multiple resource types

chenyulin0719:
### What is this PR for?
Add new rest endpoint to return multiple resource types per partition
- New API: /ws/v1/scheduler/node-utilizations 
- Mark old API deprecated: /ws/v1/scheduler/node-utilization 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2257

1. make test should passed.
2. deploy yunikorn-core and call http://localhost:9889/ws/v1/scheduler/node-utilizations

YUNIKORN-2257 Api result v2

### Questions:
NA

### DIFF
--- a/pkg/webservice/dao/node_util.go
+++ b/pkg/webservice/dao/node_util.go
@@ -18,6 +18,12 @@
 
 package dao
 
+type PartitionNodesUtilDAOInfo struct {
+	ClusterId     string              `json:"clusterId,omitempty"`
+	Partition     string              `json:"partition,omitempty"`
+	NodesUtilList []*NodesUtilDAOInfo `json:"utilizationList,omitempty"`
+}
+
 type NodesUtilDAOInfo struct {
 	ResourceType string             `json:"type,omitempty"`
 	NodesUtil    []*NodeUtilDAOInfo `json:"utilization,omitempty"`
--- a/pkg/webservice/dao/node_util.go
+++ b/pkg/webservice/dao/node_util.go
@@ -19,9 +19,9 @@
 package dao
 
 type PartitionNodesUtilDAOInfo struct {
-	ClusterId     string              `json:"clusterId,omitempty"`
-	Partition     string              `json:"partition,omitempty"`
-	NodesUtilList []*NodesUtilDAOInfo `json:"utilizationList,omitempty"`
+	ClusterId     string              `json:"clusterId"` // no omitempty, cluster id should not be empty
+	Partition     string              `json:"partition"` // no omitempty, partition should not be empty
+	NodesUtilList []*NodesUtilDAOInfo `json:"utilizations,omitempty"`
 }
 
 type NodesUtilDAOInfo struct {
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -355,7 +355,7 @@ func getNodesDAO(entries []*objects.Node) []*dao.NodeDAOInfo {
 // for the default partition. Dominant resource is defined as the highest utilised resource
 // type on the root queue based on the registered resources.
 // Only check the default partition
-// Deprecated - To be removed in 1.7.0. Replaced with getNodesUtilisations
+// Deprecated - To be removed in next major release. Replaced with getNodesUtilisations
 func getNodeUtilisation(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	partitionContext := schedulerContext.GetPartitionWithoutClusterID(configs.DefaultPartition)
@@ -380,7 +380,7 @@ func getNodeUtilisation(w http.ResponseWriter, r *http.Request) {
 }
 
 // getNodesUtilJSON loads the nodes utilisation for a partition for a specific resource type.
-// Deprecated - To be removed in 1.7.0. Replaced with getPartitionNodesUtilJSON
+// Deprecated - To be removed in next major release. Replaced with getPartitionNodesUtilJSON
 func getNodesUtilJSON(partition *scheduler.PartitionContext, name string) *dao.NodesUtilDAOInfo {
 	mapResult := make([]int, 10)
 	mapName := make([][]string, 10)
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -424,8 +424,7 @@ func getNodeUtilisations(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	var result []*dao.PartitionNodesUtilDAOInfo
 	for _, part := range schedulerContext.GetPartitionMapClone() {
-		partitionNodesUtilJSON := getPartitionNodesUtilJSON(part)
-		result = append(result, partitionNodesUtilJSON)
+		result = append(result, getPartitionNodesUtilJSON(part))
 	}
 
 	if err := json.NewEncoder(w).Encode(result); err != nil {
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -354,7 +354,8 @@ func getNodesDAO(entries []*objects.Node) []*dao.NodeDAOInfo {
 // getNodeUtilisation loads the node utilisation based on the dominant resource used
 // for the default partition. Dominant resource is defined as the highest utilised resource
 // type on the root queue based on the registered resources.
-// Use the default partition until we get YUNIKORN-2088 fixed
+// Only check the default partition
+// Deprecated - To be removed in 1.7.0. Replaced with getNodesUtilisations
 func getNodeUtilisation(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 	partitionContext := schedulerContext.GetPartitionWithoutClusterID(configs.DefaultPartition)
@@ -379,6 +380,7 @@ func getNodeUtilisation(w http.ResponseWriter, r *http.Request) {
 }
 
 // getNodesUtilJSON loads the nodes utilisation for a partition for a specific resource type.
+// Deprecated - To be removed in 1.7.0. Replaced with getPartitionNodesUtilJSON
 func getNodesUtilJSON(partition *scheduler.PartitionContext, name string) *dao.NodesUtilDAOInfo {
 	mapResult := make([]int, 10)
 	mapName := make([][]string, 10)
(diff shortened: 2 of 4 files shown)


pkg/webservice/handlers.go
```diff
@@ -418,6 +420,80 @@ func getNodesUtilJSON(partition *scheduler.PartitionContext, name string) *dao.N
...
+		NodeCount []int      // 10 buckets, each bucket contains number of nodes
+		NodeList  [][]string // 10 buckets, each bucket contains node name list
+	}
+	resourceBuckets := make(map[string]*UtilizationBucket) // key is resource type, value is UtilizationBucket
+
+	// put nodes to buckets
+	for _, node := range partition.GetNodes() {
+		capacity := node.GetCapacity()
+		resourceAllocated := node.GetAllocatedResource()
+		absUsedCapacity := resources.CalculateAbsUsedCapacity(capacity, resourceAllocated)
```
chenyulin0719, line 448: Call CalculateAbsUsedCapacity() once per node

pkg/webservice/handlers.go
```diff
@@ -354,7 +354,8 @@ func getNodesDAO(entries []*objects.Node) []*dao.NodeDAOInfo {
 // getNodeUtilisation loads the node utilisation based on the dominant resource used
 // for the default partition. Dominant resource is defined as the highest utilised resource
 // type on the root queue based on the registered resources.
-// Use the default partition until we get YUNIKORN-2088 fixed
+// Only check the default partition
+// Deprecated - To be removed in 1.7.0. Replaced with getNodesUtilisations
```
chia7712, line 358: Should it be removed in next major release?

pkg/webservice/dao/node_util.go
```diff
@@ -18,6 +18,12 @@
 
 package dao
 
+type PartitionNodesUtilDAOInfo struct {
+	ClusterId     string              `json:"clusterId,omitempty"`
```
chia7712, line 22: Is it possible that we have empty cluster id?

pkg/webservice/dao/node_util.go
```diff
@@ -18,6 +18,12 @@
...
+	Partition     string              `json:"partition,omitempty"`
```
chia7712, line 23: ditto

pkg/webservice/handlers.go
```diff
@@ -418,6 +420,80 @@ func getNodesUtilJSON(partition *scheduler.PartitionContext, name string) *dao.N
...
+// getPartitionNodesUtilJSON retrieves the utilization of all resource types for nodes within a specific partition.
+func getPartitionNodesUtilJSON(partition *scheduler.PartitionContext) *dao.PartitionNodesUtilDAOInfo {
+	type UtilizationBucket struct {
+		NodeCount []int      // 10 buckets, each bucket contains number of nodes
+		NodeList  [][]string // 10 buckets, each bucket contains node name list
+	}
+	resourceBuckets := make(map[string]*UtilizationBucket) // key is resource type, value is UtilizationBucket
+
+	// put nodes to buckets
+	for _, node := range partition.GetNodes() {
```
chia7712, line 445: just curios. IIRC, the order of `partition.GetNodes` is not specified (as it is based on map). Could it be a potential issue?

chia7712 (commented):
@chenyulin0719 thanks for this patch. a couple of comments. please take a look

pkg/webservice/dao/node_util.go
```diff
@@ -18,6 +18,12 @@
...
+	NodesUtilList []*NodesUtilDAOInfo `json:"utilizationList,omitempty"`
```
chia7712, line 24: how about naming it `utilizations`?

pkg/webservice/handlers_test.go
```diff
@@ -815,6 +824,131 @@ func addAndConfirmApplicationExists(t *testing.T, partitionName string, partitio
...
+	nodeUtilisation := getNodesUtiliByType(t, partitionNodesUtilDAOInfo.NodesUtilList, resourceType)
+	assert.Equal(t, nodeUtilisation.ResourceType, resourceType, fmt.Sprintf("should have returned '%s', but got '%s'", resourceType, nodeUtilisation.ResourceType))
+	assert.Equal(t, len(nodeUtilisation.NodesUtil), 10, fmt.Sprintf("should have 10 bucket, but got %d", len(nodeUtilisation.NodesUtil)))
+	assert.Assert(t,
+		confirmNodeCount(nodeUtilisation.NodesUtil, int64(expectedNodeCount)),
+		fmt.Sprintf("unexpected number of nodes returned, should be %d", expectedNodeCount),
+	)
+}
+
+func getNodesUtiliByType(t *testing.T, nodesUtilList []*dao.NodesUtilDAOInfo, resourceType string) *dao.NodesUtilDAOInfo {
```
chia7712, line 941: `getNodesUtilisationByType`?

pkg/webservice/dao/node_util.go
chenyulin0719: It sounds good to me, renamed 'utilizationList' to 'utilizations'

pkg/webservice/handlers_test.go
chenyulin0719: I still prefer to use getNodesUtilByType() because the internal hierarchy of node_util.go is
1. Partition**NodesUtil**DAOInfo (Including []***NodesUtil**DAOInfo)
2. NodesUtilDAOInfo (Including []*NodeUtilDAOInfo)
3. NodeUtilDAOInfo

get**NodesUtil**iByType() is to find **NodesUtil**DAOInfo of partiticular type in **PartitionNodesUtil**DAOInfo.

PS: I previously struggled with whether to use "Util" or "Utilisation", but in the end, I decided to align with the previous node-utilization API

chenyulin0719: PS: fixed another typo, some abbreviations are incorrect, 'Utili' should be 'Util'.

pkg/webservice/handlers.go
chenyulin0719: Apart from nodes, the resource types are also unordered, but I believe this is acceptible. 
There shouldn't be a specific order for node names and resource types.

(One possible ordering is to sort them after calculating the node/resource loading, but I think this is unnecessary. 
The ordering could be controlled in client side.)

pkg/webservice/handlers.go
```diff
@@ -418,6 +420,80 @@ func getNodesUtilJSON(partition *scheduler.PartitionContext, name string) *dao.N
 	}
 }
 
+func getNodeUtilisations(w http.ResponseWriter, r *http.Request) {
+	writeHeaders(w)
+	var result []*dao.PartitionNodesUtilDAOInfo
+	for _, part := range schedulerContext.GetPartitionMapClone() {
+		partitionNodesUtilJSON := getPartitionNodesUtilJSON(part)
```
chia7712, line 427:
```go
result = append(result, getPartitionNodesUtilJSON(part))
```

how about this style?

pkg/webservice/handlers.go
```diff
@@ -418,6 +420,80 @@ func getNodesUtilJSON(partition *scheduler.PartitionContext, name string) *dao.N
...
+	// put nodes to buckets
+	for _, node := range partition.GetNodes() {
+		capacity := node.GetCapacity()
+		resourceAllocated := node.GetAllocatedResource()
+		absUsedCapacity := resources.CalculateAbsUsedCapacity(capacity, resourceAllocated)
+
+		// append to bucket based on resource type, only count if node advertises the resource
+		for resourceType := range capacity.Resources {
+			idx := 0
+			if absValue, ok := absUsedCapacity.Resources[resourceType]; ok {
```
chia7712, line 453: what if the keys of `absUsedCapacity` are not equal to `capacity`? the `idx` is always 0

chia7712 (commented):
@chenyulin0719 two comments left. PTAL

pkg/webservice/handlers.go
chenyulin0719: That's as expected.
If a key existing in capacity but not in absUsedCapacity(Not used before),  the node should be count in 0%-10% bucket. (idx=0)

ex:
If node have GCP/TPU, but never used those resource before.  
The absUsedCapacity won't have GCP/TPU abs value.

So we will count the node to
- resourceBuckets["GPU"].NodeCount[0]++
- resourceBuckets["CPU"].NodeCount[0]++

pkg/webservice/handlers.go
chenyulin0719: Fixed in the latest commit 0b66a545850ace2d724d7db284f3083462a87dbe

pkg/webservice/handlers_test.go
```diff
@@ -815,6 +824,131 @@ func addAndConfirmApplicationExists(t *testing.T, partitionName string, partitio
...
+	addAllocatedResource(t, node2, "alloc-1", "app-1", map[string]resources.Quantity{"memory": 1, "vcore": 1})
+	addAllocatedResource(t, node3, "alloc-1", "app-1", map[string]resources.Quantity{"memory": 1, "vcore": 1})
+	addAllocatedResource(t, node4, "alloc-1", "app-1", map[string]resources.Quantity{"gpu": 1})
+
+	// get nodes utilizations
+	getNodeUtilisations(resp, req)
+	err = json.Unmarshal(resp.outputBytes, &partitionNodesUtilDAOInfo)
+	assert.NilError(t, err, "should decode a list of *dao.PartitionNodesUtilDAOInfo")
+	assert.Equal(t, len(partitionNodesUtilDAOInfo), 2)
+	assert.Equal(t, partitionNodesUtilDAOInfo[0].ClusterId, rmID)
```
chia7712, line 908: why check only first element?

pkg/webservice/handlers_test.go
```diff
@@ -815,6 +824,131 @@ func addAndConfirmApplicationExists(t *testing.T, partitionName string, partitio
...
+
+	defaultPartitionNodesUtilDAOInfo := partitionNodesUtilDAOInfo[0]
+	gpuPartitionNodesUtilDAOInfo := partitionNodesUtilDAOInfo[1]
+	if defaultPartitionNodesUtilDAOInfo.Partition == "gpu" {
+		defaultPartitionNodesUtilDAOInfo = partitionNodesUtilDAOInfo[1]
+		gpuPartitionNodesUtilDAOInfo = partitionNodesUtilDAOInfo[0]
+	}
+
+	expectedResourceTypeCount := 2
+	assert.Equal(t, len(defaultPartitionNodesUtilDAOInfo.NodesUtilList), expectedResourceTypeCount)
```
chia7712, line 918: `assert.Equal(t, len(defaultPartitionNodesUtilDAOInfo.NodesUtilList), 2)` 

how about using constant directly?

pkg/webservice/handlers_test.go
```diff
@@ -815,6 +824,131 @@ func addAndConfirmApplicationExists(t *testing.T, partitionName string, partitio
...
+	expectedResourceTypeCount = 1
+	assert.Equal(t, len(gpuPartitionNodesUtilDAOInfo.NodesUtilList), expectedResourceTypeCount)
```
chia7712, line 920: ditto

chia7712 (commented):
@chenyulin0719 thanks for this nice patch. some trivial comments.

pkg/webservice/handlers_test.go
chenyulin0719: Sorry for causing confusion. Just added the checking for second element.  8440789aeb44ac0bccf81f57afad5e6d8015b2af

My original thought is that the rmID value is merely a pass-in value and there is no difference among the two elements,  so I just did a simple check.

pkg/webservice/handlers_test.go
chenyulin0719: Changed to constant.

pkg/webservice/handlers_test.go
chenyulin0719: Changed to constant.

chia7712 (approved):
LGTM

PR #763, opened 2023-12-26
[YUNIKORN-2260] Partition limit is NOT equivalent with the root limit

brandboat:
### What is this PR for?
Apply partition limits to root queue if the limits for the root queue are not defined.

### What type of PR is it?
* [x] - Bug Fix

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2260

covered by unit test

N/A

### Questions:
N/A

### DIFF
--- a/pkg/common/configs/config_test.go
+++ b/pkg/common/configs/config_test.go
@@ -22,6 +22,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strings"
 	"testing"
 
@@ -910,12 +911,12 @@ partitions:
 	if len(conf.Partitions[0].Queues) != 1 || len(conf.Partitions[0].Queues[0].Limits) != 3 {
 		t.Errorf("partition limits not correctly applied to root queue %v", conf)
 	}
-	if len(conf.Partitions[0].Limits) != 0 {
-		t.Errorf("partition limits not wiped out %v", conf)
+	if len(conf.Partitions[0].Limits) != 3 {
+		t.Errorf("partition users linked not correctly loaded %v", conf)
 	}
 
 	// limit 1 check
-	limit := conf.Partitions[0].Queues[0].Limits[0]
+	limit := conf.Partitions[0].Limits[0]
 	if len(limit.Users) != 1 || limit.Users[0] != "user1" || len(limit.Groups) != 0 {
 		t.Errorf("failed to load max apps %v", limit)
 	}
@@ -928,7 +929,7 @@ partitions:
 	}
 
 	// limit 2 check
-	limit = conf.Partitions[0].Queues[0].Limits[1]
+	limit = conf.Partitions[0].Limits[1]
 	if len(limit.Users) != 1 || limit.Users[0] != "user2" ||
 		len(limit.Groups) != 1 || limit.Groups[0] != "prod" {
 		t.Errorf("user and group list have incorrect entries (limit 2): %v", limit)
@@ -938,7 +939,7 @@ partitions:
 	}
 
 	// limit 3 check
-	limit = conf.Partitions[0].Queues[0].Limits[2]
+	limit = conf.Partitions[0].Limits[2]
 	if len(limit.Groups) != 2 || limit.Groups[0] != "dev" || limit.Groups[1] != "test" {
 		t.Errorf("failed to load groups from config (limit 3): %v", limit)
 	}
@@ -947,6 +948,10 @@ partitions:
 	}
 	if limit.MaxApplications != 20 || limit.MaxResources != nil || len(limit.MaxResources) != 0 {
 		t.Errorf("loaded resource limits incorrectly (limit 3): %v", limit)
+	}
+
+	if !reflect.DeepEqual(conf.Partitions[0].Limits, conf.Partitions[0].Queues[0].Limits) {
+		t.Errorf("partition limits and root queue limits are not equivalent : %v", conf)
 	}
 }
 
@@ -1244,38 +1249,42 @@ partitions:
 	if len(conf.Partitions[0].Queues) != 1 || len(conf.Partitions[0].Queues[0].Limits) != 5 {
 		t.Errorf("partition limits not correctly applied to root queue: %v", conf)
 	}
-	if len(conf.Partitions[0].Limits) != 0 {
-		t.Errorf("partition limits not wiped out: %v", conf)
+	if len(conf.Partitions[0].Limits) != 5 {
+		t.Errorf("failed to load partition limits from config: %v", conf)
 	}
 
 	// user with dot check
-	limit := conf.Partitions[0].Queues[0].Limits[0]
+	limit := conf.Partitions[0].Limits[0]
 	if len(limit.Users) != 1 || limit.Users[0] != "user.lastname" {
 		t.Errorf("failed to load 'dot' user from config: %v", limit)
 	}
 
 	// user with @ check
-	limit = conf.Partitions[0].Queues[0].Limits[1]
+	limit = conf.Partitions[0].Limits[1]
 	if len(limit.Users) != 1 || limit.Users[0] != "user@domain" {
 		t.Errorf("failed to load '@' user from config: %v", limit)
 	}
 
 	// user wildcard
-	limit = conf.Partitions[0].Queues[0].Limits[2]
+	limit = conf.Partitions[0].Limits[2]
 	if len(limit.Users) != 1 || limit.Users[0] != "*" {
 		t.Errorf("failed to load wildcard user from config: %v", limit)
 	}
 
 	// group
-	limit = conf.Partitions[0].Queues[0].Limits[3]
+	limit = conf.Partitions[0].Limits[3]
 	if len(limit.Groups) != 1 || limit.Groups[0] != "test" {
 		t.Errorf("failed to load group from config: %v", limit)
 	}
 
 	// wildcard group
-	limit = conf.Partitions[0].Queues[0].Limits[4]
+	limit = conf.Partitions[0].Limits[4]
 	if len(limit.Groups) != 1 || limit.Groups[0] != "*" {
 		t.Errorf("failed to load wildcard group from config: %v", limit)
+	}
+
+	if !reflect.DeepEqual(conf.Partitions[0].Limits, conf.Partitions[0].Queues[0].Limits) {
+		t.Errorf("partition limits and root queue limits are not equivalent : %v", conf)
 	}
 }
 
--- a/pkg/common/configs/config_test.go
+++ b/pkg/common/configs/config_test.go
@@ -907,15 +907,15 @@ partitions:
 	conf, err := CreateConfig(data)
 	assert.NilError(t, err, "partition user parsing should not have failed")
 	// gone through validation: 1 top level queues
-	if len(conf.Partitions[0].Queues) != 1 {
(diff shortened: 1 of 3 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -573,6 +573,19 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
 	return nil
 }
 
+func checkLimitsStructure(rootQueue *QueueConfig, partitionLimits []Limit) error {
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) > 0 {
+		return fmt.Errorf("partition limits and root queue limits both exist")
+	}
+
+	// if root queue limits not defined, apply partition limits
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) == 0 {
```
chia7712, line 582: if `rootQueue` is NOT root, we don't need this check, right?

brandboat: `rootQueue` will always be the root, as `checkLimitsStructure` will be triggered after `checkQueuesStructure`, and `checkQueuesStructure` will ensure that the root element in queues is the root queue.
https://github.com/apache/yunikorn-core/blob/526a8db4b61c0db48a3fa6a41c9f4d59481862e7/pkg/common/configs/configvalidator.go#L662-L667

pkg/common/configs/configvalidator.go
```diff
@@ -573,6 +573,19 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
 	return nil
 }
 
+func checkLimitsStructure(rootQueue *QueueConfig, partitionLimits []Limit) error {
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) > 0 {
```
chia7712, line 577: this change is a kind of behavior change, so it deserves a docs. Could you please file a PR as follow-up to reminder user?

pkg/common/configs/configvalidator.go
```diff
@@ -573,6 +573,19 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
...
+		return fmt.Errorf("partition limits and root queue limits both exist")
+	}
+
+	// if root queue limits not defined, apply partition limits
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) == 0 {
+		rootQueue.Limits = partitionLimits
```
chia7712, line 583: this solution is simple and it should work I think. However, this method is not idempotent as the `PartitionConfig` changed by this method will get failed if we pass `PartitionConfig` to this method again. It seems to me that is error-prone. In other words, the validator hate `PartitionConfig` has both partition-level limits and root-level limits, but it may produce such case itself.

chia7712 (commented):
@brandboat thanks for this patch. two comments left.

pkg/common/configs/configvalidator.go
brandboat: Thanks for the detailed review, indeed we should wipe out partition limits after applied them to root queue, already addressed that in the latest commit.

pkg/common/configs/configvalidator.go
brandboat: I'll raised another pr after this pull-request merged. Thanks !

pkg/common/configs/configvalidator.go
```diff
@@ -734,7 +756,8 @@ func Validate(newConfig *SchedulerConfig) error {
 
 	// check uniqueness
 	partitionMap := make(map[string]bool)
-	for i, partition := range newConfig.Partitions {
+	for i := range newConfig.Partitions {
+		partition := newConfig.Partitions[i]
```
chia7712, line 760: It seems this change is redundant?

pkg/common/configs/configvalidator.go
```diff
@@ -706,7 +728,7 @@ func checkQueuesStructure(partition *PartitionConfig) error {
 	if rootQueue.Resources.Guaranteed != nil || rootQueue.Resources.Max != nil {
 		return fmt.Errorf("root queue must not have resource limits set")
 	}
-	return checkQueues(&rootQueue, 1)
+	return nil
```
chia7712, line 731: pardon, is this change related to this PR?

pkg/common/configs/configvalidator.go
```diff
@@ -595,6 +595,28 @@ func checkLimits(limits []Limit, obj string, queue *QueueConfig) error {
...
+
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) > 0 {
+		return fmt.Errorf("partition limits and root queue limits both exist")
+	}
+
+	// if root queue limits not defined, apply partition limits
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) == 0 {
+		partitionConfig.Queues[0].Limits = partitionLimits
+		// clear partition limits
+		partitionConfig.Limits = nil
```
chia7712, line 614: Could you add comment to https://github.com/apache/yunikorn-core/blob/master/pkg/common/configs/config.go#L52 to reminder readers that the field could be empty. Also, we should honor the root-level limits.

pkg/common/configs/config_test.go
```diff
@@ -907,15 +907,15 @@ partitions:
...
 	}
-	if len(conf.Partitions[0].Limits) != 3 {
-		t.Errorf("partition users linked not correctly loaded  %v", conf)
+	if len(conf.Partitions[0].Limits) != 0 {
+		t.Errorf("partition limits not wiped out %v", conf)
 	}
 
 	// limit 1 check
-	limit := conf.Partitions[0].Limits[0]
+	limit := conf.Partitions[0].Queues[0].Limits[0]
```
chia7712, line 918: ok, that changes are the side effect.

or could we accept the root-level limits is equal with partition-level limits? that may be a good way, as the caller can call either root-level limits or partition-level limits.

chia7712 (commented):
@brandboat a couple of comments. please take a look.

pkg/common/configs/configvalidator.go
brandboat: Doing this to avoid lint error `G601: Implicit memory aliasing in for loop.`

pkg/common/configs/configvalidator.go
brandboat: Yes, we need the `checkLimits` inside `checkQueues` to check root queue limits (after partition limits is applied to root queue), so here extracted the method out of `checkQueuesStructure`

pkg/common/configs/configvalidator.go
chia7712: this lint error is caused by calling reference of `partition`, right? Is this lint error existent before?

pkg/common/configs/config_test.go
brandboat: > or could we accept the root-level limits is equal with partition-level limits? that may be a good way, as the caller can call either root-level limits or partition-level limits.

I've considered this before, and at that time, I thought making configuration changes was a relatively simple approach without the need to modify manager.go. Anyway, this is a completely different approach. I will submit another PR and let the community decide.

pkg/common/configs/configvalidator.go
chia7712: ok, you want to the root queue gets all fixes (limits and inserts) before calling `checkQueues`.

for another, you change the behavior or `checkQueuesStructure` without changing test-related code. It seems `checkQueuesStructure` has no tests, right? If so, could you file jira to complete units for each `checkXXX`

pkg/common/configs/configvalidator.go
brandboat: Yes it's exist from the beginning, you can try `./tools/golangci-lint run` and search error `G601: Implicit memory aliasing in for loop`.

pkg/common/configs/config_test.go
chia7712: ummm, you are right. please ignore my previous comments :)

pkg/common/configs/configvalidator.go
chia7712: Or we should make sure that partition limits are always equal to root limits? We do NOT undefine the partition limits. Instead, we allow both limits if they are identical.

pkg/common/configs/configvalidator.go
brandboat: filed https://issues.apache.org/jira/browse/YUNIKORN-2297

pkg/common/configs/configvalidator.go
brandboat: > Instead, we allow both limits if they are identical.

Sounds good, I'll adopt your approach. Many thanks !

manirajv06 (approved):
looks good

pkg/common/configs/configvalidator.go
```diff
@@ -595,6 +596,26 @@ func checkLimits(limits []Limit, obj string, queue *QueueConfig) error {
 	return nil
 }
 
+func checkLimitsStructure(partitionConfig *PartitionConfig) error {
+	partitionLimits := partitionConfig.Limits
+	rootQueue := partitionConfig.Queues[0]
+
+	if len(partitionConfig.Queues) < 1 || strings.ToLower(partitionConfig.Queues[0].Name) != RootQueue {
```
chia7712, line 603: `partitionConfig.Queues[0]` can be replaced by `rootQueue`

pkg/common/configs/configvalidator.go
```diff
@@ -595,6 +596,26 @@ func checkLimits(limits []Limit, obj string, queue *QueueConfig) error {
...
+		return fmt.Errorf("top queue is not root")
+	}
+
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) > 0 && !reflect.DeepEqual(partitionLimits, rootQueue.Limits) {
+		return fmt.Errorf("partition limits and root queue limits are not equivalent")
+	}
+
+	// if root queue limits not defined, apply partition limits
+	if len(partitionLimits) > 0 && len(rootQueue.Limits) == 0 {
+		partitionConfig.Queues[0].Limits = partitionLimits
```
chia7712, line 613: ditto

pkg/common/configs/configvalidator.go
```diff
@@ -595,6 +596,26 @@ func checkLimits(limits []Limit, obj string, queue *QueueConfig) error {
 	return nil
 }
 
+func checkLimitsStructure(partitionConfig *PartitionConfig) error {
+	partitionLimits := partitionConfig.Limits
+	rootQueue := partitionConfig.Queues[0]
+
+	if len(partitionConfig.Queues) < 1 || strings.ToLower(partitionConfig.Queues[0].Name) != RootQueue {
+		return fmt.Errorf("top queue is not root")
```
chia7712, line 604: the error message can include the name of current top queue.

chia7712 (commented):
@brandboat a couple of minor comments. PTAL

pkg/common/configs/configvalidator.go
brandboat: Actually this can't be replaced with `rootQueue`, otherwise the partition limits won't change.

chia7712: oh, I get it now.

brandboat: ping @chia7712 , already addressed comments in the latest comment. Many thanks~

chia7712 (approved):
LGTM. will merge it after QA gets passed.

PR #769, opened 2024-01-02
[YUNIKORN-2303] Data race during queue sorting

pbacsko:
### What is this PR for?
We must not access `Queue.pending` directly.  Use `GetPendingResource()` because it can be modified from different goroutines.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2303

### Questions:


chia7712 (approved):
nice finding. LGTM

PR #774, opened 2024-01-17
[YUNIKORN-2331] Add Preemption Queue tests

manirajv06:
### What is this PR for?
Added new Preemption Queue test suites to assert various Preemption behaviour especially focussed on important methods like IsAtorAbove, IsWithIn & GetRemaining. These tests required some changes in those methods as well.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2331

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -698,6 +698,12 @@ func (qps *QueuePreemptionSnapshot) IsAtOrAboveGuaranteedResource() bool {
 	if qps == nil {
 		return false
 	}
+
+	// check the parent, as violations at any level mean we are not within limits
+	if qps.Parent.IsAtOrAboveGuaranteedResource() {
+		return true
+	}
+
 	guaranteed := qps.GetGuaranteedResource()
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/preemption.go
```diff
@@ -698,6 +698,12 @@ func (qps *QueuePreemptionSnapshot) IsAtOrAboveGuaranteedResource() bool {
 	if qps == nil {
 		return false
 	}
+
+	// check the parent, as violations at any level mean we are not within limits
+	if qps.Parent.IsAtOrAboveGuaranteedResource() {
+		return true
+	}
+
```
pbacsko, line 706: This and line 755 are behavioral changes. The JIRA says the type is "Test", but then that's a mistake, isn't it?

manirajv06:
@craigcondit I verified this change in my local cluster setup by running preemption test suite. I don't see any failures and test was completed successfully couple of times.

PR #775, opened 2024-01-18
[YUNIKORN-2333] Redundant min and max computation in resources.go

manirajv06:
### What is this PR for?

Methods like `ComponentWiseMin` & `ComponentWiseMax` can avoid the redundant min and max calculations for resource computed already.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2333

Existing tests should cover this.

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -800,7 +800,9 @@ func ComponentWiseMin(left, right *Resource) *Resource {
 			out.Resources[k] = MinQuantity(v, right.Resources[k])
 		}
 		for k, v := range right.Resources {
-			out.Resources[k] = MinQuantity(v, left.Resources[k])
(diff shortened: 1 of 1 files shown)


pbacsko (approved):
+1 LGTM

manirajv06:
@craigcondit I do agree with some of the trade-offs being discussed here. @pbacsko and I touched upon more or less same points lightly in one of our offline discussions, but not in detail. I will do one more round and take it accordingly.

pbacsko:
Do at least a benchmark test to see if there's any improvement with Go's benchmark framework using different resource pairs.

manirajv06:
Benchmark results are not good. So, we don't want to proceed with the changes further.

PR #776, opened 2024-01-19
[YUNIKORN-2329] Remove goroutine in RMProxy.UpdateNode after benchmark

chenyulin0719:
### What is this PR for?
[2024-01-20 Update version 2]
As a solution for [YUNIKORN-2327](https://issues.apache.org/jira/browse/YUNIKORN-2327). Remove goroutine in [RMProxy.UpdateNode()](https://github.com/apache/yunikorn-core/blob/master/pkg/rmproxy/rmproxy.go#L378-L389) could make sure the function put si.NodeRequest in channel in the same order from shim.

Also, remove some trivial parallelism in RMProxy.go because it is considered "premature optimization." The parallelism is not really necessary. Remove goroutine in below functions:
1. func (rmp *RMProxy) UpdateApplication(request *si.ApplicationRequest)
2. func (rmp *RMProxy) UpdateAllocation(request *si.AllocationRequest)
3. func (rmp *RMProxy) UpdateNode(request *si.NodeRequest)

Also remove unnecessary empty list checks.

Once this PR is merged, I'll trigger another PR for shim to upgrade go.mod version.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
Update yunikorn-core version in shim

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2329

Rebuild shim with this core version, all existing e2e test should pass.
The e2e test passed in my forked shim:
https://github.com/chenyulin0719/yunikorn-k8shim/actions/runs/7587158325

NA

### Questions:
NA

### DIFF
--- a/pkg/rmproxy/rmproxy.go
+++ b/pkg/rmproxy/rmproxy.go
@@ -314,37 +314,26 @@ func (rmp *RMProxy) UpdateAllocation(request *si.AllocationRequest) error {
 	if rmp.GetResourceManagerCallback(request.RmID) == nil {
 		return fmt.Errorf("received AllocationRequest, but RmID=\"%s\" not registered", request.RmID)
 	}
-	go func() {
-		// Update allocations
-		if len(request.Allocations) > 0 {
-			for _, alloc := range request.Allocations {
-				alloc.PartitionName = common.GetNormalizedPartitionName(alloc.PartitionName, request.RmID)
-			}
-		}
+	// Update allocations
+	for _, alloc := range request.Allocations {
+		alloc.PartitionName = common.GetNormalizedPartitionName(alloc.PartitionName, request.RmID)
+	}
 
-		// Update asks
-		if len(request.Asks) > 0 {
-			for _, ask := range request.Asks {
-				ask.PartitionName = common.GetNormalizedPartitionName(ask.PartitionName, request.RmID)
-			}
-		}
+	// Update asks
+	for _, ask := range request.Asks {
+		ask.PartitionName = common.GetNormalizedPartitionName(ask.PartitionName, request.RmID)
+	}
 
-		// Update releases
-		if request.Releases != nil {
-			if len(request.Releases.AllocationsToRelease) > 0 {
-				for _, rel := range request.Releases.AllocationsToRelease {
-					rel.PartitionName = common.GetNormalizedPartitionName(rel.PartitionName, request.RmID)
-				}
-			}
-
-			if len(request.Releases.AllocationAsksToRelease) > 0 {
-				for _, rel := range request.Releases.AllocationAsksToRelease {
-					rel.PartitionName = common.GetNormalizedPartitionName(rel.PartitionName, request.RmID)
-				}
-			}
+	// Update releases
+	if request.Releases != nil {
+		for _, rel := range request.Releases.AllocationsToRelease {
+			rel.PartitionName = common.GetNormalizedPartitionName(rel.PartitionName, request.RmID)
 		}
-		rmp.EventHandlers.SchedulerEventHandler.HandleEvent(&rmevent.RMUpdateAllocationEvent{Request: request})
-	}()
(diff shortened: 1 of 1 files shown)


pkg/rmproxy/rmproxy.go
```diff
@@ -375,18 +375,18 @@ func (rmp *RMProxy) UpdateNode(request *si.NodeRequest) error {
...
-	go func() {
-		if len(request.Nodes) > 0 {
-			for _, node := range request.Nodes {
-				if len(node.GetAttributes()) == 0 {
-					node.Attributes = map[string]string{}
-				}
-				partition := node.Attributes[siCommon.NodePartition]
-				node.Attributes[siCommon.NodePartition] = common.GetNormalizedPartitionName(partition, request.RmID)
+
+	if len(request.Nodes) > 0 {
```
chia7712, line 379: Could you please leave some comments? Otherwise, it looks weird since other requests are sent in parallel ...

craigcondit: We can also remove the extra empty check in the node handler loop (iterating an empty map is a no-op and completely safe, plus we really shouldn't have UpdateNode being called with zero nodes).

chia7712: > We can then say that the SI contract is that events are processed in order.

Do we address this by adding comments to code of yunikorn-interface? Or we can add UTs to make sure the order of requests from shim is preserved

craigcondit: I don't think we need to do anything really. My intent was just to point out that after these changes we basically have FIFO event processing which should be much more reliable.

chenyulin0719: >  I think for all 3 handlers (app/task/node) we should be good to simply remove the goroutine and execute the code there inline.

Hi @craigcondit, 
Let me conduct a benchmark for app/task before removal. (I assume the "task" you mentioned here refers to allocation)

Below are all the possible functions that could remove their goroutine.
- func (rmp *RMProxy) UpdateApplication(request *si.ApplicationRequest)
- func (rmp *RMProxy) UpdateAllocation(request *si.AllocationRequest) 
- func (rmp *RMProxy) UpdateNode(request *si.NodeRequest) 

Below are functions need to keep their goroutine as they require waiting for a callback:
- func (rmp *RMProxy) RegisterResourceManager(request *si.RegisterResourceManagerRequest, callback api.ResourceManagerCallback) 
- func (rmp *RMProxy) UpdateConfiguration(request *si.UpdateConfigurationRequest) 

> Could you please leave some comments? Otherwise, it looks weird since other requests are sent in parallel ...

@chia7712 If we could remove the goroutine for app/task/node, then the behaviors would aligne and I think the comments will be not necessary. Let the confirm whether we can remove it first.

craigcondit: Yes, those three functions are the ones we should update. I think we can even skip the benchmarks. The work being done in these goroutines is extremely trivial, and I believe it was a "premature optimization" to build them this way to begin with. These routines would have to be executed millions of times per second before the effects of removing these goroutines would be visible as anything other than noise. Since these are largely called in response to the receipt of individual K8s events on pods and nodes, such a frequency will never occur.

chenyulin0719: Agree, the heavy frequency will never occur. Will update this commit and run e2e test.

chenyulin0719: e2e test passed: https://github.com/chenyulin0719/yunikorn-k8shim/actions/runs/7587158325

> We can also remove the extra empty check in the node handler loop (iterating an empty map is a no-op and completely safe, plus we really shouldn't have UpdateNode being called with zero nodes).

Also remove other empty list check in UpdateApplication/UpdateAllocation.

craigcondit (approved):
+1 LGTM.

PR #777, opened 2024-01-25
[YUNIKORN-2147] Limit the number of concurrent event streams

pbacsko:
### What is this PR for?
The number of concurrent event streams should be limited.
The PR introduces 2 new properties and two sensible default values.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2147

### Questions:


pkg/webservice/stream_limit.go
```diff
@@ -0,0 +1,115 @@
...
+
+	count, ok := sl.perHost[host]
+	if !ok {
+		log.Log(log.REST).Warn("Tried to remove a non-existing host from tracking",
+			zap.String("host", host))
+		return
+	}
+
+	sl.total--
+	if count == 1 {
```
craigcondit, line 78: Minor thing, but just for safety it's probably best to check for "if count <= 1".

pkg/webservice/stream_limit.go
```diff
@@ -0,0 +1,115 @@
...
+		delete(sl.perHost, host)
+		return
+	}
+	sl.perHost[host]--
+}
+
+func (sl *StreamingLimiter) getLimits() (uint64, uint64) {
```
craigcondit, line 85: Instead of calculating this on every connection attempt, it's better to register for configmap changes and parse out at that point and cache the results.

pkg/common/configs/configs.go
```diff
@@ -33,15 +33,19 @@ const (
 	HealthCheckInterval = PrefixHealth + "checkInterval"
 
 	// events
-	CMEventTrackingEnabled    = PrefixEvent + "trackingEnabled"    // Application Tracking
-	CMEventRequestCapacity    = PrefixEvent + "requestCapacity"    // Request Capacity
-	CMEventRingBufferCapacity = PrefixEvent + "ringBufferCapacity" // Ring Buffer Capacity
+	CMEventTrackingEnabled        = PrefixEvent + "trackingEnabled"    // Application Tracking
+	CMEventRequestCapacity        = PrefixEvent + "requestCapacity"    // Request Capacity
+	CMEventRingBufferCapacity     = PrefixEvent + "ringBufferCapacity" // Ring Buffer Capacity
+	CMMaxStreamConnectionsTotal   = PrefixEvent + "maxStreamConnectionsTotal"
```
craigcondit, line 39: Minor nit: This is already a pretty long var, maybe just "maxStreamConnections" (leaving out Total) and "maxStreamConnectionsPerHost". The 'Total' suffix seems redundant.

craigcondit (changes requested):
See comments.

pkg/webservice/stream_limit.go
```diff
@@ -0,0 +1,128 @@
...
+			log.Log(log.SchedHealth).Warn("Failed to parse configuration value",
+				zap.String("key", configs.CMMaxEventStreamsPerHost),
+				zap.String("value", value),
+				zap.Error(err))
+		} else {
+			maxStreamsPerHost = parsed
+		}
+	}
+
+	sl.maxStreams = maxStreams
```
craigcondit, line 126: These updates need to be wrapped in a lock/unlock block. Otherwise they may not be updated consistently (AddHost / RemoveHost could be called in between the updates for example).

pkg/webservice/stream_limit.go
```diff
@@ -0,0 +1,128 @@
...
+
+	sync.Mutex
+}
+
+func NewStreamingLimiter() *StreamingLimiter {
+	sl := &StreamingLimiter{
+		perHostStreams: make(map[string]uint64),
+	}
+
+	configs.AddConfigMapCallback("stream-limiter", func() {
```
craigcondit, line 47: Will there only ever be a single streaming limiter? If so, this should be ok, but if there could be multiple, you need to generate a unique ID per limiter `stream-limiter-{seqnum}` or something would work. Also, you should probably have some sort of `Stop()` function that unregisters the callback.

craigcondit (changes requested):
Minor locking issue, some configmap registration questions, but otherwise looks good.

pkg/webservice/stream_limit.go
pbacsko: Correct, I updated the code.

pkg/webservice/stream_limit.go
pbacsko: Yes, single instance. It's meant to live until Yunikorn's termination (it is created inside `init()`). I think it's fine, it's also not an issue for the MockScheduler. However, just to be safe, I do create a new callback id during each call and added a `Stop()` method for the unit tests.

manirajv06 (commented):
Overall looks good.

Minor nits and suggestions:

1. Should we rename to streaming_limiter?
2. Can we have stop() inside streaming_limiter?
3. In case of setting lesser value than current value, do we need to try to close the active connection (ordered based on each connection activity) gracefully as intent would be to reduce the load at any moment?

pbacsko:
> Overall looks good.
> 
> Minor nits and suggestions:
> 
> 1. Should we rename to streaming_limiter?
> 2. Can we have stop() inside streaming_limiter?
> 3. In case of setting lesser value than current value, do we need to try to close the active connection (ordered based on each connection activity) gracefully as intent would be to reduce the load at any moment?

1. We can rename it.
2. It will not be called from the production code. It's initialized in `init()` and it's not tied to the lifecycle of the WebApp. That would be ideal, but the handlers are not part of a type so they have to reference global variables. This is definitely not good, but it's a big change. I'd be happy if this wasn't the case, but it's a bigger change for later. For now, it's acceptable.
3. That's a logic which is too complicated. It can happen after config refresh. We'd need to address the select loop inside `getStream()`, I'd rather not do that.

manirajv06 (approved):

PR #778, opened 2024-01-26
[YUNIKORN-2284] ERROR message when stopping Service context

FrankYang0529:
### What is this PR for?
The tracker object no longer exist when `PartitionContext.removeApplication()` is called. At this point the app is also in Completed state, so it's not necessary to decrement any resource.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2284

Run `go test ./... -tags deadlock -v` and check there is no error message like `user tracker must be available`.

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1847,7 +1847,9 @@ func (sa *Application) RemoveAllAllocations() []*Allocation {
 		sa.appEvents.sendRemoveAllocationEvent(alloc, si.TerminationType_STOPPED_BY_RM)
 	}
 
-	if resources.IsZero(sa.pending) {
+	// if an app doesn't have any allocations and the user doesn't have other applications,
+	// the user tracker is nonexistent. We don't want to decrease resource usage in this case.
+	if ugm.GetUserManager().GetUserTracker(sa.user.User) != nil && resources.IsZero(sa.pending) {
 		sa.decUserResourceUsage(resources.Add(sa.allocatedResource, sa.allocatedPlaceholder), true)
 	}
 	// cleanup allocated resource for app (placeholders and normal)


pkg/scheduler/objects/application.go
```diff
@@ -1845,7 +1845,7 @@ func (sa *Application) RemoveAllAllocations() []*Allocation {
 		sa.appEvents.sendRemoveAllocationEvent(alloc, si.TerminationType_STOPPED_BY_RM)
 	}
 
-	if resources.IsZero(sa.pending) {
+	if resources.IsZero(sa.pending) && !resources.IsZero(resources.Add(sa.allocatedResource, sa.allocatedPlaceholder)) {
 		sa.decUserResourceUsage(resources.Add(sa.allocatedResource, sa.allocatedPlaceholder), true)
```
pbacsko, line 1849: Nit: save the result to a variable so that we don't need to calculate it again.

pbacsko (changes requested):

pkg/scheduler/objects/application.go
```diff
@@ -1845,7 +1845,7 @@ func (sa *Application) RemoveAllAllocations() []*Allocation {
 		sa.appEvents.sendRemoveAllocationEvent(alloc, si.TerminationType_STOPPED_BY_RM)
 	}
 
-	if resources.IsZero(sa.pending) {
+	if resources.IsZero(sa.pending) && !resources.IsZero(resources.Add(sa.allocatedResource, sa.allocatedPlaceholder)) {
```
chia7712, line 1848: why this condition check can fix that issue? or is it more suitable that we just check the existence of user tracker here?

pbacsko: It fixes because we might not have an user tracker at all at this point if we call `RemoveApplication()`.

chia7712: @pbacsko thanks for response. I had another question:

why we don't check existence of tracker directly? for example: 

```golang
	if resources.IsZero(sa.pending) && ugm.GetUserManager().GetUserTracker(sa.user.User) != nil {
		sa.decUserResourceUsage(resources.Add(sa.allocatedResource, sa.allocatedPlaceholder), true)
	}
```

pbacsko: That's a solution, too. Although I feel that in both cases, we need comments to explain why we do that because it's not obvious why we do it (one can definitely think: "what can possibly happen to the user tracker?"). I'm not against that change. But I don't think we need to return the tracker instance, just a boolean flag which indicates if it exists or not.

Another thing is that, IMO it makes sense to move this check to `decUserResourceUsage()`.

chia7712: > Another thing is that, IMO it makes sense to move this check to decUserResourceUsage().

that makes sense to me.

wilfred-s: My preference would be to use the user tracker retrieval. Overhead of the lookup including the lock is far less than doing the add and a compare to zero for the resource.

wilfred-s (changes requested):
use the user tracker not the resource calculation

manirajv06:
Depending on user tracker retrieval again in application.go is as good as repeating the check already happening inside manager.go#DecreaseTrackedResource()

Checks inside manager.go has been introduced to be on bit safer side. But this jira is very legit case where single user has completed all this activities and cleaned properly, finally partition removal is trying to do the same job again through app.removeAllAllocations(). 

Do we really need to worry about this? Just changing the log message from "user tracker must be available in userTrackers map" to something like "user tracker is not available in userTrackers map. Might have been gone through cleaner removal because user would have completed all this activities?" and level to "debug" is good enough?

wilfred-s:
> Do we really need to worry about this? Just changing the log message from "user tracker must be available in userTrackers map" to something like "user tracker is not available in userTrackers map. Might have been gone through cleaner removal because user would have completed all this activities?" and level to "debug" is good enough?

We really want to see this message if it happens while the scheduler is normally running. At that point it is an ERROR and shows that there is some system inconsistency. Hiding the message in the user group manager is thus I think not a good idea. The `RemoveAllAllocations()` call is the best place to prevent the message from being logged.

manirajv06:
> The RemoveAllAllocations() call is the best place to prevent the message from being logged.

Yes, this would be the right place if we need to keep the log as is.

chia7712:
It seems there are two changes which need to be fixed before merging.

1. check user tracker instead of comparing resource https://github.com/apache/yunikorn-core/pull/778#discussion_r1473714580
2. the check should be in `RemoveAllAllocations` (https://github.com/apache/yunikorn-core/pull/778#issuecomment-1922566574 and https://github.com/apache/yunikorn-core/pull/778#issuecomment-1922790177)

@FrankYang0529 could you please address below comments? thanks

pkg/scheduler/objects/application.go
```diff
@@ -1847,7 +1847,7 @@ func (sa *Application) RemoveAllAllocations() []*Allocation {
 		sa.appEvents.sendRemoveAllocationEvent(alloc, si.TerminationType_STOPPED_BY_RM)
 	}
 
-	if resources.IsZero(sa.pending) {
+	if ugm.GetUserManager().GetUserTracker(sa.user.User) != nil && resources.IsZero(sa.pending) {
```
chia7712, line 1850: Could you please add comment for this condition? I feel the description of this PR is good example :)

chia7712 (approved):
LGTM

PR #780, opened 2024-01-30
[YUNIKORN-2335] Use go standard library min and max functions

g1geordie:
### What is this PR for?
Use go standard library min and max functions

### What type of PR is it?
* [x] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2335

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -699,8 +699,8 @@ func (qps *QueuePreemptionSnapshot) IsAtOrAboveGuaranteedResource() bool {
 		return false
 	}
 	guaranteed := qps.GetGuaranteedResource()
-	max := qps.GetMaxResource()
-	absGuaranteed := resources.ComponentWiseMinPermissive(guaranteed, max)
+	maxResource := qps.GetMaxResource()
+	absGuaranteed := resources.ComponentWiseMinPermissive(guaranteed, maxResource)
 	used := resources.Sub(qps.AllocatedResource, qps.PreemptingResource)
 
 	// if we don't fit, we're clearly above
@@ -727,8 +727,8 @@ func (qps *QueuePreemptionSnapshot) IsWithinGuaranteedResource() bool {
 	if qps.Leaf && guaranteed.IsEmpty() {
 		return false
 	}
-	max := qps.GetMaxResource()
-	absGuaranteed := resources.ComponentWiseMinPermissive(guaranteed, max)
+	maxResource := qps.GetMaxResource()
+	absGuaranteed := resources.ComponentWiseMinPermissive(guaranteed, maxResource)
 	used := resources.Sub(qps.AllocatedResource, qps.PreemptingResource)
 	return absGuaranteed.FitIn(used)
 }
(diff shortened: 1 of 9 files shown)


chia7712:
@g1geordie thanks for this patch. it looks great. The following links could use standard function also. please take a look.

1. https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/objects/preemption.go#L876
2. https://github.com/apache/yunikorn-core/blob/master/pkg/common/resources/resources.go#L453

g1geordie:
@chia7712  Thanks review, I fixed it. Please take a look :)

pbacsko (dismissed):
+1 LGTM

pkg/scheduler/objects/preemption.go
```diff
@@ -727,8 +727,8 @@ func (qps *QueuePreemptionSnapshot) IsWithinGuaranteedResource() bool {
 	if qps.Leaf && guaranteed.IsEmpty() {
 		return false
 	}
-	max := qps.GetMaxResource()
-	absGuaranteed := resources.ComponentWiseMinPermissive(guaranteed, max)
+	maxResource := qps.GetMaxResource()
```
chia7712, line 730: this change is related to IDE warnings ?

pkg/scheduler/objects/preemption.go
```diff
@@ -873,10 +873,7 @@ func preemptPredicateCheck(plugin api.ResourceManagerCallback, ch chan<- *predic
 func batchPreemptionChecks(checks []*si.PreemptionPredicatesArgs, batchSize int) [][]*si.PreemptionPredicatesArgs {
 	var result [][]*si.PreemptionPredicatesArgs
 	for i := 0; i < len(checks); i += batchSize {
-		end := i + batchSize
-		if end > len(checks) {
-			end = len(checks)
-		}
+		end := min(i+batchSize, len(checks))
```
chia7712, line 876: `i + batchSize`

chia7712 (commented):
@g1geordie thanks for this patch. two trivial comments below. please take a look.

pkg/scheduler/objects/preemption.go
craigcondit: This change is removing a call to ComponentWiseMinPermissive, which is incorrect.

craigcondit (changes requested):
See comments on preemption changes.

pkg/scheduler/objects/preemption.go
g1geordie: yes, it's IDE warning.
```
Warning:(730, 2) Variable 'max' collides with the 'builtin' function
```

g1geordie: > This change is removing a call to ComponentWiseMinPermissive, which is incorrect.

I just do a rename by IDE, the call is still like before.

pkg/scheduler/objects/preemption.go
g1geordie: I wanna ask. Is there some format command we can call?
I use Intellij format command to format and also notice it is a littlte different. 
So I try to search `i+` , I see 
image
The format don't have space too, so I ..

chia7712: it seems there is no strict format :_

chia7712 (approved):
LGTM

will merge it later

craigcondit:
@chia7712 Do not merge this as-is. It breaks preemption semantics (see previous comments).

pkg/scheduler/objects/preemption.go
craigcondit: Sorry, missed that it was showing up under the comments.

craigcondit (approved):
+1 LGTM. Sorry about the false alarm on conflicting changes.

chia7712:
> Sorry about the false alarm on conflicting changes.

thanks for confirmation!

PR #782, opened 2024-02-02
[YUNIKORN-2375] Refactor mock callback plugin

wilfred-s:
### What is this PR for?
Move the mock code for the resource manager callback out of the test directory into its own top level directory.
Removes imports of scheduler test code outside the scheduler package.

### What type of PR is it?
* [X] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2375

Tests are using the new mock, no change in coverage

### DIFF
--- a/pkg/events/event_publisher_test.go
+++ b/pkg/events/event_publisher_test.go
@@ -19,58 +19,16 @@
 package events
 
 import (
-	"fmt"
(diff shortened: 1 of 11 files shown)


manirajv06 (approved):
looks good.

PR #784, opened 2024-02-02
[YUNIKORN-2370] Proper event handling for failed headroom checks

pbacsko:
### What is this PR for?
Send proper events whenever a queue/user quota check fails. Currently, we send only one per application per second. So, in case of multiple pods, only the first one will be updated with a new event. The message is also incorrect, because it's a request that exceeds the quota, not the application.

New events:
- Queue headroom check failed (once per pod/ask)
- User quota check failed (once per pod/ask)
- Queue headroom check passed if previously failed (once per pod/ask) 
- User quota check passed if previously failed (once per pod/ask) 
- Predicate failed (once in every 15 seconds)

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2370

### Questions:

### DIFF
--- a/pkg/scheduler/objects/ask_events.go
+++ b/pkg/scheduler/objects/ask_events.go
@@ -35,7 +35,7 @@ type askEvents struct {
 	limiter     *rate.Limiter
 }
 
-func (ae *askEvents) sendRequestDoesNotFitInQueue(headroom *resources.Resource, queuePath string) {
+func (ae *askEvents) sendRequestExceedsQueueHeadroom(headroom *resources.Resource, queuePath string) {
 	if !ae.eventSystem.IsEventTrackingEnabled() {
 		return
 	}
--- a/pkg/scheduler/objects/ask_events.go
+++ b/pkg/scheduler/objects/ask_events.go
@@ -1,0 +1,93 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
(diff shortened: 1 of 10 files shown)


pbacsko:
Note to self: enhance coverage.

1. TestRequestDoesNotFitQueueEvents: implement the same for user quota
2. TestPreconditions: check allocation log as well

manirajv06 (approved):
Looks good. Minor nit.

pkg/scheduler/objects/ask_events.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// Ask-specific events. These events are of REQUEST type, so they are eventually sent to the respective pods in K8s.
+type askEvents struct {
+	eventSystem events.EventSystem
+	ask         *AllocationAsk
+	limiter     *rate.Limiter
+}
+
+func (ae *askEvents) sendRequestDoesNotFitInQueue(headroom *resources.Resource, queuePath string) {
```
manirajv06, line 38: Can we rename this to align with other method names? ex, sendRequestExceeds..
or we can follow FitIn or DoesNotFitIn for all methods.

manirajv06 (approved):

pkg/scheduler/objects/ask_events.go
pbacsko: Done

PR #791, opened 2024-02-07
[YUNIKORN-2397] Improve plugin registration unit test coverage

wilfred-s:
### What is this PR for?
New tests for plugin registration to cover all use cases:
- no interface
- single interface
- both interfaces
- multiple registrations each with one interface

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2397

change is new tests only

### DIFF
--- a/pkg/plugins/plugins_test.go
+++ b/pkg/plugins/plugins_test.go
@@ -23,11 +23,76 @@ import (
 
 	"gotest.tools/v3/assert"
 
-	"github.com/apache/yunikorn-core/pkg/mock"
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
+
+type NoPluginImplemented struct{}
+
+type RMPluginImplemented struct{}
+
+type StateDumpImplemented struct{}
+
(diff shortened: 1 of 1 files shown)


pkg/plugins/plugins_test.go
```diff
@@ -23,11 +23,76 @@ import (
...
 )
 
+type NoPluginImplemented struct{}
+
+type RMPluginImplemented struct{}
+
+type StateDumpImplemented struct{}
+
+type AllPluginsImplemented struct {
+	RMPluginImplemented
```
chia7712, line 36: why not reuse `mock.ResourceManagerCallback`?

wilfred-s: That will cause circular imports when we do that in other pieces of the code besides the fact that we need different implementations for each interface so we can register one interface at a time.

pbacsko (approved):
+1 LGTM

chia7712 (approved):
LGTM

PR #795, opened 2024-02-10
[YUNIKORN-2404] getApplicationHistory and getContainerHistory return empty array instead of null

lowc1012:
### What is this PR for?
ws/v1/history/apps and /ws/v1/history/containers should return empty array instead of null

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2404

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -532,6 +532,9 @@ func getApplicationHistory(w http.ResponseWriter, r *http.Request) {
 	// start and we cannot shortcut the loop using a break, we must finish iterating
 	records := imHistory.GetRecords()
 	result := getAppHistoryDAO(records)
+	if result == nil {
+		result = make([]*dao.ApplicationHistoryDAOInfo, 0)
+	}
 	if err := json.NewEncoder(w).Encode(result); err != nil {
 		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
 	}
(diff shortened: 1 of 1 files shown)


pkg/webservice/handlers.go
```diff
@@ -532,6 +532,9 @@ func getApplicationHistory(w http.ResponseWriter, r *http.Request) {
 	// start and we cannot shortcut the loop using a break, we must finish iterating
 	records := imHistory.GetRecords()
 	result := getAppHistoryDAO(records)
```
chia7712, line 534: how about changing the return value of `getAppHistoryDAO` from `nil` to empty array? The advantages of using empty array are shown below:

1. avoid ugly nil check
2. easy to test (just test the return value of function `getAppHistoryDAO`)

chia7712 (commented):
@lowc1012 thanks for this patch. one design suggestion please take a look. also please add unit test

pbacsko (approved):
+1

pbacsko (commented):
Nit: I already gave a +1 but this is something to consider:
```
result := make([]*dao.ApplicationHistoryDAOInfo, 0)
result = append(result, getAppHistoryDAO(records)...)
```
No branching, perhaps a bit more elegant.

EDIT: actually there's a one-liner:
```
result := append([]*dao.ApplicationDAOInfo{}, getAppHistoryDAO(records)...)
```

pkg/webservice/handlers.go
chia7712: > This can be accomplished in a follow-up I believe.

that is ok to me

@lowc1012 please file a jira to trace that refactor :)

chia7712:
> result := append([]*dao.ApplicationDAOInfo{}, getAppHistoryDAO(records)...)

It will create a new array and copy all elements into it, right? That should be one (major) reason that I preferred to refactor `getAppHistoryDAO` to avoid returning `nil` :)

pkg/webservice/handlers.go
lowc1012: @chia7712 @pbacsko
I created a new ticket.
https://issues.apache.org/jira/browse/YUNIKORN-2412

chia7712 (approved):
LGTM

PR #798, opened 2024-02-13
[YUNIKORN-2412] Avoid returning potential nil in handlers.go

lowc1012:
### What is this PR for?
Avoid returning potential nil for functions in handlers.go

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
[YUNIKORN-2412
](https://issues.apache.org/jira/browse/YUNIKORN-2412)
### Questions:


chia7712 (commented):
Could you add UT for them? Also please update docs for those functions

lowc1012:
> Could you add UT for them? Also please update docs for those functions

okay, working on it

lowc1012:
@chia7712 Regarding these changes, I think no need to modify the REST APIs docs.

chia7712:
@lowc1012 I just merge #795. Can you please rebase code and add UT for `getApplicationHistory` and `getContainerHistory`?

lowc1012:
> @lowc1012 I just merge #795. Can you please rebase code and add UT for `getApplicationHistory` and `getContainerHistory`?

Thanks @chia7712 for your review.
IIUC, Do the following codes in tests cover my code changes already?

```go
// init should return null and thus no records
imHistory = history.NewInternalMetricsHistory(5)
resp = &MockResponseWriter{}
getApplicationHistory(resp, req)
var appHist []dao.ApplicationHistoryDAOInfo
err = json.Unmarshal(resp.outputBytes, &appHist)
assert.Equal(t, resp.statusCode, 0, "app response should have no status")
assert.Equal(t, len(appHist), 0, "empty response must have no records")

```
https://github.com/apache/yunikorn-core/blob/master/pkg/webservice/handlers_test.go#L385-L393
```go
// init should return null and thus no records
imHistory = history.NewInternalMetricsHistory(5)
resp = &MockResponseWriter{}
getContainerHistory(resp, req)
var contHist []dao.ContainerHistoryDAOInfo
err = json.Unmarshal(resp.outputBytes, &contHist)
assert.Equal(t, resp.statusCode, 0, "container response should have no status")
assert.Equal(t, len(contHist), 0, "empty response must have no records")
```
https://github.com/apache/yunikorn-core/blob/master/pkg/webservice/handlers_test.go#L440-L448

chia7712:
> IIUC, Do the following codes in tests cover my code changes already?

I don't think so. We should prevent `getContainerHistory`  or `getAppHistoryDAO` from returning `nil`. The test you attached does not check that. `len(contHist)` return 0 even though `contHist` is nil

pkg/webservice/handlers_test.go
```diff
@@ -374,6 +374,8 @@ func TestApplicationHistory(t *testing.T) {
 	resp := &MockResponseWriter{}
 	// no init should return nothing
 	getApplicationHistory(resp, req)
+	var appHist []dao.ApplicationHistoryDAOInfo
+	assert.Equal(t, len(appHist), 0)
```
chia7712, line 378: This `assert` is redundant as it test only golang itself :)

pkg/webservice/handlers_test.go
```diff
@@ -386,8 +388,7 @@ func TestApplicationHistory(t *testing.T) {
 	imHistory = history.NewInternalMetricsHistory(5)
 	resp = &MockResponseWriter{}
 	getApplicationHistory(resp, req)
-	var appHist []dao.ApplicationHistoryDAOInfo
-	err = json.Unmarshal(resp.outputBytes, &appHist)
```
chia7712, line 390: this is required test since we should make sure `appHist` is not nil. 

`assert.Equal(t, len(appHist), 0, "empty response must have no records")` this test pass even though `appHist` is nil, and so we should add new `assert` for nil case

pkg/webservice/handlers_test.go
```diff
@@ -921,6 +925,19 @@ func getNodesUtilByType(t *testing.T, nodesUtilList []*dao.NodesUtilDAOInfo, res
...
+	var req *http.Request
+	req, err := http.NewRequest("GET", "/ws/v1/partitions", strings.NewReader(""))
+	assert.NilError(t, err, "App Handler request failed")
+
+	resp := &MockResponseWriter{}
+	var partitionInfo []*dao.PartitionInfo
+	getPartitions(resp, req)
+	err = json.Unmarshal(resp.outputBytes, &partitionInfo)
+	assert.NilError(t, err, unmarshalError)
+	assert.Equal(t, len(partitionInfo), 0)
```
chia7712, line 939: ditto, please ensure `partitionInfo` is not nil

chia7712 (commented):
@lowc1012 thanks for updating patch. a couple of comments below

pkg/webservice/handlers_test.go
ryanatfazz: done. check contHist too

pkg/webservice/handlers_test.go
ryanatfazz: done

chia7712 (approved):
LGTM

PR #800, opened 2024-02-13
[YUNIKORN-2116] Track user/group events

pbacsko:
### What is this PR for?
Generate group and user quota related events.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2116

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -2108,7 +2108,7 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
 
 	if sa.runnableByUserLimit != runnableByUserLimit {
 		if runnableByUserLimit {
-			log.Log(log.SchedApplication).Info("Application is now runnable based on user quota",
+			log.Log(log.SchedApplication).Info("Application is now runnable based on user/group quota",
 				zap.String("appID", sa.ApplicationID),
 				zap.String("queue", sa.queuePath),
 				zap.String("user", sa.user.User),
@@ -2117,7 +2117,9 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
 		} else {
 			log.Log(log.SchedApplication).Info("Maximum number of running applications reached the user/group limit",
 				zap.String("appID", sa.ApplicationID),
-				zap.String("queue", sa.queuePath))
+				zap.String("queue", sa.queuePath),
+				zap.String("user", sa.user.User),
+				zap.Strings("groups", sa.user.Groups))
 			sa.appEvents.sendAppNotRunnableQuotaEvent()
 		}
 	}
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -112,6 +112,8 @@ type Application struct {
 	placeholderData      map[string]*PlaceholderData // track placeholder and gang related info
 	askMaxPriority       int32                       // highest priority value of outstanding asks
 	hasPlaceholderAlloc  bool                        // Whether there is at least one allocated placeholder
+	runnableInQueue      bool                        // whether the application is runnable/schedulable in the queue. Default is true.
+	runnableByUserLimit  bool                        // whether the application is runnable/schedulable based on user/group quota. Default is true.
 
 	rmEventHandler        handler.EventHandler
 	rmID                  string
@@ -171,6 +173,8 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
 		askMaxPriority:        configs.MinPriority,
 		sortedRequests:        sortedRequests{},
 		sendStateChangeEvents: true,
+		runnableByUserLimit:   true,
+		runnableInQueue:       true,
 	}
 	placeholderTimeout := common.ConvertSITimeoutWithAdjustment(siApp, defaultPlaceholderTimeout)
 	gangSchedStyle := siApp.GetGangSchedulingStyle()
@@ -2082,4 +2086,41 @@ func getRateLimitedAppLog() *log.RateLimitedLogger {
 		rateLimitedAppLog = log.NewRateLimitedLogger(log.SchedApplication, time.Second)
 	})
 	return rateLimitedAppLog
+}
+
+func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit bool) {
+	sa.Lock()
+	defer sa.Unlock()
+	if sa.runnableInQueue != runnableInQueue {
+		sa.runnableInQueue = runnableInQueue
+		if sa.runnableInQueue {
+			log.Log(log.SchedApplication).Info("Application is now runnable in queue",
+				zap.String("appID", sa.ApplicationID),
+				zap.String("queue", sa.queuePath))
+			sa.appEvents.sendAppRunnableInQueueEvent()
+		} else {
+			log.Log(log.SchedApplication).Info("Maximum number of running applications reached the queue limit",
+				zap.String("appID", sa.ApplicationID),
+				zap.String("queue", sa.queuePath))
+			sa.appEvents.sendAppNotRunnableInQueueEvent()
+		}
+	}
+
+	if sa.runnableByUserLimit != runnableByUserLimit {
+		sa.runnableByUserLimit = runnableByUserLimit
+		if sa.runnableByUserLimit {
+			log.Log(log.SchedApplication).Info("Application is now runnable based on user quota",
+				zap.String("appID", sa.ApplicationID),
+				zap.String("queue", sa.queuePath))
+			sa.appEvents.sendAppRunnableQuotaEvent()
+		} else {
+			log.Log(log.SchedApplication).Info("Maximum number of running applications reached the user/group limit",
+				zap.String("appID", sa.ApplicationID),
+				zap.String("queue", sa.queuePath))
+			sa.appEvents.sendAppNotRunnableQuotaEvent()
+		}
+	}
+
+	sa.runnableInQueue = runnableInQueue
+	sa.runnableByUserLimit = runnableByUserLimit
 }
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -2092,8 +2092,7 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
 	sa.Lock()
 	defer sa.Unlock()
 	if sa.runnableInQueue != runnableInQueue {
-		sa.runnableInQueue = runnableInQueue
-		if sa.runnableInQueue {
+		if runnableInQueue {
 			log.Log(log.SchedApplication).Info("Application is now runnable in queue",
 				zap.String("appID", sa.ApplicationID),
 				zap.String("queue", sa.queuePath))
@@ -2105,10 +2104,10 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
 			sa.appEvents.sendAppNotRunnableInQueueEvent()
 		}
 	}
+	sa.runnableInQueue = runnableInQueue
 
 	if sa.runnableByUserLimit != runnableByUserLimit {
-		sa.runnableByUserLimit = runnableByUserLimit
-		if sa.runnableByUserLimit {
+		if runnableByUserLimit {
 			log.Log(log.SchedApplication).Info("Application is now runnable based on user quota",
 				zap.String("appID", sa.ApplicationID),
 				zap.String("queue", sa.queuePath),
@@ -2122,7 +2121,5 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
 			sa.appEvents.sendAppNotRunnableQuotaEvent()
 		}
 	}
-
-	sa.runnableInQueue = runnableInQueue
 	sa.runnableByUserLimit = runnableByUserLimit
 }
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -2111,7 +2111,9 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
 		if sa.runnableByUserLimit {
 			log.Log(log.SchedApplication).Info("Application is now runnable based on user quota",
 				zap.String("appID", sa.ApplicationID),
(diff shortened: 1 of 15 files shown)


pbacsko:
TODO:
* clarify limit/set remove events
* manager unit test to verify events

pkg/scheduler/ugm/group_tracker_test.go
```diff
@@ -172,32 +174,49 @@ func TestGTSetMaxLimits(t *testing.T) {
...
+
+	groupTracker.increaseTrackedResource(path1, TestApp1, usage, user.User)
+	assert.Equal(t, 1, len(eventSystem.Events))
+	assert.Equal(t, si.EventRecord_UG_GROUP_RESOURCE, eventSystem.Events[0].EventChangeDetail)
+
+	eventSystem.Reset()
+	groupTracker.decreaseTrackedResource(path1, TestApp1, usage, true)
+	assert.Equal(t, 1, len(eventSystem.Events))
+	assert.Equal(t, si.EventRecord_UG_GROUP_RESOURCE, eventSystem.Events[0].EventChangeDetail)
+}
```
manirajv06, line 218: These asserts(Just 2 lines !) can be added to the existing test cases itself rather than introducing a new method.

pkg/scheduler/ugm/manager.go
```diff
@@ -477,13 +480,15 @@ func (m *Manager) clearEarlierSetUserLimits(newUserLimits map[string]map[string]
...
-func (m *Manager) resetUserEarlierUsage(ut *UserTracker, hierarchy []string) {
+func (m *Manager) resetUserEarlierUsage(ut *UserTracker, queuePath string) {
 	// Is this user already tracked for the queue path?
+	hierarchy := strings.Split(queuePath, configs.DOT)
 	if ut.IsQueuePathTrackedCompletely(hierarchy) {
 		log.Log(log.SchedUGM).Debug("Need to clear earlier set configs for user",
 			zap.String("user", ut.userName),
 			zap.Strings("queue path", hierarchy))
 		ut.setLimits(hierarchy, nil, 0, false, false)
+		m.events.sendLimitSetForUser(ut.userName, queuePath)
```
manirajv06, line 491: As discussed earlier, limits would be removed or reset to the default values. Earlier configured values are cleared now. So, limits removal event.

pkg/scheduler/ugm/manager.go
```diff
@@ -537,6 +542,7 @@ func (m *Manager) resetGroupEarlierUsage(gt *GroupTracker, hierarchy []string) {
 			delete(ut.appGroupTrackers, app)
 		}
 		gt.setLimits(hierarchy, nil, 0)
+		m.events.sendLimitSetForGroup(gt.groupName, queuePath)
```
manirajv06, line 545: Same comment

pkg/scheduler/ugm/user_tracker_test.go
```diff
@@ -196,32 +204,59 @@ func TestSetMaxLimits(t *testing.T) {
...
+	assert.Equal(t, 1, len(eventSystem.Events))
+
+	// removing app
+	eventSystem.Reset()
+	userTracker.decreaseTrackedResource(path1, TestApp1, usage, true)
+	assert.Equal(t, 2, len(eventSystem.Events))
+	assert.Equal(t, si.EventRecord_UG_USER_RESOURCE, eventSystem.Events[0].EventChangeDetail)
+	assert.Equal(t, si.EventRecord_UG_APP_LINK, eventSystem.Events[1].EventChangeDetail)
+}
+
```
manirajv06, line 259: Same comment like group tracker test

manirajv06 (changes requested):

pkg/scheduler/ugm/manager.go
pbacsko: All events are sent from the trackers from now on.

pkg/scheduler/ugm/user_tracker_test.go
pbacsko: All events are sent from the trackers from now on.

pbacsko: Done

pkg/scheduler/ugm/group_tracker_test.go
pbacsko: Done

pkg/scheduler/ugm/group_tracker_test.go
```diff
@@ -158,44 +171,98 @@ func TestGTDecreaseTrackedResource(t *testing.T) {
...
 	if !decreased {
-		t.Fatalf("unable to decrease tracked resource: queuepath %+q, app %s, res %v, error %t", hierarchy2, TestApp2, usage2, err)
+		t.Fatalf("unable to decrease tracked resource: queuepath %s, app %s, res %v, error %t", path2, TestApp2, usage2, err)
 	}
 	assert.Equal(t, 0, len(groupTracker.getTrackedApplications()))
 	assert.Equal(t, removeQT, true, "wrong remove queue tracker value")
 }
 
-func TestGTSetMaxLimits(t *testing.T) {
+func TestGTSetAndClearMaxLimits(t *testing.T) {
```
pbacsko, line 182: Note: original unit test did not properly verify the behavior of the tracker, it only checked the return value of the increment which is always true (this should be changed in a follow-up JIRA). Now we canRunApp & headroom is validated with different settings.

pkg/scheduler/ugm/user_tracker_test.go
```diff
@@ -171,55 +189,115 @@ func TestDecreaseTrackedResource(t *testing.T) {
...
 	if !decreased {
-		t.Fatalf("unable to increase tracked resource: queuepath %+q, app %s, res %v, error %t", hierarchy2, TestApp2, usage2, err)
+		t.Fatalf("unable to increase tracked resource: queuepath %s, app %s, res %v, error %t", path2, TestApp2, usage2, err)
 	}
 	assert.Equal(t, 0, len(userTracker.getTrackedApplications()))
 	assert.Equal(t, removeQT, true, "wrong remove queue tracker value")
 }
 
-func TestSetMaxLimits(t *testing.T) {
+func TestSetAndClearMaxLimits(t *testing.T) {
```
pbacsko, line 217: Note: original unit test did not properly verify the behavior of the tracker, it only checked the return value of the increment which is always true (this should be changed in a follow-up JIRA). Now both canRunApp & headroom is validated with different settings.

pkg/scheduler/ugm/manager.go
```diff
@@ -414,7 +416,7 @@ func (m *Manager) clearEarlierSetUserWildCardLimits(newUserWildCardLimits map[st
 					log.Log(log.SchedUGM).Debug("Need to clear earlier set configs for user because wild card limit has been applied earlier",
 						zap.String("user", ut.userName),
 						zap.String("queue path", queuePath))
-					ut.setLimits(hierarchy, nil, 0, false, true)
+					ut.clearLimits(queuePath)
```
manirajv06, line 419: 4th & 5th arguments in setLimits() has significance. It can't be ignored.

pkg/scheduler/ugm/manager.go
```diff
@@ -477,13 +478,14 @@ func (m *Manager) clearEarlierSetUserLimits(newUserLimits map[string]map[string]
...
-func (m *Manager) resetUserEarlierUsage(ut *UserTracker, hierarchy []string) {
+func (m *Manager) resetUserEarlierUsage(ut *UserTracker, queuePath string) {
 	// Is this user already tracked for the queue path?
+	hierarchy := strings.Split(queuePath, configs.DOT)
 	if ut.IsQueuePathTrackedCompletely(hierarchy) {
 		log.Log(log.SchedUGM).Debug("Need to clear earlier set configs for user",
 			zap.String("user", ut.userName),
 			zap.Strings("queue path", hierarchy))
-		ut.setLimits(hierarchy, nil, 0, false, false)
+		ut.clearLimits(queuePath)
```
manirajv06, line 488: Same comment

manirajv06 (changes requested):
BTW, Do we need to send event when user or group cleared (once done with the last application) from the manager?

pkg/scheduler/ugm/manager.go
pbacsko: In both cases, `useWildCard` = false, is that correct?

pkg/scheduler/ugm/manager.go
pbacsko: I believe this should have been caught by unit tests. I'll file a JIRA later to properly cover setLimits/clearLimits.

pkg/scheduler/objects/application_events.go
```diff
@@ -83,31 +83,63 @@ func (evt *applicationEvents) sendRemoveAskEvent(request *AllocationAsk, detail
...
+	}
+	event := events.CreateAppEventRecord(evt.app.ApplicationID, common.Empty, common.Empty, si.EventRecord_NONE, si.EventRecord_APP_CANNOTRUN_QUEUE, nil)
+	evt.eventSystem.AddEvent(event)
+}
+
+func (evt *applicationEvents) sendAppRunnableInQueueEvent() {
+	if !evt.eventSystem.IsEventTrackingEnabled() {
+		return
+	}
+	event := events.CreateAppEventRecord(evt.app.ApplicationID, common.Empty, common.Empty, si.EventRecord_NONE, si.EventRecord_APP_RUNNABLE_QUEUE, nil)
```
chia7712, line 126: how about replacing `nil` by `resources.Zero`?

pkg/scheduler/objects/application.go
```diff
@@ -2083,3 +2087,40 @@ func getRateLimitedAppLog() *log.RateLimitedLogger {
 	})
 	return rateLimitedAppLog
 }
+
+func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit bool) {
+	sa.Lock()
+	defer sa.Unlock()
+	if sa.runnableInQueue != runnableInQueue {
+		sa.runnableInQueue = runnableInQueue
+		if sa.runnableInQueue {
```
chia7712, line 2096: we can use `runnableInQueue` to replace `sa.runnableInQueue` here, and then we can remove line #2095 to avoid duplicate assignment (`sa.runnableInQueue = runnableInQueue`)

pkg/scheduler/ugm/user_tracker.go
```diff
@@ -97,6 +109,9 @@ func (ut *UserTracker) hasGroupForApp(applicationID string) bool {
 func (ut *UserTracker) setGroupForApp(applicationID string, groupTrack *GroupTracker) {
 	ut.Lock()
 	defer ut.Unlock()
+	if groupTrack != nil {
+		ut.events.sendAppGroupLinked(applicationID, groupTrack.groupName)
```
chia7712, line 113: the first parameter should be `group`, and second is `applicationID`

chia7712 (commented):
@pbacsko thanks for this nice feature! a couple of comments below. please take a look.

pkg/scheduler/ugm/manager.go
manirajv06: Correct, but 5th argument differs for a reason. These changes has been added very recently, might lack some test coverage. You can file a jira for unit test coverage later, but not for the actual code changes. We should retain the behaviour as is to avoid unnecessary issues.

pkg/scheduler/objects/application_events.go
pbacsko: nil is used consistently everywhere, but both result in the same when EventRecord is constructed, ie. empty map.

pkg/scheduler/ugm/user_tracker.go
pbacsko: Good catch... extended the test to validate this

pkg/scheduler/objects/application.go
```diff
@@ -2083,3 +2087,40 @@ func getRateLimitedAppLog() *log.RateLimitedLogger {
...
+			sa.appEvents.sendAppNotRunnableInQueueEvent()
+		}
+	}
+
+	if sa.runnableByUserLimit != runnableByUserLimit {
+		sa.runnableByUserLimit = runnableByUserLimit
+		if sa.runnableByUserLimit {
+			log.Log(log.SchedApplication).Info("Application is now runnable based on user quota",
+				zap.String("appID", sa.ApplicationID),
+				zap.String("queue", sa.queuePath))
```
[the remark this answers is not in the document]

pkg/scheduler/objects/application.go
chia7712: for example:
```go
func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit bool) {
	sa.Lock()
	defer sa.Unlock()
	if sa.runnableInQueue != runnableInQueue {
		if runnableInQueue {
			log.Log(log.SchedApplication).Info("Application is now runnable in queue",
				zap.String("appID", sa.ApplicationID),
				zap.String("queue", sa.queuePath))
			sa.appEvents.sendAppRunnableInQueueEvent()
		} else {
			log.Log(log.SchedApplication).Info("Maximum number of running applications reached the queue limit",
				zap.String("appID", sa.ApplicationID),
				zap.String("queue", sa.queuePath))
			sa.appEvents.sendAppNotRunnableInQueueEvent()
		}
	}

	if sa.runnableByUserLimit != runnableByUserLimit {
		if runnableByUserLimit {
			log.Log(log.SchedApplication).Info("Application is now runnable based on user quota",
				zap.String("appID", sa.ApplicationID),
				zap.String("queue", sa.queuePath),
				zap.String("user", sa.user.User),
				zap.Strings("groups", sa.user.Groups))
			sa.appEvents.sendAppRunnableQuotaEvent()
		} else {
			log.Log(log.SchedApplication).Info("Maximum number of running applications reached the user/group limit",
				zap.String("appID", sa.ApplicationID),
				zap.String("queue", sa.queuePath))
			sa.appEvents.sendAppNotRunnableQuotaEvent()
		}
	}
	sa.runnableInQueue = runnableInQueue
	sa.runnableByUserLimit = runnableByUserLimit
}
```

chia7712 (approved):
LGTM

wilfred-s (approved):
LGTM +1

PR #801, opened 2024-02-13
[YUNIKORN-2406] Ensure shim is notified of existing allocations being added

craigcondit:
### What is this PR for?

When existing allocations are registered from the shim to the core, they are added to the partition, but the shim is never notified that the allocations were successfully registered. This manifests in the core as missing history for container allocations, and in the shim by tasks not transitioning to a bound state.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2406

E2e testing only; this code is deeply tied to the shim implementation.

### Questions:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ module github.com/apache/yunikorn-core
 go 1.21
 
 require (
-	github.com/apache/yunikorn-scheduler-interface v0.0.0-20231211235204-ec7bfad7d00e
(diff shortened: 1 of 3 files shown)


chia7712 (approved):
ship it!!!

PR #804, opened 2024-02-16
[YUNIKORN-2326] get apps by Internal state

priyansh17:
### What is this PR for?
`getAppsByState`, `getRejectedAppsByState`, and `getCompletedAppsByState` do the similar job that get apps according to specific state. Hence, it would be better to have a helper method and then these three methods can reuse the code.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [x] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2326

The same older tests should work for this.

### Questions:

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -1070,7 +1070,7 @@ func (pc *PartitionContext) GetRejectedApplications() []*objects.Application {
 	return appList
 }
 
-func (pc *PartitionContext) getAppsByInternalState(appMap map[string]*objects.Application, state string) []string {
+func (pc *PartitionContext) getAppsState(appMap map[string]*objects.Application, state string) []string {
 	pc.RLock()
 	defer pc.RUnlock()
 	var apps []string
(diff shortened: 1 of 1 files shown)


pkg/scheduler/partition.go
```diff
@@ -1070,44 +1070,32 @@ func (pc *PartitionContext) GetRejectedApplications() []*objects.Application {
 	return appList
 }
 
-// getAppsByState returns a slice of applicationIDs for the current applications filtered by state
-// Completed and Rejected applications are tracked in a separate map and will never be included.
-func (pc *PartitionContext) getAppsByState(state string) []string {
+func (pc *PartitionContext) getAppsByInternalState(appMap map[string]*objects.Application, state string) []string {
 	pc.RLock()
 	defer pc.RUnlock()
 	var apps []string
```
chia7712, line 1076: Could we avoid returning `nil`?

pkg/scheduler/partition.go
```diff
@@ -1070,44 +1070,32 @@ func (pc *PartitionContext) GetRejectedApplications() []*objects.Application {
 	return appList
 }
 
-// getAppsByState returns a slice of applicationIDs for the current applications filtered by state
-// Completed and Rejected applications are tracked in a separate map and will never be included.
-func (pc *PartitionContext) getAppsByState(state string) []string {
+func (pc *PartitionContext) getAppsByInternalState(appMap map[string]*objects.Application, state string) []string {
```
chia7712, line 1073: the name `getAppsState`  should be good here since there is no such internal state.

chia7712 (commented):
@priyansh17 thanks for contribution. two minor comments below. please take a look!

pkg/scheduler/partition.go
priyansh17: Sure, updated the function name to ```getAppsState```.

pkg/scheduler/partition.go
priyansh17: declaring ```apps := []string{}``` should always return an empty list of string in cases appMap is empty.

chia7712 (approved):
LGTM

PR #806, opened 2024-02-18
[YUNIKORN-2413] Initialism fixes for allocationId, clusterId

priyansh17:
### What is this PR for?
Yunikorn-core has mixed up "Id" and "ID" in code base, and it's better to standardize the use of acronyms and initialisms according to [this doc.](https://go.dev/wiki/CodeReviewComments#initialisms)

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [X] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2413
https://issues.apache.org/jira/browse/YUNIKORN-2405

The older tests written would work for this.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/node.go
+++ b/pkg/scheduler/objects/node.go
@@ -360,8 +360,8 @@ func (sn *Node) ReplaceAllocation(allocationID string, replace *Allocation, delt
 	sn.availableResource.SubFrom(delta)
 	if !before.FitIn(sn.allocatedResource) {
 		log.Log(log.SchedNode).Warn("unexpected increase in node usage after placeholder replacement",
-			zap.String("placeholder allocationId", allocationID),
-			zap.String("allocation allocationId", replace.GetAllocationID()),
+			zap.String("placeholder allocationID", allocationID),
+			zap.String("allocation allocationID", replace.GetAllocationID()),
 			zap.Stringer("delta", delta))
 	}
 }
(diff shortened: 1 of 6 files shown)


chia7712:
@priyansh17 thanks for this patch. The renaming looks good to me. btw, could you rename functions also? for example:

https://github.com/apache/yunikorn-core/blob/master/pkg/events/event_ringbuffer.go#L206
https://github.com/apache/yunikorn-core/blob/master/pkg/log/logger_test.go#L38
https://github.com/apache/yunikorn-core/blob/master/pkg/common/security/usergroup_os_resolver.go#L39

thanks!

wilfred-s (approved):

wilfred-s:
Lets tackle the remaining two function names, log and ringbuffer, in one of our next changes. We will have updates to those files regularly and can just fix it when we do that.

For [/pkg/common/security/usergroup_os_resolver.go#L39](https://github.com/apache/yunikorn-core/blob/master/pkg/common/security/usergroup_os_resolver.go#L39) I would not change it. The OS library uses the same upper case setup as we have now. Seems to me that we should align with that.

chia7712:
> Lets tackle the remaining two function names, log and ringbuffer, in one of our next changes.

https://issues.apache.org/jira/browse/YUNIKORN-2450

PR #807, opened 2024-02-20
[YUNIKORN-2427] Use r-lock instead of rw-lock in user_tracker.go#getGroupForApp

chenyulin0719:
### What is this PR for?
Use r-lock instead of rw-lock in user_tracker.go#getGroupForApp.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2427

Run e2e test, it passed in my forked repo
- https://github.com/chenyulin0719/yunikorn-k8shim/pull/36

NA

### Questions:
NA

### DIFF
--- a/pkg/scheduler/ugm/group_tracker.go
+++ b/pkg/scheduler/ugm/group_tracker.go
@@ -111,8 +111,8 @@ func (gt *GroupTracker) IsUnlinkRequired(hierarchy []string) bool {
 }
 
 func (gt *GroupTracker) UnlinkQT(hierarchy []string) bool {
-	gt.RLock()
-	defer gt.RUnlock()
+	gt.Lock()
+	defer gt.Unlock()
 	return gt.queueTracker.UnlinkQT(hierarchy)
 }
 
--- a/pkg/scheduler/ugm/user_tracker.go
+++ b/pkg/scheduler/ugm/user_tracker.go
@@ -101,8 +101,8 @@ func (ut *UserTracker) setGroupForApp(applicationID string, groupTrack *GroupTra
 }
 
 func (ut *UserTracker) getGroupForApp(applicationID string) string {
-	ut.Lock()
-	defer ut.Unlock()
+	ut.RLock()
+	defer ut.RUnlock()
 	if ut.appGroupTrackers[applicationID] != nil {
 		return ut.appGroupTrackers[applicationID].groupName
 	}
--- a/pkg/scheduler/ugm/user_tracker.go
+++ b/pkg/scheduler/ugm/user_tracker.go
@@ -156,8 +156,8 @@ func (ut *UserTracker) IsUnlinkRequired(hierarchy []string) bool {
 }
 
 func (ut *UserTracker) UnlinkQT(hierarchy []string) bool {
-	ut.RLock()
-	defer ut.RUnlock()
+	ut.Lock()
+	defer ut.Unlock()
 	return ut.queueTracker.UnlinkQT(hierarchy)
 }
 


manirajv06:
Can you do a general check in ugm module? I can see r lock being used in UnlinkQT(), but should be using rw lock.

chenyulin0719:
@manirajv06,
OK, let me take a review for the ugm module.

pbacsko (approved):
+1

manirajv06:
@chenyulin0719 We are planning to include this in upcoming release 1.5.0. Since we are nearing the date, Can you please prioritise this?

chenyulin0719:
Hi @manirajv06,
I've reviewed the ugm locks and have a question for the current locks hierarchy.

For below functions and the related locks:
- group_tracker.canRunApp() -> WLock
- group_tracker.headroom() -> WLock
- group_tracker.UnlinkQT() -> RLock, should change to WLock
- user_tracker.canRunApp() -> WLock
- user_tracker.headroom() -> WLock
- user_tracker.UnlinkQT() -> RLock, should change to WLock

Those functions mutated queue_tracker's value.  But locks are controlled in the higher level(group_tracker/user_tracker) instead of queue_tracker. So my first thought is we should put down the locks to queueTracker.

However, after rethink about it. group_tracker/user_tracker is the only entrance to motidy it's own queueTracker.  
So I guess that the reason to put locks in higer level is to centralize the locks controller. 
It'll be more easier for managing locks.

Please correct me if I'm wrong.
Have checked other locks under ugm module are correct, will add commit to fix UnlinkQT by today.

manirajv06:
Your understanding is correct. Please make changes as discussed.

chenyulin0719:
Just pushed a commit to fix UnlinkQT() lock. Waiting for e2e test pass:
- https://github.com/chenyulin0719/yunikorn-k8shim/pull/36/checks

chia7712:
> group_tracker.canRunApp() -> WLock
user_tracker.canRunApp() -> WLock

Maybe it is worth having a jira to trace it. It seems to me `canRunApp` is a read-only function.

manirajv06:
> Maybe it is worth having a jira to trace it. It seems to me canRunApp is a read-only function.

@chenyulin0719 Can you take care of this? How about headroom? I think we need to change there as well.

chia7712 (approved):
LGTM

chia7712:
@manirajv06 please merge it if you have no objection :)

pbacsko (approved):
+1

manirajv06:
Addendum commit - https://github.com/apache/yunikorn-core/pull/809

chenyulin0719:
> Maybe it is worth having a jira to trace it. It seems to me canRunApp is a read-only function.
> @chenyulin0719 Can you take care of this? How about headroom? I think we need to change there as well.

Sorry for late reply. 
headroom() and canRunApp() are not read-only function because they will recursively create children queueTracker if it does not exist. 

The function name is confusing, we should create a Jira to refactor it or rename the function.

chenyulin0719:
> We are on the same page. Please file jira for it :)

After reviewed the queueTracker implementation, it's quite difficult to rewrite canRunApp/headroom since they are currently implemented recursively. (I can't find a better name for it either. The main purpose is to check if we can run app and check Headroom size.)

So I'll add comment for group/user tracker to ensure other reviewer won't get confused about it in [YUNIKORN-2445](https://issues.apache.org/jira/browse/YUNIKORN-2445).

PR #809, opened 2024-02-21
[YUNIKORN-2427] addendum commit

manirajv06:
### What is this PR for?
Addendum commit   

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2427

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/user_tracker.go
+++ b/pkg/scheduler/ugm/user_tracker.go
@@ -122,8 +122,8 @@ func (ut *UserTracker) setLimits(hierarchy []string, resource *resources.Resourc
 }
 
 func (ut *UserTracker) headroom(hierarchy []string) *resources.Resource {
-	ut.Lock()
-	defer ut.Unlock()
+	ut.RLock()
+	defer ut.RUnlock()
 	return ut.queueTracker.headroom(hierarchy, user)
 }
 
@@ -168,7 +168,7 @@ func (ut *UserTracker) canBeRemoved() bool {
 }
 
 func (ut *UserTracker) canRunApp(hierarchy []string, applicationID string) bool {
-	ut.Lock()
-	defer ut.Unlock()
+	ut.RLock()
+	defer ut.RUnlock()
 	return ut.queueTracker.canRunApp(hierarchy, applicationID, user)
 }


pbacsko (approved):
+1

pkg/scheduler/ugm/user_tracker.go
```diff
@@ -168,7 +168,7 @@ func (ut *UserTracker) canBeRemoved() bool {
 }
 
 func (ut *UserTracker) canRunApp(hierarchy []string, applicationID string) bool {
-	ut.Lock()
-	defer ut.Unlock()
+	ut.RLock()
+	defer ut.RUnlock()
 	return ut.queueTracker.canRunApp(hierarchy, applicationID, user)
```
chia7712, line 173: https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/ugm/queue_tracker.go#L101

it would change something internally, so is it ok to use read-lock? if using read lock is safe, should we update group_tacker also ? see https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/ugm/group_tracker.go#L149

pbacsko: Ouch! Actually both `headroom()` and `canRunApp()` ends up modifying the map if the child tracker does not exist :( I was in a hurry to approve this change... @manirajv06 what do you think? As I can see it, `canRunApp()` is likely the first call which ends up creating the tracker hierarchy, unless there are limits. We need to go back to the Lock/Unlock pair.

chia7712: `headroom` has same issue. https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/ugm/queue_tracker.go#L210

Maybe we should either consider another naming or add some comments for `headroom` and `canRunApp`. It is error-prone to me.

manirajv06: reverted.

wilfred-s: Logged jira for documenting lock behaviour. When I designed the tracker objects I explicitly kept the QueueTracker lock free.

PR #812, opened 2024-02-21
[YUNIKORN-2432] Add unit test coverage for UserTracker/GroupTracker/QueueTracker.canRunApp()

pbacsko:
### What is this PR for?
Enhance unit test coverage for CanRunApp().

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2432

### Questions:


pkg/scheduler/ugm/user_tracker_test.go
```diff
@@ -302,6 +302,34 @@ func TestSetAndClearMaxLimits(t *testing.T) {
...
+
+	hierarchy1 := strings.Split(path1, configs.DOT)
+	assert.Assert(t, userTracker.canRunApp(hierarchy1, TestApp1))
+	// make sure wildcard limits are applied
+	assert.Equal(t, uint64(3), userTracker.queueTracker.childQueueTrackers["parent"].maxRunningApps)
+	assert.Assert(t, resources.Equals(maxRes, userTracker.queueTracker.childQueueTrackers["parent"].maxResources))
+
+	// limit hit
+	userTracker.setLimits(path1, resources.Zero, 1, false, false)
+	userTracker.increaseTrackedResource(path1, TestApp1, resources.NewResourceFromMap(map[string]resources.Quantity{
```
chia7712, line 326: This is unrelated to this PR, but It seems `increaseTrackedResource` always returns `true`, right? If so, we can remove the return value and simplify the impls for `QueueTracker`, `UserTracker`, and `GroupTracker`

pkg/scheduler/ugm/user_tracker_test.go
```diff
@@ -302,6 +302,34 @@ func TestSetAndClearMaxLimits(t *testing.T) {
...
+		maxApplications: 3,
+	} // manipulate map directly to avoid hierarchy creation
+
+	hierarchy1 := strings.Split(path1, configs.DOT)
+	assert.Assert(t, userTracker.canRunApp(hierarchy1, TestApp1))
+	// make sure wildcard limits are applied
+	assert.Equal(t, uint64(3), userTracker.queueTracker.childQueueTrackers["parent"].maxRunningApps)
+	assert.Assert(t, resources.Equals(maxRes, userTracker.queueTracker.childQueueTrackers["parent"].maxResources))
+
+	// limit hit
```
chia7712, line 324: Could you complete the comment about "which limit is hit"?

pkg/scheduler/ugm/user_tracker_test.go
```diff
@@ -302,6 +302,34 @@ func TestSetAndClearMaxLimits(t *testing.T) {
...
+	userTracker.setLimits(path1, resources.Zero, 1, false, false)
+	userTracker.increaseTrackedResource(path1, TestApp1, resources.NewResourceFromMap(map[string]resources.Quantity{
+		"cpu": 1000,
```
chia7712, line 327: The cpu resource is not a part of limit in this test case, right? If so, I'm a bit confused why `canRunApp` does not check max resource?

pkg/scheduler/ugm/user_tracker_test.go
```diff
@@ -302,6 +302,34 @@ func TestSetAndClearMaxLimits(t *testing.T) {
...
+	manager.userWildCardLimitsConfig["root.parent"] = &LimitConfig{
+		maxResources:    maxRes,
+		maxApplications: 3,
+	} // manipulate map directly to avoid hierarchy creation
+
+	hierarchy1 := strings.Split(path1, configs.DOT)
+	assert.Assert(t, userTracker.canRunApp(hierarchy1, TestApp1))
+	// make sure wildcard limits are applied
+	assert.Equal(t, uint64(3), userTracker.queueTracker.childQueueTrackers["parent"].maxRunningApps)
+	assert.Assert(t, resources.Equals(maxRes, userTracker.queueTracker.childQueueTrackers["parent"].maxResources))
```
chia7712, line 322: maybe we can check the `useWildCard` also?

pkg/scheduler/ugm/group_tracker_test.go
```diff
@@ -265,6 +265,33 @@ func TestGTSetAndClearMaxLimits(t *testing.T) {
...
+	user := security.UserGroup{User: "test", Groups: []string{"test"}}
+	groupTracker := newGroupTracker(user.User, newUGMEvents(mock.NewEventSystemDisabled()))
+	manager.userWildCardLimitsConfig["root.parent"] = &LimitConfig{
+		maxResources: resources.NewResourceFromMap(map[string]resources.Quantity{
+			"cpu": 1000,
+		}),
+		maxApplications: 3,
+	} // manipulate map directly to avoid hierarchy creation
+
+	hierarchy1 := strings.Split(path1, configs.DOT)
```
chia7712, line 280: this test does not care for `child1`, so using `path5` is more suitable.

pkg/scheduler/ugm/user_tracker_test.go
```diff
@@ -302,6 +302,34 @@ func TestSetAndClearMaxLimits(t *testing.T) {
...
+	userTracker := newUserTracker(user.User, newUGMEvents(mock.NewEventSystemDisabled()))
+	maxRes := resources.NewResourceFromMap(map[string]resources.Quantity{
+		"cpu": 1000,
+	})
+	manager.userWildCardLimitsConfig["root.parent"] = &LimitConfig{
+		maxResources:    maxRes,
+		maxApplications: 3,
+	} // manipulate map directly to avoid hierarchy creation
+
+	hierarchy1 := strings.Split(path1, configs.DOT)
```
chia7712, line 318: this test does not care for `child1`, so using `path5` is more suitable.

pkg/scheduler/ugm/group_tracker_test.go
```diff
@@ -265,6 +265,33 @@ func TestGTSetAndClearMaxLimits(t *testing.T) {
...
+	assert.Assert(t, groupTracker.canRunApp(hierarchy1, TestApp1))
+	// make sure wildcard limits are not applied due to the tracker type
+	assert.Equal(t, uint64(0), groupTracker.queueTracker.childQueueTrackers["parent"].maxRunningApps)
+	assert.Assert(t, groupTracker.queueTracker.childQueueTrackers["parent"].maxResources == nil)
```
chia7712, line 284: maybe we can check the useWildCard also?

chia7712 (commented):
@pbacsko nice patch. a couple of comments below. please take a look, thanks!

pkg/scheduler/ugm/user_tracker_test.go
pbacsko: We submit applications to leaf queues, so that's why it has to be `path1`. It's possible that `canRunApp()` for a parent queue returns true, but false for a leaf. That's not the case here, but it's more realistic this way.

pkg/scheduler/ugm/user_tracker_test.go
pbacsko: The limit is maxApps = 1 set by `setLimits()`. You "register" the running application with `increaseTrackedResource()`, there's no separate method for that. Although using `resources.Zero` is a bit misleading as a maximum resource, I'll fix that (it just happens to work right now because we don't check this during increment and I even created a JIRA to remove the boolean in YUNIKORN-2423, we can add this increment verification later).

pkg/scheduler/ugm/user_tracker_test.go
pbacsko: correct, we can do that

pkg/scheduler/ugm/group_tracker_test.go
pbacsko: see answer below

chia7712 (approved):
LGTM

PR #813, opened 2024-02-22
[YUNIKORN-2441] Wildcard limits are not applied to the root tracker during creation

pbacsko:
### What is this PR for?
When the "root" tracker is created for a user, the wildcard limits are not applied.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2441

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager_test.go
+++ b/pkg/scheduler/ugm/manager_test.go
@@ -332,7 +332,7 @@ func TestUpdateConfig(t *testing.T) {
 func TestUseWildCard(t *testing.T) {
 	setupUGM()
 	manager := GetUserManager()
-	user := security.UserGroup{User: "user1", Groups: []string{"group1"}}
+	user1 := security.UserGroup{User: "user1", Groups: []string{"group1"}}
 
 	expectedResource, err := resources.NewResourceFromConf(map[string]string{"memory": "50", "vcores": "50"})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/scheduler/ugm/manager_test.go
```diff
@@ -332,7 +332,7 @@ func TestUpdateConfig(t *testing.T) {
 func TestUseWildCard(t *testing.T) {
 	setupUGM()
 	manager := GetUserManager()
-	user := security.UserGroup{User: "user1", Groups: []string{"group1"}}
+	user1 := security.UserGroup{User: "user1", Groups: []string{"group1"}}
```
pbacsko, line 335: Did a little cleanup here, variables were inconsistent.
user -> user1
user1 -> user2

It's a problem in multiple tests, which can be taken care of in a separate JIRA.

pkg/scheduler/ugm/manager_test.go
```diff
@@ -348,70 +348,72 @@ func TestUseWildCard(t *testing.T) {
...
-	assert.Equal(t, manager.GetUserTracker(user.User).queueTracker.childQueueTrackers["parent"].childQueueTrackers["child1"].maxRunningApps, uint64(0))
-	assert.Equal(t, resources.Equals(manager.GetUserTracker(user.User).queueTracker.maxResources, nil), true)
-	assert.Equal(t, resources.Equals(manager.GetUserTracker(user.User).queueTracker.childQueueTrackers["parent"].maxResources, expectedHeadroom), true)
-	assert.Equal(t, resources.Equals(manager.GetUserTracker(user.User).queueTracker.childQueueTrackers["parent"].childQueueTrackers["child1"].maxResources, nil), true)
-	assert.Equal(t, manager.GetUserTracker(user.User).queueTracker.useWildCard, false)
-	assert.Equal(t, manager.GetUserTracker(user.User).queueTracker.childQueueTrackers["parent"].useWildCard, true)
-	assert.Equal(t, manager.GetUserTracker(user.User).queueTracker.childQueueTrackers["parent"].childQueueTrackers["child1"].useWildCard, false)
-
-	// user2 uses its own settings.
 	assert.Equal(t, manager.GetUserTracker(user1.User).queueTracker.maxRunningApps, uint64(20))
```
pbacsko, line 364: Now both user1 & user2 has maxApps 20 for "root".

pkg/scheduler/ugm/manager_test.go
wilfred-s: It is a preference: Equals is slightly easier to use as you specify the exact value to compare against and fail the test if it is not the value. Assert fails if the 2nd argument after evaluation is False.

The case for using Equals with boolean is when comparing to False to prevent the negation of the 2nd argument. Usage of `!` in the Assert is easily overlooked, especially if the term becomes more complex.

wilfred-s (approved):
LGTM

PR #818, opened 2024-02-27
[YUNIKORN-2454] Event streaming: send instanceUUID before the events

pbacsko:
### What is this PR for?
Send Yunikorn Instance UUID when creating a streaming connection. This way, Yunikorn restart can be detected easily.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2454

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -65,6 +65,10 @@ var allowedActiveStatusMsg string
 var allowedAppActiveStatuses map[string]bool
 var streamingLimiter *StreamingLimiter
 
+type YunikornID struct {
+	InstanceUUID string
+}
+
 func init() {
 	allowedAppActiveStatuses = make(map[string]bool)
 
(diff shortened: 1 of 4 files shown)


pkg/webservice/handlers.go
```diff
@@ -65,6 +65,10 @@ var allowedActiveStatusMsg string
 var allowedAppActiveStatuses map[string]bool
 var streamingLimiter *StreamingLimiter
 
+type YunikornID struct {
```
chia7712, line 68: Should it be located at `dao` package?

pkg/webservice/handlers.go
```diff
@@ -1172,6 +1176,14 @@ func getStream(w http.ResponseWriter, r *http.Request) {
 	stream := eventSystem.CreateEventStream(r.Host, count)
 	defer eventSystem.RemoveStream(stream)
 
+	if err := enc.Encode(YunikornID{
+		InstanceUUID: schedulerContext.GetUUID(),
+	}); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusInternalServerError)
```
chia7712, line 1182: The reason of using `500` instead of `200` is that this is the first event at this point?

chia7712 (commented):
@pbacsko two minor comments below. please take a look.

pkg/webservice/handlers.go
pbacsko: Yes. We could use 500 later on as well, but they wouldn't have any impact. Any write to the response body automatically sets 200 OK, because that must be sent beforehand.

chia7712 (approved):
LGTM

PR #819, opened 2024-03-05
[YUNIKORN-2469] Upgrade google.golang.org/protobuf to v1.33.0

craigcondit:
### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2469

### Questions:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -59,4 +59,5 @@ replace (
 	golang.org/x/sys => golang.org/x/sys v0.17.0
 	golang.org/x/text => golang.org/x/text v0.14.0
 	golang.org/x/tools => golang.org/x/tools v0.17.0
+	google.golang.org/protobuf => google.golang.org/protobuf v1.33.0
 )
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -59,4 +59,5 @@ replace (
 	golang.org/x/sys => golang.org/x/sys v0.17.0
 	golang.org/x/text => golang.org/x/text v0.14.0
 	golang.org/x/tools => golang.org/x/tools v0.17.0
+	google.golang.org/protobuf => google.golang.org/protobuf v1.33.0
```
chia7712, line 62: Should we update interface repository first, and then update version of interface by this PR?

craigcondit: We can't update SI, as there are not binary builds of the protobuf compiler for 1.33.0 available yet. The issue this update resolves is only in decoding, so it's not crucial that we do that.

craigcondit: Also, for the master branch, we don't typically update SI / core refs on every change, just when needed. For branch 1.5, we need to re-tag for release anyway.

chia7712 (approved):

PR #822, opened 2024-03-19
[YUNIKORN-2494] Revisit IsAtorAbove, WithIn, GetRemaining Guaranteed …

manirajv06:
…resources calculation

### What is this PR for?
Methods like IsAtorAbove, WithIn etc doesn't expose the actual guaranteed values and returns boolean value based on the calculation. There are cases, where these boolean values are not correct and also there is a need to know the actual guaranteed values. For example, How much is remaining in Guaranteed? How much can be preempted? etc.

Introduced new methods to expose the actual values and incorporated the same into existing preemption flow. Added tests cases to cover the new methods as well.  

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2494

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -749,6 +749,61 @@ func (qps *QueuePreemptionSnapshot) GetRemainingGuaranteed() *resources.Resource
 	return resources.ComponentWiseMin(remaining, parentResult)
 }
 
+func (qps *QueuePreemptionSnapshot) GetPreemptableResource() *resources.Resource {
+	if qps == nil {
+		return nil
+	}
+	parentPreemptableResource := qps.Parent.GetPreemptableResource()
+	used := qps.AllocatedResource.Clone()
+
+	// No usage, so nothing to preempt
+	if used.IsEmpty() {
+		return nil
+	}
+	used.SubOnlyExisting(qps.PreemptingResource)
+
+	// Calculate preemptable resource. +ve means Over utilized, -ve means Under utilized, 0 means correct utilization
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/preemption.go
```diff
@@ -749,6 +749,64 @@ func (qps *QueuePreemptionSnapshot) GetRemainingGuaranteed() *resources.Resource
...
+	used := qps.AllocatedResource.Clone()
+
+	// No usage, so nothing to preempt
+	if used.IsEmpty() {
+		return nil
+	}
+	used.SubOnlyExisting(qps.PreemptingResource)
+
+	// Calculate preemptable resource. +ve means Over utilized, -ve means Under utilized, 0 means correct utilization
+	guaranteed := qps.GuaranteedResource.Clone()
```
wilfred-s, line 766: don't need this clone use `qps.GetGuaranteedResource()`

pkg/scheduler/objects/preemption.go
```diff
@@ -749,6 +749,64 @@ func (qps *QueuePreemptionSnapshot) GetRemainingGuaranteed() *resources.Resource
...
+	actualPreemptableResource := used
```
wilfred-s, line 767: not needed, can be used directly
maybe rename `used` to `actual` and drop `actualPreemptableResource`

wilfred-s (changes requested):
perf improvments

wilfred-s (approved):
LGTM

PR #823, opened 2024-03-19
[YUNIKORN-2472] REST API returns subtree by default

steinsgateted:
subtree missing(/ws/v1/partition/default/queue/root): return one level

- `GetPartitionQueueDAOInfo(include bool)` The default include is true, which means all subtrees are included.
`GetPartitionQueueDAOInfo(include bool)` include is false, which means that only the specified queue is included.

- `ChildrenNames` changes to `ChildNames`

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2472

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -624,14 +624,14 @@ func (sq *Queue) CheckAdminAccess(user security.UserGroup) bool {
 }
 
 // GetPartitionQueueDAOInfo returns the queue hierarchy as an object for a REST call.
-// Exclude is true, which means that returns the specified queue object, but does not return the children of the specified queue.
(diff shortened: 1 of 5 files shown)


pbacsko (changes requested):
@steinsgateted could you take care of the missing test coverage?

steinsgateted:
@pbacsko Thanks for the review, add some tests in `TestGetPartitionQueueDAOInfo`.

pbacsko (approved):
LGTM

PR #827, opened 2024-03-23
[YUNIKORN-2511] Cleanup state in test cases

targetoee:
### What is this PR for?
In the Ubuntu env, when running `make test` for the second time, `TestApplicationHistoryTracking` would fail.
In the log, there's an error when calling `StartWebApp`, with the error message 'address already in use'.
This is because in the previous test, some state was not properly cleaned up.
This PR aims to fix it so that `make test` can pass every time.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
[YUNIKORN-2511](https://issues.apache.org/jira/browse/YUNIKORN-2511)

It's been tested by `make test` for multiple times and passed.


chia7712 (approved):
LGTM

chia7712:
pending for reviews from @chenyulin0719

chenyulin0719 (approved):
+1 LGTM

PR #829, opened 2024-03-27
[YUNIKORN-2519] Remove bypass ACL check from placement rules

wilfred-s:
### What is this PR for?
Instead of returning a flag to not bypass the ACL check by all rules except for the recovery rule special case the recovery rule to bypass checks.

Change logger handler to SchedApplication for placements

### What type of PR is it?
* [X] - Refactoring

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2519

Existing unit tests covers the changes


pkg/scheduler/partition.go
```diff
@@ -289,7 +289,9 @@ func (pc *PartitionContext) getPlacementManager() *placement.AppPlacementManager
 	return pc.placementManager
 }
 
-// Add a new application to the partition.
+// AddApplication adds a new application to the partition.
+// Runs the placement rules for the queue resolution. Creates a new dynamic queue if the queue does bit yet
```
craigcondit, line 293: Nit: bit -> not

craigcondit (approved):
+1 LGTM. Minor comment typo found - feel free to adjust on commit.

pbacsko (approved):
+1 LGTM

We have a couple of uncovered branches, I think it's worthwile to create a follow-up JIRA to address those.

pkg/scheduler/partition.go
wilfred-s: done

PR #830, opened 2024-03-29
[YUNIKORN-2500] Use GetPreemptableResource, GetRemainingGuaranteedRes…

manirajv06:
…ource in Preemption flow

### What is this PR for?

Use GetPreemptableResource, GetRemainingGuaranteedResource instead of old methods in preemption flow. In addition,  there are some other changes to ensure the whole flow works as earlier.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2500

### Questions:


manirajv06:
Ran the e2e test in my local setup to verify the changes. It is running with out any failures.

craigcondit (changes requested):
-1. I know that there is a spreadsheet of behavioral cases that has been created for this to document how the preemption logic should behave in a variety of circumstances. I would absolutely like to see a PR consisting only of test cases *for each of the identified scenarios*. We can disable the failing ones for now, but that would give us a baseline for how "compliant" the existing preemption logic is.  Then YUNIKORN-2500 can be rebased on top of that, re-enabling the tests that it fixes. That will ensure we don't regress compared to the existing behavior.

manirajv06:
> -1. I know that there is a spreadsheet of behavioral cases that has been created for this to document how the preemption logic should behave in a variety of circumstances. I would absolutely like to see a PR consisting only of test cases _for each of the identified scenarios_. We can disable the failing ones for now, but that would give us a baseline for how "compliant" the existing preemption logic is. Then YUNIKORN-2500 can be rebased on top of that, re-enabling the tests that it fixes. That will ensure we don't regress compared to the existing behavior.

PR https://github.com/apache/yunikorn-core/pull/847 covers all the different test cases but skipped for now. Once the same has been merged, this pr will fix all those test cases with necessary changes.

craigcondit (approved):
+1, would like Wilfred's input as well before commit.

pkg/common/resources/resources.go
```diff
@@ -751,6 +768,47 @@ func StrictlyGreaterThanOrEquals(larger, smaller *Resource) bool {
 	return true
 }
 
+// StrictlyGreaterThanOnlyExisting Return true if all quantities present or existing only in larger > smaller
+// Two resources that are equal are not considered strictly larger than each other.
+// Resource present in smaller but not in larger are not even considered.
+func StrictlyGreaterThanOnlyExisting(larger, smaller *Resource) bool {
```
wilfred-s, line 774: I think this signature should be `func (r *Resource) StrictlyGreaterThanOnlyExisting(smaller *Resource) bool`
With the description:
```
// StrictlyGreaterThanOnlyExisting returns true if all quantities for types in the defined resource are greater than
// the quantity for the same type in smaller.
// Types defined in smaller that are not in the defined resource are ignored.
// Two resources that are equal are not considered strictly larger than each other.
```
For consistency we should add and implement: if smaller is nil or empty all quantities in the defined resource must be greater than 0 (see below comment)

pkg/common/resources/resources.go
```diff
@@ -751,6 +768,47 @@ func StrictlyGreaterThanOrEquals(larger, smaller *Resource) bool {
...
+	// keep track of the number of not equal values
+	notEqual := false
+
+	// Is larger and smaller completely disjoint?
+	atleastOneResourcePresent := false
+	for k, v := range larger.Resources {
+		// even when smaller is empty, at least one of the resource type in larger should be greater than zero
+		if smaller.IsEmpty() && v > 0 {
+			return true
+		}
```
wilfred-s, line 791: if I have `{larger: {first: -10, second: 5}}` with an empty smaller this call will return `true`
That does not seem correct. It is not "strict" in that case.

pkg/common/resources/resources.go
```diff
@@ -816,6 +874,29 @@ func ComponentWiseMinPermissive(left, right *Resource) *Resource {
 	return out
 }
 
+// ComponentWiseMinOnlyExisting Returns a new Resource with the smallest value for resource type
+// existing only in left but not vice versa.
+func ComponentWiseMinOnlyExisting(left, right *Resource) *Resource {
```
wilfred-s, line 879: This is the 3rd version of `ComponentWiseMin...` need to think about a refactor in a follow up jira

wilfred-s (changes requested):
preemption side looks good
StrictlyGreaterThanOnlyExisting needs clarification

pkg/common/resources/resources.go
```diff
@@ -645,6 +645,25 @@ func Equals(left, right *Resource) bool {
...
+// MatchAny returns true if at least one type in the defined resource exists in the other resource.
+// False if none of the types exist in the other resource.
+// A nil resource is treated as an empty resource (no types defined) and returns false
+// Values are not considered during the checks
+func (r *Resource) MatchAny(other *Resource) bool {
+	if r == nil || other == nil {
+		return false
+	}
+	if r == other {
+		return true
```
pbacsko, line 657: Need coverage for L#657.

pkg/scheduler/objects/preemption.go
```diff
@@ -442,25 +465,38 @@ func (p *Preemptor) calculateAdditionalVictims(nodeVictims []*Allocation) ([]*Al
...
+						break
+					}
+					// check to see if the shortfall on the queue has changed
+					if !resources.EqualsOrEmpty(askQueueRemainingAfterVictimRemoval, askQueueNewRemaining) {
 						// remaining capacity changed, so we should keep this task
 						victims = append(victims, victim)
+					} else {
+						// remaining guaranteed amount in ask queue did not change, so preempting task won't help
+						askQueue.RemoveAllocation(victim.GetAllocatedResource())
+						queueSnapshot.AddAllocation(victim.GetAllocatedResource())
```
pbacsko, line 499: Can we add coverage to this branch?

pkg/scheduler/objects/preemption.go
```diff
@@ -915,6 +935,11 @@ func preemptPredicateCheck(plugin api.ResourceManagerCallback, ch chan<- *predic
 		}); err == nil {
 			result.success = true
 			result.index = -1
+		} else {
+			log.Log(log.SchedPreemption).Debug("Normal predicate check failed",
```
pbacsko, line 939: Shouldn't we log this an error?

pbacsko (changes requested):
Two coverage related comment + logging.

pkg/scheduler/objects/preemption.go
manirajv06: No, not required as it may flood the logs when we have many predicates to run. At times, there could be. Moreover, predicates check is not an error as it indicates that there is no resources even after the preemption.

wilfred-s (approved):
LGTM
@pbacsko I think we're OK, please have a last check

pbacsko (approved):
LGTM

PR #831, opened 2024-04-02
[YUNIKORN-2527] Reset state of configured queue on re-add

wilfred-s:
### What is this PR for?
When a queue is removed and re-added the configured queue status is not reset. If that happens within the cleanup cycle, the cleanup still gets triggered. The queue is removed causing an inconsistency between the config and the queues available in the core.

The object state needs to allow for moving from Draining back to Active. Updated tests to cover the new state transition.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-2527

unit tests included


pbacsko (approved):
+1 LGTM

PR #836, opened 2024-04-08
[YUNIKORN-2545] Eliminate multiple lock calls from Queue

pbacsko:
### What is this PR for?
Do not obtain a second `RLock()` if we already have a lock instance.

Also, the call to `getCurrentPriority()` on the child should be made after the properties have been updated with `UpdateQueueProperties()`.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2545

Enable deadlock detection with the env var `DEADLOCK_DETECTION_ENABLED`.
Run some tests inside `smoke_test.go`.
**Note**: it still prints potential deadlocks, those will be solved in follow-up PRs.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -131,6 +131,7 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
 	// add to the parent, we might have an overall lock already
 	// still need to make sure we lock the parent so we do not interfere with scheduling
 	if parent != nil {
+		// pull the properties from the parent that should be set on the child
 		sq.mergeProperties(parent.getProperties(), conf.Properties)
 		sq.UpdateQueueProperties()
 		sq.QueuePath = parent.QueuePath + configs.DOT + sq.Name
@@ -138,7 +139,6 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
 		if err != nil {
 			return nil, fmt.Errorf("configured queue creation failed: %w", err)
 		}
-		// pull the properties from the parent that should be set on the child
 	} else {
 		sq.UpdateQueueProperties()
 	}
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -131,15 +131,17 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
...
 	if parent != nil {
+		sq.mergeProperties(parent.getProperties(), conf.Properties)
+		sq.UpdateQueueProperties()
 		sq.QueuePath = parent.QueuePath + configs.DOT + sq.Name
 		err := parent.addChildQueue(sq)
 		if err != nil {
 			return nil, fmt.Errorf("configured queue creation failed: %w", err)
 		}
 		// pull the properties from the parent that should be set on the child
-		sq.mergeProperties(parent.getProperties(), conf.Properties)
```
craigcondit, line 140: Nit: I think this comment should go just before mergeProperties() above.

pkg/scheduler/objects/queue.go
```diff
@@ -131,15 +131,17 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
...
 	// still need to make sure we lock the parent so we do not interfere with scheduling
 	if parent != nil {
+		sq.mergeProperties(parent.getProperties(), conf.Properties)
+		sq.UpdateQueueProperties()
 		sq.QueuePath = parent.QueuePath + configs.DOT + sq.Name
 		err := parent.addChildQueue(sq)
 		if err != nil {
 			return nil, fmt.Errorf("configured queue creation failed: %w", err)
 		}
 		// pull the properties from the parent that should be set on the child
```
craigcondit, line 141: Nit: This comment should probably be moved to just before sq.mergeProperties above.

craigcondit (approved):
+1 LGTM. See minor nit regarding comment (feel free to fix on commit). Please also backport to branch-1.5.

PR #837, opened 2024-04-08
[YUNIKORN-2543] Fix locking in RMProxy

pbacsko:
### What is this PR for?
Reduce the scopes of read locks in RMProxy to avoid potential deadlocks which can be caused by an interference with `cache.Context`.

Also simplify the event handler references, we only need `schedulerEventHandler` which is now set once during creation.

### What type of PR is it?
* [x] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2543

### Questions:

### DIFF
--- a/pkg/entrypoint/entrypoint.go
+++ b/pkg/entrypoint/entrypoint.go
@@ -80,8 +80,7 @@ func startAllServicesWithParameters(opts startupOptions) *ServiceContext {
 	events.GetEventSystem().StartService()
 
 	sched := scheduler.NewScheduler()
-	proxy := rmproxy.NewRMProxy()
-
+	proxy := rmproxy.NewRMProxy(sched)
 	eventHandler := handler.EventHandlers{
 		SchedulerEventHandler: sched,
 		RMProxyEventHandler:   proxy,
@@ -90,7 +89,7 @@ func startAllServicesWithParameters(opts startupOptions) *ServiceContext {
 	// start services
 	log.Log(log.Entrypoint).Info("ServiceContext start scheduling services")
 	sched.StartService(eventHandler, opts.manualScheduleFlag)
-	proxy.StartService(eventHandler)
+	proxy.StartService()
 
 	context := &ServiceContext{
 		RMProxy:   proxy,
--- a/pkg/rmproxy/rmproxy.go
+++ b/pkg/rmproxy/rmproxy.go
@@ -39,8 +39,8 @@ import (
 
 // Gateway to talk to ResourceManager (behind grpc/API of scheduler-interface)
 type RMProxy struct {
-	EventHandlers handler.EventHandlers
-	stop          chan struct{}
+	schedulerEventHandler handler.EventHandler // read-only, no lock needed to access it
+	stop                  chan struct{}
 
 	// Internal fields
 	pendingRMEvents chan interface{}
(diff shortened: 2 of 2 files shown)


chenyulin0719:
Hi @pbacsko, 
May I ask what test workload you're using to reproduce the lock detection warnings?
(I've followed [Craig's guide](https://issues.apache.org/jira/browse/YUNIKORN-2521?focusedCommentId=17834392&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17834392) to enable deadlock detection and can see below logs in scheduler pod: 
- "=== Deadlock detection enabled (timeout: 60 seconds) ==="

pbacsko:
@chenyulin0719 yes, enable the deadlock detection and run a smoke test. If the tests in the core do not trigger these, run the smokes from the shim, that will definitely do the job.

pkg/entrypoint/entrypoint.go
```diff
@@ -90,7 +89,7 @@ func startAllServicesWithParameters(opts startupOptions) *ServiceContext {
 	// start services
 	log.Log(log.Entrypoint).Info("ServiceContext start scheduling services")
 	sched.StartService(eventHandler, opts.manualScheduleFlag)
-	proxy.StartService(eventHandler)
+	proxy.StartService()
```
chenyulin0719, line 92: I'm thinking about the necessity of having handler.EventHandlers struct.

Not sure why we need it...  To me, I think it would also works if we remove  handler.EventHandlers and change the startService() function to:
```
	sched.StartService(proxy, opts.manualScheduleFlag)
	proxy.StartService()
```

I suppose the reason for having have handler.EventHandlers is to group all EventHandler objects.  However, if we implement this change... the handler.EventHandlers will be reduntant since it's only existing for RMProxyEventHandler .

pbacsko:
@chenyulin0719 IMO there shouldn't be any RMProxy related potential deadlocks. Could you share your `deadlock-log.txt` with us to see what's going on?

pkg/entrypoint/entrypoint.go
pbacsko: Yeah that's correct. I think we can fix this in a follow-up JIRA.

chenyulin0719:
Hi @pbacsko, 

Just recheck the [go-deadlock doc.](https://github.com/sasha-s/go-deadlock?tab=readme-ov-file#deadlocks) 
I think I mixup below log categories:
1. happened before
2. happened after
3. Other goroutines holding locks

After this PR, all the 'rmp.RLock' related logs are under "Other goroutines holding locks".
It's merely a information for touble shooting, and doesn't indicate that there have potential deadlock.

Attached the logs file: [deadlock-log-after-YUNIKORN-2543.txt](https://github.com/apache/yunikorn-core/files/14931175/deadlock-log-after-YUNIKORN-2543.txt)

pbacsko:
@chenyulin0719 thanks, that's perfectly normal, so it's good news. It just shouldn't show up in the happened before/after call stack.

chenyulin0719 (approved):
+1 LGTM

craigcondit (approved):
+1 LGTM. Please also backport to branch-1.5.

PR #838, opened 2024-04-08
[YUNIKORN-2547] Queue: clean up logic when adding application

pbacsko:
### What is this PR for?
Clean up tag parsing logic in `Queue.AddApplication()` because that code belongs to `objects.NewApplication`.
Also, don't add the application to the queue then remove it if the placeholder doesn't fit, because it generates an unnecessary application event. Instead, calculate the new maximum resource without adding it.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2547

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -2120,12 +2120,12 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
 	sa.runnableByUserLimit = runnableByUserLimit
 }
 
-// GetGuaranteedResource returns the guaranteed resource that was set in the application tags
+// GetGuaranteedResource returns the guaranteed resource that is set in the application tags
 func (sa *Application) GetGuaranteedResource() *resources.Resource {
 	return sa.getResourceFromTags(siCommon.AppTagNamespaceResourceGuaranteed)
 }
 
-// GetMaxResource returns the max resource that was set in the application tags
+// GetMaxResource returns the max resource that is set in the application tags
 func (sa *Application) GetMaxResource() *resources.Resource {
 	return sa.getResourceFromTags(siCommon.AppTagNamespaceResourceQuota)
 }
@@ -2144,7 +2144,7 @@ func (sa *Application) getResourceFromTags(tag string) *resources.Resource {
 			zap.Error(err))
 	} else if !resources.StrictlyGreaterThanZero(resource) {
 		log.Log(log.SchedQueue).Warn("resource quantities should be greater than zero",
-			zap.Stringer("maxResource", resource))
+			zap.Stringer("resource", resource))
 		resource = nil
 	}
 
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1529,7 +1529,6 @@ func (sa *Application) SetQueue(queue *Queue) {
 	defer sa.Unlock()
 	sa.queuePath = queue.QueuePath
 	sa.queue = queue
-	sa.queue.UpdateApplicationPriority(sa.ApplicationID, sa.askMaxPriority)
 }
 
 // remove the leaf queue the application runs in, used when completing the app
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -40,6 +40,7 @@ import (
 	"github.com/apache/yunikorn-core/pkg/metrics"
 	"github.com/apache/yunikorn-core/pkg/rmproxy/rmevent"
 	"github.com/apache/yunikorn-core/pkg/scheduler/ugm"
+	siCommon "github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
@@ -75,10 +76,12 @@ type StateLogEntry struct {
 }
 
 type Application struct {
-	ApplicationID  string            // application ID
-	Partition      string            // partition Name
-	SubmissionTime time.Time         // time application was submitted
-	tags           map[string]string // application tags used in scheduling
+	ApplicationID  string              // application ID
+	Partition      string              // partition Name
+	SubmissionTime time.Time           // time application was submitted
+	tags           map[string]string   // application tags used in scheduling
+	maxResource    *resources.Resource // max resource setting from namespace based on tags
+	guaranteed     *resources.Resource // guaranteed resource setting from namespace based on tags
 
 	// Private mutable fields need protection
 	queuePath         string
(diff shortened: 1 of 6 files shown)


chia7712 (commented):
@pbacsko thanks for fixing lock issue. please take a look at following comments. thanks

wilfred-s (changes requested):
If the app gets rejected we still need to set the quota/guaranteed. There should not be a revert to the old values if the application does not get added.
Should simplify things.

pkg/scheduler/partition.go
```diff
@@ -339,26 +339,28 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
 		// check the queue sorting
 		if !queue.SupportTaskGroup() {
-			queue.RemoveApplication(app)
 			return fmt.Errorf("queue %s cannot run application %s with task group request: unsupported sort type", queueName, appID)
 		}
-		// retrieve the max set
-		if maxQueue := queue.GetMaxQueueSet(); maxQueue != nil {
+		// the new maxResource is going to be either app.GetMaxResource() from the tags or the maxResource which already set on the queue
+		maxRes := queue.GetMaxResource()
+		if app.GetMaxResource() != nil {
```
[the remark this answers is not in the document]
pbacsko: This will be completely rewritten. In fact, this part will be restored because max/guaranteed must be set regardless of whether the application can be accepted or not.

pkg/scheduler/objects/queue.go
```diff
@@ -1281,6 +1238,37 @@ func (sq *Queue) internalGetMax(parentLimit *resources.Resource) *resources.Reso
...
+	if sq.parent == nil {
+		return nil
+	}
+	limit = sq.parent.GetMaxQueueSet()
+	return sq.internalGetProposedMax(limit, maxRes)
+}
+
+// internalGetProposedMax does the real max calculation.
+func (sq *Queue) internalGetProposedMax(parentLimit, maxRes *resources.Resource) *resources.Resource {
+	sq.RLock()
```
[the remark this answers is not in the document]
pbacsko: These new methods will be removed.

pkg/scheduler/objects/queue.go
```diff
@@ -1281,6 +1238,37 @@ func (sq *Queue) internalGetMax(parentLimit *resources.Resource) *resources.Reso
 	return resources.ComponentWiseMin(parentLimit, sq.maxResource)
 }
 
+// GetProposedMaxQueueResource calculates the maximum resource for the queue based on a proposed new value.
+// This is similar to GetMaxQueueSet but uses a value that has not been applied as a maxResource for the queue.
+func (sq *Queue) GetProposedMaxQueueResource(maxRes *resources.Resource) *resources.Resource {
+	// get the limit for the parent first and check against the queue's own
+	var limit *resources.Resource
```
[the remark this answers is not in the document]
pbacsko: This new method will be removed.

pkg/scheduler/objects/application.go
```diff
@@ -75,10 +76,12 @@ type StateLogEntry struct {
...
-	ApplicationID  string            // application ID
-	Partition      string            // partition Name
-	SubmissionTime time.Time         // time application was submitted
-	tags           map[string]string // application tags used in scheduling
+	ApplicationID  string              // application ID
+	Partition      string              // partition Name
+	SubmissionTime time.Time           // time application was submitted
+	tags           map[string]string   // application tags used in scheduling
+	maxResource    *resources.Resource // max resource setting from namespace based on tags
+	guaranteed     *resources.Resource // guaranteed resource setting from namespace based on tags
```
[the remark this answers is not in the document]

pkg/scheduler/objects/application.go
```diff
@@ -186,11 +189,57 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
 	app.user = ugi
 	app.rmEventHandler = eventHandler
 	app.rmID = rmID
+	app.setResourcesFromTags()
```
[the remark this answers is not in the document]
pbacsko: Check new code, it's been changed

pkg/scheduler/objects/queue_test.go
```diff
@@ -415,136 +415,6 @@ func TestAddApplication(t *testing.T) {
 	assert.Equal(t, len(leaf.applications), 1, "Application was not replaced in the queue as expected")
 }
 
-func TestAddApplicationWithTag(t *testing.T) {
```
pbacsko, line 418: This is no longer relevant as we don't set resources in `AddApplication()`. The resource logic is now tested in `TestAddTGApplication()` and `TestAddTGAppDynamic()`.

pkg/scheduler/objects/application.go
```diff
@@ -2118,3 +2119,34 @@ func (sa *Application) updateRunnableStatus(runnableInQueue, runnableByUserLimit
...
+
+	resource, err := resources.NewResourceFromString(value)
+	if err != nil {
+		log.Log(log.SchedQueue).Warn("application resource conversion failure",
+			zap.String("tag", tag),
+			zap.String("json string", value),
+			zap.Error(err))
+	} else if !resources.StrictlyGreaterThanZero(resource) {
+		log.Log(log.SchedQueue).Warn("resource quantities should be greater than zero",
+			zap.Stringer("maxResource", resource))
```
chia7712, line 2147: `resource` could be guaranteed resource now.

pkg/scheduler/partition.go
```diff
@@ -339,26 +339,27 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
 		return fmt.Errorf("failed to find queue %s for application %s", queueName, appID)
 	}
 
-	// add the app to the queue to set the quota on the queue if needed
-	queue.AddApplication(app)
+	// set resources based on tags, but only if the queue is dynamic
+	if !queue.IsManaged() {
+		queue.SetResources(app.GetGuaranteedResource(), app.GetMaxResource())
```
chia7712, line 344: Should we keep the warnings "Trying to set max resources on a queue that is not an unmanaged leaf"?

chia7712 (commented):
@pbacsko thanks for updated PR. two trivial comments left. please take a look :smile:

pkg/scheduler/objects/application.go
pbacsko: Fixed

pkg/scheduler/partition.go
pbacsko: It's printed now

chia7712 (approved):
LGTM

pkg/scheduler/partition.go
```diff
@@ -339,26 +339,29 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
 		}
-		// retrieve the max set
 		if maxQueue := queue.GetMaxQueueSet(); maxQueue != nil {
 			if !maxQueue.FitInMaxUndef(placeHolder) {
-				queue.RemoveApplication(app)
 				return fmt.Errorf("queue %s cannot fit application %s: task group request %s larger than max queue allocation %s", queueName, appID, placeHolder.String(), maxQueue.String())
 			}
 		}
 	}
+	queue.AddApplication(app)
```
wilfred-s, line 364: Can we move this till after the queue and callback are set on the app. That means the app object cannot be reached by the scheduler via the queue or partition lowering any change of lock order issues in the future.
It would also allow the removal of the locking in both `SetQueue` and `SetTerminatedCallback`.

pkg/scheduler/partition_test.go
```diff
@@ -2804,14 +2804,22 @@ func TestAddTGApplication(t *testing.T) {
 	partition, err := newLimitedPartition(limit)
 	assert.NilError(t, err, "partition create failed")
 	// add a app with TG that does not fit in the queue
-	var tgRes *resources.Resource
-	tgRes, err = resources.NewResourceFromConf(map[string]string{"vcore": "10"})
+	tgRes, err := resources.NewResourceFromConf(map[string]string{"vcore": "10"})
```
wilfred-s, line 2807: NIT: this change introduces shadowing for the err variable. Not needed

pkg/scheduler/partition.go
```diff
@@ -339,26 +339,29 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
 
 	// all is OK update the app and add it to the partition
 	app.SetQueue(queue)
```
wilfred-s, line 367: This has not changed but the call to set the queue makes an unneeded call to `Queue.UpdateApplicationPriority()` There are no asks defined on the application yet and thus the priority will default to `configs.MinPriority`. To be overridden again as soon as the first ask gets added. Cheaper to just wait until the first ask gets added.

wilfred-s (changes requested):
Almost there, Minor things that allow further cleanup.

pkg/scheduler/partition.go
wilfred-s: See below for another `SetQueue` remark.

pbacsko: Done

pkg/scheduler/partition_test.go
pbacsko: Done

wilfred-s (approved):
LGTM

PR #839, opened 2024-04-09
[YUNIKORN-2548] Potential deadlock during concurrent bottom-up/top-down queue traversal

pbacsko:
### What is this PR for?
Avoid potential deadlock while removing queues during config refresh. `MarkQueueForRemoval()` performs a top-down traversal of the hierarchy, while resource increment/decrement does the opposite.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2548

Run test from `smoke_test.go` with the deadlock detector enabled.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -963,16 +963,22 @@ func (sq *Queue) MarkQueueForRemoval() {
 	if sq.IsManaged() {
 		log.Log(log.SchedQueue).Info("marking managed queue for deletion",
 			zap.String("queue", sq.QueuePath))
-		if err := sq.handleQueueEvent(Remove); err != nil {
-			log.Log(log.SchedQueue).Warn("failed to mark managed queue for deletion",
-				zap.String("queue", sq.QueuePath),
-				zap.Error(err))
-		}
+		sq.doRemoveQueue()
 		if len(sq.children) > 0 {
 			for _, child := range children {
 				child.MarkQueueForRemoval()
 			}
 		}
+	}
+}
+
+func (sq *Queue) doRemoveQueue() {
+	sq.Lock()
+	defer sq.Unlock()
+	if err := sq.handleQueueEvent(Remove); err != nil {
+		log.Log(log.SchedQueue).Warn("failed to mark managed queue for deletion",
+			zap.String("queue", sq.QueuePath),
+			zap.Error(err))
 	}
 }
 
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -957,17 +957,18 @@ func (sq *Queue) addChildQueue(child *Queue) error {
 // This can be executed multiple times and is only effective the first time.
 // This is a noop on an unmanaged queue.
 func (sq *Queue) MarkQueueForRemoval() {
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -957,12 +957,10 @@ func (sq *Queue) addChildQueue(child *Queue) error {
 // This can be executed multiple times and is only effective the first time.
 // This is a noop on an unmanaged queue.
 func (sq *Queue) MarkQueueForRemoval() {
-	// need to lock for write as we don't want to add a queue while marking for removal
```
pbacsko, line 960: Note: this comment is not necessary. Config refresh is done under a partition lock, so the hierarchy cannot be modified while queues are being marked.

pkg/scheduler/objects/queue.go
```diff
@@ -957,12 +957,10 @@ func (sq *Queue) addChildQueue(child *Queue) error {
...
-	sq.Lock()
-	defer sq.Unlock()
+	children := sq.GetCopyOfChildren()
 	// Mark the managed queue for deletion: it is removed from the config let it drain.
 	// Also mark all the managed children for deletion.
-	if sq.isManaged {
+	if sq.IsManaged() {
 		log.Log(log.SchedQueue).Info("marking managed queue for deletion",
 			zap.String("queue", sq.QueuePath))
 		if err := sq.handleQueueEvent(Remove); err != nil {
```
chia7712, line 966: we don't hold lock now, but `handleQueueEvent` can modify inner variable (https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/objects/queue.go#L557)

chia7712 (commented):
@pbacsko thanks for fixing potential deadlock. one question left. PTAL

pkg/scheduler/objects/queue.go
pbacsko: Good catch... gotta figure this out.

pbacsko: Done

pkg/scheduler/objects/queue.go
```diff
@@ -957,27 +957,31 @@ func (sq *Queue) addChildQueue(child *Queue) error {
 // This can be executed multiple times and is only effective the first time.
 // This is a noop on an unmanaged queue.
 func (sq *Queue) MarkQueueForRemoval() {
-	// need to lock for write as we don't want to add a queue while marking for removal
-	sq.Lock()
-	defer sq.Unlock()
+	children := sq.GetCopyOfChildren()
```
craigcondit, line 960: This is probably best done after the sq.IsManaged() check, otherwise it's just useless work. In fact, might be better to just start this function with:

```
if !sq.IsManaged() {
    return
}
```

craigcondit (changes requested):
Minor changes requested, overall approach seems fine.

craigcondit (approved):
+1 LGTM. Also please backport to branch-1.5.

PR #840, opened 2024-04-11
[YUNIKORN-2550] Fix locking in PartitionContext

pbacsko:
### What is this PR for?
Potential deadlock was detected between `PartitionContext` and `AppPlacementManager`.
Config update for the placement manager must run without holding any `PartitionContext` lock.

We also don't need any locking to access `pc.nodes` because the type is internally synchronized and the field is set once.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2550

Run `smoke_test.go` from the core with deadlock detection enabled.

### Questions:

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -170,11 +170,11 @@ func (pc *PartitionContext) updatePartitionDetails(conf configs.PartitionConfig)
 		log.Log(log.SchedPartition).Info("New placement rules not activated, config reload failed", zap.Error(err))
 		return err
 	}
+	pc.updateNodeSortingPolicy(conf)
 
 	pc.Lock()
 	defer pc.Unlock()
 	pc.rules = &conf.PlacementRules
-	pc.updateNodeSortingPolicy(conf)
 	pc.updatePreemption(conf)
 	// start at the root: there is only one queue
 	queueConf := conf.Queues[0]
@@ -587,13 +587,21 @@ func (pc *PartitionContext) updatePartitionResource(delta *resources.Resource) {
 	}
 }
 
-// Update the partition details when removing a node.
-// This locks the partition. The partition may not be locked when we process the allocation
-// additions to the node as that takes further app, queue or node locks
+// Update the partition details when adding a node.
 func (pc *PartitionContext) addNodeToList(node *objects.Node) error {
+	// we don't grab a lock here because we only update pc.nodes which is internally protected
 	if err := pc.nodes.AddNode(node); err != nil {
 		return fmt.Errorf("failed to add node %s to partition %s, error: %v", node.NodeID, pc.Name, err)
 	}
+
+	pc.addNodeResources(node)
+	return nil
+}
+
+// Update metrics & resource tracking information.
+// This locks the partition. The partition may not be locked when we process the allocation
+// additions to the node as that takes further app, queue or node locks.
+func (pc *PartitionContext) addNodeResources(node *objects.Node) {
 	pc.Lock()
 	defer pc.Unlock()
 	metrics.GetSchedulerMetrics().IncActiveNodes()
@@ -608,7 +616,6 @@ func (pc *PartitionContext) addNodeToList(node *objects.Node) error {
 		zap.String("partitionName", pc.Name),
 		zap.String("nodeID", node.NodeID),
 		zap.Stringer("partitionResource", pc.totalPartitionResource))
-	return nil
 }
 
 // removeNodeFromList removes the node from the list of partition nodes.
@@ -623,7 +630,6 @@ func (pc *PartitionContext) removeNodeFromList(nodeID string) *objects.Node {
 
 	// Remove node from list of tracked nodes
 	metrics.GetSchedulerMetrics().DecActiveNodes()
-
 	log.Log(log.SchedPartition).Info("Removed node from available partition nodes",
 		zap.String("partitionName", pc.Name),
 		zap.String("nodeID", node.NodeID))
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -159,7 +159,8 @@ func (pc *PartitionContext) updatePreemption(conf configs.PartitionConfig) {
 }
 
 func (pc *PartitionContext) updatePartitionDetails(conf configs.PartitionConfig) error {
-	// this must be performed without locking to avoid lock order differences between PartitionContext and AppPlacementManager
+	// the following piece of code (before pc.Lock()) must be performed without locking
+	// to avoid lock order differences between PartitionContext and AppPlacementManager
 	if len(conf.Queues) == 0 || conf.Queues[0].Name != configs.RootQueue {
 		return fmt.Errorf("partition cannot be created without root queue")
 	}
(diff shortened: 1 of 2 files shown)


pbacsko:
Note: the original scope of the JIRA ticket got extended. This now PR actually solves two (potential) deadlocks by adding changes to `PartitionContext`.

pkg/scheduler/partition.go
```diff
@@ -616,8 +613,6 @@ func (pc *PartitionContext) addNodeToList(node *objects.Node) error {
 // removeNodeFromList removes the node from the list of partition nodes.
 // This locks the partition.
 func (pc *PartitionContext) removeNodeFromList(nodeID string) *objects.Node {
```
chenyulin0719, line 615: The function comment should be updated. The lock responsibility has been moved down to pc.nodes. No partition lock here.

pkg/scheduler/partition.go
```diff
@@ -591,14 +590,12 @@ func (pc *PartitionContext) updatePartitionResource(delta *resources.Resource) {
...
 // additions to the node as that takes further app, queue or node locks
 func (pc *PartitionContext) addNodeToList(node *objects.Node) error {
-	pc.Lock()
-	defer pc.Unlock()
-	// Node can be added to the system to allow processing of the allocations
 	if err := pc.nodes.AddNode(node); err != nil {
 		return fmt.Errorf("failed to add node %s to partition %s, error: %v", node.NodeID, pc.Name, err)
 	}
+	pc.Lock()
+	defer pc.Unlock()
```
chenyulin0719, line 597: We postponed the partition lock to this point. 

The reason is that we don't want any go routine holding partition lock before changing  pc.nodes?

chenyulin0719: I just ran a test for this PR but move the pc.Lock() back to the first line of addNodeToList(). 

There is no POTENTIAL DEADLOCK called "Inconsistent locking". Looks like this change is unnecessary.

pkg/scheduler/partition.go
```diff
@@ -159,17 +159,19 @@ func (pc *PartitionContext) updatePreemption(conf configs.PartitionConfig) {
 }
 
 func (pc *PartitionContext) updatePartitionDetails(conf configs.PartitionConfig) error {
-	pc.Lock()
-	defer pc.Unlock()
+	// this must be performed without locking to avoid lock order differences between PartitionContext and AppPlacementManager
```
chenyulin0719, line 162: "this must be performed without lock" is not 100% correct.  We "**postpone**" the partition locks to avoid lock order differences between PartitionContext and AppPlacementManager.

pbacsko: That comments refer to the piece of code which are executed outside the critical section. I'll try to make it more clear.

pkg/scheduler/partition.go
pbacsko: It's possible, however, we'd like to be consistent. If we decide not to grab the PartitionContext lock in other cases, it's better to do the same here, becase we avoid locks that are overlapping.

chenyulin0719: Oh, I think I misunderstood.
This is an improvement to decouple locks of pc.nodes and partition.

It's not a change to fix deadlock issue.

chenyulin0719: "decouple" -> I mean remove overlapping between pc.nodes and partition.

It's not quite straightforward to me when I see the lock postponed here, maybe we can add a comment here?

pbacsko: Actually we can move it back. 

1) We already do something similar in `updatePartitionDetails()`. In `updateNodeSortingPolicy()` we make reference to `pc.nodes` while having a partition lock.
2) This can give an inconsistent view: node is already in the data structure, however, partition information (`totalPartitionResource`, `root.maxResource`) has been not updated yet. I don't think it's critical though, but it can trigger the health checker in the wrong moment.

The most important thing is NOT to have partition lock when calling back from the Application. We already achieved this task.

I'll let @wilfred-s share his opinion on this.

pkg/scheduler/partition.go
pbacsko: Updated comment.

pkg/scheduler/partition.go
wilfred-s: To make it clear why don't we follow the pattern of the node removal?
Split the function in two:
* new function `addNodeResources(node)` that locks the partition and does everything after the metrics update.
* existing function `addNodeToList(node)` that does everything including the metrics update and then calls `addNodeResources(node)`

Add proper comments to the functions and it is not up for debate etc.
BTW: the small delay between the adding of the node and the update of the resource if it exists is not an issue.

pkg/scheduler/partition.go
```diff
@@ -159,17 +159,20 @@ func (pc *PartitionContext) updatePreemption(conf configs.PartitionConfig) {
...
-	err := pc.placementManager.UpdateRules(conf.PlacementRules)
+	err := pc.getPlacementManager().UpdateRules(conf.PlacementRules)
 	if err != nil {
 		log.Log(log.SchedPartition).Info("New placement rules not activated, config reload failed", zap.Error(err))
 		return err
 	}
+
+	pc.Lock()
+	defer pc.Unlock()
 	pc.rules = &conf.PlacementRules
```
wilfred-s, line 176: We should get rid of this. It does not serve any use anymore. When we moved to use placement rules always this field became obsolete. The content is also out of sync with the real set of rules in the placement manager.
Filed a follow up to cleanup and expose the placement rules in the statedump.

pkg/scheduler/partition.go
```diff
@@ -159,17 +159,20 @@ func (pc *PartitionContext) updatePreemption(conf configs.PartitionConfig) {
...
 	pc.updateNodeSortingPolicy(conf)
```
wilfred-s, line 177: This does not need the partition lock it updates the nodes object only.

wilfred-s (changes requested):

pkg/scheduler/partition.go
chenyulin0719: +1 for below 2 options
1. Spliting addNodeToList() into two function as Wilfred suggested
2. Add a comment to explain why we postponed the pc.Lock

+0 for below 1 option
1. moving pc.Lock back to first line

I agree it has become a matter of preference. I prefer the idea to remove lock overlapping. Add a comment here might be the simplest option.

All options are OK to me.

pkg/scheduler/partition.go
pbacsko: Done

pkg/scheduler/partition.go
pbacsko: Filed https://issues.apache.org/jira/browse/YUNIKORN-2554

chenyulin0719 (approved):
+1 LGTM, validated the "POTENTIAL DEADLOCK: Inconsistent locking" warnings are gone after this PR.

wilfred-s (approved):
LGTM

PR #841, opened 2024-04-11
[YUNIKORN-2552] Recursive locking when sending remove queue event

pbacsko:
### What is this PR for?
Recursive locking (ie acquiring the same lock twice) was detected by the go-deadlock tool. The PR removes the `objects.Queue` references from `objects.queueEvents` and necessary input are passed as  arguments.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2552

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -142,10 +142,10 @@ func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error)
 	} else {
 		sq.UpdateQueueProperties()
 	}
-	sq.queueEvents = newQueueEvents(sq, events.GetEventSystem())
+	sq.queueEvents = newQueueEvents(events.GetEventSystem())
 	log.Log(log.SchedQueue).Info("configured queue added to scheduler",
 		zap.String("queueName", sq.QueuePath))
-	sq.queueEvents.sendNewQueueEvent()
+	sq.queueEvents.sendNewQueueEvent(sq.QueuePath, sq.isManaged)
 
 	return sq, nil
 }
@@ -200,10 +200,10 @@ func newDynamicQueueInternal(name string, leaf bool, parent *Queue) (*Queue, err
 	}
 
 	sq.UpdateQueueProperties()
-	sq.queueEvents = newQueueEvents(sq, events.GetEventSystem())
+	sq.queueEvents = newQueueEvents(events.GetEventSystem())
 	log.Log(log.SchedQueue).Info("dynamic queue added to scheduler",
 		zap.String("queueName", sq.QueuePath))
-	sq.queueEvents.sendNewQueueEvent()
+	sq.queueEvents.sendNewQueueEvent(sq.QueuePath, sq.isManaged)
 
 	return sq, nil
 }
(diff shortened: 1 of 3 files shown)


chenyulin0719 (commented):
Just a discussion for another possible solution.
Maybe we can simply change queue.IsManaged() to lock free cal. 

Below are the assignment for queue.isManaged

1. NewConfiguredQueue()          // It lock free as it cannot be referenced yet
2.  (sq *Queue) applyConf()       // lock free call, must be called holding the queue lock or during create only
3.  newDynamicQueueInternal() <- NewDynamicQueue()  //  lock free as it cannot be referenced yet
4. newDynamicQueueInternal() <- NewRecoveryQueue()  //  cannot be directly referenced

pbacsko:
> Just a discussion for another possible solution. Maybe we can simply change queue.IsManaged() to lock free call.
> 
> Below are the assignment for queue.isManaged
> 
> 1. NewConfiguredQueue()          // It lock free as it cannot be referenced yet
> 2. (sq *Queue) applyConf()       // lock free call, must be called holding the queue lock or during create only
> 3. newDynamicQueueInternal() <- NewDynamicQueue()  //  lock free as it cannot be referenced yet
> 4. newDynamicQueueInternal() <- NewRecoveryQueue()  //  cannot be directly referenced

Not sure I get the idea. Managed status can change during runtime during config update. So read-write operations must be performed while holding a lock.

chenyulin0719:
Below 2 function won't inference each other because they're both holding queue lock before or after this PR.

1. (sq *Queue) RemoveQueue() -> RLock (Before this PR, it called sq.IsManaged() in sendRemoveQueueEvent(), which also hold RLock)
2. (sq *Queue) ApplyConf() -> WLock

https://github.com/apache/yunikorn-core/blob/16d2c9cd79e73be51d23f0ce5c5d9f153be140bc/pkg/scheduler/objects/queue.go#L303-L307

After rethinking it, I realized that I was wrong.
Some functions call queue.IsManaged() (e.g. sendRemoveQueueEvent) while holding the queue lock, but others do not.

I suddenly have a crazy idea.  How about creating lock-free version of IsManaged()?
1. IsManaged() 
2. IsManagedNoLock() -> Lock free, for sendRemoveQueueEvent()

The pros is that we can pass a *Queue to queueEvents again, which eliminates the need for repeated parameters. 
However, I haven't seen any similar implementations before, maybe it's not a good idea for dealing lock.

(I know the benefit is quite small, it's a matter of preference. I'm good with this PR.)

pbacsko:
> The pros is that we can pass a *Queue to queueEvents again, which eliminates the need for repeated parameters. 

Actually I don't want to do that, I'll explain why. The advantage of this solution is that `queueEvents` (and other `XXXX_events.go` with some changes) can be moved to a separate package in "objects", eg "objects/events". We already have a lot of files inside "objects". That way, we can clean it up a bit. This idea had already been discussed with Wilfred last year, but we were in a hurry with the whole event system so eventually it didn't happen. But in the long run, I feel it's a good thing to do.

chenyulin0719 (approved):
+1 LGTM, validated the Potential Lock warnings are gone after this changes.

chia7712 (approved):
LGTM

one question: should we do similar fix to `application_events` and `node_events` for consistency and avoid potential recursive locking?

wilfred-s (approved):
LGTM

PR #844, opened 2024-04-16
[YUNIKORN-2559] DATA RACE: EventStore.Store() and Context.PublishEvents()

chenyulin0719:
### What is this PR for?
Fix the racing issue.
-  Return a new []*si.EventRecord in EventStore.CollectEvents() instead of reference of EventStore.events

### What type of PR is it?
* [x] - Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2559

Run make test in shim.

NA

### Questions:
NA

### DIFF
--- a/pkg/events/event_store.go
+++ b/pkg/events/event_store.go
@@ -81,14 +81,6 @@ func (es *EventStore) CollectEvents() []*si.EventRecord {
 	return messages
 }
 
-// test only
-func (es *EventStore) CollectInternalEvents(idx uint64) []*si.EventRecord {
-	es.RLock()
-	defer es.RUnlock()
-	messages := es.events[:idx]
-	return messages
-}
-
 func (es *EventStore) CountStoredEvents() uint64 {
 	es.RLock()
 	defer es.RUnlock()
--- a/pkg/events/event_store.go
+++ b/pkg/events/event_store.go
@@ -43,8 +43,9 @@ type EventStore struct {
 
 func newEventStore(size uint64) *EventStore {
 	return &EventStore{
-		events: make([]*si.EventRecord, size),
-		size:   size,
+		events:   make([]*si.EventRecord, size),
+		size:     size,
+		lastSize: size,
 	}
 }
 
@@ -77,6 +78,14 @@ func (es *EventStore) CollectEvents() []*si.EventRecord {
 	es.lastSize = es.size
 
 	metrics.GetEventMetrics().AddEventsCollected(len(messages))
+	return messages
+}
+
+// test only
+func (es *EventStore) CollectInternalEvents(idx uint64) []*si.EventRecord {
+	es.RLock()
+	defer es.RUnlock()
+	messages := es.events[:idx]
 	return messages
 }
 
--- a/pkg/events/event_store.go
+++ b/pkg/events/event_store.go
@@ -66,7 +66,9 @@ func (es *EventStore) CollectEvents() []*si.EventRecord {
 	es.Lock()
 	defer es.Unlock()
 
-	messages := es.events[:es.idx]
+	messages := make([]*si.EventRecord, len(es.events[:es.idx]))
+	copy(messages, es.events[:es.idx])
+
 	if es.size != es.lastSize {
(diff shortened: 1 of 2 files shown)


pkg/events/event_store.go
```diff
@@ -66,7 +66,9 @@ func (es *EventStore) CollectEvents() []*si.EventRecord {
 	es.Lock()
 	defer es.Unlock()
 
-	messages := es.events[:es.idx]
+	messages := make([]*si.EventRecord, len(es.events[:es.idx]))
```
chia7712, line 69: Could you add test to validate the copy?

chia7712 (commented):
@chenyulin0719 nice finding. one small comment left.

pkg/events/event_store.go
pbacsko: +1, let's make sure that changes to the original are not reflected in the copy

pbacsko (changes requested):
Add a small test or extend an existing one as requested.

pkg/events/event_store.go
```diff
@@ -43,8 +43,9 @@ type EventStore struct {
 
 func newEventStore(size uint64) *EventStore {
 	return &EventStore{
-		events: make([]*si.EventRecord, size),
-		size:   size,
+		events:   make([]*si.EventRecord, size),
+		size:     size,
+		lastSize: size,
```
chenyulin0719, line 48: Set initial lastSize value to prevent uncessarily resize in  https://github.com/apache/yunikorn-core/blob/634380f109a0ad724cabcbdb5e2e7d78ec2debd5/pkg/events/event_store.go#L70-L73

pkg/events/event_store.go
chenyulin0719: Done

pkg/events/event_store.go
```diff
@@ -80,6 +81,14 @@ func (es *EventStore) CollectEvents() []*si.EventRecord {
 	return messages
 }
 
+// test only
+func (es *EventStore) CollectInternalEvents(idx uint64) []*si.EventRecord {
```
chia7712, line 85: As it is used by testing only, we should use lowercase letter to expose it to same package only.

pkg/events/event_store_test.go
```diff
@@ -55,6 +56,16 @@ func TestStoreAndRetrieve(t *testing.T) {
 	assert.DeepEqual(t, records[0], event1, cmpopts.IgnoreUnexported(si.EventRecord{}))
 	assert.DeepEqual(t, records[1], event2, cmpopts.IgnoreUnexported(si.EventRecord{}))
 
+	// ensure that the underlying array of the return slice of CollectEvents() isn't the same as the one in EventStore.events
```
chia7712, line 59: Maybe we can add more records after calling `CollectEvents`, and then we do following checks

1.  the two slices have different points
2.  the elements of first slice from `CollectEvents` are not changed after we put more records.

with those checks we don't need to expose a internal method

chia7712 (commented):
@chenyulin0719 thanks for updated PR. a couple of comments left

pkg/events/event_store.go
chenyulin0719: Removed the test only method.

pkg/events/event_store_test.go
chenyulin0719: Change to another version, I think we can directly access the internal event in the test.

chenyulin0719: Change to the third version that accesses store.events directly in the test.

pkg/events/event_store.go
chia7712: that is a good finding. Could you add UT for that?

chenyulin0719: Done. 74b9c1ee4e8ba7638e62c2c1b4647940419ac18d
Extend the existing test: TestSetStoreSize()

pbacsko (approved):
+1 LGTM

chia7712 (approved):
LGTM

PR #846, opened 2024-04-18
WIP [YUNIKORN-2562] Nil pointer panic in Application.ReplaceAllocation()

pbacsko:
### What is this PR for?
`Application.placeholderData` is not populated properly during recovery. Panic occurs when placeholder allocations are restored and a new placeholder is added with a new taskgroup.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2562

### Questions:


pkg/scheduler/objects/application.go
```diff
@@ -695,6 +695,10 @@ func (sa *Application) RecoverAllocationAsk(ask *AllocationAsk) {
 				zap.Error(err))
 		}
 	}
+
+	if ask.IsPlaceholder() {
```
chia7712, line 699: Could we extract those behavior to a new method and then let `RecoverAllocationAsk` and `AddAllocationAsk` to reuse that method? That can avoid that `RecoverAllocationAsk` and `AddAllocationAsk` have diverged in the future.

pbacsko: Done.

wilfred-s (approved):
LGTM
code coverage misses already existed and are tagged as new as the code was moved

PR #847, opened 2024-04-19
[YUNIKORN-2570] Add test cases to break the current preemption flow

manirajv06:
### What is this PR for?
 
Add various test cases to break the current preemption flow. These test would fail now. Hence added skipped these test cases. Follow up jira's https://issues.apache.org/jira/browse/YUNIKORN-2500 should fix the problems in current preemption flow so that these test cases should pass.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2570

### Questions:


manirajv06:
As expected, all newly added tests are failing. Added skip statement to get a clean run.

manirajv06:
As expected, all newly added tests are failing. Added skip statement to get a clean run.

pkg/scheduler/objects/preemption_test.go
```diff
@@ -226,6 +231,1194 @@ func TestTryPreemption(t *testing.T) {
...
+// Guaranteed and Max resource set on both victim queue path and preemptor queue path. victim and preemptor queue are siblings.
+// Ask (Preemptor) resource type matches with 1 resource type of the victim. Guaranteed also set on specific resource type. 2 Victims are available, but total resource usage is lesser than ask requirement.
+// Setup:
+// Nodes are Node1 and Node2. Both Nodes are almost full. No space to accommodate the ask. Node 2 won't even fit because max capacity itself is lesser than ask requirement. Hence, Node 1 only is eligible.
+// root.parent.Max set on parent, first: 6
+// root.parent.child1. Guaranteed not set. 2 Allocations (belongs to single app) are running. 1st Allocation usage is first:4, pods: 1. 2nd Allocation usage is first:2, pods: 1. Total usage is first:6, pods: 2.
+// root.parent.child2. Guaranteed set, first: 5. Ask of first:5 is waiting for resources.
+// 2 Allocation on root.parent.child1 has been found and considered as victims and preempted to free up resources for ask.
+func TestTryPreemptionWithVictimsAvailableButOnDifferentNodes(t *testing.T) {
```
pbacsko, line 471: nit: function name suggestion: `TestTryPreemption_VictimsAvailableOnDifferentNodes`.

pkg/scheduler/objects/preemption_test.go
```diff
@@ -226,6 +231,1194 @@ func TestTryPreemption(t *testing.T) {
...
+// TestPreemptionOnNodeWithAskResTypesDifferedFromGuaranteedSetOnPreemptorSide Test try preemption with 2 level queue hierarchy. Since Node doesn't have enough resources to accomodate, preemption happens because of node resource constraint.
+// Guaranteed set only on preemptor queue path, but not on the victim queue path.
+// Ask (Preemptor) resource type matches with one of the victim's resource types.  Still, needs to be preempted because matching resource type has been configured as guaranteed.
+// Setup:
+// Nodes are Node1. Node is full, doesn't enough space to accommodate the ask.
+// root.parent.parent1.child1. Guaranteed set on root.parent.parent1, vcores: 1. Ask of vcores: 1 is waiting for resources.
+// root.parent.parent2.child2. 2 Allocations (belongs to single app) are running. Each Allocation usage is vcores:1, mem: 200. Total usage is vcores:2, mem: 400
+// root.parent.parent2.child3. No usage, no guaranteed set
+// 1 Allocation on root.parent.parent1.child2 should be preempted to free up resources for ask arrived in root.parent.parent1.child1.
+func TestPreemptionOnNodeWithAskResTypesDifferedFromGuaranteedSetOnPreemptorSide(t *testing.T) {
```
pbacsko, line 768: nit: naming: TestTryPreemptionOnNode_AskResTypesDifferent_GuaranteedOnPreemptorSide`.

pbacsko (dismissed):
I have very little context, so for now I only suggested some renamings. I realized that there are more tests than I had thought so I stopped after a while but it should make sense after that point.

Another idea: you can consider moving these tests to a different file, because they're much more complicated and thorough than normal preemption unit tests. For example, `preemption_scenarios_test.go`. In the header, you can explain what types of tests they are.

manirajv06:
Renamed all new test methods. Very few tests were there before. So, Let us keep it in same file.

craigcondit (approved):
+1 LGTM.

pbacsko (approved):
LGTM

PR #850, opened 2024-04-21
[YUNIKORN-2572]delete unused conditional

ryankert01:
### What is this PR for?
This [conditional](https://github.com/apache/yunikorn-core/blob/55e9fe650760bdf30d27a7ecf85b34d5f96c492f/pkg/common/resources/resources.go#L98) is not used because intValue will never be negative, since it's before filtered by [regular expression](https://github.com/apache/yunikorn-core/blob/55e9fe650760bdf30d27a7ecf85b34d5f96c492f/pkg/common/resources/quantity.go#L38).
```go
if intValue < 0 {
            return nil, fmt.Errorf("negative resources not permitted: %v", configMap)
}
```

Thus, it will never omit error about negative resource, instead it will raise invalid quantity error at https://github.com/apache/yunikorn-core/blob/55e9fe650760bdf30d27a7ecf85b34d5f96c492f/pkg/common/resources/quantity.go#L74.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2572

```sh
make test
```

### Questions:
N/A

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -95,9 +95,6 @@ func NewResourceFromConf(configMap map[string]string) (*Resource, error) {
 		if err != nil {
 			return nil, err
 		}
-		if intValue < 0 {
-			return nil, fmt.Errorf("negative resources not permitted: %v", configMap)
-		}
 		res.Resources[key] = intValue
 	}
 	return res, nil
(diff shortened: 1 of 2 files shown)


chenyulin0719 (changes requested):
Hi @ryankert01,
I have checked that negative resources are handled in ParseVCore() and ParseQuantity. Could you add a test case for the CPU to prove that this removal did not break NewResourceFromConf()?

https://github.com/plsmaop/yunikorn-core/blob/7e962e4b3e23024a3137894160d64343a8254543/pkg/common/resources/resources_test.go#L121

chenyulin0719 (approved):
+1 LGTM

PR #853, opened 2024-04-22
[YUNIKORN-2574] totalPartitionResource should not be mutated with AddTo/SubFrom

pbacsko:
### What is this PR for?
Do not mutate the `totalPartitionResource` field with `AddTo()` and `SubFrom()` methods. These should be only used for local variables for quick, in-place updates. Struct fields that are often read by other goroutines must use the immutable-based approach (eg. `resources.Add()`, `resources.Sub()`, etc).

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2574

### Questions:

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -580,7 +580,7 @@ func (pc *PartitionContext) updatePartitionResource(delta *resources.Resource) {
 		if pc.totalPartitionResource == nil {
 			pc.totalPartitionResource = delta.Clone()
 		} else {
-			pc.totalPartitionResource.AddTo(delta)
+			pc.totalPartitionResource = resources.Add(pc.totalPartitionResource, delta)
 		}
 		pc.root.SetMaxResource(pc.totalPartitionResource)
 	}
@@ -608,7 +608,7 @@ func (pc *PartitionContext) addNodeResources(node *objects.Node) {
 	if pc.totalPartitionResource == nil {
 		pc.totalPartitionResource = node.GetCapacity().Clone()
 	} else {
-		pc.totalPartitionResource.AddTo(node.GetCapacity())
+		pc.totalPartitionResource = resources.Add(pc.totalPartitionResource, node.GetCapacity())
 	}
 	pc.root.SetMaxResource(pc.totalPartitionResource)
 	log.Log(log.SchedPartition).Info("Updated available resources from added node",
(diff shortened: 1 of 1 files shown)


pkg/scheduler/partition.go
```diff
@@ -608,7 +608,7 @@ func (pc *PartitionContext) addNodeResources(node *objects.Node) {
 	if pc.totalPartitionResource == nil {
 		pc.totalPartitionResource = node.GetCapacity().Clone()
 	} else {
-		pc.totalPartitionResource.AddTo(node.GetCapacity())
```
chia7712, line 611: Should we replace `AddTo` by `resources.Add` for all? After that, we can remove `AddTo`

chia7712 (approved):
LGTM. one question is left.

pkg/scheduler/partition.go
pbacsko: `AddTo()` & its pairs have their use, see `total` in `removeAllocation()`. That's a local resource object which will not be stored anywhere and its lifecycle ends after the method returns. So this is how it's supposed to be used. But for variables which are potentially read&processed by other goroutines, using the immutable approach is safer because it just works without `Clone()`.

chia7712: `removeAllocation` it seems to me `total` can be initialized to `nil`, and then we just assign the new value (`delta`/`alloc.GetAllocatedResource()`) to `total`. Hence, `AddTo` can get removed from `removeAllocation`

craigcondit: I think in this case, the mutation is happening under lock within updatePartitionResource(), which is fine. Instead, we should add Clone() to GetTotalPartitionResource(). We should never return a mutable reference from that.

craigcondit (changes requested):
-1. See comments.

wilfred-s:
`GetTotalPartitionResource()` should return a cloned version of the current value. The value should not be mutable outside of the partition. A clone would prevent that.
The call is only made in the REST handlers or the health check code, outside any test code.Overhead would be minimal and it is the safest solution.

pkg/scheduler/partition.go
```diff
@@ -582,7 +582,7 @@ func (pc *PartitionContext) updatePartitionResource(delta *resources.Resource) {
 		} else {
 			pc.totalPartitionResource.AddTo(delta)
 		}
-		pc.root.SetMaxResource(pc.totalPartitionResource)
+		pc.root.SetMaxResource(pc.totalPartitionResource.Clone())
```
chia7712, line 585: not sure whether this `clone` is required since `SetMaxResource` do clone also (https://github.com/apache/yunikorn-core/blob/78862645c7e3f790b4bf2543865ba5a85c115163/pkg/scheduler/objects/queue.go#L1279)

chia7712 (commented):
@pbacsko thanks for updated PR. one question is left.

pkg/scheduler/partition.go
pbacsko: Ah cool. Good catch. I'll undo this.

craigcondit (approved):
+1 LGTM.

chia7712 (approved):
LGTM

PR #854, opened 2024-04-23
[YUNIKORN-2458] Remove ask repeats from AllocationAsk

craigcondit:
### What is this PR for?
Simplify ask and allocation handling by removing support for repeated requests in a single ask. This is functionality that is not used by the shim. By removing support for repeated asks, we also ensure that there is a 1:1 relationship between ask and allocation.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [x] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2458

Updated existing unit tests as this is largely a refactoring.

### Questions:


pkg/scheduler/objects/application.go
```diff
@@ -695,44 +697,60 @@ func (sa *Application) addAllocationAskInternal(ask *AllocationAsk) {
...
-func (sa *Application) updateAskRepeatInternal(ask *AllocationAsk, delta int32) (*resources.Resource, error) {
-	// updating with delta does error checking internally
-	if !ask.updatePendingAskRepeat(delta) {
-		return nil, fmt.Errorf("ask repaeat not updated resulting repeat less than zero for ask %s on app %s", ask.GetAllocationKey(), sa.ApplicationID)
+func (sa *Application) allocateAsk(ask *AllocationAsk) (*resources.Resource, error) {
+	if !ask.allocate() {
+		return nil, fmt.Errorf("unable to allocate previously allocated ask %s on app %s", ask.GetAllocationKey(), sa.ApplicationID)
+	}
+
+	if ask.GetPriority() >= sa.askMaxPriority {
```
chia7712, line 723: Pardon me, why it is `>=` rather than `>`?

pkg/scheduler/objects/application.go
```diff
@@ -695,44 +697,60 @@ func (sa *Application) addAllocationAskInternal(ask *AllocationAsk) {
 	}
 }
 
-func (sa *Application) UpdateAskRepeat(allocKey string, delta int32) (*resources.Resource, error) {
+func (sa *Application) AllocateAsk(allocKey string) (*resources.Resource, error) {
```
chia7712, line 700: It seems this method is used by testing only, so should we keep it in production code?

chia7712 (commented):
@craigcondit nice cleanup. two questions are left. PTAL

pkg/scheduler/objects/application.go
craigcondit: Because if we're removing an ask (by allocating it), the max priority of the app may decrease. If the priority is < the app's current max, then another ask exists that is of greater priority, so that calculation is not needed.

craigcondit: The equal check is needed because this ask may be the one that is holding the app priority high (i.e. equal).

chia7712: Got it. thanks!

pkg/scheduler/objects/application.go
craigcondit: It's needed (unfortunately) as there are multiple test packages that call it (pkg/scheduler and pkg/scheduler/objects). Therefore it needs to be here in order to be available to both.

pbacsko (approved):
+1 LGTM.

chia7712 (approved):
LGTM

PR #855, opened 2024-04-24
[YUNIKORN-2583] Possible log spew on DEBUG level from objects.Node

pbacsko:
### What is this PR for?
Large amount of log is written to the console in certain situations from `objects.Node` on DEBUG level. Reduce it with `RateLimitedLogger`.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2583

### Questions:

### DIFF
--- a/pkg/scheduler/objects/node.go
+++ b/pkg/scheduler/objects/node.go
@@ -21,6 +21,8 @@ package objects
 import (
 	"fmt"
 	"strings"
+	"sync"
+	"time"
 
 	"go.uber.org/zap"
 
@@ -32,6 +34,9 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/common"
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
+
+var initNodeLogOnce sync.Once
+var rateLimitedNodeLog *log.RateLimitedLogger
 
 const (
 	UnknownInstanceType = "UNKNOWN"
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/node.go
```diff
@@ -412,7 +417,7 @@ func (sn *Node) preConditions(ask *AllocationAsk, allocate bool) bool {
 func (sn *Node) preAllocateCheck(res *resources.Resource, resKey string) bool {
 	// cannot allocate zero or negative resource
 	if !resources.StrictlyGreaterThanZero(res) {
-		log.Log(log.SchedNode).Debug("pre alloc check: requested resource is zero",
+		getRateLimitedNodeLog().Debug("pre alloc check: requested resource is zero",
```
pbacsko, line 420: Note: it's not in the PR (yet), but we might do something with L#425-430:
```
	if sn.IsReserved() {
		if !sn.isReservedForApp(resKey) {
			log.Log(log.SchedNode).Debug("pre alloc check: node reserved for different app or ask",
				zap.String("nodeID", sn.NodeID),
				zap.String("resKey", resKey))
			return false
		}
	}
```
Not sure how much output this generates. We can just log the successful reservation (which is NOT logged at all right now, see `Node.Reserve()`) and ignore this.

craigcondit:
I'm really not a fan of this. DEBUG is not used by default so there is no log spew in the normal case. However, when debugging we want all this detail. RateLimitedLogger doesn't take into account the parameters to the logger so it hides messages which are not duplicates of each other. This makes actual debugging much more difficult. Most loggers that do rate limiting only collapse completely duplicated messages, and only for a short time, so you'd see something like "last message repeated 28 times", etc. Our rate-limited logger doesn't do this. I don't think it's appropriate to rate-limit DEBUG messages.

pbacsko:
Update: I had a discussion w/ @wilfred-s and looks like it's a bit more complicated. Counting is nice, but we can do that only if the logged output is exactly the same. In our case, the NodeID keeps chaning. It would still help somewhat because we run the scheduling cycle multiple times per second, but the net effect is less stellar than I originally thought. I'm dropping this for the time being, maybe we can get back to this later.

PR #863, opened 2024-05-01
[YUNIKORN-2598] Update the unit test for CheckQueues and checkNodeSortingPolicy functions

SophieTech88:
### What is this PR for?
There are many methods not covered by configvalidator_test.go. This commit works on unit Test for checkQueues function.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* Update the unit Test for checkQueues function.

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2598

This is a unite test. All existing unit tests should pass.

### Questions:

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -506,7 +506,7 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
 		}
 
 		if existedGroupName[name] {
-			return fmt.Errorf("duplicated group name %s , already existed", name)
+			return fmt.Errorf("duplicated group name '%s' , already existed", name)
 		}
 		existedGroupName[name] = true
 
@@ -654,7 +654,7 @@ func checkQueues(queue *QueueConfig, level int) error {
 	queueMap := make(map[string]bool)
 	for _, child := range queue.Queues {
 		if !QueueNameRegExp.MatchString(child.Name) {
-			return fmt.Errorf("invalid child name %s, a name must only have alphanumeric characters,"+
+			return fmt.Errorf("invalid child name '%s', a name must only have alphanumeric characters,"+
 				" - or _, and be no longer than 64 characters", child.Name)
 		}
 		if queueMap[strings.ToLower(child.Name)] {
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -489,7 +489,7 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
 		}
 
 		if existedUserName[name] {
-			return fmt.Errorf("duplicated user name %s , already existed", name)
+			return fmt.Errorf("duplicated user name '%s', already exists", name)
 		}
 		existedUserName[name] = true
 
@@ -658,7 +658,7 @@ func checkQueues(queue *QueueConfig, level int) error {
 				" - or _, and be no longer than 64 characters", child.Name)
 		}
 		if queueMap[strings.ToLower(child.Name)] {
-			return fmt.Errorf("duplicate child name found with name %s, level %d", child.Name, level)
+			return fmt.Errorf("duplicate child name found with name '%s', level %d", child.Name, level)
 		}
 		queueMap[strings.ToLower(child.Name)] = true
 	}
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -632,7 +632,7 @@ func checkNodeSortingPolicy(partition *PartitionConfig) error {
 // Check the queue names configured for compliance and uniqueness
 // - no duplicate names at each branched level in the tree
 // - queue name is alphanumeric (case ignore) with - and _
-// - queue name is maximum 16 char long
+// - queue name is maximum 64 char long
 func checkQueues(queue *QueueConfig, level int) error {
 	// check the ACLs (if defined)
 	err := checkACL(queue.AdminACL)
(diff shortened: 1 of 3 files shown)


pkg/common/configs/configvalidator_test.go
```diff
@@ -2012,3 +2012,227 @@ func TestCheckQueuesStructure(t *testing.T) {
...
+				Limits: []Limit{
+					{
+						Limit: "user-limit",
+						Users: []string{"user1", "user2", "user1"},
+					},
+				},
+			},
+			level:            0,
+			errorExpected:    true,
+			expectedErrorMsg: "duplicated user name user1 , already existed",
```
pbacsko, line 2106: This error message needs to be fixed in the config validator. 
It should be "duplicate user name user1, already exists". Or just simply `user name "user1" already exists`. There's no need for the extra whitespaces either.

pkg/common/configs/configvalidator_test.go
```diff
@@ -2012,3 +2012,227 @@ func TestCheckQueuesStructure(t *testing.T) {
...
+				AdminACL:  "admin",
+				SubmitACL: "submit",
+				Queues: []QueueConfig{
+					{Name: "duplicateQueue"},
+					{Name: "duplicateQueue"},
+				},
+			},
+			level:            0,
+			errorExpected:    true,
+			expectedErrorMsg: "duplicate child name found with name duplicateQueue, level 0",
```
pbacsko, line 2062: Let's put duplicateQueue in single quotes for consistency: `duplicate child name found with name 'duplicateQueue', level 0`

pkg/common/configs/configvalidator_test.go
```diff
@@ -2012,3 +2012,227 @@ func TestCheckQueuesStructure(t *testing.T) {
...
+						Queues: []QueueConfig{
+							{Name: "duplicateQueue"},
+							{Name: "duplicateQueue"},
+						},
+					},
+				},
+			},
+			level:            0,
+			errorExpected:    true,
+			expectedErrorMsg: "duplicate child name found with name duplicateQueue, level 1",
```
pbacsko, line 2084: same here

pbacsko (changes requested):
Minor comments, they are related to the existing error messages. The PR itself looks good but please fix the error messages in the config validator.

Also add `//nolint:funlen` to the test function.

pkg/common/configs/configvalidator_test.go
SophieTech88: for checkLimit function, it check the user name is duplicate or not. 
The original code is [here](https://github.com/apache/yunikorn-core/blob/master/pkg/common/configs/configvalidator.go#L491):
```
if existedUserName[name] {
    return fmt.Errorf("duplicated user name %s , already existed", name)
}
```
And I just want to trigger the error of `checkLimits` for `chckQueues` function. 
So the expectedErrorMsg is `duplicated user name user1 , already existed`. The whitespaces exist . So I need to match the expectedErrorMsg to the return error.
Do I need to change the both error message?

pkg/common/configs/configvalidator_test.go
SophieTech88: Something similar here, the expectedErrorMsg need to match the return error when it call the function `checkQueues`, the error message to duplicate name is [here](https://github.com/apache/yunikorn-core/blob/master/pkg/common/configs/configvalidator.go#L660):
```
if queueMap[strings.ToLower(child.Name)] {
    return fmt.Errorf("duplicate child name found with name %s, level %d", child.Name, level)
}
```
Do I need to add single quote `'` to the error messages for both?

pkg/common/configs/configvalidator_test.go
pbacsko: Yes, you need to change both, that was my request. The error message is weird, so why not fix it?

SophieTech88: Sure. Let me update both. Thanks.

pkg/common/configs/configvalidator_test.go
```diff
@@ -2012,3 +2012,227 @@ func TestCheckQueuesStructure(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckQueues(t *testing.T) { //nolint:funlen
+	testCases := []struct {
+		name             string
+		queue            *QueueConfig
+		level            int
+		errorExpected    bool
```
chia7712, line 2021: It seems `validateFunc` is not called if `errorExpected` is true. Maybe we can merge them to simplify code.

chia7712 (commented):
@SophieTech88 thanks for your patch. one small comment left.

pkg/common/configs/configvalidator_test.go
SophieTech88: Update the code to remove the errorExpected variable and simplify code. Many thanks for the suggestion.

pkg/common/configs/configvalidator_test.go
SophieTech88: Update the code.

pkg/common/configs/configvalidator.go
```diff
@@ -489,7 +489,7 @@ func checkLimit(limit Limit, existedUserName map[string]bool, existedGroupName m
 		}
 
 		if existedUserName[name] {
```
pbacsko, line 491: We need some spell check here as well... let's do it in a different JIRA.

SophieTech88: Sure.

SophieTech88:
@pbacsko Many thanks for the comments and suggestions. Just make a follow-up to check whether there is any thing I need to do for this PR. Thanks

chia7712 (approved):
LGTM

chia7712:
@SophieTech88 thanks for your contribution!

PR #865, opened 2024-05-10
[YUNIKORN-2590] Handler tests should check for nil request on create

ryankert01:
### What is this PR for?
In the handler_test.go file we have an anti pattern showing a large number (40+) warnings in an IDE:

'req' might have 'nil' or other unexpected value as its corresponding error variable might be not 'nil'

The warning are due to the fact that we have the following pattern:

```
req, err = http.NewRequest("GET", "path", strings.NewReader(""))
req = req.WithContext(context.WithValue(req.Context(), httprouter.ParamsKey, httprouter.Params{}))
```
There is no error assertion after the request creation. We should add a simple `assert.NilError(t, err, "HTTP request create failed")` inserted between creating and using the request.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
n/a

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2590

Github CI

### Questions:

### DIFF
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -1070,6 +1070,7 @@ func TestGetPartitionQueuesHandler(t *testing.T) {
 
 	var req *http.Request
 	req, err = http.NewRequest("GET", "/ws/v1/partition/default/queues", strings.NewReader(""))
+	assert.NilError(t, err, "HTTP request create failed")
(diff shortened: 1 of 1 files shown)


chia7712 (approved):
LGTM

PR #872, opened 2024-05-23
[YUNIKORN-2633] Unnecessary warning from Partition when adding an application

pbacsko:
### What is this PR for?
Prevent misleading WARN message from Partition when adding an application.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2633

### Questions:

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -339,13 +339,18 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
 		return fmt.Errorf("failed to find queue %s for application %s", queueName, appID)
 	}
 
-	// set resources based on tags, but only if the queue is dynamic (unmanaged)
-	if queue.IsManaged() {
-		log.Log(log.SchedQueue).Warn("Trying to set resources on a queue that is not an unmanaged leaf",
-			zap.String("queueName", queue.QueuePath))
-	} else {
-		queue.SetResources(app.GetGuaranteedResource(), app.GetMaxResource())
+	guaranteedRes := app.GetGuaranteedResource()
+	maxRes := app.GetMaxResource()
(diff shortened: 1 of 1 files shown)


pkg/scheduler/partition.go
```diff
@@ -339,13 +339,18 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
-		queue.SetResources(app.GetGuaranteedResource(), app.GetMaxResource())
+	guaranteedRes := app.GetGuaranteedResource()
+	maxRes := app.GetMaxResource()
+	if guaranteedRes != nil || maxRes != nil {
+		// set resources based on tags, but only if the queue is dynamic (unmanaged)
+		if queue.IsManaged() {
+			log.Log(log.SchedQueue).Warn("Trying to set resources on a queue that is not an unmanaged leaf",
+				zap.String("queueName", queue.QueuePath))
+		} else {
+			queue.SetResources(guaranteedRes, maxRes)
```
chia7712, line 350: This is unrelated to this PR, but what happens if this dynamic queue is `@recovery@` queue? IIRC, recovery queue has no quota

chia7712 (approved):
@pbacsko thanks for this patch. LGTM if my comment is invalid.

pkg/scheduler/partition.go
pbacsko: Yes, you're right. Perhaps we should filter for the recovery queue. I think I'll create a separate JIRA for that.

PR #875, opened 2024-05-25
[YUNIKORN-2643] utils.go WaitForCondition  improvement

mean-world:
### What is this PR for?
test case for WaitForCondition in utils.go

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring
* [x] - test

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2643

make test

### Questions:


pbacsko (commented):
Actually I think `WaitForCondition()` should be removed. We already have `WaitFor()` which is essentially the same (the deadline check is performer earlier). Existing stuff should be migrated to `WaitFor()` instead of adding new code.

chia7712:
> Actually I think WaitForCondition() should be removed. We already have WaitFor() which is essentially the same (the deadline check is performer earlier). Existing stuff should be migrated to WaitFor() instead of adding new code.

agree

chia7712:
k8shim module has a same method 

https://github.com/apache/yunikorn-k8shim/blob/58adfe941d2d8dae5544af8b49e435f304678807/pkg/common/utils/utils.go#L245

maybe we can use core's version to replace it.

pkg/common/utils.go
```diff
@@ -224,6 +227,9 @@ func ZeroTimeInUnixNano(t time.Time) *int64 {
 }
 
 func WaitForCondition(eval func() bool, interval time.Duration, timeout time.Duration) error {
+	if interval > timeout {
```
chia7712, line 230: Personally, changing the `interval` is a bit awkward since that breaks the definition. Maybe we can rewrite this method.

```go
	deadline := time.Now().Add(timeout)
	for {
		if eval() {
			return nil
		}

		if time.Now().After(deadline) || time.Now().Add(interval).After(deadline) {
			return fmt.Errorf("timeout waiting for condition")
		}

		time.Sleep(interval)
	}
```

In short, we return timeout error early if it will be timeout due to sleeping.

pkg/common/utils.go
```diff
@@ -63,6 +63,9 @@ func GetPartitionNameWithoutClusterID(partitionName string) string {
 }
 
 func WaitFor(interval time.Duration, timeout time.Duration, condition func() bool) error {
```
chia7712, line 65: Could we keep one of `WaitForCondition` or `WaitFor`? The k8shim module has `WaitForCondition`, and so maybe we can keep `WaitForCondition` for consistency.

https://github.com/apache/yunikorn-k8shim/blob/58adfe941d2d8dae5544af8b49e435f304678807/pkg/common/utils/utils.go#L245

chia7712 (commented):
@mean-world thanks for updated PR

pkg/common/utils.go
pbacsko: I'm OK with either approach.

pkg/common/utils.go
wilfred-s: Can we define a fixed error in `common.errors.go` (new file) like this:
```
package common

import (
	"errors"
)

var (
	// ErrorTimeout returned if waiting for a condition times out
	ErrorTimeout = errors.New("timeout waiting for condition")
)
```
That will make error chaining easy and allow using `errors.Is()` in all kinds of places. It would be inline with the constants definition we already have

pbacsko (changes requested):
Please remove `WaitFor()` or  `WaitForCondition()` because we don't need both.

pkg/common/utils.go
```diff
@@ -223,15 +209,15 @@ func ZeroTimeInUnixNano(t time.Time) *int64 {
 	return &tInt
 }
 
-func WaitForCondition(eval func() bool, interval time.Duration, timeout time.Duration) error {
+func WaitForCondition(interval time.Duration, timeout time.Duration, eval func() bool) error {
```
chia7712, line 212: Just curious, why we change the argument order? I assume we can do similar cleanup for k8hime module. And this changes of order will make difference between core and k8shim.

mean-world: Since the argument order of the WaitFor function is different from that of the WaitForCondition function, I need to replace the WaitFor function with the WaitForCondition function. However, since the WaitFor function is more frequently used in the project than the WaitForCondition function, I have decided to change the argument order of the WaitForCondition function.

pkg/webservice/handler_mock_test.go
```diff
@@ -1,19 +1,19 @@
...
- limitations under the License.
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
```
chia7712, line 10: Could you please revert those unrelated changes?

pkg/common/utils_test.go
```diff
@@ -234,36 +235,29 @@ func TestConvertSITimestamp(t *testing.T) {
...
+	tests := []struct {
+		input    bool
+		interval time.Duration
+		timeout  time.Duration
+		output   error
 	}{
-		{"Timeout case", 10000, true},
-		{"Fullfilling case", 10, false},
+		{true, time.Duration(1) * time.Second, time.Duration(2) * time.Second, nil},
+		{false, time.Duration(1) * time.Second, time.Duration(2) * time.Second, fmt.Errorf("timeout waiting for condition")},
```
chia7712, line 250: `fmt.Errorf("timeout waiting for condition")` should be replaced by `ErrorTimeout`, right?

chia7712 (commented):
@mean-world thanks for updated PR. two small questions are left.

pkg/common/utils.go
chia7712: that makes sense to me

pkg/common/utils_test.go
```diff
@@ -234,36 +234,28 @@ func TestConvertSITimestamp(t *testing.T) {
...
+	tests := []struct {
+		input    bool
+		interval time.Duration
+		timeout  time.Duration
+		output   error
 	}{
-		{"Timeout case", 10000, true},
-		{"Fullfilling case", 10, false},
+		{true, time.Duration(1) * time.Second, time.Duration(2) * time.Second, nil},
+		{false, time.Duration(1) * time.Second, time.Duration(2) * time.Second, ErrorTimeout},
```
chia7712, line 249: Could you please add test for "interval > timeout"? It should call `eval` only once

pkg/common/error.go
```diff
@@ -0,0 +1,28 @@
+/*
```
pbacsko, line 1: No need to create a new file just for a single error variable. You can move it to `utils.go`.

pbacsko (changes requested):
See comment - move `ErrorTimeout` to `utils.go`.

PR #877, opened 2024-05-29
[YUNIKORN-2646] Allow lock order detection to be turned off

wilfred-s:
### What is this PR for?
Most of the false positive deadlock detections are linked to the preemption cycle. In that cycle we have a lock on the application being scheduled and acquire read locks on other applications to look for victims.
In the next scheduling cycle the applications involved might be reversed triggering an inconsistent lock detection to be logged. That causes log spew.
The lock order detection can be turned off by setting an option exposed via a new environment variable.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
* [YUNIKORN-2646](https://issues.apache.org/jira/browse/YUNIKORN-2646)

Unit test for the new option and deadlock detection added

### DIFF
--- a/pkg/locking/locking_race_test.go
+++ b/pkg/locking/locking_race_test.go
@@ -22,7 +22,6 @@
 package locking
 
 import (
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/locking/locking_race_test.go
```diff
@@ -43,3 +44,46 @@ func TestDeadlockDetection(t *testing.T) {
...
+			}()
+			wg.Wait()
+			wg.Add(1)
+			go func() { // lock ordering: b, a, a, b
+				defer wg.Done()
+				b.Lock()
+				a.RLock()
+				a.RUnlock()
+				b.Unlock()
+			}()
```
pbacsko, line 84: Potential simplification: I tried this without goroutines and the result is the same.

wilfred-s: done

pbacsko (approved):
+1

PR #880, opened 2024-05-29
[YUNIKORN-2649] Improve CalculateAbsUsedCapacity & CompUsageRatio funtion's test coverage

SP12893678:
### What is this PR for?
Improve CalculateAbsUsedCapacity & CompUsageRatio funtion's test coverage

### What type of PR is it?
* [x] - Test

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2649

### DIFF
--- a/pkg/common/resources/resources_test.go
+++ b/pkg/common/resources/resources_test.go
@@ -1205,6 +1205,110 @@ func TestGetShares(t *testing.T) {
 	}
 }
 
+func TestCompUsageRatio(t *testing.T) {
+	tests := []struct {
+		left     *Resource
+		right    *Resource
+		total    *Resource
+		expected int
+		message  string
+	}{
+		{
+			left:     nil,
+			right:    nil,
+			total:    nil,
+			expected: 0,
+			message:  "nil resources",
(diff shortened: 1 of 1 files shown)


pkg/common/resources/resources_test.go
```diff
@@ -1205,6 +1205,110 @@ func TestGetShares(t *testing.T) {
...
+		{
+			left:     NewResource(),
+			right:    NewResource(),
+			total:    NewResource(),
+			expected: 0,
+			message:  "empty resources",
+		},
+		{
+			left:     &Resource{Resources: map[string]Quantity{"zero": 0}},
+			right:    &Resource{Resources: map[string]Quantity{"zero": 0}},
```
pbacsko, line 1239: Is there a particular reason why you directly create the type instead of calling `NewResourceXXX()`?

SP12893678: I refer other test case use this way.
Thanks for the suggestion, I will change to use method call.

pbacsko (approved):
+1

PR #885, opened 2024-06-04
[YUNIKORN-2663] Improve ACL struct funtion's test coverage

SP12893678:
### What is this PR for?
Remove unreachable code in NewACL func
Improve the following funtion's test coverage in acl.go
- TestSetUsers
- TestSetGroups

### What type of PR is it?
* [x] - Test

### What is the Jira issue?
https://[issues.apache.org/jira/browse/YUNIKORN-2663](https://issues.apache.org/jira/browse/YUNIKORN-2663)

### DIFF
--- a/pkg/common/security/acl.go
+++ b/pkg/common/security/acl.go
@@ -49,10 +49,6 @@ func (a *ACL) setAllAllowed(part string) {
 // set the user list in the ACL, invalid user names are ignored
 func (a *ACL) setUsers(userList []string) {
 	a.users = make(map[string]bool)
-	// list could be empty
-	if len(userList) == 0 {
-		return
-	}
 	// special case if the user list is just the wildcard
 	if len(userList) == 1 && userList[0] == common.Wildcard {
 		log.Log(log.Security).Info("user list is wildcard, allowing all access")
@@ -78,10 +74,6 @@ func (a *ACL) setUsers(userList []string) {
 // set the group list in the ACL, invalid group names are ignored
 func (a *ACL) setGroups(groupList []string) {
 	a.groups = make(map[string]bool)
-	// list could be empty
-	if len(groupList) == 0 {
-		return
-	}
 	// special case if the wildcard was already set
 	if a.allAllowed {
 		log.Log(log.Security).Info("ignoring group list in ACL: wildcard set")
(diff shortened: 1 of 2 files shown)


pkg/common/security/acl.go
```diff
@@ -123,10 +123,6 @@ func NewACL(aclStr string) (ACL, error) {
 	}
 	// trim and check for wildcard
 	acl.setAllAllowed(aclStr)
-	// an empty ACL is a deny
```
SP12893678, line 126: The above code `fields := strings.Split(aclStr, common.Space)` will must lead to `fields` length more than 0
So it will be unreachable for the code block.

SP12893678, line 126: The above code `fields := strings.Split(aclStr, common.Space)` will must lead to `fields` length more than 0
So it will be unreachable for the condition expression.

SP12893678 (commented):
The above code `fields := strings.Split(aclStr, common.Space)` will must lead to `fields` length more than 0
So it will be unreachable for the code block.

pkg/common/security/acl.go
```diff
@@ -123,10 +123,6 @@ func NewACL(aclStr string) (ACL, error) {
...
-	if len(fields) == 0 {
-		return acl, nil
```
SP12893678, line 128: The above code `fields := strings.Split(aclStr, common.Space)` will must lead to `fields` length more than 0
So it will be unreachable for the code block.

pbacsko (changes requested):
Both `setUsers()` and `setGroups()` are called from `NewACL()`, so there's no need to test them separately. Make sure that we have coverage though and instead of adding new test, extend `TestACLCreate()`.

As I can see, we miss three cases:
- `setUsers()`: when `if len(userList) == 0` true 
- `setUsers()`: when `if userNameRegExp.MatchString(user)` false
- `setGroups()`: when if `len(groupList) == 0` true

pbacsko:
> However, due to strings.Split, which results in an array with a length of at least 1, the following two scenarios are unreachable. 

Then just simply remove the `if` branch. In fact, even if an empty slice is passed, we do very little extra work, so nothing happens. This is not hot code, so just go ahead and remove the path which is never taken. This also increases coverage in %. In both functions, the early return is an unnecessary optimization.

SP12893678:
> Then just simply remove the `if` branch. In fact, even if an empty slice is passed, we do very little extra work, so nothing happens. This is not hot code, so just go ahead and remove the path which is never taken. This also increases coverage in %. In both functions, the early return is an unnecessary optimization.

Sure, not performing the empty list check will not affect the current logic.

pbacsko (approved):
+1

PR #886, opened 2024-06-05
[YUNIKORN-2515] Add property event.RESTResponseSize to the batch event handler

pbacsko:
### What is this PR for?
Update the code with the property `event.RESTResponseSize` as per the design document.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2515

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -87,20 +87,7 @@ func init() {
 	streamingLimiter = NewStreamingLimiter()
 
 	configs.AddConfigMapCallback("rest-response-size", func() {
-		newSize := configs.DefaultRESTResponseSize
-		configMap := configs.GetConfigMap()
-		if value, ok := configMap[configs.CMRESTResponseSize]; ok {
-			parsed, err := strconv.ParseUint(value, 10, 64)
-			if err != nil {
-				log.Log(log.REST).Warn("Failed to parse configuration value",
-					zap.String("key", configs.CMRESTResponseSize),
-					zap.String("value", value),
-					zap.Error(err))
-			} else {
-				newSize = parsed
-			}
-		}
-
+		newSize := common.GetConfigurationUint(configs.GetConfigMap(), configs.CMRESTResponseSize, configs.DefaultRESTResponseSize)
 		if newSize == 0 {
 			log.Log(log.REST).Warn("Illegal value `0` for config key, using default",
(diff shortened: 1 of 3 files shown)


pbacsko:
I think `Test_fixedRule_ruleDAO` needs to be corrected, the map iteration order sometimes changes.

pbacsko:
Created #887 to fix the test case. Will rebase this PR as soon as that one gets merged.

chia7712:
I have another idea of implementing this new property.

 `event.RESTResponseSize` can be applied to `EventSystemImpl` and thus `GetEventsFromID` can pick up `min(count, RESTResponseSize)` to generate `EventRecord`. That is similar to `requestCapacity` and `ringBufferCapacity` WDYT?

https://github.com/apache/yunikorn-core/blob/d46f216e2fa85a38edff11d48e82f481a0c03f3c/pkg/events/event_ringbuffer.go#L104

pbacsko:
> `event.RESTResponseSize` can be applied to `EventSystemImpl` and thus `GetEventsFromID` can pick up `min(count, > RESTResponseSize)` to generate `EventRecord`. That is similar to `requestCapacity` and `ringBufferCapacity` WDYT?

Hm... I'd rather keep this separate. This property concerns the REST output, I think it's better to handle it separately from the event system.

pkg/webservice/handlers.go
```diff
@@ -83,6 +85,35 @@ func init() {
 	allowedActiveStatusMsg = fmt.Sprintf("Only following active statuses are allowed: %s", strings.Join(activeStatuses, ","))
 
 	streamingLimiter = NewStreamingLimiter()
+
+	configs.AddConfigMapCallback("rest-response-size", func() {
+		newSize := configs.DefaultRESTResponseSize
+		configMap := configs.GetConfigMap()
+		if value, ok := configMap[configs.CMRESTResponseSize]; ok {
```
chia7712, line 92: How about reusing `common.GetConfigurationUint`?

chia7712 (commented):
@pbacsko thanks for updated PR. overall LGTM. one last comment :)

chia7712 (approved):
LGTM

PR #891, opened 2024-06-13
[YUNIKORN-2657] Validate queue generated as part of the placement rules

manirajv06:
### What is this PR for?

Queue name being generated as part of the various placement rule should follow similar validation checks already being carried out for queuename coming through config.

### What type of PR is it?
* [ ] - Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2657

### Questions:

### DIFF
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -69,7 +69,11 @@ var DefaultPreemptionDelay = 30 * time.Second
 
 // A queue can be a username with the dot replaced. Most systems allow a 32 character user name.
 // The queue name must thus allow for at least that length with the replacement of dots.
-var QueueNameRegExp = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,64}$`)
+var QueueNameRegExp = regexp.MustCompile(`^[a-zA-Z0-9_:#/@-]{1,64}$`)
+
+// A queue can be a username with the dot replaced. Most systems allow a 32 character user name.
+// The queue name must thus allow for at least that length with the replacement of dots.
+var QueuePathRegExp = regexp.MustCompile(`^[a-zA-Z0-9_.:#/@-]{1,64}$`)
 
 // User and group name check: systems allow different things POSIX is the base but we need to be lenient and allow more.
 // allow upper and lower case, add the @ and . (dot) and officially no length.
@@ -653,9 +657,9 @@ func checkQueues(queue *QueueConfig, level int) error {
 	// check this level for name compliance and uniqueness
 	queueMap := make(map[string]bool)
 	for _, child := range queue.Queues {
-		if !QueueNameRegExp.MatchString(child.Name) {
-			return fmt.Errorf("invalid child name '%s', a name must only have alphanumeric characters,"+
-				" - or _, and be no longer than 64 characters", child.Name)
+		err = IsQueueNameValid(child.Name)
+		if err != nil {
+			return err
 		}
 		if queueMap[strings.ToLower(child.Name)] {
 			return fmt.Errorf("duplicate child name found with name '%s', level %d", child.Name, level)
@@ -669,6 +673,22 @@ func checkQueues(queue *QueueConfig, level int) error {
 		if err != nil {
 			return err
 		}
+	}
+	return nil
+}
+
+func IsQueueNameValid(queueName string) error {
+	if !QueueNameRegExp.MatchString(queueName) {
+		return fmt.Errorf("invalid child name '%s', a name must only have alphanumeric characters,"+
+			" - or _, and be no longer than 64 characters", queueName)
+	}
+	return nil
+}
+
+func IsQueuePathValid(queuePath string) error {
+	if !QueuePathRegExp.MatchString(queuePath) {
+		return fmt.Errorf("invalid queue path '%s', a queue path must only have alphanumeric characters,"+
+			" - or _, and be no longer than 64 characters", queuePath)
 	}
 	return nil
 }
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -675,8 +675,7 @@ func checkQueues(queue *QueueConfig, level int) error {
 
 func IsQueueNameValid(queueName string) error {
 	if !QueueNameRegExp.MatchString(queueName) {
-		return fmt.Errorf("invalid child name '%s', a name must only have alphanumeric characters,"+
-			" - or _, and be no longer than 64 characters", queueName)
+		return fmt.Errorf("invalid queue name '%s', max 64 characters consisting of alphanumeric characters and '-', '_', '#', '@', '/', ':' allowed", queueName)
 	}
 	return nil
 }
--- a/pkg/common/configs/configvalidator.go
+++ b/pkg/common/configs/configvalidator.go
@@ -71,10 +71,6 @@ var DefaultPreemptionDelay = 30 * time.Second
 // The queue name must thus allow for at least that length with the replacement of dots.
 var QueueNameRegExp = regexp.MustCompile(`^[a-zA-Z0-9_:#/@-]{1,64}$`)
 
-// A queue can be a username with the dot replaced. Most systems allow a 32 character user name.
-// The queue name must thus allow for at least that length with the replacement of dots.
-var QueuePathRegExp = regexp.MustCompile(`^[a-zA-Z0-9_.:#/@-]{1,64}$`)
-
 // User and group name check: systems allow different things POSIX is the base but we need to be lenient and allow more.
 // allow upper and lower case, add the @ and . (dot) and officially no length.
 var UserRegExp = regexp.MustCompile(`^[_a-zA-Z][a-zA-Z0-9:#/_.@-]*[$]?$`)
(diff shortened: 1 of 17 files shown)


pkg/common/configs/configvalidator.go
```diff
@@ -69,7 +69,11 @@ var DefaultPreemptionDelay = 30 * time.Second
 
 // A queue can be a username with the dot replaced. Most systems allow a 32 character user name.
 // The queue name must thus allow for at least that length with the replacement of dots.
-var QueueNameRegExp = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,64}$`)
+var QueueNameRegExp = regexp.MustCompile(`^[a-zA-Z0-9_:#/@-]{1,64}$`)
+
+// A queue can be a username with the dot replaced. Most systems allow a 32 character user name.
+// The queue name must thus allow for at least that length with the replacement of dots.
+var QueuePathRegExp = regexp.MustCompile(`^[a-zA-Z0-9_.:#/@-]{1,64}$`)
```
pbacsko, line 76: Fix the comment, that is a copy-paste from `QueueNameRegExp`.

The maximum length of 64 is a bit too low, not sure what's a reasonable limit here, I'd simply not validate the length of the path at all.

pbacsko (changes requested):
Fix lintern issues + see comment regarding the path regexp.

pkg/common/configs/configvalidator.go
wilfred-s: Not sure that this will even work at all...

pkg/common/configs/configvalidator.go
```diff
@@ -673,6 +677,22 @@ func checkQueues(queue *QueueConfig, level int) error {
...
 
+func IsQueueNameValid(queueName string) error {
+	if !QueueNameRegExp.MatchString(queueName) {
+		return fmt.Errorf("invalid child name '%s', a name must only have alphanumeric characters,"+
+			" - or _, and be no longer than 64 characters", queueName)
+	}
+	return nil
+}
+
+func IsQueuePathValid(queuePath string) error {
```
wilfred-s, line 688: Not sure how this would work as we allow dots which means the length cannot be limited as each part can be 64 characters and I could have 10, 20 or more levels in a queue.

These valid queue fails this check:
`root.parent_lvl1.parent_lvl2.parent_lvl3.parent_lvl4.parent_lvl5.parent_lvl6.parent_lvl7.parent_lvl8.parent_lvl9.parent_lvl10.leaf`
or
`root.company_name.a-long-namespace-name.wilfred_dot_spiegelenburg`

Checking each part before building the path should be enough to handle this. See further comments below

pkg/scheduler/placement/fixed_rule.go
```diff
@@ -121,12 +121,19 @@ func (fr *fixedRule) placeApplication(app *objects.Application, queueFn func(str
 		if parentName == "" {
 			parentName = configs.RootQueue
 		}
-		queueName = parentName + configs.DOT + fr.queue
+		childQueueName := replaceDot(fr.queue)
+		if err = configs.IsQueueNameValid(childQueueName); err != nil {
+			return "", err
+		}
+		queueName = parentName + configs.DOT + childQueueName
```
wilfred-s, line 128: This breaks a use case. I can set the fixed part to a partial path and not qualify it.
So I can set "prod.high" as the fixed part, not qualified. Use a parent rule of `User` to finally create a queue `root.wilfred.prod.high`, after this change it would create `root.wilfred.prod_dot_high`

The fixed part compliance for all parts of the possible unqualified path should be added to the `initialise()` code

pkg/scheduler/placement/fixed_rule.go
```diff
@@ -121,12 +121,19 @@ func (fr *fixedRule) placeApplication(app *objects.Application, queueFn func(str
...
 	}
 	// Log the result before we check the create flag
 	log.Log(log.SchedApplication).Debug("Fixed rule intermediate result",
 		zap.String("application", app.ApplicationID),
 		zap.String("queue", queueName))
+	if err := configs.IsQueuePathValid(queueName); err != nil {
+		return "", err
+	}
```
wilfred-s, line 136: If each part is valid the path must be valid. We should have checked each part before we get to the final queue path assembly.
At this point we cannot distinguish anymore between valid and invalid `.` dot characters so it adds nothing.

wilfred-s (changes requested):

pkg/scheduler/placement/fixed_rule.go
wilfred-s: Did not repeat the comment for every rule this was part of.

pkg/scheduler/placement/fixed_rule.go
manirajv06: Yes, we should not replace dot in this rule. https://issues.apache.org/jira/browse/YUNIKORN-2660 has been filed to make the changes like you explained.

pkg/common/configs/configvalidator.go
manirajv06: Yes, length check is not required. However, I still see a need to do validate that code path because there is an option to pass the full queue path in the config. If we really want to avoid full queuePath check, we should split the string based on dot and apply queueName check itself for each part.

pkg/scheduler/placement/testrule.go
```diff
@@ -70,6 +71,12 @@ func (tr *testRule) placeApplication(app *objects.Application, queueFn func(stri
 		return "", fmt.Errorf("nil app passed in")
 	}
 	if queuePath := app.GetQueuePath(); queuePath != "" {
+		parts := strings.Split(queuePath, configs.DOT)
+		for _, part := range parts {
+			if err := configs.IsQueueNameValid(part); err != nil {
+				return "", err
```
pbacsko, line 77: Make sure this is covered by a testcase.

pkg/scheduler/objects/queue.go
```diff
@@ -177,8 +177,8 @@ func NewDynamicQueue(name string, leaf bool, parent *Queue) (*Queue, error) {
 		return nil, fmt.Errorf("dynamic queue can not be added without parent: %s", name)
 	}
 	// name might not be checked do it here
-	if !configs.QueueNameRegExp.MatchString(name) {
-		return nil, fmt.Errorf("invalid queue name '%s', a name must only have alphanumeric characters, - or _, and be no longer than 64 characters", name)
+	if !configs.QueueNameRegExp.MatchString(name) || name == common.RecoveryQueue {
+		return nil, fmt.Errorf("invalid queue name '%s', a name must only have alphanumeric characters, - or _, and be no longer than 64 characters except the recovery queue root.@recovery@", name)
```
pbacsko, line 181: Error message is not exact. Now we also allow `:`, `#`, `/`, and `@`.
I suggest returning different, shorter error messages. This message is too long.
1) If recovery queue: `the queue name cannot be root.@recovery@`
2) Longer than 64 chars: `queue name is longer than 64 characters`
3) If regexp fails: `invalid queue name '%s', a name must consists of alphanumeric characters and '-', '_', '#', '@', '/', ':'`

No.2) obviously affects the regexp, but I prefer simpler regexps anyway. Checking the length separately is not a big deal.

pkg/scheduler/objects/queue.go
```diff
@@ -177,8 +177,8 @@ func NewDynamicQueue(name string, leaf bool, parent *Queue) (*Queue, error) {
 		return nil, fmt.Errorf("dynamic queue can not be added without parent: %s", name)
 	}
 	// name might not be checked do it here
-	if !configs.QueueNameRegExp.MatchString(name) {
-		return nil, fmt.Errorf("invalid queue name '%s', a name must only have alphanumeric characters, - or _, and be no longer than 64 characters", name)
+	if !configs.QueueNameRegExp.MatchString(name) || name == common.RecoveryQueue {
```
pbacsko, line 180: Re-use `IsQueueNameValid()`.

pkg/common/configs/configvalidator.go
```diff
@@ -673,6 +673,14 @@ func checkQueues(queue *QueueConfig, level int) error {
 	return nil
 }
 
+func IsQueueNameValid(queueName string) error {
+	if !QueueNameRegExp.MatchString(queueName) {
+		return fmt.Errorf("invalid child name '%s', a name must only have alphanumeric characters,"+
+			" - or _, and be no longer than 64 characters", queueName)
```
pbacsko, line 679: Simplify error message, see below.

pbacsko (changes requested):
Some smaller things.

manirajv06:
Moved queueName checks to initialise block for fixed rule so that https://issues.apache.org/jira/browse/YUNIKORN-2660 also have been taken care of.

pkg/scheduler/objects/queue.go
```diff
@@ -177,8 +177,11 @@ func NewDynamicQueue(name string, leaf bool, parent *Queue) (*Queue, error) {
 		return nil, fmt.Errorf("dynamic queue can not be added without parent: %s", name)
 	}
 	// name might not be checked do it here
-	if !configs.QueueNameRegExp.MatchString(name) {
-		return nil, fmt.Errorf("invalid queue name '%s', a name must only have alphanumeric characters, - or _, and be no longer than 64 characters", name)
+	if err := configs.IsQueueNameValid(name); err != nil {
+		return nil, err
```
pbacsko, line 181: L#181 final branch that needs coverage.

pbacsko:
Now it looks good, I just have a final comment left regarding coverage.

pbacsko (approved):
LGTM

wilfred-s (approved):
LGTM

PR #876, opened 2024-05-28
[YUNIKORN-2542] Consistent logging and tracker handling for increment/decrement

9501sam:
### What is this PR for?
1. For logging consistency, make the logs to  `applicationID` and `appGroup` consistent.
2. For handling group tracker consistency, move the call to  `GroupTracker.increaseTrackedResource()`  from `UserTracker.increaseTrackedResource()` to `Manager.IncreaseTrackedResource()` like what `Manager.DecreaseTrackedResource()` does. Also added similar log for this.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2542

### How should this be tested?
`make test`
Added unit test `TestIncreaseTrackedResourceForGroupTracker()` for this PR.

### Questions:
* [ ] - The licenses files need update.
* [ ] - There is breaking changes for older versions.
* [ ] - It needs documentation.

PR #896, opened 2024-06-19
[YUNIKORN-2683] Unnecessary error is logged when resource usage is increased

pbacsko:
### What is this PR for?
Don't log an error when there's no group tracker, this can be a perfectly normal situation.

The problem was caused by a refactor done in YUNIKORN-2542 (#876).

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2683

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -107,9 +107,6 @@ func (m *Manager) IncreaseTrackedResource(queuePath, applicationID string, usage
 		zap.Stringer("resource", usage))
 	groupTracker := m.GetGroupTracker(appGroup)
 	if groupTracker == nil {
-		log.Log(log.SchedUGM).Error("group tracker should be available in groupTrackers map",
-			zap.String("application", applicationID),
-			zap.String("group", appGroup))
 		return
 	}
 	log.Log(log.SchedUGM).Debug("Increasing resource usage for group",
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -99,6 +99,9 @@ func (m *Manager) IncreaseTrackedResource(queuePath, applicationID string, usage
 	}
 	userTracker.increaseTrackedResource(queuePath, applicationID, usage)
 	appGroup := userTracker.getGroupForApp(applicationID)
+	if appGroup == common.Empty {
+		return
+	}
 	log.Log(log.SchedUGM).Debug("Increasing resource usage for user",
 		zap.String("user", user.User),
 		zap.String("queue path", queuePath),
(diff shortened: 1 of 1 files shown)


wilfred-s (approved):

pkg/scheduler/ugm/manager.go
```diff
@@ -107,9 +107,6 @@ func (m *Manager) IncreaseTrackedResource(queuePath, applicationID string, usage
 		zap.Stringer("resource", usage))
 	groupTracker := m.GetGroupTracker(appGroup)
 	if groupTracker == nil {
-		log.Log(log.SchedUGM).Error("group tracker should be available in groupTrackers map",
```
chia7712, line 110: Should we align the behavior of `DecreaseTrackedResource`[0] with this change?

[0] https://github.com/apache/yunikorn-core/blob/97c38a97a55ee9b457c366c11a4ae32348093e2c/pkg/scheduler/ugm/manager.go#L110

chia7712 (commented):
@pbacsko overall +1 except for one small question.

pkg/scheduler/ugm/manager.go
pbacsko: In that case we're guared by:
```
	if appGroup == common.Empty {
		return
	}
```
In this code path, we already check it inside `ensureGroupTrackerForApp()`.

chia7712: That is interesting. We do the similar check/ensure for `IncreaseTrackedResource`

https://github.com/apache/yunikorn-core/blob/97c38a97a55ee9b457c366c11a4ae32348093e2c/pkg/scheduler/ugm/manager.go#L98

It seems the error message makes sense since `groupTracker` should be existent in the code path?

pbacsko: It doesn't have to be there. Before the refactor, we did not have this check because you can call `IncreaseTrackedResource()` on a `nil` group tracker. However I don't like that approach, just want to return on `nil`.

chia7712: Got it. thanks!

chia7712: Sorry for one more question:

> In that case we're guarded by:

why we not adding similar code to `IncreaseTrackedResource` for consistency? It means that is fine to return directly if `appGroup` is empty. Also, we add error log if `appGroup` is NOT empty and `ensureGroupTrackerForApp` is called?

pbacsko: Check the latest version.

pkg/scheduler/ugm/manager.go
```diff
@@ -99,6 +99,9 @@ func (m *Manager) IncreaseTrackedResource(queuePath, applicationID string, usage
 	}
 	userTracker.increaseTrackedResource(queuePath, applicationID, usage)
 	appGroup := userTracker.getGroupForApp(applicationID)
+	if appGroup == common.Empty {
```
chia7712, line 102: Maybe this check should happen after log "Increasing resource usage for user"?

chia7712 (commented):
@pbacsko thanks for updated PR!

chia7712 (approved):
LGTM

PR #892, opened 2024-06-14
[YUNIKORN-2652] Add Resource Usage details to getApplication endpoint

richscott:
### What is this PR for?
These changes enhance the `getApplication()` function in `pkg/webservice/` to include some additional fields that may be valuable to users viewing completed applications - specifically, the response now included `StartTime`, `ResourceUsage`, `PreemptedResource`, and `PlaceholderResource` members.  

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2652

### How should this be tested?
Any request to the endpoint(s) that use the `getApplication()` function can be queried, with an additional boolean query parameter of `details` set to true, e.g.
```
$ curl --silent 'http://localhost:9889/ws/v1/partition/default/queue/root.default/application/app-1?details=true' | jq
```

PR #897, opened 2024-06-19
[YUNIKORN-2652] Add Resource Usage details to getApplication endpoint

suprjinx:
### What is this PR for?

These changes enhance the `getApplication()` function in `pkg/webservice/` to include some additional fields that may be valuable to users viewing completed applications - specifically, the response now included `StartTime`, `ResourceUsage`, `PreemptedResource`, and `PlaceholderResource` members.

### What type of PR is it?                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      * [ ]  - Bug Fix
* [x]  - Improvement
* [ ]  - Feature
* [ ]  - Documentation
* [ ]  - Hot Fix
* [ ]  - Refactoring
                                                                                                                                                                                                                                                             
### Todos
* [ ]  - Task            

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2652

Any request to the endpoint(s) that use the `getApplication()` function can be queried, with an additional boolean query parameter of `details` set to true, e.g.

```
$ curl --silent 'http://localhost:9889/ws/v1/partition/default/queue/root.default/application/app-1?details=true' | jq
```


suprjinx:
This is a duplicate of PR #892 with some unit test coverage.

pbacsko:
Please fix the lintern problem, you can check locally by running "make lint".

pkg/webservice/dao/application_info.go
```diff
@@ -44,6 +48,14 @@ type ApplicationDAOInfo struct {
 	MaxRequestPriority int32                   `json:"maxRequestPriority,omitempty"`
 }
 
+type ApplicationDetailsDAOInfo struct {
```
pbacsko, line 51: Apologies for the late review, I was on vacation for 2 weeks.

I'd rather put this in `ApplicationDAOInfo`. In fact, retrieving the summary is a fast operation, so I'm not sure that we need `details=true/false` in the URL. Unless there's an objection from someone, I'd just add these new fields to the existing `ApplicationDAOInfo`

suprjinx: done!

pkg/webservice/handlers_test.go
```diff
@@ -1690,6 +1694,28 @@ func TestGetApplicationHandler(t *testing.T) {
...
+	var appDao *dao.ApplicationDAOInfo
+	getApplication(resp6, req6)
+	appSummary := app.GetApplicationSummary(partitionNameWithoutClusterID)
+	err = json.Unmarshal(resp6.outputBytes, &appDao)
+	assert.NilError(t, err, unmarshalError)
+	assert.Equal(t, "app-1", appDao.ApplicationID)
+	assert.Equal(t, app.StartTime().UnixMilli(), appDao.StartTime)
+	assert.Equal(t, appSummary.ResourceUsage.String(), appDao.ResourceUsage.String())
+	assert.Equal(t, appSummary.PreemptedResource.String(), appDao.PreemptedResource.String())
+	assert.Equal(t, appSummary.PlaceholderResource.String(), appDao.PlaceholderResource.String())
```
pbacsko, line 1718: Use `resources.Equals()` to verify equality

pbacsko (changes requested):
Just a small nit.

Also add `//nolint:funlen` to `TestGetApplicationHandler()`.

pkg/webservice/handlers_test.go
richscott: @pbacsko (background: I am collaborating with @suprjinx on this PR.) I just added the `//nolint:funlen` as you requested. I attempted to use `resources.Equals()` on those 3 checks for `ResourceUsage`, `PreemptedResource`, and `PlaceholderResource`, however `resources.Equals()` takes two `*resource.Resource` arguments, and these 3 items are actually `*resources.TrackedResource`, which the compiler won't allow.  I wasn't able to implement an analogous `Equals(a, b *TrackedResource)` func, since the `resources` package already has `Equals(a, b *Resource)` and both `Resource` and `TrackedResource` are in the same `resources` package. So, should I make a `EqualsTrackedResource()` func, and rename the first to `EqualsResource()`, or just stay with the current test logic of `String()` comparison? Thanks.

pbacsko: @richscott yes it's a different type I just found out...

I'd be happier with sth like `EqualsTracked()` created for `TrackedResource`. You can put it inside `tracked_resource.go`. The first layer is the extra code, the rest can be taken from the existing `Equals()`.

Eg.
```
func EqualsTracked(left, right *TrackedResource) bool {
	if left == right {
		return true
	}

	if left == nil || right == nil {
		return false
	}

	equals := true
	for k, v := range left.TrackedResourceMap {
		inner, ok := right.TrackedResourceMap[k]
		if !ok {
			return false
		}

		equals = equals && equalsMapContents(v, inner)
	}

	return equals
}

// basically the same as Equals()
func equalsMapContents(left, right map[string]int64) bool {
	if left == nil || right == nil {  // probably this is not needed
		return false
	}

	for k, v := range left {
		if right[k] != v {
			return false
		}
	}

	for k, v := range right {
		if left[k] != v {
			return false
		}
	}

	return true
}
```

I haven't tried this, but the main point is the approach. It would be much better if `TrackedResourceMap` was actually `map[string]*Resource` because we could re-use `Equals()` but this can be improved in a follow-up JIRA.

richscott: @pbacsko  - updated as suggested - thanks for the code sample. I agree that adjusting `TrackedResourceMap` to use `map[string]*Resource` would be a nice clean-up in the near future. Thanks.

richscott: I see that Codecov is failing its check - I will add a test for `EqualsTracked(..)` (mostly based off the `TestEquals(t)` in `resources_test.go`) to get the coverage up a bit so it passes.

pkg/common/resources/tracked_resources.go
```diff
@@ -103,3 +103,46 @@ func (tr *TrackedResource) AggregateTrackedResource(instType string,
...
+			return false
+		}
+
+		equals = equals && equalsMapContents(v, inner)
+	}
+
+	return equals
+}
+
+// basically the same as Equals()
```
pbacsko, line 129: comment can be deleted

pkg/common/resources/tracked_resources.go
```diff
@@ -103,3 +103,46 @@ func (tr *TrackedResource) AggregateTrackedResource(instType string,
...
+func equalsMapContents(left, right map[string]int64) bool {
+	if left == nil || right == nil { // probably this is not needed
+		return false
+	}
```
pbacsko, line 133: L131-133 can be deleted

pkg/common/resources/tracked_resources.go
pbacsko, line 129: comment can be deleted

pkg/common/resources/tracked_resources.go
```diff
@@ -103,3 +103,46 @@ func (tr *TrackedResource) AggregateTrackedResource(instType string,
...
+	for k, v := range left.TrackedResourceMap {
+		inner, ok := right.TrackedResourceMap[k]
+		if !ok {
+			return false
+		}
+
+		equals = equals && equalsMapContents(v, inner)
+	}
+
+	return equals
```
pbacsko, line 126: Slightly better without `equals` variable:

```
	for k, v := range left.TrackedResourceMap {
		inner, ok := right.TrackedResourceMap[k]
		if !ok {
			return false
		}

		if !equalsMapContents(v, inner) {
			return false
		}
	}

	return true
```

pkg/common/resources/tracked_resources_test.go
```diff
@@ -210,3 +210,55 @@ func TestTrackedResourceAggregateTrackedResource(t *testing.T) {
...
+			var base, compare *TrackedResource
+			if tt.input.base != nil {
+				base = NewTrackedResourceFromMap(tt.input.base)
+			}
+			if tt.input.compare != nil {
+				compare = NewTrackedResourceFromMap(tt.input.compare)
+			}
+
+			if result := EqualsTracked(base, compare); result != tt.expected {
+				t.Errorf("Equal result should be %v instead of %v, left %v, right %v", tt.expected, result, base, compare)
```
pbacsko, line 257: nit: use `assert.Assert()` instead of `t.Errorf()`

pkg/common/resources/tracked_resources_test.go
```diff
@@ -210,3 +210,55 @@ func TestTrackedResourceAggregateTrackedResource(t *testing.T) {
...
+			}
+			if result := EqualsTracked(compare, base); result != tt.expected {
+				t.Errorf("Equal result should be %v instead of %v, left %v, right %v", tt.expected, result, compare, base)
```
pbacsko, line 260: nit: use `assert.Assert()` instead of `t.Errorf()`

pbacsko (changes requested):
OK, some small things and it's OK to me

pbacsko:
Note: created [YUNIKORN-2777](https://issues.apache.org/jira/browse/YUNIKORN-2777) as a follow-up.

pkg/common/resources/tracked_resources.go
richscott: Fixed in commit 9b3c237

pkg/common/resources/tracked_resources.go
richscott: Fixed in commit 9b3c237

pkg/common/resources/tracked_resources.go
richscott: Nice improvement! Fixed in commit 9b3c237

pkg/common/resources/tracked_resources_test.go
richscott: Fixed in commit 9b3c237

pkg/common/resources/tracked_resources_test.go
richscott: Fixed in commit 9b3c237

pbacsko (approved):
+1

PR #903, opened 2024-06-26
[YUNIKORN-2568] Move all xxxEvents types to objects/events

pbacsko:
### What is this PR for?
Move the `xxxEvents` types to `objects\events`, because `objects` already contains too many files.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2568

### Questions:


pkg/scheduler/objects/allocation_ask.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/apache/yunikorn-core/pkg/events"
```
chia7712, line 30: maybe we can move the classes from `events` to `events/systems`? That can eliminate the reference conflicts.

chia7712 (commented):
@pbacsko thanks for this patch. one small comment is left

pkg/scheduler/objects/allocation_ask.go
pbacsko: I don't want to introduce too much changes. I've been thinking about a better package name inside `objects`, but everything else is too long or not specific enough.

pkg/scheduler/objects/events/application_events.go
```diff
@@ -0,0 +1,150 @@
...
+*/
+
+package events
+
+import (
+	"fmt"
+
+	"github.com/apache/yunikorn-core/pkg/common"
+	"github.com/apache/yunikorn-core/pkg/common/resources"
+	evt "github.com/apache/yunikorn-core/pkg/events"
```
chia7712, line 26: Do we need this alias? it seems there is no conflict

chia7712 (commented):
@pbacsko overall +1 except for one small question

pkg/scheduler/objects/events/application_events.go
pbacsko: Correct. Fixed.

chia7712 (approved):

PR #904, opened 2024-07-01
[YUNIKORN-2703] Core: Fallback to default queue if no placement rules match

mitdesai:
### What is this PR for?
Added a new field in the partition config to define the default queue. During the application placement, we will evaluate the rules and if no rule match, we will try to place the application in the default queue.

If the default queue config is not provided, the code will use 'root.default'

Putting the PR to review the changes.  I will make the shim changes as part of YUNIKORN-2711

### What type of PR is it?
* [X] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2703

### Questions:

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -21,11 +21,12 @@ package common
 const (
 	Empty = ""
 
-	Wildcard          = "*"
-	Separator         = ","
-	Space             = " "
-	AnonymousUser     = "nobody"
-	AnonymousGroup    = "nogroup"
-	RecoveryQueue     = "@recovery@"
-	RecoveryQueueFull = "root." + RecoveryQueue
+	Wildcard              = "*"
+	Separator             = ","
+	Space                 = " "
+	AnonymousUser         = "nobody"
+	AnonymousGroup        = "nogroup"
+	RecoveryQueue         = "@recovery@"
+	RecoveryQueueFull     = "root." + RecoveryQueue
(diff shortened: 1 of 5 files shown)


craigcondit:
-1 on the approach. Please do not add a new config field. We already have the configmap entry for this. Those can be retrieved via the extra config passed to the core (we do this for other things as well).

mitdesai:
Thanks for the review @craigcondit, I have addressed the comments.

The default queue value is taken from the configmap key that we have defined already. If the value is set, we use the defaultQueue defined in the config map otherwise fallback to 'root.default'

craigcondit (changes requested):
-1. Please don't add any new configuration (we already have a default queue). Additionally, there shouldn't be anything shim-specific in the core (such as AM).

mitdesai:
@craigcondit I have updated the design according to what we discussed today.
We still need changes in the core because that is where we will try to fall back to root.default queue if no placement queue was found after going through all the placement rules.

If we run out of rules, as a last resort, we look for 'root.default' and see if can place the app in that queue. Otherwise reject the application.

Changes to shim to remove passing the default queue: https://github.com/apache/yunikorn-k8shim/pull/874

PR #905, opened 2024-07-01
[YUNIKORN-2715] Handle special characters for params like queue, user…

manirajv06:
…name & groupname

### What is this PR for?
QueueName, UserName and GroupName may contain special characters which need to be escaped before hitting the url and the same params needs to be unescaped before processing at the server side.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2715

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"net/url"
 	"runtime"
 	"sort"
 	"strconv"
@@ -668,6 +669,10 @@ func getPartitionQueue(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	queueName := vars.ByName("queue")
+	if _, err := url.QueryUnescape(queueName); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
+		return
+	}
 	queueErr := validateQueue(queueName)
 	if queueErr != nil {
 		buildJSONErrorResponse(w, queueErr.Error(), http.StatusBadRequest)
@@ -737,6 +742,10 @@ func getQueueApplications(w http.ResponseWriter, r *http.Request) {
 	}
 	partition := vars.ByName("partition")
 	queueName := vars.ByName("queue")
+	if _, err := url.QueryUnescape(queueName); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
+		return
+	}
 	queueErr := validateQueue(queueName)
 	if queueErr != nil {
 		buildJSONErrorResponse(w, queueErr.Error(), http.StatusBadRequest)
@@ -820,6 +829,10 @@ func getApplication(w http.ResponseWriter, r *http.Request) {
 	}
 	partition := vars.ByName("partition")
 	queueName := vars.ByName("queue")
+	if _, err := url.QueryUnescape(queueName); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
+		return
+	}
 	application := vars.ByName("application")
 	partitionContext := schedulerContext.GetPartitionWithoutClusterID(partition)
 	if partitionContext == nil {
@@ -1078,6 +1091,10 @@ func getUserResourceUsage(w http.ResponseWriter, r *http.Request) {
 		buildJSONErrorResponse(w, UserNameMissing, http.StatusBadRequest)
 		return
 	}
+	if _, err := url.QueryUnescape(user); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
+		return
+	}
 	userTracker := ugm.GetUserManager().GetUserTracker(user)
 	if userTracker == nil {
 		buildJSONErrorResponse(w, UserDoesNotExists, http.StatusNotFound)
@@ -1112,6 +1129,10 @@ func getGroupResourceUsage(w http.ResponseWriter, r *http.Request) {
 	group := vars.ByName("group")
 	if group == "" {
 		buildJSONErrorResponse(w, GroupNameMissing, http.StatusBadRequest)
+		return
+	}
+	if _, err := url.QueryUnescape(group); err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 	groupTracker := ugm.GetUserManager().GetGroupTracker(group)
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -669,17 +669,17 @@ func getPartitionQueue(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	queueName := vars.ByName("queue")
-	escapedQueueName, err := url.QueryUnescape(queueName)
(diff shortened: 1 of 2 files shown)


chia7712 (commented):
@manirajv06 thanks for this patch. maybe we can add ut for it? for another, do we have docs for "all allowed characters"?

pkg/webservice/handlers.go
```diff
@@ -668,12 +669,17 @@ func getPartitionQueue(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	queueName := vars.ByName("queue")
-	queueErr := validateQueue(queueName)
+	escapedQueueName, err := url.QueryUnescape(queueName)
+	if err != nil {
+		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	queueErr := validateQueue(escapedQueueName)
```
pbacsko, line 677: This approach results in fewer changes:
```
	queueName, err := url.QueryUnescape(vars.ByName("queue"))
	if err != nil {
		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
		return
	}
	queueErr := validateQueue(queueName)
```
We don't need the unescaped string.

manirajv06: We need to throw "missing" kind of error before this unescaping step. In fact filed [jira](https://issues.apache.org/jira/browse/YUNIKORN-2712) to address the same earlier.

manirajv06:
Refactored lot of code using newly introduced helper method. Filed a [jira](https://issues.apache.org/jira/browse/YUNIKORN-2720) to use the same in all other places.

pkg/webservice/handlers.go
pbacsko: I realized that the variable name is still incorrect. It should either be called "unescapedQueueName" or rename the original to "escapedQueueName". If the unescaped is used more frequently, I'd just leave it and would call the output of `vars.ByName("queue")` as "escapedQueueName".

chia7712: Maybe we can introduce a small helper for all params?
```go
func byName(vars httprouter.Params, name string) (string, error) {
	value := vars.ByName(name)
	unescape, err := url.QueryUnescape(value)
	if err != nil {
		return "", err
	}
	return unescape, nil
}

	queueName, err := byName(vars, "queue")
	if err != nil {
		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
		return
	}
```

https://issues.apache.org/jira/browse/YUNIKORN-2712 can leverage this helper to add `missing` check

pbacsko: Sounds good to me.

manirajv06: > I realized that the variable name is still incorrect. It should either be called "unescapedQueueName" or rename the original to "escapedQueueName". If the unescaped is used more frequently, I'd just leave it and would call the output of `vars.ByName("queue")` as "escapedQueueName".

Yes, it should be "unescapedQueueName". Made the changes.

pkg/webservice/handlers_test.go
```diff
@@ -1248,13 +1224,45 @@ func TestGetPartitionQueuesHandler(t *testing.T) {
...
+		httprouter.Param{Key: "partition", Value: partitionNameWithoutClusterID},
+		httprouter.Param{Key: "queue", Value: queueName},
+	}))
+	resp = &MockResponseWriter{}
+	getPartitionQueue(resp, req)
+	assertQueueNotExists(t, resp)
+
+	// test queue name with special characters escaped not properly, catch error at request level
+	_, err = http.NewRequest("GET", partitionQueuesHandler+invalidQueueName, strings.NewReader(""))
+	assert.ErrorContains(t, err, "invalid URL escape")
```
pbacsko, line 1249: Is this needed? You're essentially just testing Go standard library itself.

pkg/webservice/handlers_test.go
```diff
@@ -1106,74 +1108,17 @@ func TestGetPartitionQueuesHandler(t *testing.T) {
...
-	assert.Equal(t, partitionQueuesDao.Parent, "")
-	assert.Assert(t, partitionQueuesDao.AbsUsedCapacity == nil)
-	assert.Equal(t, partitionQueuesDao.MaxRunningApps, uint64(0))
-	assert.Equal(t, partitionQueuesDao.RunningApps, uint64(0))
-	assert.Equal(t, partitionQueuesDao.CurrentPriority, configs.MinPriority)
-	assert.Assert(t, partitionQueuesDao.AllocatingAcceptedApps == nil)
-	assert.Equal(t, len(partitionQueuesDao.Properties), 1)
-	assert.Equal(t, partitionQueuesDao.Properties[configs.ApplicationSortPolicy], policies.FifoSortPolicy.String())
-	assert.DeepEqual(t, partitionQueuesDao.TemplateInfo, &templateInfo)
+	assertPartitionQueueDaoInfo(t, &partitionQueuesDao, configs.RootQueue, configs.DefaultPartition, nil, nil, false, true, "", &templateInfo)
```
pbacsko, line 1111: I'm all in for refactoring, but I do think this should be done in a separate ticket. This can make backporting more difficult if we mix up too much change.

pbacsko (changes requested):
Please do the refactor in a separate JIRA for better task separation. Otherwise I'm OK with the changes.

pkg/webservice/handlers_test.go
manirajv06: Yes, am aware of it. Felt it is better to do.

pkg/webservice/handlers_test.go
manirajv06: This whole things started with lint warning ( > x lines warnings) as this pr has touched those places. We need to fix these warnings to get a clean build. Then analysed the code for further improvements and thats how we landed up here.

pbacsko (approved):
LGTM after offline discussion

pkg/webservice/handlers_test.go
```diff
@@ -1976,22 +1999,43 @@ func TestSpecificUserAndGroupResourceUsage(t *testing.T) {
...
 	// Test params name missing
 	req, err = http.NewRequest("GET", "/ws/v1/partition/default/usage/group/", strings.NewReader(""))
-	assert.NilError(t, err, "Get Group Resource Usage Handler request failed")
+	assert.NilError(t, err)
 	resp = &MockResponseWriter{}
 	getGroupResourceUsage(resp, req)
 	assertParamsMissing(t, resp)
+
+	// Test group name with special characters escaped properly
+	validGroup := url.QueryEscape("test_a-b_c@#d@do:mai/n.com")
```
chia7712, line 2016: @manirajv06 this string can't be matched by `GroupRegExp`

https://github.com/apache/yunikorn-core/blob/6c8fd60b82fd0c1973a1c1aa9717fb6bf4440c30/pkg/common/configs/configvalidator.go#L79

not sure whether we should revise the RegExp or this test string is invalid

manirajv06: @chia7712 Yes, you are right. https://issues.apache.org/jira/browse/YUNIKORN-2719 has been filed for the same reason. Same applies for user name too.

PR #909, opened 2024-07-05
[YUNIKORN-2699] Preemption e2e tests fail in latest master

manirajv06:
### What is this PR for?
Incorrect victims list has been sent to shim for preemption. Fixed the issue by sending the correct list. 
Once victims has been chosen finally, there is one more round of filtering step to do further filter in TryPreemption(). After this filtering step, final list has been prepared. Instead of sending the final list to shim, earlier list has been sent. Hence, some victims has been preempted unnecessarily.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2699

### Questions:


pbacsko (approved):
+1

PR #918, opened 2024-07-18
[YUNIKORN-2262] propagate the error message when queue creation gets …

0lai0:
### What is this PR for?
The current error handling suppresses the root cause error message, making it difficult to derive insights from the generic message "failed to create rule based queue...".

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2262

make test

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -325,12 +325,12 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
 		if isRecoveryQueue {
 			queue, err = pc.createRecoveryQueue()
 			if err != nil {
-				return fmt.Errorf("failed to create recovery queue %s for application %s", common.RecoveryQueueFull, appID)
+				return fmt.Errorf("failed to create recovery queue %s for application %s : %w", common.RecoveryQueueFull, appID, err)
 			}
 		} else {
 			queue, err = pc.createQueue(queueName, app.GetUser())
 			if err != nil {
-				return fmt.Errorf("failed to create rule based queue %s for application %s", queueName, appID)
+				return fmt.Errorf("failed to create rule based queue %s for application %s : %w", queueName, appID, err)
 			}
 		}
 	}


pkg/scheduler/partition.go
```diff
@@ -324,12 +324,12 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
 		if isRecoveryQueue {
 			queue, err = pc.createRecoveryQueue()
 			if err != nil {
-				return fmt.Errorf("failed to create recovery queue %s for application %s", common.RecoveryQueueFull, appID)
+				return fmt.Errorf(strings.Join([]string{"failed to create recovery queue", queueName, "for application", appID, ":", err.Error()}, " "))
```
ryankert01, line 327: I think it should be error.join(), as mentioned in the jira comment section.

an quick example:
```go
package main

import (
	"errors"
	"fmt"
)

func main() {
	err1 := errors.New("err1")
	err2 := errors.New("err2")
	err := errors.Join(err1, err2)
	fmt.Println(err)
	if errors.Is(err, err1) {
		fmt.Println("err is err1")
	}
	if errors.Is(err, err2) {
		fmt.Println("err is err2")
	}
}
```

pkg/scheduler/partition.go
```diff
@@ -324,12 +324,12 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
 			}
 		} else {
 			queue, err = pc.createQueue(queueName, app.GetUser())
 			if err != nil {
-				return fmt.Errorf("failed to create rule based queue %s for application %s", queueName, appID)
+				return fmt.Errorf(strings.Join([]string{"failed to create rule based queue", queueName, "for application", appID, ":", err.Error()}, " "))
```
ryankert01, line 332: ditto.

ryankert01 (changes requested):

pkg/scheduler/partition.go
0lai0: Thank you for your feedback. Let me take a moment to reconsider the meaning of the message.

ryankert01 (approved):
LGTM

pkg/scheduler/partition.go
0lai0: I’ve finished modifying the PR, but I’m not sure if I have correctly expressed it in the Jira comment section. Please take a look.

pkg/scheduler/partition.go
```diff
@@ -324,12 +325,14 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
 		if isRecoveryQueue {
 			queue, err = pc.createRecoveryQueue()
 			if err != nil {
-				return fmt.Errorf("failed to create recovery queue %s for application %s", common.RecoveryQueueFull, appID)
+				outter_err := fmt.Errorf("failed to create recovery queue %s for application %s", common.RecoveryQueueFull, appID)
+				return errors.Join(outter_err, err)
```
chia7712, line 329: how about: `return errors.Join(fmt.Errorf("failed to create recovery queue %s for application %s", common.RecoveryQueueFull, appID), err)`

pkg/scheduler/partition.go
```diff
@@ -324,12 +325,14 @@ func (pc *PartitionContext) AddApplication(app *objects.Application) error {
...
 			}
 		} else {
 			queue, err = pc.createQueue(queueName, app.GetUser())
 			if err != nil {
-				return fmt.Errorf("failed to create rule based queue %s for application %s", queueName, appID)
+				outter_err := fmt.Errorf("failed to create rule based queue %s for application %s", queueName, appID)
+				return errors.Join(outter_err, err)
```
chia7712, line 335: ditto: `return errors.Join(fmt.Errorf("failed to create rule based queue %s for application %s", queueName, appID), err)`

chia7712 (commented):
@0lai0 thanks for updates

pkg/scheduler/partition.go
ryankert01: I think it does! Although I never tested it. 
Great job!

pkg/scheduler/partition.go
0lai0: I think it’s good, let me make some changes.

chia7712 (approved):
LGTM

PR #922, opened 2024-07-26
[YUNIKORN-2712] Missing specific param error for REST API

ryankert01:
### What is this PR for?
Some REST API's throw "missing specific param" kind of errors, but not all. For example, user name is missing. Similarly, all mandatory parameters in other REST API's can follow the same pattern. It is very clear, rather than saying "doesn't exists" kind of error.

Suggestion given in https://github.com/apache/yunikorn-core/pull/905#discussion_r1663068429 can be used as reference for implementation.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2712

```
make test
```

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -646,7 +646,7 @@ func getPartitionQueues(w http.ResponseWriter, r *http.Request) {
 	}
 	partitionName, err := byName(vars, "partition")
 	if err != nil {
-		buildJSONErrorResponse(w, MissingParamsName, http.StatusBadRequest)
+		buildJSONErrorResponse(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 	var partitionQueuesDAOInfo dao.PartitionQueueDAOInfo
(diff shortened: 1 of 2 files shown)


pbacsko (changes requested):
Thanks for the patch @ryankert01. The changes introduce new code paths without coverage. 

Please update the unit tests to have proper coverage.

ryankert01:
@pbacsko Thank you pointed that out. Just add some tests to fix the coverage.

wilfred-s:
In PR #530 we removed the missing parameter checks and moved to a simple nil check for the whole params array. We did that because the httprouter will not call the handler unless we have a match in the URL. The nil check is a defence in depth. In normal operation we can never call any of the handler functions without the variables set as the URL matching will fail before that. I think we forgot to clean up the user and group check.

If we can show that a call made via the httprouter, not directly calling the handler in test code, can have missing parameters we should add these checks. If they are always present, with a value or not, we should not add the check and leave it to the handled as a "Not Found" error. 
For this change I would go the other direction: remove specific param checks

ryankert01:
I agree. If httprouter does the deeds, we don't have to do them twice. 
gentle ping @manirajv06 WDYT?

ryankert01:
If there's no further objection until tomorrow, I will close this pr and open a jira to cleanup.

ryankert01:
open Jira on: https://issues.apache.org/jira/browse/YUNIKORN-2793

PR #923, opened 2024-07-30
[YUNIKORN-2769] Preemption fails between two siblings when preemptor …

manirajv06:
…is UG and parent is above OG

### What is this PR for?
Ask queue starving for resources (under guaranteed) located under parent who is over guaranteed because of another child usage should be able to preempt victims from sibling as it is deserve to get resources for its pending pods.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2769

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -615,12 +615,11 @@ func compareShares(lshares, rshares []float64) int {
 	return 0
 }
 
-// Compare the resources equal returns the specific values for following cases:
-// left  right  return
-// nil   nil    true
-// nil   <set>  false
-// <set> nil    false
-// <set> <set>  true/false  *based on the individual Quantity values
+// Equals Compare the resources based on common resource type available in both left and right Resource
+// Resource type available in left Resource but not in right Resource and vice versa is not taken into account
+// False in case anyone of the resources is nil
+// False in case resource type value differs
+// True in case when resource type values of left Resource matches with right Resource if resource type is available
 func Equals(left, right *Resource) bool {
 	if left == right {
 		return true
@@ -635,22 +634,20 @@ func Equals(left, right *Resource) bool {
 			return false
 		}
 	}
-
 	for k, v := range right.Resources {
 		if left.Resources[k] != v {
 			return false
 		}
 	}
-
 	return true
 }
 
-// DeepEquals Compare the resources equal returns the specific values for following cases:
-// left  right  return
-// nil   nil    true
-// nil   <set>  false
-// <set> nil    false
-// <set> <set>  true/false  *based on the individual resource type presence and its Quantity values
+// DeepEquals Compare the resources based on resource type existence and its values as well
+// False in case anyone of the resources is nil
+// False in case resource length differs
+// False in case resource type existed in left Resource not exist in right Resource
+// False in case resource type value differs
+// True in case when all resource type and its values of left Resource matches with right Resource
 func DeepEquals(left, right *Resource) bool {
 	if left == right {
 		return true
@@ -658,17 +655,11 @@ func DeepEquals(left, right *Resource) bool {
 	if left == nil || right == nil {
 		return false
 	}
+	if len(right.Resources) != len(left.Resources) {
+		return false
+	}
 	for k, v := range left.Resources {
 		if val, ok := right.Resources[k]; ok {
-			if val != v {
-				return false
-			}
-		} else {
-			return false
-		}
-	}
-	for k, v := range right.Resources {
-		if val, ok := left.Resources[k]; ok {
 			if val != v {
 				return false
 			}
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -645,6 +645,40 @@ func Equals(left, right *Resource) bool {
 	return true
 }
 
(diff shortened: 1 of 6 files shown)


pbacsko (changes requested):
I suspect this needs a rebase to get rid of the compilation error.

pkg/common/resources/resources.go
```diff
@@ -884,6 +918,29 @@ func ComponentWiseMinPermissive(left, right *Resource) *Resource {
 	return out
 }
 
+// ComponentWiseMinPermissiveWithPrecedence Returns a new Resource by giving higher precedence for resource type values present in left when
+// it is present at right as well. Resource type not present in left and present in right would be included too.
+// If either Resource passed in is nil the other Resource is returned
+// If a Resource type is missing from one of the Resource, it is considered empty and the quantity from the other Resource is returned
+func ComponentWiseMinPermissiveWithPrecedence(left, right *Resource) *Resource {
```
wilfred-s, line 925: not the right name we need something better.
At the moment I do not see this covering the action it does

wilfred-s (changes requested):

pbacsko (changes requested):
See comments.

manirajv06:
@craigcondit Can you take a look?

pkg/common/resources/resources.go
```diff
@@ -884,6 +909,29 @@ func ComponentWiseMinPermissive(left, right *Resource) *Resource {
...
+	if right == nil && left == nil {
+		return nil
+	}
+	if left == nil {
+		return right.Clone()
+	}
+	if right == nil {
+		return left.Clone()
+	}
+	out := left
```
craigcondit, line 926: I think this should be `out := left.Clone()`, otherwise you are mutating the original resource. This seems wrong, since you're returning a new one.

pkg/common/resources/resources_test.go
```diff
@@ -588,6 +588,46 @@ func TestComponentWiseMinOnlyExisting(t *testing.T) {
 	}
 }
 
+func TestComponentWiseMinPermissiveWithPrecedence(t *testing.T) {
```
craigcondit, line 591: This test name makes no sense. You're not calling that method from the test at all.

pkg/scheduler/objects/preemption_queue_test.go
```diff
@@ -123,113 +123,185 @@ func TestGetPreemptableResource(t *testing.T) {
...
+		assert.Assert(t, resources.IsZero(rootRemaining), "guaranteed set, used completely. usage also has extra resource types. However, no remaining")
+		assert.Assert(t, resources.DeepEquals(pRemaining, resources.Add(expectedSmallestRes1, resources.Multiply(childRes, -1))), "guaranteed set, one resource type used completely. usage also has another resource types which is used bit more. remaining should have zero for one resource type and -ve for another")
+		assert.Assert(t, resources.DeepEquals(cRemaining1, resources.Add(expectedSmallestRes1, resources.Multiply(childRes, -1))), "guaranteed set, used double than guaranteed. so remaining should be in -ve")
+		assert.Assert(t, resources.DeepEquals(cRemaining2, tt.childQ2Remaining1), "guaranteed set, used bit lesser. parent's usage also has extra resource types. remaining should be based on ask queue")
+	}
+}
+
+func setup(t *testing.T) (*Queue, *Queue, *Queue, *Queue, *Queue) {
```
craigcondit, line 241: Normally I'm not in favor of named return values, but when there's 5 of the same type... I think it's warranted.

pkg/scheduler/objects/preemption_queue_test.go
```diff
@@ -123,113 +123,185 @@ func TestGetPreemptableResource(t *testing.T) {
...
-	rootQ.maxResource = resources.Multiply(smallestRes, 4)
-	parentQ.maxResource = resources.Multiply(smallestRes, 2)
-	childQ1.maxResource = smallestRes
-	childQ2.maxResource = smallestRes
-	rootRemaining, pRemaining, cRemaining1, cRemaining2 = getRemainingGuaranteed(rootQ, parentQ, childQ1, childQ2)
+	childQ3, err = createManagedQueue(parent1Q, "child3", false, map[string]string{})
+	assert.NilError(t, err)
+	return rootQ, parentQ, childQ1, childQ2, childQ3
+}
+func assertZeroRemaining(t *testing.T, rootRemaining *resources.Resource, pRemaining *resources.Resource, cRemaining1 *resources.Resource, cRemaining2 *resources.Resource) {
```
craigcondit, line 257: Nit: Add a newline here.

craigcondit (changes requested):
Few minor points, but mostly good I think. The clack of `Clone()` in resources.go I think is significant though.

pkg/common/resources/resources.go
craigcondit: This still needs to be updated. This should never update left, and without a clone it will.

craigcondit (changes requested):
We need to clone the result in MergeIfNotPresent().

craigcondit (approved):
+1 LGTM. FYI, I merged master back to fix a one-line merge conflict.

PR #925, opened 2024-07-31
[YUNIKORN-2681] Data race in TestCheckHealthStatusNotFound

pbacsko:
### What is this PR for?
There is a data race in the unit tests due to `TestGetStream_Limit` because it starts handler functions on different goroutines. Since `schedulerContext` is a global variable, it needs to be accessed safely.

**Note**: a better solution would be to elimiate global package variables and handler functions (ie. handlers should be tied to an instance), but that's a much bigger change. This is OK until the refactor is done.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2681

### Questions:

### DIFF
--- a/pkg/webservice/handlers.go
+++ b/pkg/webservice/handlers.go
@@ -127,7 +127,7 @@ func getStackInfo(w http.ResponseWriter, r *http.Request) {
 func getClusterInfo(w http.ResponseWriter, r *http.Request) {
 	writeHeaders(w)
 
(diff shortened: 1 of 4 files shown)


craigcondit (approved):
+1 LGTM.

PR #926, opened 2024-08-01
[YUNIKORN-2528] Increase coverage for UGM code

pbacsko:
### What is this PR for?
There are some code paths in UGM which are not covered by unit tests.  

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2528

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/queue_tracker.go
+++ b/pkg/scheduler/ugm/queue_tracker.go
@@ -403,24 +403,20 @@ func (qt *QueueTracker) canRunApp(hierarchy []string, applicationID string, trac
 // it decides the removal. It returns false the moment it sees any unexpected values for any queue in any levels.
 // Note: Lock free call. The RLock of the linked tracker (UserTracker and GroupTracker) should be held before calling this function.
 func (qt *QueueTracker) canBeRemoved() bool {
-	for _, childQT := range qt.childQueueTrackers {
-		// quick check to avoid further traversal
-		if childQT.canBeRemovedInternal() {
+	if len(qt.childQueueTrackers) > 0 {
+		for _, childQT := range qt.childQueueTrackers {
+			// quick check to avoid further traversal
 			if !childQT.canBeRemoved() {
 				return false
 			}
-		} else {
-			return false
 		}
+		return true // removable because all children are removable
 	}
-	// reached leaf queues, no more to traverse
-	return qt.canBeRemovedInternal()
-}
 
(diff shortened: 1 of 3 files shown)


pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -403,24 +403,20 @@ func (qt *QueueTracker) canRunApp(hierarchy []string, applicationID string, trac
...
-		qt.maxRunningApps == 0 && resources.IsZero(qt.maxResources) {
-		return true
-	}
-	return false
+	// reached a leaf queue
+	return len(qt.runningApplications) == 0 &&
+		resources.IsZero(qt.resourceUsage) &&
+		len(qt.childQueueTrackers) == 0 &&
+		qt.maxRunningApps == 0 &&
+		resources.IsZero(qt.maxResources)
```
pbacsko, line 421: @manirajv06 please check this logic, I think this is simpler than the original and accomplishes the same.

manirajv06: In newer approach, we take decision only after reaching the leaf queue by traversing all the queues in the queue hierarchy. Whereas earlier, earlier one takes decision at very earlier stage and come out then and there.

pbacsko: Even in the old code, `canBeRemovedInternal()` only returns true if `len(qt.childQueueTrackers) == 0`. So you need to descend down the hierarchy. You can't return earlier, therefore the optimization is not really there.

manirajv06: For example,

root.parent running app is 10. root.parent has 100's of child queues (or child queue trackers) below it. In earlier approach, We can come out right away the moment we see 10 apps running in root.parent queue and confirm that it is better not to remove root.parent and any queue below it.

pbacsko: Ok I'm reverting this change, let's not spend too much time on this. Maybe I'll get back to it later.

manirajv06 (approved):
looks good.

Are there any more places to be covered? JIRA description talks about manager methods too.

PR #928, opened 2024-08-02
[YUNIKORN-2777] Improve TrackedResource type

pbacsko:
### What is this PR for?
Refactor `TrackedResource` type so that we can re-use code for `*Resource` if needed.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2777

### Questions:

### DIFF
--- a/pkg/scheduler/partition_test.go
+++ b/pkg/scheduler/partition_test.go
@@ -1924,17 +1924,18 @@ func TestRequiredNodeAllocation(t *testing.T) {
 func assertPreemptedResource(t *testing.T, appSummary *objects.ApplicationSummary, memorySeconds int64,
 	vcoresSecconds int64) {
 	detailedResource := appSummary.PreemptedResource.TrackedResourceMap["UNKNOWN"]
-	memValue, memPresent := detailedResource["memory"]
-	vcoreValue, vcorePresent := detailedResource["vcore"]
+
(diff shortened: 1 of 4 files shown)


pkg/scheduler/partition_test.go
```diff
@@ -2004,7 +2005,7 @@ func TestPreemption(t *testing.T) {
 	assertPreemptedResource(t, appSummary, -1, 5000)
 
 	appSummary = app2.GetApplicationSummary("default")
-	assertPreemptedResource(t, appSummary, -1, 0)
+	assert.Assert(t, appSummary.PreemptedResource.TrackedResourceMap["UNKNOWN"] == nil)
```
pbacsko, line 2008: Note: since it's no longer `map[string]map[string]int64` but `map[string]*Resource`, referencing a non-existing key results in a panic (this does not happen in the former case, even if `map[string]int64` for a given key was not created with `make()`).

zhuqi-lucas: Got it!

zhuqi-lucas (approved):
LGTM

PR #935, opened 2024-08-07
[YUNIKORN-2678] Improve FAIR queue sort algorithm

psantacl:
### What is this PR for?

Changes to the Fair scheduling algorithm to always produce ratio's only when sorting queues for 'fair' usage.

Jira: https://issues.apache.org/jira/browse/YUNIKORN-2678

### What type of PR is it?
* [X] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2678

Please see Jira.


pkg/common/resources/resources.go
```diff
@@ -455,6 +455,74 @@ func (r *Resource) fitIn(smaller *Resource, skipUndef bool) bool {
...
+func getShareFairForDenominator(resourceType string, allocated Quantity, denominatorResources *Resource) (float64, bool) {
+	if denominatorResources == nil {
+		return 0.0, false
+	}
+
+	denominator, ok := denominatorResources.Resources[resourceType]
+
+	switch {
+	case ok && denominator <= 0:
+		if denominator < 0 {
```
craigcondit, line 469: Can we remove the logging? This is hot-path code, even evaluating whether log level is active is pretty expensive. Also, if you have n queues to compare you're going to emit ~ n log(n) debug calls per sort() invocation. Take that times every scheduling cycle, and you're going to emit millions of log entries very quickly if someone is unfortunate enough to turn this on.

pkg/common/resources/resources.go
```diff
@@ -455,6 +455,74 @@ func (r *Resource) fitIn(smaller *Resource, skipUndef bool) bool {
...
+			log.Log(log.Resources).Debug("denominator is negative. will not compute share",
+				zap.String("resource key", resourceType),
+				zap.Int64("resource quantity", int64(denominator)))
+		}
+
+		if allocated <= 0 {
+			//explicit 0 or negative value with NO usage
+			return float64(0.0), true
```
craigcondit, line 477: I don't believe the cast is necessary here.

pkg/common/resources/resources.go
```diff
@@ -455,6 +455,74 @@ func (r *Resource) fitIn(smaller *Resource, skipUndef bool) bool {
...
+
+		} else {
+			//explicit 0 or negative value with usage
+			return float64(1.0), true
```
craigcondit, line 481: Also cast not required.

pkg/common/resources/resources.go
```diff
@@ -455,6 +455,74 @@ func (r *Resource) fitIn(smaller *Resource, skipUndef bool) bool {
...
+		//no denominator. ie. no guarantee or fairmax for resourceType
+		return 0.0, false
+	}
+}
+
+// For a given queue, produce a ratio which represents it's current 'fair' share usage.
+// Iterate over all of the allocated resource types.  For each, compute the ratio, ultimately returning the max ratio encountered.
+// The numerator will be the allocated usage.
+// If guarantees are present, they will be used for the denominator, otherwise we will fallback to the 'maxfair' capacity of the cluster.
+func getFairShare(allocated, guaranteed, fair *Resource) float64 {
```
craigcondit, line 497: Nit: Can we rename 'fair' to 'max' to align with how this is used elsewhere in the code?

pkg/common/resources/resources.go
```diff
@@ -455,6 +455,74 @@ func (r *Resource) fitIn(smaller *Resource, skipUndef bool) bool {
...
+	if allocated == nil || len(allocated.Resources) == 0 {
+		return float64(0)
```
craigcondit, line 499: Unnecessary cast.

pkg/common/resources/resources.go
```diff
@@ -455,6 +455,74 @@ func (r *Resource) fitIn(smaller *Resource, skipUndef bool) bool {
...
+
+		nextShare, found := getShareFairForDenominator(k, v, guaranteed)
+		if !found {
+			nextShare, found = getShareFairForDenominator(k, v, fair)
+		}
+		if found {
+			if nextShare > maxShare {
+				maxShare = nextShare
+			}
+		}
```
craigcondit, line 519: This whole expression can be simplified:

```
if !found {
	nextShare, found = getShareFairForDenominator(k, v, fair)
} else if nextShare > maxShare {
	maxShare = nextShare
}
```

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,63 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+	}
+
+	out := resources.NewResource()
+	for k, v := range parent.Resources {
+		out.Resources[k] = v
+	}
+
+	//child wins every resources collision
+	for k, v := range us.Resources {
+		log.Log(log.SchedQueue).Info("child overwriting parent",
```
craigcondit, line 1271: Again, skip the logging here. It will overwhelm the rest of the system.

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,63 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
 	return sq.internalGetMax(limit)
 }
 
+// Starting with the root, descend down to the target queue allowing children to override Resource values .
+// If the root includes an explicit 0 value for a Resource, do not include it in the accumulator and treat it as missing.
+// If no children provide a maximum capacity override, the resulting value will be the value found on the Root.
```
craigcondit, line 1225: Add the function name as the first word in the description per godoc conventions. I.e. "GetFairMaxResource computes the fair maximum resources ...."

pkg/scheduler/objects/sorters.go
```diff
@@ -65,8 +65,10 @@ func sortQueuesByPriorityAndFairness(queues []*Queue) {
 		if lPriority < rPriority {
 			return false
 		}
-		comp := resources.CompUsageRatioSeparately(l.GetAllocatedResource(), l.GetGuaranteedResource(),
-			r.GetAllocatedResource(), r.GetGuaranteedResource())
+
```
craigcondit, line 68: I think this function needs some work as long as we're rewriting it. We are doing the expensive recursive computation of resources on every comparison. Since we already have a slice of queues, it would be better to precompute a slice of resources computed on each one, and then use that in the comparison function so that we only create the resource maps once per sort run instead of once per comparison.

craigcondit (changes requested):
Some initial revisions to work on.

pkg/common/resources/resources.go
craigcondit: Withdrawing this one after community discussion.

pkg/scheduler/objects/sorters.go
craigcondit: Wilfred also pointed out in the community meeting that we may also be able to optimize the parent hierarchy separately as all the compared queues will have the same parent. If we do that, we do need to be sure that the parent is not the root.

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,63 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+		return us.Clone()
+	}
+	if us == nil {
+		return parent.Clone()
+	}
+
+	out := resources.NewResource()
+	for k, v := range parent.Resources {
+		out.Resources[k] = v
+	}
```
pbacsko, line 1267: Isn't this just simple cloning here?

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,63 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+
+	//child wins every resources collision
+	for k, v := range us.Resources {
+		log.Log(log.SchedQueue).Info("child overwriting parent",
+			zap.Any("k", k),
+			zap.Any("v", v))
+
+		out.Resources[k] = v
+	}
```
pbacsko, line 1276: I think this entire chain can written in a more Go-like fashion:

```
var out *resources.Resources
switch {
  case parent == nil && us == nil:
    // nop
  case parent == nil:
    out = us.Clone()
  case us == nil:
    out = parent.Clone()
  default:
    out = parent.Clone()  // see my question above
    for k, v := range us.Resources {
	  out.Resources[k] = v  // overwrite parent values with child values
    }
}

return out
```

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,63 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
 }
 
+// Starting with the root, descend down to the target queue allowing children to override Resource values .
+// If the root includes an explicit 0 value for a Resource, do not include it in the accumulator and treat it as missing.
+// If no children provide a maximum capacity override, the resulting value will be the value found on the Root.
+// It is useful for fair-scheduling to allow a ratio to be produced representing the rough utilization % of a given queue.
+func (sq *Queue) GetFairMaxResource() *resources.Resource {
+	var limit *resources.Resource
+	if sq.parent == nil {
+		cleanedRootResources := resources.NewResource()
```
pbacsko, line 1230: Nit: no need to use long variable names in tiny scopes. Just call it "ret" or "out" or even "cleaned".

pkg/common/resources/resources.go
pbacsko: +1, we need no logging here, even if it's disabled

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,63 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+		if sq.maxResource == nil {
+			return cleanedRootResources
```
pbacsko, line 1232: Please check this: if you always return an empty resource, then the `if parent == nil` will never be false in `internalGetFairMaxResource()` because it will propagate towards the lower part of the hierarchy. You can verify this with the code coverage as well - certain branches are never taken.

pkg/common/resources/resources.go
psantacl: done

pkg/common/resources/resources.go
psantacl: done

pkg/common/resources/resources.go
psantacl: done

pkg/common/resources/resources.go
psantacl: done

pkg/common/resources/resources.go
psantacl: This, I believe, is necessary.  First we check to see if we can compute a share with a guarantee.  Failing that we attempt to compute a share with FairMax resources as the denominator.  If at least one of these succeed(found == true), then we check the computed share to see if it is the new maximum.

pkg/scheduler/objects/queue.go
psantacl: done

pkg/common/resources/resources.go
craigcondit: Makes sense. I missed the reassignment to `found`.

pkg/scheduler/objects/queue.go
craigcondit: Yeah, that should probably be `out := parent.Clone()`

craigcondit:
@psantacl please run `make lint` before pushing PRs. If the linter doesn't pass, the PR is rejected before the remainder of the checks run.

pkg/scheduler/objects/queue.go
psantacl: done

pkg/scheduler/objects/queue.go
psantacl: done

pkg/scheduler/objects/queue.go
psantacl: done

pkg/scheduler/objects/queue.go
psantacl: I think your logic is sound and code coverage supports it.  i will make the change.

psantacl: done

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,50 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+		cleaned := resources.NewResource()
+		if sq.maxResource == nil {
+			return cleaned
+		}
+
+		for k, v := range sq.maxResource.Resources {
+			if v != 0 {
+				cleaned.Resources[k] = v
+			}
+		}
```
pbacsko, line 1240: If we wait a little bit for #943, then we'll be able to call `cleaned.Prune()`.

IMO this is the most readable approach:
```
if sq.maxResource == nil {
  return resources.NewResource()
}

return sq.maxResources.Clone().Prune()
```

Or perhaps we shouldn't even allow "0" resources in max in the first place and we just clone it...

pbacsko (commented):
I just have a final comment, I think it's worth waiting for #943.

pbacsko (changes requested):
@psantacl please resolve merge conflicts

pkg/scheduler/objects/queue.go
craigcondit: 0 resources need to be allowed as they may be explicitly set on queues. Only on the root queue do we ignore zero.

pkg/scheduler/objects/sorters.go
psantacl: Currently on vacation, but I will add this optimization when I return.

pkg/scheduler/objects/queue.go
pbacsko: Thx, good to know.

pkg/common/resources/resources.go
```diff
@@ -469,6 +469,64 @@ func (r *Resource) fitIn(smaller *Resource, skipUndef bool) bool {
...
+
+		nextShare, found := getShareFairForDenominator(k, v, guaranteed)
+		if !found {
+			nextShare, found = getShareFairForDenominator(k, v, fair)
+		}
+		if found {
+			if nextShare > maxShare {
+				maxShare = nextShare
+			}
```
pbacsko, line 524: Nit: this can be written as

```
if found && nextShare > maxShare {
  maxShare = nextShare
}
```

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,50 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+}
+
+func (sq *Queue) internalGetFairMaxResource(parent *resources.Resource) *resources.Resource {
+	sq.RLock()
+	defer sq.RUnlock()
+
+	us := sq.maxResource
+	if us == nil {
+		return parent.Clone()
+	}
```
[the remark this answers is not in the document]
psantacl: done.

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,50 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+
+	out := parent.Clone()
+
+	// perform merge. child wins every resources collision
+	for k, v := range us.Resources {
+		out.Resources[k] = v
+	}
```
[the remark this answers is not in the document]
psantacl: gotcha.  changes made.

pkg/scheduler/objects/queue_test.go
```diff
@@ -869,6 +869,114 @@ func TestMaxHeadroomMax(t *testing.T) {
 	assert.Assert(t, resources.Equals(res, headRoom), "leaf2 queue head room not as expected %v, got: %v", res, headRoom)
 }
 
+func TestGetFairMaxResource(t *testing.T) {
+	tests := []struct {
```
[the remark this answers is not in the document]
psantacl: done

pkg/scheduler/objects/queue_test.go
```diff
@@ -869,6 +869,114 @@ func TestMaxHeadroomMax(t *testing.T) {
...
+		RootResource     map[string]string
+		ParentResource   map[string]string
+		Tier0Resource    map[string]string
+		Tier0Expectation map[string]string
+		Tier1Resource    map[string]string
+		Tier1Expectation map[string]string
+	}{
```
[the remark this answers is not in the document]
psantacl: tests have been added.

pkg/scheduler/objects/queue.go
```diff
@@ -1258,6 +1258,48 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+		cleaned := resources.NewResource()
+		if sq.maxResource == nil {
+			return cleaned
+		}
+
+		for k, v := range sq.maxResource.Resources {
+			if v != 0 {
+				cleaned.Resources[k] = v
+			}
+		}
```
craigcondit, line 1278: This block can be replaced by:

```
return sq.GetMaxResource().Clone()
```

As root queue resources are now always in a pruned state.

pkg/scheduler/objects/queue.go
```diff
@@ -1258,6 +1258,48 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+	limit = sq.parent.GetFairMaxResource()
+	return sq.internalGetFairMaxResource(limit)
+}
+
+func (sq *Queue) internalGetFairMaxResource(limit *resources.Resource) *resources.Resource {
+	sq.RLock()
+	defer sq.RUnlock()
+
+	if sq.maxResource.IsEmpty() || limit.IsEmpty() {
+		return limit
```
craigcondit, line 1292: This should be `return limit.Clone()`. `Resource` objects are not immutable, so it's better to clone since we're returning one.

pkg/scheduler/objects/queue.go
```diff
@@ -1258,6 +1258,48 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+	}
+
+	// perform merge. child wins every resources collision
+	for k, v := range sq.maxResource.Resources {
+		limit.Resources[k] = v
+	}
+
+	return limit
```
craigcondit, line 1300: Here, do your clone of `limit` and only update the clone, otherwise you're mutating an object which was passed into the function.

craigcondit (changes requested):
Some clone-safety and code tweaks. Also the big optimization work yet to-do.

pkg/scheduler/objects/queue.go
psantacl: done

pkg/scheduler/objects/queue.go
```diff
@@ -1220,6 +1220,50 @@ func (sq *Queue) GetMaxResource() *resources.Resource {
...
+			return cleaned
+		}
+
+		for k, v := range sq.maxResource.Resources {
+			if v != 0 {
+				cleaned.Resources[k] = v
+			}
+		}
+
+		return cleaned
```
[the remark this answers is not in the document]
psantacl: done

pkg/scheduler/objects/queue.go
psantacl: done

pkg/scheduler/objects/queue.go
psantacl: done

pkg/scheduler/objects/queue.go
psantacl: done

pkg/scheduler/objects/queue.go
craigcondit: Also, to address Wilfred's comment earlier about using limit directly rather than cloning -- yes, that's safe *today* but the fact we're passing in limit to this function means a future caller might be rather surprised to find that limit is both mutated and returned from the function. I'd rather sacrifice a tiny bit of performance than risk unexpected updates.

craigcondit (changes requested):
+1, other than the big refactor.

pbacsko (approved):
+1

craigcondit:
@wilfred-s thoughts on committing this version and working on the performance refactor in a followup?

pkg/scheduler/objects/sorters.go
psantacl: pushed.  This should compute GetFairMaxResource O(n) times instead of NLog(N).

craigcondit: This is better, but to Wilfred's point from the community meeting, check out the callers of `queue.sortQueues()` -- the are in queue.go under `GetQueueOutstandingRequests()`, `TryAllocate()`, `TryPlaceholderAllocate()` and `TryReservedAllocate()`. Each computes `queue.getHeadRoom()` at the start which is used for further processing. Since these methods are all recursive from the root queue down, you can compute the parent values at each step and then pass the results into the `sortQueues()` function. This avoids having to recompute all the parent queue values during sorting.

craigcondit: I'm okay with deferring that refactor until a follow-up PR. @wilfred-s what do you think?

wilfred-s: We can defer that to the next release. It is not just a question of performance but also a question about correctness. Since the queue values can change it might be the case that a parents' setting changes. If that happens between the child checks, specially when you have a larger number of children, you might not have the exact same inputs for each child computation.
The chance is low but we should try an prevent that.
The change would not be in the sorter but in the queue code when we build the `fairMaxResources` slice

wilfred-s (approved):
LGTM
Follow up for the performance and correctness change YUNIKORN-2840

craigcondit (approved):
+1 LGTM.

PR #936, opened 2024-08-08
[YUNIKORN-2789] Queue internalGetMax should use permissive calculator

wilfred-s:
In the implementation of internalGetMax on the queue, we call resources ComponentWiseMin(). This returns 0 values for each type that is not defined in both of the resources passed in.
In case there is no overlap it will return a resource with all types set to "zero".
That is contrary to the documentation on how quotas work and what we do everywhere besides the internalGetMax implementation.

The ComponentWiseMin function is only used in queue internalGetMax. Refactor to make the ComponentWiseMinPermissive version the standard (rename and remove "permissive"), and remove the existing ComponentWiseMin code.

### What is this PR for?

### What type of PR is it?
* [X] - Bug Fix
* [X] - Refactor

### What is the Jira issue?
* [YUNIKORN-2789](https://issues.apache.org/jira/browse/YUNIKORN-2789)

Unit test cover the changes no functional change

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -600,7 +600,7 @@ func (sq *Queue) GetGuaranteedResource() *resources.Resource {
 	return sq.guaranteedResource
 }
 
-// GetActualGuaranteedResources returns the actual (including parent) guaranteed resources for the queue.
+// GetActualGuaranteedResource returns the actual (including parent) guaranteed resources for the queue.
 func (sq *Queue) GetActualGuaranteedResource() *resources.Resource {
 	if sq == nil {
 		return resources.NewResource()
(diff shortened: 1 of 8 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -1242,9 +1240,6 @@ func (sq *Queue) internalGetMax(parentLimit *resources.Resource) *resources.Reso
 	defer sq.RUnlock()
 	// no parent queue limit set, not even for root
 	if parentLimit == nil {
-		if sq.maxResource == nil {
-			return nil
-		}
```
chenyulin0719, line 1247: I think we should keep the nil check to prevent panic error, even though it's not possible in the current production code. (root queue always have maxResource) 

Additionally, it seems that the nil checks for`parentLimit` and `sq.maxResource` in `internalGetMax()` are redundant, the resources.ComponentWiseMin(parentLimit, sq.maxResource) will do the same check.

https://github.com/wilfred-s/yunikorn-core/blob/a8537787fd20c65c56357d15a7219879d586c21c/pkg/scheduler/objects/queue.go#L1241-L1248

craigcondit: No need for the check, as the Clone() method is nil-safe.

PR #943, opened 2024-08-13
[YUNIKORN-2796] Prune zero resource type from root and partition

wilfred-s:
### What is this PR for?
After the last node is removed that registers a specific resource type the root queue and partition show it as a resource type with 0 maximum. A resource type that does not exist should not be shown at all. Pruning the 0 valued resource types for root and partition will fix that

Clean up of the partition code to use the same function to update partition and root settings for create, update and delete of a node.

Fix root queue check for max resource.
Fix tests that fake allocation or node registration to correctly set the root queue limits.

### What type of PR is it?
* [X] - Improvement
* [X] - Refactoring

### What is the Jira issue?
* [YUNIKORN-2796](https://issues.apache.org/jira/browse/YUNIKORN-2796)

Unit tests cover the internal checks
e2e tests cover the changes when included in the overall build


pbacsko (commented):
A question: don't we need to call `root.Prune()` somewhere?

wilfred-s:
> A question: don't we need to call `root.Prune()` somewhere?

The prune is called on the partition setting in updatePartitionResource before we set the same on the root. So it is called indirectly.

pbacsko (approved):
LGTM

pkg/scheduler/objects/queue_test.go
```diff
@@ -1052,60 +1052,58 @@ func TestIsEmpty(t *testing.T) {
 
 func TestGetOutstandingRequestMax(t *testing.T) {
 	// queue structure:
-	// root
+	// root (max.cpu = 5)
```
manirajv06, line 1055: This is bit confusing especially with "root max is irrelevant for the calculation" comment. My understanding is, root max res has been set based on the actual usage just to satisfy root Q fitIn operation.

pkg/scheduler/objects/queue_test.go
```diff
@@ -1176,10 +1173,11 @@ func TestGetOutstandingOnlyUntracked(t *testing.T) {
 	alloc, err := resources.NewResourceFromConf(map[string]string{"pods": "1"})
 	assert.NilError(t, err, "failed to create basic resource")
 	var used *resources.Resource
-	used, err = resources.NewResourceFromConf(map[string]string{"cpu": "10", "pods": "10"})
+	usedMap := map[string]string{"cpu": "10", "pods": "10"}
+	used, err = resources.NewResourceFromConf(usedMap)
 	assert.NilError(t, err, "failed to create basic resource")
 	var root, queue1 *Queue
-	root, err = createRootQueue(nil)
+	root, err = createRootQueue(usedMap)
```
manirajv06, line 1180: Can we edit root max cpu in Queue structure like earlier?

pkg/scheduler/objects/queue.go
```diff
@@ -1028,6 +1028,12 @@ func (sq *Queue) IncAllocatedResource(alloc *resources.Resource, nodeReported bo
 func (sq *Queue) allocatedResFits(alloc *resources.Resource) bool {
 	sq.RLock()
 	defer sq.RUnlock()
+	// on the root we want to reject a new allocation if it asks for resources not registered
+	// so do not use the "undefined" flag, also handles pruned max for root
+	if sq.isRoot() {
+		return sq.maxResource.FitIn(resources.AddOnlyExisting(alloc, sq.allocatedResource))
+	}
+	// any other queue undefined is always good
```
manirajv06, line 1036: Can we have a single line check for all queues? Is Internal FitIn method private? Can it be made public?

manirajv06 (changes requested):
looks good. Few minor nits.

pkg/scheduler/objects/queue.go
craigcondit: Root queue often needs special casing, as we have here. I'm fine with the code as-is.

craigcondit (approved):
+1 LGTM.

pkg/common/resources/resources.go
```diff
@@ -144,7 +144,20 @@ func (r *Resource) Clone() *Resource {
 	return ret
 }
 
-// Add additional resource to the base updating the base resource
+// Prune removes any resource type that has a zero value set.
+// Note that a zero value set and a type not being set are interpreted differently for quotas.
+func (r *Resource) Prune() {
```
zhuqi-lucas, line 149: One question about the Prune, do we want to support other resources prune, for example:

When the maxapps is zero, do we want to prune it?

wilfred-s: I have thought about the allocated resources for the queues those could be pruned to show the type is no longer there. That would clean up the UI etc nicely. Follow up jira?

maxapps is not pruneable as the "nil" value for an integer type is 0 so we do not gain anything with that.

pkg/scheduler/objects/queue_test.go
wilfred-s: not really confusing: it is important to know what the root max is. What can be accommodated by the root max is much lower than the outstanding requests list returns clearly showing that it is irrelevant. That is also why I did not set some strange high value but linked it to the fake allocation size.

pkg/scheduler/objects/queue_test.go
wilfred-s: We did not edit it. It was left as nil and we did not register nodes. The root had a max of nil. Because we did not do the hard fit check these test cases all passed. The root queue was treated as any other queue: no max FitIn would always pass.

We do the correct thing now. tests should either register nodes or create a root with a max set to fake the node registrations.

pkg/common/resources/resources.go
zhuqi-lucas: Got it @wilfred-s , thanks the clarify, i create a jira to follow up:
https://issues.apache.org/jira/browse/YUNIKORN-2802

zhuqi-lucas (approved):
LGTM

pkg/scheduler/objects/queue_test.go
wilfred-s: oh I understand now what you mean we need to update line 1166 with the root size like we did in the other tests. Will add that

manirajv06 (approved):

PR #945, opened 2024-08-14
[YUNIKORN-2803] Use FitIn for node check

manirajv06:
### What is this PR for?
Use FitIn instead of FitInMaxUndef to know whether ask fits in node or not.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2803

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -213,7 +213,6 @@ func (p *Preemptor) checkPreemptionQueueGuarantees() bool {
 //nolint:funlen
 func (p *Preemptor) calculateVictimsByNode(nodeAvailable *resources.Resource, potentialVictims []*Allocation) (int, []*Allocation) {
 	nodeCurrentAvailable := nodeAvailable.Clone()
-	allocationsByQueueSnap := p.duplicateQueueSnapshots()
 
 	// Initial check: Will allocation fit on node without preemption? This is possible if preemption was triggered due
 	// to queue limits and not node resource limits.
@@ -222,6 +221,7 @@ func (p *Preemptor) calculateVictimsByNode(nodeAvailable *resources.Resource, po
 		return -1, make([]*Allocation, 0)
 	}
 
+	allocationsByQueueSnap := p.duplicateQueueSnapshots()
 	// get the current queue snapshot
 	askQueue, ok := allocationsByQueueSnap[p.queuePath]
 	if !ok {
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -326,7 +326,7 @@ func (p *Preemptor) calculateVictimsByNode(nodeAvailable *resources.Resource, po
 					// removing task does not violate queue constraints, adjust queue and node
 					nodeCurrentAvailable.AddTo(victim.GetAllocatedResource())
 					// check if ask now fits and we haven't had this happen before
-					if nodeCurrentAvailable.FitInMaxUndef(p.ask.GetAllocatedResource()) && index < 0 {
+					if nodeCurrentAvailable.FitIn(p.ask.GetAllocatedResource()) && index < 0 {
 						index = len(results)
 					}
 					// add victim to results


pkg/scheduler/objects/preemption.go
```diff
@@ -326,7 +326,7 @@ func (p *Preemptor) calculateVictimsByNode(nodeAvailable *resources.Resource, po
 					// removing task does not violate queue constraints, adjust queue and node
 					nodeCurrentAvailable.AddTo(victim.GetAllocatedResource())
 					// check if ask now fits and we haven't had this happen before
-					if nodeCurrentAvailable.FitInMaxUndef(p.ask.GetAllocatedResource()) && index < 0 {
+					if nodeCurrentAvailable.FitIn(p.ask.GetAllocatedResource()) && index < 0 {
 						index = len(results)
 					}
```
pbacsko, line 331: Can we add a test case for this? I'm worried that such small things can change and we have no corresponding test which tells us if there's something wrong.

wilfred-s:
I am wondering why we never filter the nodes to start with: if the allocation asks for a resource type not registered by the node. Or if the node is too small for the allocation. We should remove the node if the allocation does not fit the capacity to start with. 
We do some filtering in the `initWorkingState()` call but we do not check that. If we have that check there will be no impact of the change from `FitiIMaxUndef()` to `FitIn()`.

BTW: we have an optimisation around the start of `calculateVictimsByNode()` we need to make. Creating the copy of the queue snapshot (Line 216) should not happen until after we checked the node should be looked at (Line 223) Currently we duplicate objects and then check. Which drops all the work and just costs CPU, memory and GC if we do not consider the node.

manirajv06:
> I am wondering why we never filter the nodes to start with: if the allocation asks for a resource type not registered by the node. Or if the node is too small for the allocation. We should remove the node if the allocation does not fit the capacity to start with. We do some filtering in the `initWorkingState()` call but we do not check that. If we have that check there will be no impact of the change from `FitiIMaxUndef()` to `FitIn()`.
> 

Reason for not doing the check in `initWorkingState` method is to consider that node for further steps and to avoid elimination then and there. Though fitIn check fails bit later in `calculateVictimsByNode()`, returning empty allocations list for that specific node makes it to pass through further steps. One important step among them is to running normal predicate checks to ensure node doesn't have any predicate failures. So, moving the check bit higher up in the hierarchy would eliminate the node which is something we don't want to do.

BTW, only allocations (victims) sorting between those two methods, which we won't happen anyway for this node as there are no allocations (victims). Hence, there is not much difference form performance perspective.

Since FitIn() check happens in `calculateVictimsByNode()` happens in the beginning stage as discussed above, doing the same again should not have any impact.

> BTW: we have an optimisation around the start of `calculateVictimsByNode()` we need to make. Creating the copy of the queue snapshot (Line 216) should not happen until after we checked the node should be looked at (Line 223) Currently we duplicate objects and then check. Which drops all the work and just costs CPU, memory and GC if we do not consider the node.

Taken care.

wilfred-s:
> Reason for not doing the check in initWorkingState method is to consider that node for further steps and to avoid elimination then and there. Though fitIn check fails bit later in calculateVictimsByNode(), returning empty allocations list for that specific node makes it to pass through further steps. One important step among them is to running normal predicate checks to ensure node doesn't have any predicate failures. So, moving the check bit higher up in the hierarchy would eliminate the node which is something we don't want to do.

The difference in the check that I am looking at adding is that I want to compare the size of the `ask.GetAllocatedResource()` against the capacity of the node. Not the available size. If the allocation is larger than the node has registered we can never place the allocation on that node. We might as well skip that node completely.
```
ask = {memory: 192G, vcore: 4000}
node capacity = {memory: 128G, vcore: 32000}
```
Even if I kill all allocations on that node 256G will not fit on a 128G node. That means in the end `calculateVictimsByNode()` returns a nil list as the last check to see if the allocation fits will never trigger (part of walking over the victim list and increasing available resources on the node. The `FitInMaxUndef` we are now changing..
Same if I have resources like GPU etc. We should not look a the node if the ask needs GPU and the node does not have.

pkg/scheduler/objects/preemption.go
wilfred-s: I am happy to do this without a test. We looked at adding a test but we need to rip apart processing too much to be able to check the return value of the method we are changing. The result of `calculateVictimsByNode` gets hidden too much to be easily accessible. The setup required for testing the function directly is not straight forward and fragile.

wilfred-s (approved):
LGTM, please file a follow up jira for the node filtering in the `initWorkingState` call

manirajv06:
> 
> The difference in the check that I am looking at adding is that I want to compare the size of the `ask.GetAllocatedResource()` against the capacity of the node. Not the available size. If the allocation is larger than the node has registered we can never place the allocation on that node. We might as well skip that node completely.
> 
> ```
> ask = {memory: 192G, vcore: 4000}
> node capacity = {memory: 128G, vcore: 32000}
> ```
> 
> Even if I kill all allocations on that node 256G will not fit on a 128G node. That means in the end `calculateVictimsByNode()` returns a nil list as the last check to see if the allocation fits will never trigger (part of walking over the victim list and increasing available resources on the node. The `FitInMaxUndef` we are now changing.. Same if I have resources like GPU etc. We should not look a the node if the ask needs GPU and the node does not have.

Filed a new [jira](https://issues.apache.org/jira/browse/YUNIKORN-2820) to track the same.

PR #950, opened 2024-08-19
[YUNIKORN-2802] Consider more resources and types which can be pruned when it is zero

zhuqi-lucas:
### What is this PR for?
We rethink to support more resources and types which can be pruned, see details:
https://github.com/apache/yunikorn-core/pull/943#discussion_r1716250452

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -581,6 +581,7 @@ func (sa *Application) removeAsksInternal(allocKey string, detail si.EventRecord
 			if !ask.IsAllocated() {
 				deltaPendingResource = ask.GetAllocatedResource()
 				sa.pending = resources.Sub(sa.pending, deltaPendingResource)
+				sa.pending.Prune()
 			}
 			delete(sa.requests, allocKey)
 			sa.sortedRequests.remove(ask)
@@ -648,6 +649,7 @@ func (sa *Application) AddAllocationAsk(ask *Allocation) error {
 	// Update total pending resource
 	delta.SubFrom(oldAskResource)
 	sa.pending = resources.Add(sa.pending, delta)
+	sa.pending.Prune()
 	sa.queue.incPendingResource(delta)
 
 	log.Log(log.SchedApplication).Info("ask added successfully to application",
@@ -728,6 +730,7 @@ func (sa *Application) allocateAsk(ask *Allocation) (*resources.Resource, error)
 
 	delta := resources.Multiply(ask.GetAllocatedResource(), -1)
 	sa.pending = resources.Add(sa.pending, delta)
+	sa.pending.Prune()
 	// update the pending of the queue with the same delta
 	sa.queue.incPendingResource(delta)
 
@@ -748,6 +751,7 @@ func (sa *Application) deallocateAsk(ask *Allocation) (*resources.Resource, erro
 
 	delta := ask.GetAllocatedResource()
 	sa.pending = resources.Add(sa.pending, delta)
+	sa.pending.Prune()
 	// update the pending of the queue with the same delta
 	sa.queue.incPendingResource(delta)
 
@@ -1660,6 +1664,8 @@ func (sa *Application) addAllocationInternal(allocType AllocationResultType, all
 		sa.incUserResourceUsage(alloc.GetAllocatedResource())
 		sa.allocatedPlaceholder = resources.Add(sa.allocatedPlaceholder, alloc.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedPlaceholder, sa.maxAllocatedResource)
+		sa.allocatedPlaceholder.Prune()
+		sa.maxAllocatedResource.Prune()
 
 		// If there are no more placeholder to allocate we should move state
 		if resources.Equals(sa.allocatedPlaceholder, sa.placeholderAsk) {
@@ -1686,6 +1692,8 @@ func (sa *Application) addAllocationInternal(allocType AllocationResultType, all
 		sa.incUserResourceUsage(alloc.GetAllocatedResource())
 		sa.allocatedResource = resources.Add(sa.allocatedResource, alloc.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedResource, sa.maxAllocatedResource)
+		sa.allocatedResource.Prune()
+		sa.maxAllocatedResource.Prune()
 	}
 	sa.appEvents.SendNewAllocationEvent(sa.ApplicationID, alloc.allocationKey, alloc.allocatedResource)
 	sa.allocations[alloc.GetAllocationKey()] = alloc
@@ -1797,6 +1805,7 @@ func (sa *Application) removeAllocationInternal(allocationKey string, releaseTyp
 		// as and when every ph gets removed (for replacement), resource usage would be reduced.
 		// When real allocation happens as part of replacement, usage would be increased again with real alloc resource
 		sa.allocatedPlaceholder = resources.Sub(sa.allocatedPlaceholder, alloc.GetAllocatedResource())
+		sa.allocatedPlaceholder.Prune()
 
 		// if all the placeholders are replaced, clear the placeholder timer
 		if resources.IsZero(sa.allocatedPlaceholder) {
@@ -1821,6 +1830,7 @@ func (sa *Application) removeAllocationInternal(allocationKey string, releaseTyp
 		sa.decUserResourceUsage(alloc.GetAllocatedResource(), removeApp)
 	} else {
 		sa.allocatedResource = resources.Sub(sa.allocatedResource, alloc.GetAllocatedResource())
+		sa.allocatedResource.Prune()
 
 		// Aggregate the resources used by this alloc to the application's resource tracker
 		sa.trackCompletedResource(alloc)
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -728,11 +728,11 @@ func (sa *Application) allocateAsk(ask *Allocation) (*resources.Resource, error)
 		sa.updateAskMaxPriority()
 	}
 
-	delta := resources.Multiply(ask.GetAllocatedResource(), -1)
-	sa.pending = resources.Add(sa.pending, delta)
+	delta := ask.GetAllocatedResource()
+	sa.pending = resources.Sub(sa.pending, delta)
 	sa.pending.Prune()
 	// update the pending of the queue with the same delta
-	sa.queue.incPendingResource(delta)
+	sa.queue.decPendingResource(delta)
 
 	return delta, nil
 }
@@ -751,7 +751,6 @@ func (sa *Application) deallocateAsk(ask *Allocation) (*resources.Resource, erro
 
 	delta := ask.GetAllocatedResource()
 	sa.pending = resources.Add(sa.pending, delta)
-	sa.pending.Prune()
 	// update the pending of the queue with the same delta
 	sa.queue.incPendingResource(delta)
 
@@ -1664,8 +1663,6 @@ func (sa *Application) addAllocationInternal(allocType AllocationResultType, all
 		sa.incUserResourceUsage(alloc.GetAllocatedResource())
 		sa.allocatedPlaceholder = resources.Add(sa.allocatedPlaceholder, alloc.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedPlaceholder, sa.maxAllocatedResource)
-		sa.allocatedPlaceholder.Prune()
-		sa.maxAllocatedResource.Prune()
 
 		// If there are no more placeholder to allocate we should move state
 		if resources.Equals(sa.allocatedPlaceholder, sa.placeholderAsk) {
@@ -1692,8 +1689,6 @@ func (sa *Application) addAllocationInternal(allocType AllocationResultType, all
 		sa.incUserResourceUsage(alloc.GetAllocatedResource())
 		sa.allocatedResource = resources.Add(sa.allocatedResource, alloc.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedResource, sa.maxAllocatedResource)
-		sa.allocatedResource.Prune()
-		sa.maxAllocatedResource.Prune()
 	}
 	sa.appEvents.SendNewAllocationEvent(sa.ApplicationID, alloc.allocationKey, alloc.allocatedResource)
 	sa.allocations[alloc.GetAllocationKey()] = alloc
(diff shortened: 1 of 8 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -1660,6 +1664,8 @@ func (sa *Application) addAllocationInternal(allocType AllocationResultType, all
 		sa.incUserResourceUsage(alloc.GetAllocatedResource())
 		sa.allocatedPlaceholder = resources.Add(sa.allocatedPlaceholder, alloc.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedPlaceholder, sa.maxAllocatedResource)
+		sa.allocatedPlaceholder.Prune()
+		sa.maxAllocatedResource.Prune()
```
wilfred-s, line 1668: i would be really surprised if that would ever have a zero in any type. We can only add a type to this aggregate if it is in the request of an allocation as a non zero value. If we prune `allocatedPlaceholder` before we pass it into the calculation we only need one prune call.
move `allocatedPlaceholder.Prune()` upone line and remove `maxAllocatedResource.Prune()`

pkg/scheduler/objects/application.go
```diff
@@ -1686,6 +1692,8 @@ func (sa *Application) addAllocationInternal(allocType AllocationResultType, all
 		sa.incUserResourceUsage(alloc.GetAllocatedResource())
 		sa.allocatedResource = resources.Add(sa.allocatedResource, alloc.GetAllocatedResource())
 		sa.maxAllocatedResource = resources.ComponentWiseMax(sa.allocatedResource, sa.maxAllocatedResource)
+		sa.allocatedResource.Prune()
+		sa.maxAllocatedResource.Prune()
```
wilfred-s, line 1696: Same as above

pkg/scheduler/objects/node.go
```diff
@@ -312,6 +313,8 @@ func (sn *Node) RemoveAllocation(allocationKey string) *Allocation {
 		delete(sn.allocations, allocationKey)
 		sn.allocatedResource.SubFrom(alloc.GetAllocatedResource())
 		sn.availableResource.AddTo(alloc.GetAllocatedResource())
+		sn.allocatedResource.Prune()
+		sn.availableResource.Prune()
```
wilfred-s, line 317: The alloc should not have zeros in it. The only prune that is needed is when we do the `SubFrom`

pkg/scheduler/objects/node.go
```diff
@@ -353,6 +356,8 @@ func (sn *Node) addAllocationInternal(alloc *Allocation, force bool) bool {
 		sn.allocations[alloc.GetAllocationKey()] = alloc
 		sn.allocatedResource.AddTo(res)
 		sn.availableResource.SubFrom(res)
+		sn.allocatedResource.Prune()
+		sn.availableResource.Prune()
```
wilfred-s, line 360: The alloc should not have zeros in it. The only prune that is needed is when we do the `SubFrom`

pkg/scheduler/objects/node.go
```diff
@@ -377,6 +382,8 @@ func (sn *Node) ReplaceAllocation(allocationKey string, replace *Allocation, del
 	// The allocatedResource and availableResource should be updated in the same way
 	sn.allocatedResource.AddTo(delta)
 	sn.availableResource.SubFrom(delta)
+	sn.allocatedResource.Prune()
+	sn.availableResource.Prune()
```
wilfred-s, line 386: The alloc should not have zeros in it. The only prune that is needed is when we do the `SubFrom`

pkg/scheduler/objects/application.go
wilfred-s: On top of that alloc should never have zero values in it so not sure we need to prune at all here.

pkg/scheduler/objects/application.go
```diff
@@ -728,6 +730,7 @@ func (sa *Application) allocateAsk(ask *Allocation) (*resources.Resource, error)
 
 	delta := resources.Multiply(ask.GetAllocatedResource(), -1)
 	sa.pending = resources.Add(sa.pending, delta)
+	sa.pending.Prune()
 	// update the pending of the queue with the same delta
 	sa.queue.incPendingResource(delta)
```
wilfred-s, line 735: This construct looks weird: multiply by -1 and then call Add. We should just call `Sub` and `decPendingResource`. 
The delta that gets returned is also dropped by all callers or expected to be nil. Except for 1 tests which does not really need it. Not sure we should keep it.

pkg/scheduler/objects/application.go
```diff
@@ -748,6 +751,7 @@ func (sa *Application) deallocateAsk(ask *Allocation) (*resources.Resource, erro
 
 	delta := ask.GetAllocatedResource()
 	sa.pending = resources.Add(sa.pending, delta)
+	sa.pending.Prune()
```
wilfred-s, line 754: ask or alloc should not have zero types, no need to prune
The delta that gets returned is also dropped by all callers or expected to be nil. Except for 1 tests which does not really need it. Not sure we should keep it.

pkg/scheduler/objects/queue.go
```diff
@@ -1609,20 +1615,36 @@ func (sq *Queue) updateMaxResourceMetrics() {
 
 // updateAllocatedResourceMetrics updates allocated resource metrics for all queue types.
 func (sq *Queue) updateAllocatedResourceMetrics() {
+	// Special case when we prune the allocated resources to zero, we need to update the metrics
+	// here only keep basic types memory and vcores to make sure we have updated metrics
+	if sq.allocatedResource == nil || len(sq.allocatedResource.Resources) == 0 {
```
wilfred-s, line 1620: use `sq.allocatedResource.IsZero()`

pkg/scheduler/objects/queue.go
```diff
@@ -1609,20 +1615,36 @@ func (sq *Queue) updateMaxResourceMetrics() {
...
 	for k, v := range sq.allocatedResource.Resources {
 		metrics.GetQueueMetrics(sq.QueuePath).SetQueueAllocatedResourceMetrics(k, float64(v))
 	}
 }
 
 // updatePendingResourceMetrics updates pending resource metrics for all queue types.
 func (sq *Queue) updatePendingResourceMetrics() {
+	// Special case when we prune the allocated resources to zero, we need to update the metrics
+	// here only keep basic types memory and vcores to make sure we have updated metrics
+	if sq.pending == nil || len(sq.pending.Resources) == 0 {
```
wilfred-s, line 1633: use `sq.pending.IsZero()`

pkg/scheduler/objects/queue.go
```diff
@@ -1609,20 +1615,36 @@ func (sq *Queue) updateMaxResourceMetrics() {
...
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueuePendingResourceMetrics("vcores", float64(0))
+	}
 	for k, v := range sq.pending.Resources {
 		metrics.GetQueueMetrics(sq.QueuePath).SetQueuePendingResourceMetrics(k, float64(v))
 	}
 }
 
 // updatePreemptingResourceMetrics updates preempting resource metrics for all queue types.
 func (sq *Queue) updatePreemptingResourceMetrics() {
+	if sq.pending == nil || len(sq.pending.Resources) == 0 {
```
wilfred-s, line 1644: use `sq.pending.IsZero()`

pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -113,6 +113,7 @@ func (qt *QueueTracker) increaseTrackedResource(hierarchy []string, applicationI
 		qt.resourceUsage = resources.NewResource()
 	}
 	qt.resourceUsage.AddTo(usage)
+	qt.resourceUsage.Prune()
```
wilfred-s, line 116: The alloc should not have zeros in it. The only prune that is needed is when we do the `Sub` or `SubFrom`

wilfred-s (changes requested):
In general: we should only need prune if we do a `Sub` or `SubFrom` unless we know that we call the `Add` or `AddTo` function with a negative value.

pkg/scheduler/objects/application.go
zhuqi-lucas: Thanks @wilfred-s , it makes sense to me, we only need prune for sub subFrom if we add positive value.

pkg/scheduler/ugm/queue_tracker.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/queue.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/queue.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/queue.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/application.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/node.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/node.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/node.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/application.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/scheduler/objects/application.go
```diff
@@ -648,6 +649,7 @@ func (sa *Application) AddAllocationAsk(ask *Allocation) error {
 	// Update total pending resource
 	delta.SubFrom(oldAskResource)
 	sa.pending = resources.Add(sa.pending, delta)
+	sa.pending.Prune()
```
zhuqi-lucas, line 652: Keep this for add, because delta seems will be negative in some case.

pkg/scheduler/objects/node.go
```diff
@@ -312,6 +313,7 @@ func (sn *Node) RemoveAllocation(allocationKey string) *Allocation {
 		delete(sn.allocations, allocationKey)
 		sn.allocatedResource.SubFrom(alloc.GetAllocatedResource())
 		sn.availableResource.AddTo(alloc.GetAllocatedResource())
+		sn.allocatedResource.Prune()
```
wilfred-s, line 316: NIT: should move it one line up for clarity. Can be fixed on commit.

wilfred-s, line 316: NIT should move this one line up for clarity

pkg/scheduler/objects/queue.go
```diff
@@ -1609,20 +1612,38 @@ func (sq *Queue) updateMaxResourceMetrics() {
 
 // updateAllocatedResourceMetrics updates allocated resource metrics for all queue types.
 func (sq *Queue) updateAllocatedResourceMetrics() {
+	// Special case when we prune the allocated resources to zero, we need to update the metrics
+	// here only keep basic types memory and vcores to make sure we have updated metrics
+	if sq.allocatedResource.IsEmpty() {
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueueAllocatedResourceMetrics("memory", float64(0))
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueueAllocatedResourceMetrics("vcores", float64(0))
+	}
```
wilfred-s, line 1620: If we prune after we call the update for the metrics we keep the metrics clean and correct. We do not have to make a guess to what was zero. The queue also might not have used memory or vcores ever. Setting them here to zero in the metrics could throw things off.

This would need a comment in the code to why we first call the metric update and then prune.

pkg/scheduler/objects/queue.go
```diff
@@ -1609,20 +1612,38 @@ func (sq *Queue) updateMaxResourceMetrics() {
...
 }
 
 // updatePendingResourceMetrics updates pending resource metrics for all queue types.
 func (sq *Queue) updatePendingResourceMetrics() {
+	// Special case when we prune the pending resources to zero, we need to update the metrics
+	// here only keep basic types memory and vcores to make sure we have updated metrics
+	if sq.pending.IsEmpty() {
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueuePendingResourceMetrics("memory", float64(0))
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueuePendingResourceMetrics("vcores", float64(0))
+	}
```
wilfred-s, line 1633: Same as for the allocated metrics update.

pkg/scheduler/objects/queue.go
```diff
@@ -1609,20 +1612,38 @@ func (sq *Queue) updateMaxResourceMetrics() {
...
 }
 
 // updatePreemptingResourceMetrics updates preempting resource metrics for all queue types.
 func (sq *Queue) updatePreemptingResourceMetrics() {
+	// Special case when we prune the preempting resources to zero, we need to update the metrics
+	// here only keep basic types memory and vcores to make sure we have updated metrics
+	if sq.preemptingResource.IsEmpty() {
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueuePreemptingResourceMetrics("memory", float64(0))
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueuePreemptingResourceMetrics("vcores", float64(0))
+	}
```
wilfred-s, line 1646: Same as for the allocated metrics update.

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
zhuqi-lucas: This is a better and clear solution @wilfred-s , addressed in latest PR.

pkg/scheduler/objects/queue.go
zhuqi-lucas: Addressed in latest PR.

pkg/scheduler/objects/queue.go
zhuqi-lucas: Addressed in latest PR.

pkg/scheduler/objects/node.go
zhuqi-lucas: Addressed in latest PR, thanks!

craigcondit (approved):
+1 LGTM.

PR #951, opened 2024-08-20
[YUNIKORN-2818] Fix state tracking metrics for app and queue

zhuqi-lucas:
…d should be fixed

### What is this PR for?
State of appMetrics of Queue Metrics is incomplete and should be fixed

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-2818

### Questions:

### DIFF
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -203,10 +203,6 @@ func (m *QueueMetrics) IncQueueApplicationsRejected() {
 	m.incQueueApplications(AppRejected)
 }
 
-func (m *QueueMetrics) DecQueueApplicationsRejected() {
-	m.decQueueApplications(AppRejected)
-}
-
 func (m *QueueMetrics) GetQueueApplicationsRejected() (int, error) {
 	metricDto := &dto.Metric{}
 	err := m.appMetricsLabel.WithLabelValues(AppRejected).Write(metricDto)
@@ -254,10 +250,6 @@ func (m *QueueMetrics) IncQueueApplicationsFailed() {
 	m.incQueueApplications(AppFailed)
 }
 
-func (m *QueueMetrics) DecQueueApplicationsFailed() {
-	m.decQueueApplications(AppFailed)
-}
-
 func (m *QueueMetrics) GetQueueApplicationsFailed() (int, error) {
 	metricDto := &dto.Metric{}
 	err := m.appMetricsLabel.WithLabelValues(AppFailed).Write(metricDto)
@@ -286,10 +278,6 @@ func (m *QueueMetrics) GetQueueApplicationsCompleting() (int, error) {
 
 func (m *QueueMetrics) IncQueueApplicationsCompleted() {
 	m.incQueueApplications(AppCompleted)
-}
-
-func (m *QueueMetrics) DecQueueApplicationsCompleted() {
-	m.decQueueApplications(AppCompleted)
 }
 
 func (m *QueueMetrics) GetQueueApplicationsCompleted() (int, error) {
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -27,11 +27,16 @@ import (
 )
 
 const (
-	AppAccepted  = "accepted"
-	AppRunning   = "running"
-	AppFailed    = "failed"
-	AppRejected  = "rejected"
-	AppCompleted = "completed"
+	AppNew        = "new"
+	AppAccepted   = "accepted"
+	AppRunning    = "running"
+	AppFailing    = "failing"
+	AppFailed     = "failed"
+	AppRejected   = "rejected"
+	AppResuming   = "resuming"
+	AppCompleting = "completing"
+	AppCompleted  = "completed"
+	AppExpired    = "expired"
 
 	ContainerReleased  = "released"
 	ContainerAllocated = "allocated"
@@ -160,8 +165,29 @@ func (m *QueueMetrics) GetQueueApplicationsRunning() (int, error) {
 	return -1, err
 }
 
+func (m *QueueMetrics) IncQueueApplicationsNew() {
+	m.incQueueApplications(AppNew)
+}
+
+func (m *QueueMetrics) DecQueueApplicationsNew() {
+	m.decQueueApplications(AppNew)
+}
+
+func (m *QueueMetrics) GetQueueApplicationsNew() (int, error) {
+	metricDto := &dto.Metric{}
+	err := m.appMetricsLabel.WithLabelValues(AppNew).Write(metricDto)
+	if err == nil {
+		return int(*metricDto.Gauge.Value), nil
+	}
+	return -1, err
+}
+
 func (m *QueueMetrics) IncQueueApplicationsAccepted() {
 	m.incQueueApplications(AppAccepted)
+}
+
+func (m *QueueMetrics) DecQueueApplicationsAccepted() {
+	m.decQueueApplications(AppAccepted)
 }
 
 func (m *QueueMetrics) GetQueueApplicationsAccepted() (int, error) {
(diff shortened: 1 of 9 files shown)


pkg/metrics/scheduler.go
```diff
@@ -241,27 +241,49 @@ func (m *SchedulerMetrics) GetSchedulingErrors() (int, error) {
...
+	metricDto := &dto.Metric{}
+	err := m.applicationSubmission.WithLabelValues(AppAccepted).Write(metricDto)
+	if err == nil {
+		return int(*metricDto.Gauge.Value), nil
+	}
+	return -1, err
 }
 
-func (m *SchedulerMetrics) AddTotalApplicationsRejected(value int) {
-	m.applicationSubmission.WithLabelValues(ContainerRejected).Add(float64(value))
```
zhuqi-lucas, line 257: The name was wrong, it should be applicationReject, and also remove this logic because it will be all handled in state machine changes, it will be duplicated.

pkg/metrics/scheduler.go
```diff
@@ -274,10 +296,6 @@ func (m *SchedulerMetrics) DecTotalApplicationsRunning() {
 	m.application.WithLabelValues(AppRunning).Dec()
 }
 
-func (m *SchedulerMetrics) SubTotalApplicationsRunning(value int) {
-	m.application.WithLabelValues(AppRunning).Sub(float64(value))
```
zhuqi-lucas, line 278: Same with above, remove this duplicated logic.

pkg/metrics/scheduler.go
```diff
@@ -84,8 +84,8 @@ func InitSchedulerMetrics() *SchedulerMetrics {
 			Help:      "Total number of attempts to allocate containers. State of the attempt includes `allocated`, `rejected`, `error`, `released`",
 		}, []string{"state"})
 
-	s.applicationSubmission = prometheus.NewCounterVec(
-		prometheus.CounterOpts{
+	s.applicationSubmission = prometheus.NewGaugeVec(
```
zhuqi-lucas, line 87: Change to gauge because, we want to support dec also

craigcondit:
Please run `make lint` before submitting PRs.

zhuqi-lucas:
> Please run `make lint` before submitting PRs.

Fix go lint now.

pkg/metrics/queue.go
```diff
@@ -177,6 +203,10 @@ func (m *QueueMetrics) IncQueueApplicationsRejected() {
 	m.incQueueApplications(AppRejected)
 }
 
+func (m *QueueMetrics) DecQueueApplicationsRejected() {
+	m.decQueueApplications(AppRejected)
+}
```
chenyulin0719, line 208: DecQueueApplicationsRejected() is not called in "leave_Rejected" state? Is it intentional?

pkg/metrics/queue.go
```diff
@@ -199,10 +267,31 @@ func (m *QueueMetrics) GetQueueApplicationsFailed() (int, error) {
...
+	return -1, err
+}
+
 func (m *QueueMetrics) IncQueueApplicationsCompleted() {
 	m.incQueueApplications(AppCompleted)
 }
 
+func (m *QueueMetrics) DecQueueApplicationsCompleted() {
+	m.decQueueApplications(AppCompleted)
+}
```
chenyulin0719, line 293: DecQueueApplicationsCompleted() is not called in "leave_completed" state too.

chenyulin0719 (commented):
Looks like "Rejected","Completed","Failed" to "Expired" app states are not tracked, neither in queue nor sheduler. Is it intentional?

chenyulin0719 (changes requested):
The queue metrices description is not updated: 
- https://github.com/zhuqi-lucas/yunikorn-core/blob/0d21b97470d3ca748e54ad075231d85d038f5dcc/pkg/metrics/queue.go#L73
- https://github.com/zhuqi-lucas/yunikorn-core/blob/0d21b97470d3ca748e54ad075231d85d038f5dcc/pkg/metrics/queue.go#L81

image

pkg/metrics/scheduler.go
wilfred-s: A submission can not be changed. When an application is submitted it will always be submitted. It should be a counter that keeps increasing. It is not a state that we track it is a pure counter from start to shutdown.

pkg/metrics/queue.go
wilfred-s: Rejected applications should only increase, never decrease: a simple counter.
It is a final state that the application only gets removed from as part of the cleanup. We keep it around for a while to make it traceable for the submitter. i.e. submit an application that gets rejected: find it is the list as such. If we would drop it immediately the submitter has no idea what happened.

pkg/metrics/scheduler.go
zhuqi-lucas: Thanks @wilfred-s for clarify for this field,it makes sense to me.

pkg/metrics/queue.go
zhuqi-lucas: @chenyulin0719 @wilfred-s 
If i make sense right, we don't need to add decrease for following state?
1. Rejected
2. Failed
3. Completed

And for expired, i don't think we need to tracking it?

pkg/metrics/queue.go
```diff
@@ -186,10 +216,48 @@ func (m *QueueMetrics) GetQueueApplicationsRejected() (int, error) {
...
+	return -1, err
+}
+
 func (m *QueueMetrics) IncQueueApplicationsFailed() {
 	m.incQueueApplications(AppFailed)
 }
 
+func (m *QueueMetrics) DecQueueApplicationsFailed() {
+	m.decQueueApplications(AppFailed)
+}
```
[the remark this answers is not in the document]
zhuqi-lucas: Same as above.

pkg/metrics/queue.go
zhuqi-lucas: Same as above.

pkg/metrics/scheduler.go
```diff
@@ -287,16 +305,49 @@ func (m *SchedulerMetrics) GetTotalApplicationsRunning() (int, error) {
...
+	err := m.application.WithLabelValues(AppResuming).Write(metricDto)
+	if err == nil {
+		return int(*metricDto.Gauge.Value), nil
+	}
+	return -1, err
+}
+
+func (m *SchedulerMetrics) DecTotalApplicationsFailed() {
+	m.application.WithLabelValues(AppFailed).Dec()
+}
```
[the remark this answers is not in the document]
zhuqi-lucas: Removed DecTotalApplicationsFailed, it was not used.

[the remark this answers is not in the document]
zhuqi-lucas: Addressed in latest PR.

pkg/metrics/queue.go
zhuqi-lucas: Addressed in latest PR.

pkg/metrics/queue.go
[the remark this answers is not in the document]
zhuqi-lucas: Addressed in latest PR.

pkg/metrics/queue.go
zhuqi-lucas: Addressed in latest PR.

chenyulin0719: @wilfred-s Understood, thanks for the explanation.
@zhuqi-lucas Not tracking expired metrics make sense to me. 
To avoid any confusion, I think we should remove all the decrement function for Rejected/Failed/Completed.  WDYT?

zhuqi-lucas: Got it @chenyulin0719 , remove those unused code in latest PR!

chenyulin0719 (approved):
+1 LGTM

craigcondit (approved):
+1 LGTM.

PR #957, opened 2024-08-28
[YUNIKORN-2837] Log & Send Events, Improve logging

manirajv06:
### What is this PR for?
Send events in appropriate places, enhanced logging etc to help understand the preemption flow while debugging preemption related issues.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2837

### Questions:

### DIFF
--- a/pkg/common/errors.go
+++ b/pkg/common/errors.go
@@ -20,8 +20,10 @@ package common
 
 import "errors"
 
-// Common errors
-var (
-	// InvalidQueueName returned when queue name is invalid
-	InvalidQueueName = errors.New("invalid queue name, max 64 characters consisting of alphanumeric characters and '-', '_', '#', '@', '/', ':' allowed")
-)
+// InvalidQueueName returned when queue name is invalid
+var InvalidQueueName = errors.New("invalid queue name, max 64 characters consisting of alphanumeric characters and '-', '_', '#', '@', '/', ':' allowed")
+
+const PreemptionPreconditionsFailed = "Preemption preconditions failed"
+const PreemptionDoesNotGuarantee = "Preemption queue guarantees check failed"
+const PreemptionShortfall = "Preemption helped but short of resources"
+const PreemptionDoesNotHelp = "Preemption does not help"
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1056,6 +1056,7 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
 						// preemption occurred, and possibly reservation
 						return result
 					}
+					request.LogAllocationFailure(common.PreemptionDoesNotHelp, true)
 				}
 			}
 			request.LogAllocationFailure(NotEnoughQueueQuota, true) // error message MUST be constant!
@@ -1122,6 +1123,7 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
 						return result
 					}
 				}
+				request.LogAllocationFailure(common.PreemptionDoesNotHelp, true)
 			}
 		}
 	}
@@ -1394,6 +1396,7 @@ func (sa *Application) tryPreemption(headRoom *resources.Resource, preemptionDel
 
 	// validate prerequisites for preemption of an ask and mark ask for preemption if successful
 	if !preemptor.CheckPreconditions() {
+		ask.LogAllocationFailure(common.PreemptionPreconditionsFailed, true)
 		return nil, false
 	}
 
--- a/pkg/scheduler/objects/preemption_test.go
+++ b/pkg/scheduler/objects/preemption_test.go
@@ -25,6 +25,7 @@ import (
 
 	"gotest.tools/v3/assert"
 
+	"github.com/apache/yunikorn-core/pkg/common"
 	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/apache/yunikorn-core/pkg/mock"
 	"github.com/apache/yunikorn-core/pkg/plugins"
@@ -85,6 +86,13 @@ func TestCheckPreconditions(t *testing.T) {
 	// verify that recently checked ask is disqualified
 	ask.preemptCheckTime = time.Now()
 	assert.Assert(t, !preemptor.CheckPreconditions(), "preconditions succeeded with recently tried ask")
+	getNode := func(nodeID string) *Node {
+		return node
+	}
+	preemptionAttemptsRemaining := 1
+	result := app.tryAllocate(resources.NewResourceFromMap(map[string]resources.Quantity{"first": 2}), true, 1*time.Second, &preemptionAttemptsRemaining, iterator, iterator, getNode)
+	assert.Check(t, result == nil, "unexpected result")
+	assertAllocationLog(t, ask)
 	ask.preemptCheckTime = time.Now().Add(-1 * time.Minute)
 	assert.Assert(t, preemptor.CheckPreconditions(), "preconditions failed")
 	assert.Assert(t, !preemptor.CheckPreconditions(), "preconditions succeeded on successive run")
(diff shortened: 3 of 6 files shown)


craigcondit (changes requested):
I'm -1 on this entire approach (see comments on JIRA for a detailed explanation).

pkg/scheduler/objects/allocation.go
```diff
@@ -565,3 +567,39 @@ func (a *Allocation) setUserQuotaCheckPassed() {
...
+}
+
+func (a *Allocation) setPreemptionQueueGuaranteesCheckFailed() {
+	a.Lock()
+	defer a.Unlock()
+	if !a.preemptionQueueGuaranteesCheckFailed {
+		a.preemptionQueueGuaranteesCheckFailed = true
+		a.askEvents.SendPreemptionQueueGuaranteesCheckFailed(a.allocationKey, a.applicationID, a.allocatedResource)
+	}
+}
```
pbacsko, line 605: These are all `REQUEST` events which means they will be sent to the respective pods. 

I think the only message which might be useful is an indication that a particular pod successfully triggered preemption. Or the other way around: we send a message to the terminated pod that it's been preempted to make room for pod X.

Everything else (guarantees passed/failed, precondition passed/failed) is just too much noise which should not be sent to K8s as a pod event.

pbacsko (changes requested):
I'm also -1, we need to be careful about what we send & how often.

pkg/scheduler/objects/allocation.go
craigcondit: I agree we should not send events except in these cases:

Request pod: "Preempting pod {namespace}/{pod} to make room"  (there could be multiple of these if we preempt multiple pods)

Victim pods: "Preempting pod to make room for {namespace}/{pod}" (only one of these per victim)

This allows users to see that their pod triggered preemption, or that their pod was preempted and by whom.

manirajv06: As of now, there is only one log line "Reserving node for ask after preemption" being logged at the end of the preemption process. Other than this, we don't have any info. 

We need below info to debug the preemption related issues:

1. How do we know whether preemption has been even attempted or not?
2. If attempted, Is there any problem in passing through important steps?
3. If attempted and successful case, how do we know this case? Answer: Above log line helps

I went through all these questions while trying to find the reason for test failures logged in https://issues.apache.org/jira/browse/YUNIKORN-2808.

> I think the only message which might be useful is an indication that a particular pod successfully triggered preemption.

Yes. It definitely helps. We need something like this. I had also similar thought, but doing this way would always sends an event. Hence, I followed the other usual model of 1. Sending only one event and that too only when failed. Does not repeat. 2. Sending only one event when succeeded (Only If 1st event happens). Does not repeat.

> Request pod: "Preempting pod {namespace}/{pod} to make room" (there could be multiple of these if we preempt multiple pods)
> 
> Victim pods: "Preempting pod to make room for {namespace}/{pod}" (only one of these per victim)

Yes, it helps but only in successful case and that too at the end.

craigcondit: We *must not* log for every attempt as this happens far too frequently. It will overwhelm the system. Currently in the scheduler we only log *when action is taken or the state of the system changes*.

craigcondit: Pod events definitely should only be sent once, and only when action is taken.

pbacsko: I think an acceptable approach is recording stuff like the allocation log for each request. That looks reasonable to me. We can have a type like `AllocationLog` and we define fixed causes about why the preemption failed:
* "Preconditions checks failed" (`preemptor.CheckPreconditions()` false)
* "No guarantee to free up resources" (`p.checkPreemptionQueueGuarantees()` false)
* "Preemption shortfall" (`p.ask.GetAllocatedResource().StrictlyGreaterThanOnlyExisting()` false)

Every time something fails, you increase a counter for it. You also count the number of total attempts.

So I'm thinking sth like:
```
// Mutable fields which need protection
allocated           bool
allocLog            map[string]*AllocationLogEntry
preemptionLog map[string]*AllocationLogEntry  // new field
preemptionAttempts int64  // new field
preemptionTriggered bool
preemptCheckTime    time.Time
...

var (
  ErrPreemptionPrecondFailed = errors.New("Preconditions checks failed")
  ErrPreemptionNoGuarantee = errors.New("No guarantee to free up resources")
  ErrPreemptionShortfall = errors.New("Preemption shortfall")
)

```
Then in `application.go`:
```
				if fullIterator != nil {
                                        request.IncPreemptionAttempts()
					if result, err := sa.tryPreemption(headRoom, preemptionDelay, request, fullIterator, false); err != nil {
						// preemption occurred, and possibly reservation
						return result
					}
					request.LogPreemptionFailure(err) 
				}
			}
```

Obviously you need to make sure that problems propagate back so you can log it. Then this whole thing can be exposed in the REST API just like the allocation log.

It's obviously not as detailed as a separate log entry printed to the stdout, but I see this as a good compromise.

craigcondit: Why not just reuse `AllocationLog`? It already contains events pertaining to things like predicate failures; preemption attempts can be tracked there easily as well.

pbacsko: Yeah, that's fine too.

manirajv06: Have used allocationLog earlier too in few places in addition with events & logging. Now using only allocationLog, removed events and logs. Still seeing a gap to add debug logs in few places especially after guarantee checks passes. Anyways, we can  address this later depends on the need.

pkg/common/errors.go
```diff
@@ -24,4 +24,9 @@ import "errors"
 var (
 	// InvalidQueueName returned when queue name is invalid
 	InvalidQueueName = errors.New("invalid queue name, max 64 characters consisting of alphanumeric characters and '-', '_', '#', '@', '/', ':' allowed")
+
+	PreemptionPreconditionsFailed = errors.New("PreemptionPreconditionsFailed")
+	PreemptionDoesNotGuarantee    = errors.New("PreemptionQueueGuaranteesCheckFailed")
+	PreemptionShortfall           = errors.New("PreemptionHelpedButShortfallOfResources")
+	PreemptionDoesNotHelp         = errors.New("PreemptionDoesNotHelp")
```
pbacsko, line 31: Just use string constants

pkg/scheduler/objects/application.go
```diff
@@ -1122,6 +1123,7 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
 						return result
 					}
 				}
+				request.LogAllocationFailure(common.PreemptionDoesNotHelp.Error(), true)
```
pbacsko, line 1126: Please add proper coverage to this line

pkg/scheduler/objects/preemption_test.go
```diff
@@ -85,6 +86,25 @@ func TestCheckPreconditions(t *testing.T) {
 	// verify that recently checked ask is disqualified
 	ask.preemptCheckTime = time.Now()
 	assert.Assert(t, !preemptor.CheckPreconditions(), "preconditions succeeded with recently tried ask")
+	nodeMap := map[string]*Node{"node1": node}
+	getNode := func(nodeID string) *Node {
+		return nodeMap[nodeID]
```
pbacsko, line 91: nit: do we need this map? Can't we just return a "node" object?

pbacsko (changes requested):
Looks good, just minor things

pkg/common/errors.go
pbacsko: Actually I just realized that it's better if the string literals are complete sentences like "Preemption Preconditions Failed", "Preemption queue guarantees check failed", etc because that will be exposed to YK users.

pkg/scheduler/objects/utilities_test.go
```diff
@@ -292,6 +293,21 @@ func assertUserResourcesAndGroupResources(t *testing.T, userGroup security.UserG
...
+	PreemptionDoesNotHelp := false
+	for _, l := range log {
+		if l.Message == common.PreemptionPreconditionsFailed {
+			preemptionPreconditionsFailed = true
+		} else if l.Message == common.PreemptionDoesNotHelp {
+			PreemptionDoesNotHelp = true
+		}
+	}
+	assert.Assert(t, preemptionPreconditionsFailed, true)
+	assert.Assert(t, PreemptionDoesNotHelp, true)
```
pbacsko, line 308: nit: "true" is not needed at the end (can be fixed on commit)

pbacsko (approved):
LGTM, I found only a nit

PR #960, opened 2024-09-06
[YUNIKORN-2855] Rethink the call for update metrics with nil resource

zhuqi-lucas:
### What is this PR for?
We will not update the nil resource to resource metrics, we need to fix it.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-2855

### Questions:

### DIFF
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -19,6 +19,7 @@
 package metrics
 
 import (
+	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"go.uber.org/zap"
@@ -57,6 +58,8 @@ type QueueMetrics struct {
 	resourceMetricsLabel *prometheus.GaugeVec
 	// Deprecated - To be removed in 1.7.0. Replaced with queue label Metrics
 	resourceMetricsSubsystem *prometheus.GaugeVec
+	// Track known resource types.
+	knownResourceTypes map[string]struct{}
 }
 
 // InitQueueMetrics to initialize queue metrics
@@ -123,6 +126,7 @@ func InitQueueMetrics(name string) *QueueMetrics {
 		}
 	}
 
+	q.knownResourceTypes = make(map[string]struct{})
 	return q
 }
 
@@ -146,6 +150,7 @@ func (m *QueueMetrics) Reset() {
 	m.appMetricsSubsystem.Reset()
 	m.resourceMetricsLabel.Reset()
 	m.resourceMetricsSubsystem.Reset()
+	m.knownResourceTypes = make(map[string]struct{})
 }
 
 func (m *QueueMetrics) IncQueueApplicationsRunning() {
@@ -301,12 +306,20 @@ func (m *QueueMetrics) AddReleasedContainers(value int) {
 	m.containerMetrics.WithLabelValues(ContainerReleased).Add(float64(value))
 }
 
-func (m *QueueMetrics) SetQueueGuaranteedResourceMetrics(resourceName string, value float64) {
-	m.setQueueResource(QueueGuaranteed, resourceName, value)
-}
+func (m *QueueMetrics) UpdateQueueResourceMetrics(state string, newResources map[string]resources.Quantity) {
+	// Iterate over new resource types and set their values.
+	for resourceName, value := range newResources {
+		m.setQueueResource(state, resourceName, float64(value))
+		// Add new resources to the known list
+		m.knownResourceTypes[resourceName] = struct{}{}
+	}
 
-func (m *QueueMetrics) SetQueueMaxResourceMetrics(resourceName string, value float64) {
-	m.setQueueResource(QueueMax, resourceName, value)
+	// Emit old resource types that are missing in the new collection with zero.
+	for resourceName := range m.knownResourceTypes {
+		if _, exists := newResources[resourceName]; !exists {
+			m.setQueueResource(state, resourceName, float64(0))
+		}
+	}
 }
 
 func (m *QueueMetrics) SetQueueAllocatedResourceMetrics(resourceName string, value float64) {
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -19,11 +19,11 @@
 package metrics
 
 import (
-	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"go.uber.org/zap"
 
+	"github.com/apache/yunikorn-core/pkg/common/resources"
 	"github.com/apache/yunikorn-core/pkg/log"
 )
 
--- a/pkg/metrics/queue.go
+++ b/pkg/metrics/queue.go
@@ -19,13 +19,12 @@
 package metrics
 
 import (
-	"sync"
-
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"go.uber.org/zap"
 
 	"github.com/apache/yunikorn-core/pkg/common/resources"
+	"github.com/apache/yunikorn-core/pkg/locking"
 	"github.com/apache/yunikorn-core/pkg/log"
 )
 
@@ -62,7 +61,7 @@ type QueueMetrics struct {
 	resourceMetricsSubsystem *prometheus.GaugeVec
 	// Track known resource types
 	knownResourceTypes map[string]struct{}
-	lock               sync.Mutex
+	lock               locking.RWMutex
 }
 
 // InitQueueMetrics to initialize queue metrics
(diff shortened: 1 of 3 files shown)


craigcondit (changes requested):
I'm -1 on this approach. This adds significant complexity to the code and requires all callers to do the calculations... make the metrics functions handle the nil case properly themselves.

zhuqi-lucas:
Thanks @craigcondit for review, i addressed the code to a more reasonable way and setting zero resource for nil update case.

craigcondit (commented):
Maybe I'm missing something entirely... Why are we setting arbitrary resources to zero here? What is it we're trying to accomplish?

zhuqi-lucas:
@craigcondit , we support setting maxResource/ guaranteed resource to nil, but the metrics update is wrong because we don't update it from previous value update to zero/nil.

craigcondit:
Then it seems the proper solution is to retrieve the previous value of the metric and update the values there (an "unprune" so to speak). Missing values n the new map should be set to zero. Setting arbitrary resources is just wrong. Doing it this way also has the advantage of ensuring that any resource types that were emitted by the metric previously are still emitted, but with zero instead of missing.

zhuqi-lucas:
Thanks, @craigcondit i like the idea unprune, i will address this way.

craigcondit (changes requested):
Just... no. You already have the metric in the collector itself. Read it out to get the list of values that are previously set and build a new metric using a combination of the old and new values. There should be no code changes outside the collector at all.

zhuqi-lucas:
I see... @craigcondit let me change my code.

zhuqi-lucas:
Addressed in latest PR:

1. Getting the existed metrics.
2. Setting those field to zero.

pkg/scheduler/objects/queue.go
```diff
@@ -1659,17 +1659,21 @@ func (sq *Queue) SupportTaskGroup() bool {
 func (sq *Queue) updateGuaranteedResourceMetrics() {
 	if sq.guaranteedResource != nil {
 		for k, v := range sq.guaranteedResource.Resources {
```
craigcondit, line 1661: There's no reason for any of this logic to be here. Encapsulate all of it within pkg/metrics/queue.go.

pkg/scheduler/objects/queue.go
```diff
@@ -1659,17 +1659,21 @@ func (sq *Queue) SupportTaskGroup() bool {
...
+	} else {
+		metrics.GetQueueMetrics(sq.QueuePath).SetQueueNilResourceMetrics(metrics.QueueGuaranteed)
 	}
 }
 
 // updateMaxResourceMetrics updates max resource metrics.
 func (sq *Queue) updateMaxResourceMetrics() {
 	if sq.maxResource != nil {
 		for k, v := range sq.maxResource.Resources {
-			metrics.GetQueueMetrics(sq.QueuePath).SetQueueMaxResourceMetrics(k, float64(v))
```
craigcondit, line 1671: Same - move logic into pkg/metrics/queue.go.

pkg/metrics/queue.go
```diff
@@ -301,12 +301,46 @@ func (m *QueueMetrics) AddReleasedContainers(value int) {
...
 
-func (m *QueueMetrics) SetQueueGuaranteedResourceMetrics(resourceName string, value float64) {
-	m.setQueueResource(QueueGuaranteed, resourceName, value)
+func (m *QueueMetrics) SetQueueResourceMetrics(state string, resourceName string, value float64) {
+	m.setQueueResource(state, resourceName, value)
 }
 
-func (m *QueueMetrics) SetQueueMaxResourceMetrics(resourceName string, value float64) {
-	m.setQueueResource(QueueMax, resourceName, value)
+func (m *QueueMetrics) SetQueueNilResourceMetrics(state string) {
```
craigcondit, line 308: This should be made an internal function once logic is moved here from scheduler/objects/queue.go.

pkg/metrics/queue.go
```diff
@@ -301,12 +301,46 @@ func (m *QueueMetrics) AddReleasedContainers(value int) {
...
+}
+
+func (m *QueueMetrics) GetAllQueueStateMetrics(state string) (map[string]float64, error) {
+	metrics := make(map[string]float64)
+	metricFamilies, err := prometheus.DefaultGatherer.Gather()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, metricFamily := range metricFamilies {
```
craigcondit, line 327: Can this be simplified? This seems... extreme. Perhaps we can simply keep (as part of the QueueMetrics object) a list of known resource types for the current queue. Then we can pass in all the resources types at once into SetQueueResourceMetrics(). Iterate both old and new types. Old types that don't exist in the new collection get emitted with zero; new types get added to the known list. Values in the new collection get emitted as-is. Be sure to us appropriate locking.

This allows a much more efficient lookup that doesn't involve traversing the entirety of the metrics, and provides a much more convenient interface for callers.

craigcondit (changes requested):
The code is still too complex. See review comments.

pkg/metrics/queue.go
zhuqi-lucas: Good idea, i will address this in new PR.

pkg/scheduler/objects/queue.go
zhuqi-lucas: Addressed in latest PR, thanks.

pkg/metrics/queue.go
zhuqi-lucas: It seems we have a lock for metric level, but not queue level:
```
func init() {
	once.Do(func() {
		m = &Metrics{
			scheduler: InitSchedulerMetrics(),
			queues:    make(map[string]*QueueMetrics),
			event:     initEventMetrics(),
			lock:      locking.RWMutex{},
			runtime:   initRuntimeMetrics(),
		}
	})
}
```

Do we need to create a queue level lock to lock the knownResourceType?

pkg/metrics/queue.go
```diff
@@ -57,6 +58,8 @@ type QueueMetrics struct {
 	resourceMetricsLabel *prometheus.GaugeVec
 	// Deprecated - To be removed in 1.7.0. Replaced with queue label Metrics
 	resourceMetricsSubsystem *prometheus.GaugeVec
+	// Track known resource types.
+	knownResourceTypes map[string]struct{}
```
craigcondit, line 62: Need a lock here to guard access to knownResourceTypes.

pkg/metrics/queue.go
```diff
@@ -146,6 +150,7 @@ func (m *QueueMetrics) Reset() {
 	m.appMetricsSubsystem.Reset()
 	m.resourceMetricsLabel.Reset()
 	m.resourceMetricsSubsystem.Reset()
+	m.knownResourceTypes = make(map[string]struct{})
```
craigcondit, line 153: Guard this with a lock (see previous comment).

pkg/metrics/queue.go
```diff
@@ -301,12 +306,20 @@ func (m *QueueMetrics) AddReleasedContainers(value int) {
...
 
-func (m *QueueMetrics) SetQueueGuaranteedResourceMetrics(resourceName string, value float64) {
-	m.setQueueResource(QueueGuaranteed, resourceName, value)
-}
+func (m *QueueMetrics) UpdateQueueResourceMetrics(state string, newResources map[string]resources.Quantity) {
+	// Iterate over new resource types and set their values.
+	for resourceName, value := range newResources {
+		m.setQueueResource(state, resourceName, float64(value))
+		// Add new resources to the known list
+		m.knownResourceTypes[resourceName] = struct{}{}
```
craigcondit, line 314: Guard this with a lock (see previous comment).

craigcondit (changes requested):
Add locking per comments.

pkg/metrics/queue.go
craigcondit: The individual metric instance needs a lock (see review comments).

pkg/metrics/queue.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/metrics/queue.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/metrics/queue.go
zhuqi-lucas: Addressed in latest PR, thanks!

pkg/metrics/queue.go
```diff
@@ -58,8 +59,9 @@ type QueueMetrics struct {
 	resourceMetricsLabel *prometheus.GaugeVec
 	// Deprecated - To be removed in 1.7.0. Replaced with queue label Metrics
 	resourceMetricsSubsystem *prometheus.GaugeVec
-	// Track known resource types.
+	// Track known resource types
 	knownResourceTypes map[string]struct{}
+	lock               locking.RWMutex
```
craigcondit, line 64: This doesn't need to be an RWMutex as we never lock for reads.

craigcondit (changes requested):
Change lock type.

pkg/metrics/queue.go
zhuqi-lucas: Good catch, addressed in latest PR, thanks!

craigcondit (approved):
+1 LGTM.

PR #965, opened 2024-09-07
[YUNIKORN-2859] fix: missing assertion in TestGetApplicationHandler

ryankert01:
### What is this PR for?
our current code cannot verify *dao.ApplicationDAOInfo, both empty and legit app can pass the test. we should fix that.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2859

### Questions:

### DIFF
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -1741,6 +1741,9 @@ func TestGetApplicationHandler(t *testing.T) {
 	getApplication(resp, req)
 	err = json.Unmarshal(resp.outputBytes, &appsDao)
 	assert.NilError(t, err, unmarshalError)
+	assert.Equal(t, appsDao.ApplicationID, "app-1")
+	assert.Equal(t, appsDao.Partition, "default")
+	assert.Equal(t, appsDao.QueueName, "root.default")
 
 	if !appsDao.HasReserved {
 		assert.Equal(t, len(appsDao.Reservations), 0)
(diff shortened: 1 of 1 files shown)


pkg/webservice/handlers_test.go
```diff
@@ -1781,6 +1784,15 @@ func TestGetApplicationHandler(t *testing.T) {
...
 	assert.NilError(t, err, unmarshalError)
+	assert.Assert(t, appsDao4 != nil)
+	assert.Equal(t, appsDao4.ApplicationID, "app-1")
+	assert.Equal(t, appsDao4.Partition, "default")
+	assert.Equal(t, appsDao4.QueueName, "root.default")
+	if !appsDao4.HasReserved {
+		assert.Equal(t, len(appsDao4.Reservations), 0)
+	} else {
+		assert.Check(t, len(appsDao4.Reservations) > 0, "app should have at least 1 reservation")
+	}
```
pbacsko, line 1795: What determines which branch is taken? This is not a re-used code as far as I can see, so this is supposed to be predictable.

pbacsko: My question also applies to the existing code at L#1748.

ryankert01: Indeed, and looks like there's no reservation there, so I assert it to be the default value.
```
assert.Equal(t, len(appsDao4.Reservations), 0)
```

ryankert01: Done.

craigcondit (approved):
+1 LGTM.

PR #968, opened 2024-09-13
[YUNIKORN-2876] Initialize queue metrics for app after queue is set

zhuqi-lucas:
…later after the queue placed

### What is this PR for?
Currently, we push the queue app new state metrics after application objects are created, but when testing, the dynamic queue should be empty at that time, we need to do this later when we can make sure the queue placed for dynamic case.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-2876

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -52,7 +52,6 @@ var (
 	terminatedTimeout         = 3 * 24 * time.Hour
 	defaultPlaceholderTimeout = 15 * time.Minute
 )
-var once sync.Once
 var initAppLogOnce sync.Once
 var rateLimitedAppLog *log.RateLimitedLogger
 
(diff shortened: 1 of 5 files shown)


pkg/scheduler/objects/application.go
```diff
@@ -191,13 +190,15 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
...
-	app.setNewMetrics()
 	return app
 }
 
-func (sa *Application) setNewMetrics() {
-	metrics.GetSchedulerMetrics().IncTotalApplicationsNew()
-	metrics.GetQueueMetrics(sa.GetQueuePath()).IncQueueApplicationsNew()
+func (sa *Application) SetNewMetrics() {
+	// ensure the queue is set before calling this function
+	if sa.GetQueuePath() != "" {
```
craigcondit, line 198: Don't add a conditional, just execute `setNewMetrics()` inside `SetQueue()`. Callers should not need to be updated at all.

craigcondit (changes requested):
See comments.

pkg/scheduler/objects/application.go
zhuqi-lucas: Thanks @craigcondit , good suggestion, addressed in latest PR.

craigcondit (approved):
+1 LGTM.

PR #970, opened 2024-09-16
[YUNIKORN-2797] Increase handlers.go test coverage

ryankert01:
### What is this PR for?
Increase test coverage and some refactor.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2797

### Questions:

### DIFF
--- a/pkg/webservice/handlers_test.go
+++ b/pkg/webservice/handlers_test.go
@@ -318,6 +318,18 @@ func newApplication(appID, partitionName, queueName, rmID string, ugi security.U
 	return objects.NewApplication(siApp, userGroup, nil, rmID)
 }
 
+func TestGetStackInfo(t *testing.T) {
+	req, err := http.NewRequest("GET", "", strings.NewReader(baseConf))
+	assert.NilError(t, err, "Error creating request")
+	resp := &MockResponseWriter{}
+	getStackInfo(resp, req)
+
+	// assert stack trace
+	assert.Assert(t, strings.Contains(string(resp.outputBytes), "goroutine"), "Stack trace should be present in the response")
+	assert.Assert(t, strings.Contains(string(resp.outputBytes), "test"), "Stack trace should be present in the response")
+	assert.Assert(t, strings.Contains(string(resp.outputBytes), "github.com/apache/yunikorn-core/pkg/webservice.getStackInfo"), "Stack trace should be present in the response")
+}
+
 func TestValidateConf(t *testing.T) {
 	confTests := []struct {
 		content          string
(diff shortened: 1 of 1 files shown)


pkg/webservice/handlers_test.go
```diff
@@ -318,6 +318,18 @@ func newApplication(appID, partitionName, queueName, rmID string, ugi security.U
...
+	req, err := http.NewRequest("GET", "", strings.NewReader(baseConf))
+	assert.NilError(t, err, "Error creating request")
+	resp := &MockResponseWriter{}
+	getStackInfo(resp, req)
+
+	// assert stack trace
+	assert.Assert(t, strings.Contains(string(resp.outputBytes), "goroutine"), "Stack trace should be present in the response")
+	assert.Assert(t, strings.Contains(string(resp.outputBytes), "test"), "Stack trace should be present in the response")
+	assert.Assert(t, strings.Contains(string(resp.outputBytes), "github.com/apache/yunikorn-core/pkg/webservice.getStackInfo"), "Stack trace should be present in the response")
+}
```
manirajv06, line 331: Can you cover -ve cases?

pkg/webservice/handlers_test.go
```diff
@@ -1358,17 +1364,42 @@ func TestGetPartitionNodes(t *testing.T) {
...
+	getPartitionNode(resp, req)
+	assertParamsMissing(t, resp)
+
+	// Test partition not exist
+	req, err = createRequest(t, "/ws/v1/partition/notexists/node/node-1", map[string]string{"partition": "notexists"})
+	assert.NilError(t, err, "Get Nodes for PartitionNode Handler request failed")
+	resp = &MockResponseWriter{}
+	getPartitionNodes(resp, req)
+	assertPartitionNotExists(t, resp)
+}
```
manirajv06, line 1396: It is growing big. Probably, time to split into two separate methods, one for Nodes and another one for specific node.

manirajv06 (changes requested):

pkg/webservice/handlers_test.go
```diff
@@ -1358,17 +1364,42 @@ func TestGetPartitionNodes(t *testing.T) {
...
 	assertNodeIDNotExists(t, resp)
+
+	// Test param missing
+	req, err = http.NewRequest("GET", "/ws/v1/partition/default/node", strings.NewReader(""))
+	assert.NilError(t, err, "Get Node for PartitionNode Handler request failed")
+	resp = &MockResponseWriter{}
+	getPartitionNode(resp, req)
+	assertParamsMissing(t, resp)
+
+	// Test partition not exist
```
pbacsko, line 1390: nit: "does not exist"

pkg/webservice/handlers_test.go
ryankert01: Done.

pkg/webservice/handlers_test.go
ryankert01: Done.

pkg/webservice/handlers_test.go
ryankert01: Is there a negative case for getStackInfo? It seems to be cannot make runtime.Stack() failed. 

Ps. I added a deep stack trace case.

manirajv06: I meant, asserting cases like StatusInternalServerError etc by creating such situation.

ryankert01: @manirajv06 Just to clarify, I'm actually asking how to create a negative case (such as StatusInternalServerError). The whole API doesn't need any parameter, and the function mainly use `runtime.Stacktrace`, not sure if there's a way to fail it...

manirajv06: StatusInternalServerError is returned only when go built in write() method fails. Yes, not required to cover the case. Just to be safer, we can assert status code against valid response code (200).

ryankert01: Got it, done.

manirajv06 (approved):
looks good. Will merge after successful run.

ryankert01:
I got it! We shouldn't assert statusOK in this function's test because it will not set statusOK in the slope.

manirajv06:
> I got it! We shouldn't assert statusOK in this function's test because the tested function will not set statusOK in the slope.

It should other way around, not equal to StatusInternalServerError. Anyways, not a problem.

PR #971, opened 2024-09-20
[YUNIKORN-2883] Improve sorters & queue_tracker funtion's test coverage

SP12893678:
### What is this PR for?
Improve the following funtion's test coverage:

- canBeRemoved (child queue Tracker canBeRemovedInternal false case)
- String
- sortApplicationsBySubmissionTimeAndPriority (left SubmissionTime before right)

### What type of PR is it?
* [x] - Test

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2883

### DIFF
--- a/pkg/scheduler/ugm/queue_tracker.go
+++ b/pkg/scheduler/ugm/queue_tracker.go
@@ -410,11 +410,7 @@ func (qt *QueueTracker) canRunApp(hierarchy []string, applicationID string, trac
 func (qt *QueueTracker) canBeRemoved() bool {
 	for _, childQT := range qt.childQueueTrackers {
 		// quick check to avoid further traversal
-		if childQT.canBeRemovedInternal() {
-			if !childQT.canBeRemoved() {
-				return false
-			}
-		} else {
+		if !childQT.canBeRemovedInternal() {
 			return false
 		}
 	}
--- a/pkg/scheduler/objects/sorters_test.go
+++ b/pkg/scheduler/objects/sorters_test.go
@@ -236,15 +236,15 @@ func TestSortAppsFifo(t *testing.T) {
 
 	input["app-1"].askMaxPriority = 3
 	input["app-3"].askMaxPriority = 5
-	input["app-2"].SubmissionTime = input["app-3"].SubmissionTime
+	input["app-2"].SubmissionTime = input["app-3"].SubmissionTime.Add(time.Nanosecond * -5)
 	input["app-1"].SubmissionTime = input["app-3"].SubmissionTime
 	list = sortApplications(input, policies.FifoSortPolicy, false, nil)
 	/*
 	* apps order: 0, 3, 1, 2
-	* the resultType of app index is [0, 2, 3, 1]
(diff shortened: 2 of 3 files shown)


pkg/scheduler/ugm/queue_tracker.go
```diff
@@ -410,11 +410,7 @@ func (qt *QueueTracker) canRunApp(hierarchy []string, applicationID string, trac
 func (qt *QueueTracker) canBeRemoved() bool {
 	for _, childQT := range qt.childQueueTrackers {
 		// quick check to avoid further traversal
-		if childQT.canBeRemovedInternal() {
-			if !childQT.canBeRemoved() {
```
ryankert01, line 414: Isn't it meant to be recursively?

ryankert01 (commented):
Hi, thank you for the patch. A little question left.

pkg/scheduler/ugm/queue_tracker.go
SP12893678: That place seems unreachable.

ryankert01: Could you verify it by adding test case with tree that depth >= 3?

ryankert01: Did you mean `len(qt.childQueueTrackers) == 0` cause this?

ryankert01: I see how this is going. This function can only remove tree that height=2.

pbacsko: I've already discussion about this code with @manirajv06 please involve him as well.

ryankert01: gentle ping @manirajv06 , are you available to give us some insight with this code?

craigcondit:
@manirajv06 can you take a look at this? Some of this logic looks suspect to me...

pkg/scheduler/ugm/queue_tracker.go
wilfred-s: The check works even for deeper trees. What is not supported is a recursive delete so that is what it prevents from happening.
root -> parent1 .to. parent5 -> child1 .to. child5 -> leaf1 .to. leaf5
The .to. signifies a repeat of the queues. So we have 5 parents under the root. Each with 5 child queues etc.
When we check for a removal we traverse the tree and would do a depth first check. To remove parent1 it has to be empty and have no quota set, that is the original call. Before we go all the way to the leaf at any point we check if the child queues at the current level can be removed.
If any child has a quota set, even without any usage, we cannot remove parent and would return false. 
I think the check `len(qt.childQueueTrackers) == 0` in `canBeRemovedInternal()` should not be there as well as it stops going down the tree. I think @SP12893678 and @ryankert01 made the correct remarks to start of the conversation. I think we need to fix that vunction and make it work recursively as it should.

wilfred-s:
closing stale PR
Before improving test coverage `canBeRemoved()`should be assessed to work correctly

PR #975, opened 2024-09-26
[YUNIKORN-2832] [core] Add non-Yunikorn allocation tracking logic

pbacsko:
### What is this PR for?
Add foreign pod tracking logic to the scheduler core.

**Note**: existing code which updates `occupiedResources` is not removed in this PR. Removal will be done as a follow-up task.
Updating the allocation resource usage be implemented in YUNIKORN-1765.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [x] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2832

### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation.go
+++ b/pkg/scheduler/objects/allocation.go
@@ -156,14 +156,6 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
 	}
 }
 
-// NewAllocationFromSIAllocated creates an Allocation where the "allocated" flag is always true,
-// regardless whehether the NodeID if empty or not. Used for testing.
-func NewAllocationFromSIAllocated(siAlloc *si.Allocation) *Allocation {
-	alloc := NewAllocationFromSI(siAlloc)
-	alloc.allocated = true
-	return alloc
-}
-
 // NewSIFromAllocation converts the Allocation into a SI object. This is a limited set of values that gets copied into
 // the SI. This is only used to communicate *back* to the RM. All other fields are considered incoming fields from
 // the RM into the core. The limited set of fields link the Allocation to an Application and Node.
--- a/pkg/scheduler/objects/allocation.go
+++ b/pkg/scheduler/objects/allocation.go
@@ -49,6 +49,8 @@ type Allocation struct {
 	originator        bool
 	tags              map[string]string
 	resKeyWithoutNode string // the reservation key without node
+	foreign           bool
+	preemptable       bool
 
 	// Mutable fields which need protection
 	allocated            bool
@@ -106,6 +108,20 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
 		createTime = time.Unix(siCreationTime, 0)
 	}
 
+	foreign := false
+	preemptable := true
+	if foreignType, ok := alloc.AllocationTags[siCommon.Foreign]; ok {
+		foreign = true
+		switch foreignType {
+		case siCommon.AllocTypeStatic:
+			preemptable = false
+		case siCommon.AllocTypeDefault:
+		default:
+			log.Log(log.SchedAllocation).Warn("Foreign tag has illegal value, using default",
+				zap.String("value", foreignType))
+		}
+	}
+
 	var allocated bool
 	var nodeID string
 	var bindTime time.Time
@@ -135,7 +151,17 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
 		allocated:         allocated,
 		nodeID:            nodeID,
 		bindTime:          bindTime,
+		foreign:           foreign,
+		preemptable:       preemptable,
 	}
+}
+
+// NewAllocationFromSIAllocated creates an Allocation where the "allocated" flag is always true,
+// regardless whehether the NodeID if empty or not. Used for testing.
+func NewAllocationFromSIAllocated(siAlloc *si.Allocation) *Allocation {
+	alloc := NewAllocationFromSI(siAlloc)
+	alloc.allocated = true
+	return alloc
 }
 
 // NewSIFromAllocation converts the Allocation into a SI object. This is a limited set of values that gets copied into
@@ -572,4 +598,12 @@ func (a *Allocation) setUserQuotaCheckPassed() {
 		a.userQuotaCheckFailed = false
 		a.askEvents.SendRequestFitsInUserQuota(a.allocationKey, a.applicationID, a.allocatedResource)
 	}
+}
+
+func (a *Allocation) IsForeign() bool {
+	return a.foreign
+}
+
+func (a *Allocation) IsPreemptable() bool {
+	return a.preemptable
 }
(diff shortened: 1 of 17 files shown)


pkg/scheduler/partition.go
```diff
@@ -1286,12 +1291,54 @@ func (pc *PartitionContext) UpdateAllocation(alloc *objects.Allocation) (request
...
 }
 
+// getOrSetNodeIDForAlloc returns the nodeID for a given foreign allocation, or sets is if it's unset
+func (pc *PartitionContext) getOrSetNodeIDForAlloc(allocKey, nodeID string) string {
+	pc.Lock()
+	defer pc.Unlock()
+	id := pc.foreignAllocs[allocKey]
+	if id != common.Empty {
+		return id
+	}
```
pbacsko, line 1337: Notes: lines 1316-1321, 1336 will be covered as part of YUNIKORN-1765.

pkg/scheduler/objects/allocation.go
```diff
@@ -135,9 +151,19 @@ func NewAllocationFromSI(alloc *si.Allocation) *Allocation {
...
 		nodeID:            nodeID,
 		bindTime:          bindTime,
+		foreign:           foreign,
+		preemptable:       preemptable,
 	}
 }
 
+// NewAllocationFromSIAllocated creates an Allocation where the "allocated" flag is always true,
+// regardless whehether the NodeID if empty or not. Used for testing.
+func NewAllocationFromSIAllocated(siAlloc *si.Allocation) *Allocation {
```
craigcondit, line 161: I understand the temptation to add this, but I'd rather we didn't.

pkg/scheduler/objects/node.go
```diff
@@ -243,6 +243,31 @@ func (sn *Node) GetAllAllocations() []*Allocation {
...
+func (sn *Node) GetForeignAllocations() []*Allocation {
+	sn.RLock()
+	defer sn.RUnlock()
+	return sn.getAllocations(true)
+}
+
+func (sn *Node) getAllocations(foreign bool) []*Allocation {
+	arr := make([]*Allocation, 0)
+	for _, v := range sn.allocations {
+		if (v.IsForeign() && foreign) || (!v.IsForeign() && !foreign) {
```
craigcondit, line 263: This can be simplified to: `if v.IsForeign() == foreign {`

pkg/scheduler/partition.go
```diff
@@ -1286,12 +1291,54 @@ func (pc *PartitionContext) UpdateAllocation(alloc *objects.Allocation) (request
...
+	return false, false, nil
+}
+
 func (pc *PartitionContext) convertUGI(ugi *si.UserGroupInformation, forced bool) (security.UserGroup, error) {
 	pc.RLock()
 	defer pc.RUnlock()
 	return pc.userGroupCache.ConvertUGI(ugi, forced)
 }
 
+// getOrSetNodeIDForAlloc returns the nodeID for a given foreign allocation, or sets is if it's unset
```
craigcondit, line 1330: How is this meant to be used?

pkg/scheduler/partition.go
```diff
@@ -1496,6 +1547,28 @@ func (pc *PartitionContext) removeAllocation(release *si.AllocationRelease) ([]*
 	return released, confirmed
 }
 
+func (pc *PartitionContext) removeForeignAllocation(allocID string) {
+	nodeID := pc.foreignAllocs[allocID]
```
craigcondit, line 1551: Do we really need all these warning cases?

craigcondit (changes requested):
A few minor things, and some questions.

pkg/scheduler/partition.go
pbacsko: Might not be necessary... perhaps convert them to DEBUG?

pkg/scheduler/partition.go
pbacsko: it's a combination of a get + set so we grab the lock only once, but if it's difficult to read, I can split it to two parts

pkg/scheduler/objects/allocation.go
pbacsko: Found a way to work it around

chenyulin0719 (commented):
Some questions.

pkg/scheduler/partition.go
```diff
@@ -1286,12 +1291,54 @@ func (pc *PartitionContext) UpdateAllocation(alloc *objects.Allocation) (request
...
+	allocated := alloc.IsAllocated()
+	if !allocated {
+		return false, false, fmt.Errorf("trying to add a foreign request (non-allocation) %s", allocationKey)
+	}
+	if alloc.GetNodeID() == "" {
+		return false, false, fmt.Errorf("node ID is empty for allocation %s", allocationKey)
+	}
+	if node == nil {
+		return false, false, fmt.Errorf("failed to find node %s for allocation %s", nodeID, allocationKey)
+	}
```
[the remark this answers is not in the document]
pbacsko: Correct, it's sent back and it needs to be handled properly.
Currently, this happens inside `scheduler_callback.go`:

```
	for _, reject := range response.RejectedAllocations {
		// request rejected by the scheduler, reject it
		log.Log(log.ShimRMCallback).Debug("callback: response to rejected allocation",
			zap.String("allocationKey", reject.AllocationKey))
		dispatcher.Dispatch(NewRejectTaskEvent(reject.ApplicationID, reject.AllocationKey,
			fmt.Sprintf("task %s allocation from application %s is rejected by scheduler",
				reject.AllocationKey, reject.ApplicationID)))
	}
```

This doesn't look right, we don't need a task reject event for a non-Yunikorn allocation. Therefore, we need changes in this code path (my feeling is that there's not much we can do other than logging something) but, this is yet to be done.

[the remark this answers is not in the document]
chenyulin0719: Agree, there is nothing we can do except logging.

 If we want to log the information back to the Pod (Pod event? or other places?), we need to determine whether "This rejected allocation is YuniKorn or Non-YuniKorn." How do we identify it?

Will we keep the Non-YuniKorn allocation ID in shim?
My previous question assumes we don't keep the Non-YuniKorn allocation ID list in shim.

> Besides, should we add a flag of allocation type to &si.RejectedAllocation?

 Therefor, we might need a flag from si.RejectedAllocation.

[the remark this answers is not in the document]
craigcondit: The only way we would *ever* reject a non-YK alloc in the core is if the node was not found. This would indicate a bug in the shim, and so there's no appropriate way to handle it. So logging loudly is probably sufficient.

pkg/scheduler/partition.go
craigcondit: Yeah, DEBUG seems good.

craigcondit (changes requested):
Looks good except for debug instead of warning in logging.

craigcondit (approved):
+1 LGTM.

chenyulin0719 (approved):
+1 LGTM

PR #976, opened 2024-10-01
[YUNIKORN-2895] Don't add duplicated allocation to node when the allo…

zhuqi-lucas:
…cation already allocated before.

When i try to revisit the new update allocation logic, the potential duplicated allocation to node will happen if the allocation already allocated.  And we try to add the allocation to the node again and don't revert it.

**Note:**

We have unwind logic for another side call for placeholder allocate:

https://github.com/apache/yunikorn-core/blob/fe067b770c25c238449e0b429da92400da8556a4/pkg/scheduler/objects/application.go#L1285

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-2895

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1594,12 +1594,24 @@ func (sa *Application) tryNode(node *Node, ask *Allocation) (*AllocationResult, 
 			node.RemoveAllocation(ask.GetAllocationKey())
 			return nil, nil
 		}
+
 		// mark this ask as allocated
+		// If the allocation already exists, we should not add it to the node again, it will potentially cause the orphan allocation issue.
 		_, err := sa.allocateAsk(ask)
 		if err != nil {
 			log.Log(log.SchedApplication).Warn("allocation of ask failed unexpectedly",
 				zap.Error(err))
+			// revert the node update
+			node.RemoveAllocation(ask.GetAllocationKey())
+			// revert the queue update
+			err = sa.queue.DecAllocatedResource(ask.GetAllocatedResource())
+			if err != nil {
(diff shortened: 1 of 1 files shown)


craigcondit (changes requested):
This will break horribly. If allocateAsk() fails, there is nothing to unwind as the allocation result won't be returned and the final allocation will it happen. Your logic will remove something that was not added in the first place.

zhuqi-lucas:
Thanks @craigcondit for review:
We have inconsistent logic for allocate failed, here is the call with unwind, so i am assuming it may wrong:
https://github.com/apache/yunikorn-core/blob/fe067b770c25c238449e0b429da92400da8556a4/pkg/scheduler/objects/application.go#L1285


It seems both two cases call TryNodeAllocation, so i am confused.

wilfred-s:
-1 on this approach. I think the issue is way more basic. I opened a discussion on the jira as I see multiple issues in the way now track allocations and asks.

zhuqi-lucas:
Thanks @wilfred-s @craigcondit , i will try do add unit test and check the comments, look into the root cause of those issues.

craigcondit:
The `allocateAsk()` call can only fail if the ask was previously allocated. This is checked for in all code paths that call `allocateAsk()` earlier, and always with the application lock held. Therefore, if we ever hit the error case here, we have a serious BUG elsewhere.

craigcondit:
Closing this after determining that original issue is not a problem.

PR #983, opened 2024-10-14
[YUNIKORN-2848] Refactor preemption_queue_test.go

ryankert01:
### What is this PR for?
Refactor TestGetPreemptableResource test based on variables and syntax constructs used in TestGetRemainingGuaranteedResource.

For example, variables defined for test resources like res1, res2 etc and t.Run({})

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2848

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption_queue_test.go
+++ b/pkg/scheduler/objects/preemption_queue_test.go
@@ -65,6 +65,7 @@ func setupQueueStructure(t *testing.T) (*Queue, *Queue, *Queue, *Queue) {
 	return rootQ, parentQ, childQ1, childQ2
 }
 
+//nolint:funlen
 func TestGetPreemptableResource(t *testing.T) {
 	tests := []struct {
 		testName   string


pkg/scheduler/objects/preemption_queue_test.go
```diff
@@ -53,101 +53,156 @@ type res struct {
...
+		resources res
+	}{
+		{
+			// no guaranteed and no usage. so nothing to preempt
+			testName: "NoGuaranteedNoUsage",
+			resources: res{
+				guaranteed:     resource{nil, nil, nil, nil},
+				allocated:      allocated1,
+				remaining:      preemptable1,
+				assertMessages: assertMessageStruct1,
```
craigcondit, line 103: This is ugly. It's difficult to see what is being passed in. Even though there's a bit of code duplication here, setting the values explicitly here makes the tests clearer.

pkg/scheduler/objects/preemption_queue_test.go
```diff
@@ -53,101 +53,156 @@ type res struct {
...
+				remaining:      preemptable4,
+				assertMessages: assertMessageStruct4,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			rootQ, parentQ, childQ1, childQ2 := setupQueueStructure(t)
+			assertPreemptable(t, rootQ, parentQ, childQ1, childQ2, tt.resources)
```
craigcondit, line 168: Don't extract assertPreemptable() here, but leave the logic inline. This makes it *much* easier to see where failures occur, as stack traces will show this instead of the generic line in assertPreemptable().

pkg/scheduler/objects/preemption_queue_test.go
ryankert01: Make sense, Done.

pkg/scheduler/objects/preemption_queue_test.go
ryankert01: Agree, Done.

craigcondit (approved):
+1 LGTM.

PR #991, opened 2024-11-01
[YUNIKORN-2953] Placeholder release count incorrect

wilfred-s:
### What is this PR for?
While releasing placeholders the allocated placeholders are counted twice in the tracking information. With YUNIKORN-2926 in place this happens if some but not all placeholders are allocated only.

Mark unallocated placeholders that timeout as released to prevent scheduling issues.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* [YUNIKORN-2953](https://issues.apache.org/jira/browse/YUNIKORN-2953)

This fixes should make a failing e2e test pass: gangscheduling_test


craigcondit (approved):
+1 LGTM.

PR #993, opened 2024-11-13
[YUNIKORN-2972] Remove internal object from user and group REST info

wilfred-s:
### What is this PR for?
Remove the Resource object from the REST response for the user and group trackers. Removed the usage of the DAO objects in the internal unit tests. The unit tests use new functions which directly expose the same structures without using the DAO.

Cleanup:
- remove export from internal functions (UserTracker, GroupTracker)
- renamed functions to match their results (Manager)
- update handler to call renamed functions in Manager
- reimplemented test functions to not use DAO objects (Manager, UserTracker, GroupTracker, QueueTracker)
- assert function changes to use new functions removing DAO usage (scheduler/utilites_test, scheduler/objects/utilites_test)

### What type of PR is it?
* [X] - Bug Fix
* [X] - Improvement
* [X] - Refactoring

### What is the Jira issue?
* [YUNIKORN-2972](https://issues.apache.org/jira/browse/YUNIKORN-2972)

Existing unit test cover the code

NOTE: this will break e2e test, the `user_group_limit_test` uses the DAO directly

### Questions:
* [X] - There is breaking changes for older versions.

### DIFF
--- a/pkg/scheduler/objects/utilities_test.go
+++ b/pkg/scheduler/objects/utilities_test.go
@@ -284,19 +284,23 @@ func getUserGroup(userName string, groupNameList []string) security.UserGroup {
 }
 
 func assertUserGroupResource(t *testing.T, userGroup security.UserGroup, expected *resources.Resource) {
-	ugm := ugm.GetUserManager()
-	userResource := ugm.GetUserResources(userGroup)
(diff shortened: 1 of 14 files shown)


craigcondit (approved):
+1 LGTM.

PR #995, opened 2024-11-15
[YUNIKORN-2974] Expose preemption and priority settings via REST

wilfred-s:
### What is this PR for?
Add preemption and priority details to the queue REST information. Currently the configuration is exposed via the properties. The properties contain the raw configuration values which need to be interpreted.
Invalid configuration values are ignored and not set on the queue. This could cause an incorrect view of what is or should be active on the queue if based on the properties. Using the active values from the object to show the real state.

Expose Partition level preemption flag as part of this change to complement the queue details.

### What type of PR is it?
* [X] - Improvement

### What is the Jira issue?
[YUNIKORN-2976](https://issues.apache.org/jira/browse/YUNIKORN-2976)

new unit tests added

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -696,7 +696,7 @@ func (sq *Queue) GetPartitionQueueDAOInfo(include bool) dao.PartitionQueueDAOInf
 	queueInfo.AbsUsedCapacity = resources.CalculateAbsUsedCapacity(sq.maxResource, sq.allocatedResource).DAOMap()
 	queueInfo.SortingPolicy = sq.sortType.String()
 	queueInfo.PrioritySorting = sq.prioritySortEnabled
-	queueInfo.PreemptionEnabled = sq.preemptionPolicy == policies.DisabledPreemptionPolicy
+	queueInfo.PreemptionEnabled = sq.preemptionPolicy != policies.DisabledPreemptionPolicy
 	queueInfo.IsPreemptionFence = sq.preemptionPolicy == policies.FencePreemptionPolicy
 	queueInfo.PreemptionDelay = sq.preemptionDelay.String()
 	queueInfo.IsPriorityFence = sq.priorityPolicy == policies.FencePriorityPolicy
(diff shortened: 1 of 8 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -693,8 +693,14 @@ func (sq *Queue) GetPartitionQueueDAOInfo(include bool) dao.PartitionQueueDAOInf
 	queueInfo.IsManaged = sq.isManaged
 	queueInfo.CurrentPriority = sq.getCurrentPriority()
 	queueInfo.TemplateInfo = sq.template.GetTemplateInfo()
-	queueInfo.AbsUsedCapacity = resources.CalculateAbsUsedCapacity(
-		sq.maxResource, sq.allocatedResource).DAOMap()
+	queueInfo.AbsUsedCapacity = resources.CalculateAbsUsedCapacity(sq.maxResource, sq.allocatedResource).DAOMap()
+	queueInfo.SortingPolicy = sq.sortType.String()
+	queueInfo.PrioritySorting = sq.prioritySortEnabled
+	queueInfo.PreemptionEnabled = sq.preemptionPolicy == policies.DisabledPreemptionPolicy
```
brandboat, line 699: I am a bit misleading here, does `PreemptionEnabled = true` means the preemptionPolicy is disabled ?

craigcondit: I believe this should be `sq.preemptionPolicy != policies.DisabledPreemptionPolicy`

craigcondit (changes requested):
Looks good except for inverted logic on PreemptionEnabled.

pbacsko (approved):
LGTM after the "preemptionEnabled" change

pkg/scheduler/objects/queue.go
wilfred-s: correct fixed in the next commit

brandboat (approved):
LGTM, thanks !

pbacsko (approved):

craigcondit (approved):
+1 LGTM.

PR #996, opened 2024-11-15
[YUNIKORN-2978] Fix handling of reserved allocations where node differs

craigcondit:
### What is this PR for?
YUNIKORN-2700 introduced a bug where allocations of previously-reserved tasks were not handled correctly in the case where we schedule on a different node than the reservation. Ensure that we unreserve and allocate using the proper node in both cases.

Also introduce additional logging of allocations on nodes to make finding issues like this easier in the future.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2978

Verified successful processing of 1000-pod job on autoscaled cluster where previously this would fail.

### Questions:

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -871,17 +871,6 @@ func (pc *PartitionContext) allocate(result *objects.AllocationResult) *objects.
 	// the node ID is set when a reservation is allocated on a non-reserved node
 	alloc := result.Request
 	targetNodeID := result.NodeID
-	var reservedNodeID string
-	if result.ReservedNodeID == "" {
-		reservedNodeID = result.NodeID
-	} else {
-		reservedNodeID = result.ReservedNodeID
-		log.Log(log.SchedPartition).Debug("Reservation allocated on different node",
-			zap.String("current node", result.NodeID),
-			zap.String("reserved node", reservedNodeID),
-			zap.String("appID", appID))
-	}
-
 	targetNode := pc.GetNode(targetNodeID)
 	if targetNode == nil {
 		log.Log(log.SchedPartition).Info("Target node was removed while allocating",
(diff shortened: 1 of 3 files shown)


pbacsko (commented):
Is it possible to have at least a single unit test that fails with the old code and passes with this PR?

pkg/scheduler/partition.go
```diff
@@ -869,44 +869,54 @@ func (pc *PartitionContext) allocate(result *objects.AllocationResult) *objects.
 	// find the node make sure it still exists
 	// if the node was passed in use that ID instead of the one from the allocation
 	// the node ID is set when a reservation is allocated on a non-reserved node
-	var nodeID string
 	alloc := result.Request
+	targetNodeID := result.NodeID
+	var srcNodeID string
```
wilfred-s, line 874: Rename to `reservedNodeID` is more logical than `srcNodeID`
The reserved node is only used if the result is `Unreserved` or `AllocatedReserved` This retrieval and the logging should be moved inside that check (line 911) and not clutter the rest of the code.

wilfred-s (changes requested):

wilfred-s:
> Is it possible to have at least a single unit test that fails with the old code and passes with this PR?

I think the missing bit is just a single line:
```
928       alloc.SetNodeID(targetNodeID)
```
We need a unit tests, and it should be doable to create one:
* fill up a node with allocations
* create a request that does not fit on the used node
* manually create a reservation for that request on that filled up node
* run the normal allocation and get it to allocate on the "other" node.
* the new allocation should show the correct node.

Before the fix the allocation will show the reserved node ID or none at all.

craigcondit:
Addressed review comments. Reservation test updated to verify node assignment -- verified this test fails prior to this PR but passes now.

pbacsko (approved):
LGTM

PR #998, opened 2024-11-20
[YUNIKORN-2980] DaemonSet preemption: don't flood the logs if victim selection fails

pbacsko:
### What is this PR for?
We generate too much output if requiredNode preemption is not successful. Instead, do the following:
1. Add an entry to the allocation log (or increment the counter)
2. Send a REQUEST event to the DS pod in every 15 seconds
3. Only log if we can successfully preempt a victim

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2977

### Questions:


pkg/common/errors.go
```diff
@@ -27,3 +27,4 @@ const PreemptionPreconditionsFailed = "Preemption preconditions failed"
 const PreemptionDoesNotGuarantee = "Preemption queue guarantees check failed"
 const PreemptionShortfall = "Preemption helped but short of resources"
 const PreemptionDoesNotHelp = "Preemption does not help"
+const NoVictimForRequiredNode = "RequiredNode preemption failed, no victim found"
```
craigcondit, line 30: Can we improve the wording to not use internal camelcase names? Maybe "No fit on required node, preemption does not help" or something?

craigcondit (changes requested):
Mostly good, minor nit on message wording.

craigcondit (approved):
+1 LGTM, pending test run.

PR #1000, opened 2024-11-25
[YUNIKORN-2982] Send event when preemption occurs

rhh777:
### What is this PR for?
When a preemption occurs, send an event to be displayed in the victim's event list.

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
[YUNIKORN-2982](https://issues.apache.org/jira/browse/YUNIKORN-2982)

Trigger preemption

### Screenshots
image

### Questions:
* [x] - At present, the default interval for event sending is 2 seconds. When sending a preemptive event, the task may have already been removed, triggering an exception that the corresponding task cannot be found, resulting in the failure of event sending.
Narrowing the time interval can be a temporary solution, but I haven't thought of a more suitable solution at the moment.
image

### DIFF
--- a/pkg/scheduler/objects/events/application_events.go
+++ b/pkg/scheduler/objects/events/application_events.go
@@ -79,14 +79,6 @@ func (ae *ApplicationEvents) SendRemoveAllocationEvent(appID, allocKey string, a
 	ae.eventSystem.AddEvent(event)
 }
 
-func (ae *ApplicationEvents) SendPreemptAllocationEvent(appID, allocKey string, allocated *resources.Resource, message string) {
-	if !ae.eventSystem.IsEventTrackingEnabled() {
-		return
-	}
-	event := events.CreateAppEventRecord(appID, message, allocKey, si.EventRecord_REMOVE, si.EventRecord_ALLOC_PREEMPT, allocated)
-	ae.eventSystem.AddEvent(event)
-}
-
 func (ae *ApplicationEvents) SendRemoveAskEvent(appID, allocKey string, allocated *resources.Resource, detail si.EventRecord_ChangeDetail) {
 	if !ae.eventSystem.IsEventTrackingEnabled() {
 		return
--- a/pkg/scheduler/objects/events/application_events.go
+++ b/pkg/scheduler/objects/events/application_events.go
@@ -79,6 +79,14 @@ func (ae *ApplicationEvents) SendRemoveAllocationEvent(appID, allocKey string, a
 	ae.eventSystem.AddEvent(event)
 }
 
+func (ae *ApplicationEvents) SendPreemptAllocationEvent(appID, allocKey string, allocated *resources.Resource, message string) {
+	if !ae.eventSystem.IsEventTrackingEnabled() {
+		return
+	}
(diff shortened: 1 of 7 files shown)


pkg/scheduler/objects/events/application_events.go
```diff
@@ -79,6 +79,14 @@ func (ae *ApplicationEvents) SendRemoveAllocationEvent(appID, allocKey string, a
 	ae.eventSystem.AddEvent(event)
 }
 
+func (ae *ApplicationEvents) SendPreemptAllocationEvent(appID, allocKey string, allocated *resources.Resource, message string) {
+	if !ae.eventSystem.IsEventTrackingEnabled() {
+		return
+	}
+	event := events.CreateAppEventRecord(appID, message, allocKey, si.EventRecord_REMOVE, si.EventRecord_ALLOC_PREEMPT, allocated)
```
pbacsko, line 86: This must be a `REQUEST` type. That will result in a real K8s event being sent to the target pod.

The removal of the allocation will be a natural consequence of that, which will flow through the scheduler, so an app-specific event is not needed here.

pbacsko (changes requested):
See comments.

pkg/scheduler/objects/preemption.go
```diff
@@ -635,6 +636,9 @@ func (p *Preemptor) TryPreemption() (*AllocationResult, bool) {
 				zap.String("victimNodeID", victim.GetNodeID()),
 				zap.String("victimQueue", victimQueue.Name),
 			)
+			// send event
+			message := fmt.Sprintf("Preempted by %v from application %v in %v", p.ask.GetTag("kubernetes.io/meta/podName"), p.ask.applicationID, p.application.queuePath)
```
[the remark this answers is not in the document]
pbacsko: I looked at `preemption_test.go`, the simplest test case is `TestTryPreemption`. You can add the extra code there to verify the event.

rhh777:
> See comments.

Event has been switched to Request, please continue reviewing.

pkg/scheduler/objects/preemption.go
```diff
@@ -635,6 +635,8 @@ func (p *Preemptor) TryPreemption() (*AllocationResult, bool) {
 				zap.String("victimNodeID", victim.GetNodeID()),
 				zap.String("victimQueue", victimQueue.Name),
 			)
+			// send event
```
pbacsko, line 638: very nit: unnecessary comment

pbacsko:
Overall looks good, added @wilfred-s to to have his opinion too.

pkg/scheduler/objects/preemption.go
rhh777: Updated.

pbacsko (approved):
LGTM

wilfred-s:
> This must be a type of REQUEST. That will result in a real K8s event being sent to the target pod.
>
> The removal of the allocation will be a natural consequence of that, which will flow through the scheduler, so an app-specific event is not needed here.

I can see that an application specific event is already generated as part of `removeAllocationInternal()` that handles the overall tracking nicely.
This change adds the missing piece by generating an event on the pod. Based on that I think we're good .

wilfred-s (approved):
LGTM +1

PR #1001, opened 2024-12-03
[YUNIKORN-2976] Handle multiple require node allocations per node

wilfred-s:
### What is this PR for?
If an allocation requires a specific node the scheduler should not consider any other node. We should allow multiple allocations that require the same node to reserve the node at the same time. A required node allocation must be placed on the node before anything else. If other non required node reservations are made on a node remove the existing reservations that do not require that node. Make sure that the releases are tracked correctly in the partition.

After the repeat count removal reservations can be simplified:
- track reservations using the allocation key
- removed the composite key setup
- removed collection listener call on reserve or unreserve of a node

Cleanup testing of the node collection

### What type of PR is it?
* [X] - Bug Fix
* [X] - Improvement

### Todos
 * [X] e2e test in the shim requesting multiple daemon sets at once

### What is the Jira issue?
[YUNIKORN-2976](https://issues.apache.org/jira/browse/YUNIKORN-2976)

unit tests are updated new e2e test should be added

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -796,20 +796,15 @@ func (sa *Application) HasReserved() bool {
 	return len(sa.reservations) > 0
 }
 
-// IsReservedOnNode returns true if and only if the node has been reserved by the application
-// An empty nodeID is never reserved.
-func (sa *Application) IsReservedOnNode(nodeID string) bool {
-	if nodeID == "" {
-		return false
-	}
+// NodeReservedForAsk returns the nodeID that has been reserved by the application for the ask
+// An empty nodeID means the ask is not reserved. An empty askKey is never reserved.
+func (sa *Application) NodeReservedForAsk(askKey string) string {
 	sa.RLock()
 	defer sa.RUnlock()
-	for _, reserved := range sa.reservations {
-		if reserved.nodeID == nodeID {
-			return true
-		}
+	if reserved, ok := sa.reservations[askKey]; ok {
+		return reserved.nodeID
 	}
-	return false
+	return ""
 }
 
 // Reserve the application for this node and alloc combination.
@@ -923,19 +918,6 @@ func (sa *Application) unReserveInternal(reserve *reservation) int {
 	return 0
 }
 
-// IsAllocationReserved returns true if the allocation has been reserved
-func (sa *Application) IsAllocationReserved(allocKey string) bool {
-	sa.RLock()
-	defer sa.RUnlock()
-	return sa.reservations[allocKey] != nil
-}
-
-// getAllocationReservation returns the reservation object for the allocation
-// No locking must be called while holding the lock
-func (sa *Application) getAllocationReservation(allocKey string) *reservation {
-	return sa.reservations[allocKey]
-}
-
 // canAllocationReserve Check if the allocation has already been reserved. An alloc can never reserve more than one node.
 // No locking must be called while holding the lock
 func (sa *Application) canAllocationReserve(alloc *Allocation) error {
@@ -945,7 +927,7 @@ func (sa *Application) canAllocationReserve(alloc *Allocation) error {
 			zap.String("allocationKey", allocKey))
 		return common.ErrorReservingAlloc
 	}
-	reserved := sa.getAllocationReservation(allocKey)
+	reserved := sa.reservations[allocKey]
 	if reserved != nil {
 		log.Log(log.SchedApplication).Debug("reservation already exists",
 			zap.String("allocKey", allocKey),
@@ -1100,12 +1082,13 @@ func (sa *Application) tryRequiredNode(request *Allocation, getNodeFn func(strin
 	if len(reservations) > 0 {
 		num = sa.cancelReservations(reservations)
 	}
+	_, thisReserved := sa.reservations[allocationKey]
 	// now try the request, we don't care about predicate error messages here
 	result, _ := sa.tryNode(node, request) //nolint:errcheck
 	if result != nil {
 		result.CancelledReservations = num
 		// check if the node was reserved and we allocated after a release
-		if _, ok := sa.reservations[allocationKey]; ok {
+		if thisReserved {
 			log.Log(log.SchedApplication).Debug("allocation on required node after release",
 				zap.String("appID", sa.ApplicationID),
 				zap.String("nodeID", requiredNode),
@@ -1118,6 +1101,11 @@ func (sa *Application) tryRequiredNode(request *Allocation, getNodeFn func(strin
 			zap.String("allocationKey", allocationKey),
 			zap.Stringer("resultType", result.ResultType))
 		return result
+	}
+	// if this ask was already reserved we should not have deleted any reservations
+	// we also do not need to send back a reservation result and just return nil to check the next ask
+	if thisReserved {
+		return nil
 	}
 	result = newReservedAllocationResult(node.NodeID, request)
 	result.CancelledReservations = num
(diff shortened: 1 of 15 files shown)


pbacsko (commented):
Wow, nice change :)

First, I just have a general observation: there seems to be too many uncovered paths in `application.go`. I'm sure some of those already existed before, but this change touches critical part of the allocation cycle, so I'd be happy to see those codecov reports go away, like:

#L883 - L887
#L944 - L947
#L1106 - L1120
#L1143
#L1241 (this looks weird, because the flow continues)
#L1317 (looks weird too)
#L1362

Perhaps they're covered indirectly from partition, but it's hard to reason about it.

pkg/scheduler/objects/reservation.go
```diff
@@ -61,51 +60,21 @@ func newReservation(node *Node, app *Application, ask *Allocation, appBased bool
...
+	if r == nil {
+		return "nil reservation"
+	}
 	if r.nodeID == "" {
-		return r.node.NodeID + " -> " + r.appID + "|" + r.askKey
+		return r.node.NodeID + " -> " + r.appID + "|" + r.allocKey
 	}
-	return r.app.ApplicationID + " -> " + r.nodeID + "|" + r.askKey
+	return r.app.ApplicationID + " -> " + r.nodeID + "|" + r.allocKey
 }
```
pbacsko, line 71: This is so interesting here. If `nodeID` == "" then we just take the it from `r.node.NodeID` anyway.  To me it looks like we're storing a string value for boolean purposes (we should store `appBased` instead), but it's not clear to me what happens here. Also `r.app.ApplicationID` could be just `r.appID`, correct?

pkg/scheduler/objects/reservation.go
```diff
@@ -21,37 +21,36 @@ package objects
...
+	node  *Node
+	alloc *Allocation
 }
 
 // The reservation inside the scheduler. A reservation object is never mutated and does not use locking.
 // The key depends on where the reservation was made (node or app).
 // appBased must be true for a reservation for an app and false for a reservation on a node
-func newReservation(node *Node, app *Application, ask *Allocation, appBased bool) *reservation {
-	if ask == nil || app == nil || node == nil {
+func newReservation(node *Node, app *Application, alloc *Allocation, appBased bool) *reservation {
```
pbacsko, line 41: I know it's a matter of semantics, but can we call this allocation at this point? It's funny because now we have the same object for both an ask and alloc... This occurs multiple places in this patch. Don't want to nitpick on this one, just wondering which naming convention is better.

pkg/scheduler/objects/application.go
```diff
@@ -816,144 +804,155 @@ func (sa *Application) IsReservedOnNode(nodeID string) bool {
...
-		return reservationKeys
-	}
-	for key := range sa.reservations {
-		if strings.HasSuffix(key, allocKey) {
-			reservationKeys = append(reservationKeys, key)
-		}
-	}
-	return reservationKeys
+func (sa *Application) getAllocationReservation(allocKey string) *reservation {
+	return sa.reservations[allocKey]
```
pbacsko, line 936: Do we need this method? It's just a map lookup. It's only used once in the prod code.

pbacsko (commented):
Some more comments after round #1.

pkg/scheduler/objects/reservation.go
craigcondit: My personal view on this: Use whatever name makes sense given the context. If it's known we're dealing with an unsatisified allocation, ask seems fine. It helps give some context. In the case of reservations, I think ask is correct, since once we satisfy an allocation, it shouldn't interact further with reservations.

craigcondit: I don't want to be pedantic either way though.

craigcondit:
I've installed this patch on a cluster where I've spun up a kwok-based autoscaler. I encountered a scenario where a monitoring pod (daemonset member) wouldn't schedule on a full node by triggering preemption.  The logs ping-pong back and forth between these two messages now:

```
2024-12-03T23:37:20.251Z    INFO    core.scheduler.queue    objects/queue.go:1438    allocation found on queue    {"queueName": "root.monitoring", "appID": "monitoring-ac69ed37-d4d7-4c0f-92bd-8f89a16cec74", "resultType": "Reserved", "allocation": "allocationKey ac69ed37-d4d7-4c0f-92bd-8f89a16cec74, applicationID monitoring-ac69ed37-d4d7-4c0f-92bd-8f89a16cec74, Resource map[pods:1], Allocated false"}
2024-12-03T23:37:20.251Z    INFO    core.scheduler.partition    scheduler/partition.go:957    Application is already reserved on node    {"appID": "monitoring-ac69ed37-d4d7-4c0f-92bd-8f89a16cec74", "nodeID": "kind-worker-6fhtj"}
``` 

Events for the pod:

```
Normal  Scheduling     4m13s  yunikorn  monitoring/kube-prometheus-stack-prometheus-node-exporter-5b6p7 is queued and waiting for allocation
Normal  Informational  4m12s  yunikorn  Unschedulable request 'ac69ed37-d4d7-4c0f-92bd-8f89a16cec74' with required node 'kind-worker-6fhtj', no preemption victim found
Normal  Scheduling     4m5s   yunikorn  monitoring/kube-prometheus-stack-prometheus-node-exporter-5b6p7 is queued and waiting for allocation
```

I'm using `admissionController.filtering.generateUniqueAppId: true` to use per-pod App ID generation, if that makes a difference. Restarting YK doesn't seem to help.

pkg/scheduler/objects/reservation.go
wilfred-s: if the `nodeID` stored in the reservation is not an empty string the `appID` stored is an empty string. One of the two is always empty. Instead of testing twice we use that to only test once and return the correct details for both cases.
It is the only way we can distinguish between the allocation belonging to either the application or the node object. We could have further simplified it by using a boolean and always get the IDs directly out of the linked objects. That can still be done without a large impact on the code. Might change some logging entries...

pkg/scheduler/objects/application.go
wilfred-s: removed

wilfred-s:
Sorry for the late response:

> The logs ping-pong back and forth between these two messages now:
>
> ```
> 2024-12-03T23:37:20.251Z    INFO    core.scheduler.queue    objects/queue.go:1438    allocation found on queue    {"queueName": "root.monitoring", "appID": "monitoring-ac69ed37-d4d7-4c0f-92bd-8f89a16cec74", "resultType": "Reserved", "allocation": "allocationKey ac69ed37-d4d7-4c0f-92bd-8f89a16cec74, applicationID monitoring-ac69ed37-d4d7-4c0f-92bd-8f89a16cec74, Resource map[pods:1], Allocated false"}
> 2024-12-03T23:37:20.251Z    INFO    core.scheduler.partition    scheduler/partition.go:957    Application is already reserved on node    {"appID": "monitoring-ac69ed37-d4d7-4c0f-92bd-8f89a16cec74", "nodeID": "kind-worker-6fhtj"}
> ```
One function was not updated to allow multiple daemon set pods to reserve multiple nodes. A test case was missing for that unit test also.

wilfred-s:
filed two follow up jiras to add unit tests to tryNodesNoReserve and tryPlaceholderAllocate on the application

craigcondit (approved):
+1 LGTM. I've verified the updated patch on a cluster with a kwok autoscaler, and have been unable to reproduce the original issues. This looks good.

PR #1002, opened 2024-12-06
[YUNIKORN-2991] The queue in Draining state does not accept new applications

rhh777:
### What is this PR for?
When the queue is in the Drafting state, new applications are still allowed to be scheduled after submission. 
I think we should refuse.

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2991

Unit tests included

### DIFF
--- a/pkg/scheduler/placement/placement_test.go
+++ b/pkg/scheduler/placement/placement_test.go
@@ -330,7 +330,7 @@ partitions:
 	err = man.PlaceApplication(app)
 	queueName = app.GetQueuePath()
 	if err == nil || queueName != "" {
-		t.Errorf("Cannot Placing application in draining queue: '%s', error: %v", queueName, err)
+		t.Errorf("draining queue: app should not have been placed, queue: '%s', error: %v", queueName, err)
 	}
 }
 
--- a/pkg/scheduler/placement/placement_test.go
+++ b/pkg/scheduler/placement/placement_test.go
@@ -323,6 +323,15 @@ partitions:
 	if err == nil || queueName != "" {
 		t.Errorf("parent queue: app should not have been placed, queue: '%s', error: %v", queueName, err)
 	}
+
+	// provided rule (2nd): submit to draining queue
+	app = newApplication("app1", "default", "root.testparent.testchild", user, nil, nil, "")
+	man.queueFn("root.testparent.testchild").MarkQueueForRemoval()
+	err = man.PlaceApplication(app)
+	queueName = app.GetQueuePath()
+	if err == nil || queueName != "" {
+		t.Errorf("Cannot Placing application in draining queue: '%s', error: %v", queueName, err)
+	}
 }
 
 //nolint:funlen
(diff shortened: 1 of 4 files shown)


wilfred-s:
See my comment in the jira, can we investigate doing the check in the AppPlacementManager?

Sanjayir:
that great

rhh777:
I have made the suggested modifications, and now the rejected information in the pod event information is not very clear. Should we display more rejection information?

Before modification
image
After modification
image

> See my comment in the jira, can we investigate doing the check in the AppPlacementManager?

pkg/scheduler/placement/placement_test.go
```diff
@@ -323,6 +323,15 @@ partitions:
...
 		t.Errorf("parent queue: app should not have been placed, queue: '%s', error: %v", queueName, err)
 	}
+
+	// provided rule (2nd): submit to draining queue
+	app = newApplication("app1", "default", "root.testparent.testchild", user, nil, nil, "")
+	man.queueFn("root.testparent.testchild").MarkQueueForRemoval()
+	err = man.PlaceApplication(app)
+	queueName = app.GetQueuePath()
+	if err == nil || queueName != "" {
+		t.Errorf("Cannot Placing application in draining queue: '%s', error: %v", queueName, err)
```
wilfred-s, line 333: Clarify text by replacing `Cannot Placing application in draining queue`
with: `draining queue: app should not have been placed, queue: '%s', error: %v`

wilfred-s (changes requested):

wilfred-s:
> I have made the suggested modifications, and now the rejected information in the pod event information is not very clear. Should we display more rejection information?

I think that is a problem on the shim side. When I look at the processing of the rejection there it does this:
```
func (app *Application) handleRejectApplicationEvent(reason string) {
	log.Log(log.ShimCacheApplication).Info("app is rejected by scheduler", zap.String("appID", app.applicationID))
	// for rejected apps, we directly move them to failed state
	dispatcher.Dispatch(NewFailApplicationEvent(app.applicationID,
		fmt.Sprintf("%s: %s", constants.ApplicationRejectedFailure, reason)))
}
```
It skips the rejected state and goes to failed. That is something we can follow up separately. I think the code change beside the two nits on the core side is looking good now.

pkg/scheduler/placement/placement_test.go
rhh777: done

rhh777:
> > I have made the suggested modifications, and now the rejected information in the pod event information is not very clear. Should we display more rejection information?
> 
> I think that is a problem on the shim side. When I look at the processing of the rejection there it does this:
> 
> ```
> func (app *Application) handleRejectApplicationEvent(reason string) {
> 	log.Log(log.ShimCacheApplication).Info("app is rejected by scheduler", zap.String("appID", app.applicationID))
> 	// for rejected apps, we directly move them to failed state
> 	dispatcher.Dispatch(NewFailApplicationEvent(app.applicationID,
> 		fmt.Sprintf("%s: %s", constants.ApplicationRejectedFailure, reason)))
> }
> ```
> 
> It skips the rejected state and goes to failed. That is something we can follow up separately. I think the code change beside the two nits on the core side is looking good now.

rhh777:
@wilfred-s Please continue to review 😁

craigcondit (approved):
+1 LGTM.

PR #1004, opened 2024-12-18
[YUNIKORN-2997] add unit test for tryPlaceholderAllocate

kousei47747:
### What is this PR for?
Add unit test for tryPlaceholderAllocate, including 4 test cases:
- No placeholder is allocated
- Requested resource in the real allocation is smaller than the placeholder
- Requested resource in the real allocation is larger than the placeholder
- Allocation request has a different task group than the placeholder

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2997

### DIFF
--- a/pkg/scheduler/objects/application_test.go
+++ b/pkg/scheduler/objects/application_test.go
@@ -3438,3 +3438,130 @@ func TestApplication_canAllocationReserve(t *testing.T) {
 		})
 	}
 }
+
+func TestTryPlaceHolderAllocateNoPlaceHolders(t *testing.T) {
+	const tg1 = "tg-1"
+
+	node := newNode(nodeID1, map[string]resources.Quantity{"first": 5})
+	nodeMap := map[string]*Node{nodeID1: node}
+	iterator := getNodeIteratorFn(node)
+	getNode := func(nodeID string) *Node {
+		return nodeMap[nodeID]
+	}
+
+	app := newApplication(appID0, "default", "root.default")
+
+	queue, err := createRootQueue(nil)
+	assert.NilError(t, err, "queue create failed")
+	app.queue = queue
+
+	res := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 5})
+	ask := newAllocationAsk(aKey, appID0, res)
+	ask.taskGroupName = tg1
+	err = app.AddAllocationAsk(ask)
+	assert.NilError(t, err, "ask should have been added to app")
+
+	result := app.tryPlaceholderAllocate(iterator, getNode)
+	assert.Assert(t, result == nil, "result should be nil since there are no placeholders to allocate")
+}
+
+func TestTryPlaceHolderAllocateSmallerRequest(t *testing.T) {
+	const tg1 = "tg-1"
+
+	node := newNode(nodeID1, map[string]resources.Quantity{"first": 5})
+	nodeMap := map[string]*Node{nodeID1: node}
+	iterator := getNodeIteratorFn(node)
+	getNode := func(nodeID string) *Node {
+		return nodeMap[nodeID]
+	}
+
+	app := newApplication(appID0, "default", "root.default")
+
+	queue, err := createRootQueue(nil)
+	assert.NilError(t, err, "queue create failed")
+	app.queue = queue
+
+	res := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 5})
+	ph := newPlaceholderAlloc(appID0, nodeID1, res, tg1)
+	app.AddAllocation(ph)
+	app.addPlaceholderData(ph)
+	assertPlaceholderData(t, app, tg1, 1, 0, 0, res)
+
+	// allocation request is smaller than placeholder
+	smallerRes := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 1})
+	ask := newAllocationAsk(aKey, appID0, smallerRes)
+	ask.taskGroupName = tg1
+	err = app.AddAllocationAsk(ask)
+	assert.NilError(t, err, "ask should have been added to app")
+
+	result := app.tryPlaceholderAllocate(iterator, getNode)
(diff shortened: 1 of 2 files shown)


pkg/scheduler/objects/application_test.go
```diff
@@ -3438,3 +3438,130 @@ func TestApplication_canAllocationReserve(t *testing.T) {
...
+
+	// allocation request is smaller than placeholder
+	smallerRes := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 1})
+	ask := newAllocationAsk(aKey, appID0, smallerRes)
+	ask.taskGroupName = tg1
+	err = app.AddAllocationAsk(ask)
+	assert.NilError(t, err, "ask should have been added to app")
+
+	result := app.tryPlaceholderAllocate(iterator, getNode)
+	assert.Assert(t, result != nil, "result should not be nil since the ask is smaller than the placeholder")
```
wilfred-s, line 3498: Extend the checks of the result to be more than just not nil: what is the type returned etc.
What happens with the placeholder data?

pkg/scheduler/objects/application_test.go
```diff
@@ -3438,3 +3438,130 @@ func TestApplication_canAllocationReserve(t *testing.T) {
...
+
+	// allocation request is larger than placeholder
+	largerRes := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 10})
+	ask := newAllocationAsk(aKey, appID0, largerRes)
+	ask.taskGroupName = tg1
+	err = app.AddAllocationAsk(ask)
+	assert.NilError(t, err, "ask should have been added to app")
+
+	result := app.tryPlaceholderAllocate(iterator, getNode)
+	assert.Assert(t, result == nil, "result should be nil since the ask is larger than the placeholder")
```
wilfred-s, line 3531: Extend the checks of the result to be more than just nil. What happens with the placeholder data? What happens to the placeholder?

pkg/scheduler/objects/application_test.go
```diff
@@ -3438,3 +3438,130 @@ func TestApplication_canAllocationReserve(t *testing.T) {
...
+}
+
+func TestTryPlaceHolderAllocateDifferentTaskGroups(t *testing.T) {
+	const (
+		tg1 = "tg-1"
+		tg2 = "tg-2"
+	)
```
wilfred-s, line 3538: Would be good to refactor this now as we have this same const declaration in at least 3 places in the file. Create one const for the file.
Can be done as part of this PR

wilfred-s (changes requested):

wilfred-s:
No code coverage report as the previous commit was a non code change (e2e test change) The commit did not submit a coverage report to diff against.

pkg/scheduler/objects/application_test.go
kousei47747: Thanks for reviewing! I've added some checks:

- ResultType, NodeId and Request of the result
- links between real allocation and placeholder
- placeholder data should remain unchanged

pkg/scheduler/objects/application_test.go
kousei47747: add 2 checks:
- placeholder data should remain unchanged
- placeholder should be released

pkg/scheduler/objects/application_test.go
kousei47747: add tg1, tg2 and tg3 in `utilities_test.go`

chenyulin0719 (commented):
Hi @kousei47747, thanks for the nice patch.
Could we add one more test case to cover the path "allocate placeholder to non-placeholder node"?  It happens when node.preReserveConditions(request) != nil.

- https://github.com/apache/yunikorn-core/blob/149820f2ceaa33db7a9f4ebb073df856dc5f829b/pkg/scheduler/objects/application.go#L1199

- https://github.com/apache/yunikorn-core/blob/149820f2ceaa33db7a9f4ebb073df856dc5f829b/pkg/scheduler/objects/application.go#L1261-L1276

kousei47747:
@chenyulin0719 thanks for reviewing! I've added the test case, please check.

pkg/scheduler/objects/application_test.go
```diff
@@ -3438,3 +3427,172 @@ func TestApplication_canAllocationReserve(t *testing.T) {
...
+	app.queue = queue
+
+	res := resources.NewResourceFromMap(map[string]resources.Quantity{"first": 5})
+	ph := newPlaceholderAlloc(appID0, nodeID1, res, tg1)
+	app.AddAllocation(ph)
+	app.addPlaceholderData(ph)
+	assertPlaceholderData(t, app, tg1, 1, 0, 0, res)
+
+	// predicate check fails on node1 and passes on node2
+	mockPlugin := mockCommon.NewPredicatePlugin(false, map[string]int{nodeID1: 0})
```
chenyulin0719, line 3579: At the beginning, I'm confused about why using `{nodeID1: 0}) (fail always)` instead of `{nodeID1: -1}) (fail reserve)`.  After rethinking it, I realize it's reasonable because if we set `{nodeID1: -1})`, `nodeID1` will be picked again in the following `try all nodes` check , the [preAllocateCheck()](https://github.com/apache/yunikorn-core/blob/2113cd7e439c840330db313901d8f7855ab3b532/pkg/scheduler/objects/application.go#L1241C1-L1242C1) will pass.  

So it's make sense to me now.

chenyulin0719 (approved):
+1 LGTM

pbacsko (approved):
LGTM

ryankert01 (approved):

wilfred-s (approved):
LGTM +1

PR #1007, opened 2024-12-21
[YUNIKORN-3004] Lock User Trackers and Group Trackers map during dele…

manirajv06:
…te operations

### What is this PR for?
Lock `userTrackers` and `groupTrackers` map during delete operations.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3004

### Questions:

### DIFF
--- a/pkg/scheduler/ugm/manager.go
+++ b/pkg/scheduler/ugm/manager.go
@@ -159,7 +159,9 @@ func (m *Manager) DecreaseTrackedResource(queuePath, applicationID string, usage
 	if userTracker.decreaseTrackedResource(queuePath, applicationID, usage, removeApp) {
 		log.Log(log.SchedUGM).Info("Removing user from manager",
 			zap.String("user", user.User))
+		m.Lock()
 		delete(m.userTrackers, user.User)
+		m.Unlock()
 	}
 	// if the app did not have a group we're done otherwise update the groupTracker
 	if appGroup == common.Empty {
@@ -184,7 +186,9 @@ func (m *Manager) DecreaseTrackedResource(queuePath, applicationID string, usage
 			zap.String("queue path", queuePath),
 			zap.String("application", applicationID),
 			zap.Bool("removeApp", removeApp))
+		m.Lock()
 		delete(m.groupTrackers, appGroup)
+		m.Unlock()
 	}
 }
 


pbacsko (commented):
Please also take a look at `resetGroupEarlierUsage()` and `resetGroupEarlierUsage()`. No locks there either and those are called during config refresh I believe. If that's the case, then lock/unlock pair must be added there too.

manirajv06:
> Please also take a look at `resetGroupEarlierUsage()` and `resetGroupEarlierUsage()`. No locks there either and those are called during config refresh I believe. If that's the case, then lock/unlock pair must be added there too.

I did already checked. `resetGroupEarlierUsage()` is called from `clearEarlierSetLimits` which locks the object before calling the earlier.

pbacsko (approved):

wilfred-s:
We should have a test case for the empty group tracker removal after decrease, requesting a follow up jira

Besides that fix itself is good, all access is now locked.

ryankert01:
Interest in helping create this test! 
jira link: https://issues.apache.org/jira/browse/YUNIKORN-3005

---

@wilfred-s I saw something interesting while trying to create the test: we don't have logic to decrease `qt.maxRunningApps` and `qt.maxResources` while [decreaseTrackedResource() on L159](https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/ugm/manager.go#L159) requires [it](https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/ugm/queue_tracker.go#L175) to be zero to be deleted. 
Therefore, in [decreaseTrackedResource() on L183](https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/ugm/manager.go#L183), it will not be deleted.
Maybe we should relax/add some logic with `qt.maxRunningApps` and `qt.maxResources` ?

chenyulin0719:
@ryankert01 I don't think we need to add other test-only functions for the test case. 

The `maxRunningApps` and `maxResources` in QueueTracker are created from queue configurations and will not be changed during their lifetime.  You should be able to create QueueTracker  with nil `maxRunningApps`/`maxResources`  by configureing the corresponding queue config.

ryankert01:
@chenyulin0719 Great insight! I completely think in the wrong direction (: 
PR submitted!

PR #1008, opened 2024-12-25
[YUNIKORN-2998] tryNodesNoReserve is not tested

kousei47747:
### What is this PR for?
Add unit test for tryNodesNoReserve, including 5 test cases:

- Node is the reserved node
- Node is unschedulable
- Node does not have enough resources
- Node fails predicate
- Allocation succeeds

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2998


pkg/scheduler/objects/application_test.go
```diff
@@ -3596,3 +3596,48 @@ func TestTryPlaceHolderAllocateDifferentNodes(t *testing.T) {
...
+
+	// case 2: node is unschedulable
+	node2 := newNode(nodeID2, map[string]resources.Quantity{"first": 5})
+	node2.schedulable = false
+	iterator = getNodeIteratorFn(node2)
+	result = app.tryNodesNoReserve(ask, iterator(), node1.NodeID)
+	assert.Assert(t, result == nil, "result should be nil since node2 is unschedulable")
+
+	// case 3: node does not have enough resources
+	node3 := newNode("node-3", map[string]resources.Quantity{"first": 1})
```
ryankert01, line 3630: maybe we should stay consistent by using variable as node name?

ryankert01 (commented):
overall lgtm! btw, is there a way to cover L1430?

kousei47747:
@ryankert01 thanks for reviewing! I've added the variables and another test case, the coverage is now 100%, PTAL :)
- https://github.com/apache/yunikorn-core/blob/859a7600454fc60bad9d5422d048376d21a7b9d1/pkg/scheduler/objects/utilities_test.go#L51-L53
- https://github.com/apache/yunikorn-core/blob/859a7600454fc60bad9d5422d048376d21a7b9d1/pkg/scheduler/objects/application_test.go#L3635-L3642

pkg/scheduler/objects/coverage.out
```diff
@@ -0,0 +1,1755 @@
+mode: set
```
ryankert01, line 1: ditto.

coverage.out
```diff
@@ -0,0 +1 @@
+mode: set
```
ryankert01, line 1: This looks like a redundant file! add by accident?

ryankert01 (commented):
L1431 still not covered, but I think it might be fine if there's no clear way to do it. 
截圖 2024-12-25 下午2 37 55

also leaves some comments.

coverage.out
kousei47747: ah yes, thanks for pointing it out, file has been removed

ryankert01 (approved):
lgtm

chenyulin0719 (approved):
+1 LGTM

PR #1009, opened 2024-12-26
[YUNIKORN-2907] Queue config processing log spew

kousei47747:
### What is this PR for?
Silence the logs of shadow queue structure during configuration update.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2907


pkg/scheduler/partition.go
```diff
@@ -224,7 +224,7 @@ func (pc *PartitionContext) updateQueues(config []configs.QueueConfig, parent *o
 		queue := pc.getQueueInternal(pathName)
 		var err error
 		if queue == nil {
-			queue, err = objects.NewConfiguredQueue(queueConfig, parent)
+			queue, err = objects.NewConfiguredShadowQueue(queueConfig, parent)
```
chia7712, line 227: Is this path intended to update the partition based on a configuration reload? If so, could you explain why it is referred to as a "shadow queue"?

chia7712 (commented):
@kousei47747 thanks for this patch. There is one small question remaining. Please take a look.

pkg/scheduler/objects/queue.go
```diff
@@ -113,9 +113,29 @@ func newBlankQueue() *Queue {
...
 func NewConfiguredQueue(conf configs.QueueConfig, parent *Queue) (*Queue, error) {
+	queue, err := newConfiguredQueueInternal(conf, parent)
+	if queue != nil {
+		log.Log(log.SchedQueue).Info("configured queue added to scheduler",
+			zap.String("queueName", queue.QueuePath))
+	}
+	return queue, err
+}
+
+// NewConfiguredShadowQueue creates a new queue from scratch based on the configuration and logs at debug level
```
blueBlue0102, line 127: Hi @kousei47747, thank you for your contribution!

Could you provide more details about the differences between `NewConfiguredQueue` and `NewConfiguredShadowQueue`?
Currently, it appears the only difference is the log level, which may be confusing for those unfamiliar with shadow queues.

It would be helpful to explain when developers should use `NewConfiguredQueue` versus `NewConfiguredShadowQueue`.

wilfred-s:
I dropped a large comment into the jira to provide some guidance and point out another issue.

pkg/scheduler/partition.go
kousei47747: @chia7712 thanks for reviewing! Yes, it should not be referred to as a "shadow queue", I completely misunderstood the code here. I'm now working on a new version based on the guidance.

pkg/scheduler/objects/queue.go
kousei47747: @blueBlue0102 thanks for reviewing! I will provide more detailed description in the new patch.

kousei47747:
@wilfred-s thanks for reviewing! I've replied on jira, please take a look.

kousei47747:
I’ve updated the PR based on the guidance and noticed that `NewPlacementManager` also logs. Should it be silenced too?

pkg/scheduler/partition.go
```diff
@@ -79,7 +79,29 @@ type PartitionContext struct {
 	locking.RWMutex
 }
 
+// newPartitionContextForValidation initializes a shadow partition based on the configuration.
+// The shadow partition is used to validate the configuration, it is not used for scheduling.
+func newPartitionContextForValidation(conf configs.PartitionConfig, rmID string, cc *ClusterContext) (*PartitionContext, error) {
```
pbacsko, line 84: The return value `*PartitionContext` is never used from this function, so you might as well just remove it and just call it `validateConfiguration()`. Unless of course, you create unit tests which actually use it for some kind of verification... See other comments.

pkg/scheduler/partition.go
```diff
@@ -191,17 +252,27 @@ func (pc *PartitionContext) updatePartitionDetails(conf configs.PartitionConfig)
...
+	return err
+}
+
 // Process the config structure and create a queue info tree for this partition
 func (pc *PartitionContext) addQueue(conf []configs.QueueConfig, parent *objects.Queue) error {
+	err := pc.addQueueInternal(conf, parent, objects.NewConfiguredQueue)
+	return err
+}
+
+func (pc *PartitionContext) addQueueInternal(conf []configs.QueueConfig, parent *objects.Queue, newQueueFn func(configs.QueueConfig, *objects.Queue) (*objects.Queue, error)) error {
```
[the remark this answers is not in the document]
kousei47747: @pbacsko Thanks for reviewing! I'm not sure about these method duplications either. 

My first thought is like you said that NewConfiguredQueue() has some callers, adding a flag would require all of them to include an additional parameter. But after implementing these method duplications, the code feels redundant, and it turns out that more test cases need to be covered.

I'll try the flag approach. Thanks a lot, my confusion is now cleared!

blueBlue0102:
If we need to silence all logs when creating shadow queues, it seems that some places were missed.  

For example:  

- **`ACL.setUsers()`**:  
  - Related code: [acl.go#L54](https://github.com/kousei47747/yunikorn-core/blob/36bb742f7e703efad7a0de19d066addd65fbc606/pkg/common/security/acl.go#L54)  
  - Call stack:  
    ```plaintext
    NewConfiguredQueue
      └── sq.applyConf(conf)
            └── security.NewACL()
                  └── acl.setUsers()
    ```

It seems that either adding an additional parameter or creating new functions adds quite a lot of complexity to the codebase, but I don't have a good idea for a third approach at the moment.

pkg/scheduler/partition.go
```diff
@@ -191,17 +199,18 @@ func (pc *PartitionContext) updatePartitionDetails(conf configs.PartitionConfig)
 	return ugm.GetUserManager().UpdateConfig(queueConf, conf.Queues[0].Name)
```
pbacsko, line 199: What about this one? This is still called regardless of the value of `silence`.

kousei47747: `updatePartitionDetails` is called only when performing the real update, so I think we can skip `silence` here.
https://github.com/apache/yunikorn-core/blob/34ca3b2499c6bba66651075577e200f54708e36b/pkg/scheduler/context.go#L372-L374

pkg/scheduler/context.go
```diff
@@ -365,7 +365,7 @@ func (cc *ClusterContext) updateSchedulerConfig(conf *configs.SchedulerConfig, r
 		part, ok := cc.partitions[p.Name]
 		if ok {
 			// make sure the new info passes all checks
-			_, err = newPartitionContext(p, rmID, nil)
+			_, err = newPartitionContext(p, rmID, nil, true)
 			if err != nil {
```
blueBlue0102, line 369: It seems we call `newPartitionContext` here only for validating `configs.PartitionConfig`.  
If my understanding is correct, the main purpose of this JIRA is to avoid any side effects when validating the config before `updatePartitionDetails`,  
**not only**:  
- Logging  
- Sending out events  

**but also including**:  
- Loading the user manager and updating user settings.  
  `ugm.GetUserManager().UpdateConfig()` would be called in `updatePartitionDetails()` when validation passes.  
- etc.  

So rather than adding a lot of flag parameters to each underlying function, extracting all the validation logic from `newPartitionContext` into another function or adding those validation logic into `updatePartitionDetails` might be a better approach.  
Let `newPartitionContext` do what it needs to do without consideration of "silence". What do you think?

blueBlue0102 (commented):
Thanks for the patches! It seems there are still some places that were missed. It made me think of another approach.

pkg/common/security/acl.go
```diff
@@ -116,7 +119,7 @@ func NewACL(aclStr string) (ACL, error) {
 	// trim and check for wildcard
 	acl.setAllAllowed(aclStr)
 	// parse users and groups
-	acl.setUsers(strings.Split(fields[0], common.Separator))
+	acl.setUsers(strings.Split(fields[0], common.Separator), silence)
 	if len(fields) == 2 {
 		acl.setGroups(strings.Split(fields[1], common.Separator))
```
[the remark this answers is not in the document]
kousei47747: Silence flag added to setGroups, thanks!

pkg/scheduler/context.go
kousei47747: Thanks for the nice approach! It looks like the trade-off would be duplicate logic/methods when extracting it from stacked callers. I tried a similar approach in the first few commits, but it turned out to be a bit cumbersome. So I think I still prefer adding the silence flag.
Please let me know if I’m missing something!

pbacsko (approved):
LGTM

pkg/scheduler/partition.go
```diff
@@ -79,7 +79,7 @@ type PartitionContext struct {
 	locking.RWMutex
 }
 
-func newPartitionContext(conf configs.PartitionConfig, rmID string, cc *ClusterContext) (*PartitionContext, error) {
+func newPartitionContext(conf configs.PartitionConfig, rmID string, cc *ClusterContext, silence bool) (*PartitionContext, error) {
 	if conf.Name == "" || rmID == "" {
 		log.Log(log.SchedPartition).Info("partition cannot be created",
```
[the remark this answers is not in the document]
chenyulin0719: Keeping the log for misconfiguration make sense to me.

chenyulin0719 (approved):
LGTM. Just a confirmation.

pkg/scheduler/partition.go
```diff
@@ -115,38 +116,45 @@ func (pc *PartitionContext) initialPartitionFromConfig(conf configs.PartitionCon
...
 	pc.userGroupCache = security.GetUserGroupCache("")
-	pc.updateNodeSortingPolicy(conf)
+	pc.updateNodeSortingPolicy(conf, silence)
 	pc.updatePreemption(conf)
 
 	// update limit settings: start at the root
-	return ugm.GetUserManager().UpdateConfig(queueConf, conf.Queues[0].Name)
+	if !silence {
+		return ugm.GetUserManager().UpdateConfig(queueConf, conf.Queues[0].Name)
+	}
```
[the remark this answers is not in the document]
kousei47747: @chenyulin0719 yes, thanks :)

ryankert01 (approved):
+1, lgtm

PR #1014, opened 2025-03-03
[YUNIKORN-3036] fix race prevention regression

wilfred-s:
### What is this PR for?
Re-instate the change from YUNIKORN-1993 that was reversed as part of YUNIKORN-2658. Add more explicit comment to not move the code. Minor cleanup in function naming and fix the function length of removeAllocation (lint fix)

Lint fixes to get make lint to pass locally (*_test.go files)

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* [YUNIKORN-3036](https://issues.apache.org/jira/browse/YUNIKORN-3036)

Race condition fix that was in 1.5.x adnd regressed in 1.6.0

### DIFF
--- a/pkg/scheduler/partition.go
+++ b/pkg/scheduler/partition.go
@@ -1416,7 +1416,9 @@ func (pc *PartitionContext) calculateNodesResourceUsage() map[string][]int {
 	return mapResult
 }
 
-func (pc *PartitionContext) generateReleased(release *si.AllocationRelease, app *objects.Application) []*objects.Allocation {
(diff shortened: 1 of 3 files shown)


craigcondit (approved):
+1 LGTM.

manirajv06:
looks good.

We need to check the reason for e2e test failure, that too only for K8s 1.32.0

wilfred-s:
I looked at the last PRs and all fail with the same mention around the scheduler-interface not being available for download.

PR #1021, opened 2025-05-21
[YUNIKORN-656] Add LDAP resolver for group resolution

mitdesai:
### What is this PR for?
This PR adds LDAP resolver functionality for user group resolution in YuniKorn. It enables the scheduler to resolve user groups using LDAP, which is essential for enterprise environments where user and group information is stored in LDAP directories. 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [X] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-656

### Questions:


pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+		secretCount++
+		secretKey := fileName // use the cached fileName
+		secretValueBytes, err := os.ReadFile(filepath.Join(secretsDir, secretKey))
+		if err != nil {
+			log.Log(log.Security).Warn("Could not read secret file",
+				zap.String("file", secretKey),
+				zap.Error(err))
+			continue
+		}
+		secretValue := strings.TrimSpace(string(secretValueBytes))
```
manirajv06, line 100: Is it ok to accept these values as is? Do we need to validate against defined template/format?

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+
+func GetUserGroupCacheLdap() *UserGroupCache {
+	readSecrets()
+
+	return &UserGroupCache{
+		ugs:           map[string]*UserGroup{},
+		interval:      cleanerInterval * time.Second,
+		lookup:        LdapLookupUser,
+		lookupGroupID: LdapLookupGroupID,
+		groupIds:      LDAPLookupGroupIds,
```
manirajv06, line 156: Do we need to initialize `stop` channel as well? We might to add test related based on cache clean up, interval set up etc for this new resolver

pkg/common/security/usergroup_test.go
```diff
@@ -50,9 +51,24 @@ func (c *UserGroupCache) getUGmap() map[string]*UserGroup {
...
+// UserGroupResolver Config for the os resolver
+var osResolver = configs.UserGroupResolver{
+	Type: "os",
+}
+
+// UserGroupResolver Config for the unknown resolver
+var unknownResolver = configs.UserGroupResolver{
+	Type: "unknown",
+}
+
```
manirajv06, line 68: Please see earlier comment. We need to run these tests even for this new resolver.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+func LdapLookupGroupID(gid string) (*user.Group, error) {
+	group := user.Group{Gid: gid}
+	group.Name = gid
+	return &group, nil
+}
+
+func LDAPLookupGroupIds(osUser *user.User) ([]string, error) {
+	sr, err := LDAPConn_Bind(osUser.Username)
+	if err != nil {
+		return nil, err
```
manirajv06, line 178: Debug log would help here.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+	} else {
+		LDAP_URI = "ldap"
+	}
+
+	ldapaddr := fmt.Sprintf("%s://%s:%d", LDAP_URI, ldapConf.Host, ldapConf.Port)
+
+	l, err := ldap.DialURL(ldapaddr,
+		ldap.DialWithTLSConfig(&tls.Config{InsecureSkipVerify: ldapConf.Insecure})) // #nosec G402
+	if err != nil {
+		return nil, err
```
manirajv06, line 208: Debug log would help here.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	defer l.Close()
+
+	err = l.Bind(ldapConf.BindUser, ldapConf.BindPassword)
+	if err != nil {
+		return nil, err
```
manirajv06, line 214: Debug log would help here.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+			log.Log(log.Security).Warn("Encountered unrecognized LDAP secret key",
+				zap.String("key", secretKey))
+		}
+	}
+	log.Log(log.Security).Info("Finished loading LDAP secrets",
+		zap.Int("numberOfSecretsLoaded", secretCount))
+}
+
+func GetUserGroupCacheLdap() *UserGroupCache {
+	readSecrets()
```
manirajv06, line 149: Should we go forward if there are no secrets? Could happen because of LDAP secret mount setup issues. Since we use default values, warning log would help here with values as well because some values might be available or some not.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+func LDAPLookupGroupIds(osUser *user.User) ([]string, error) {
+	sr, err := LDAPConn_Bind(osUser.Username)
+	if err != nil {
+		return nil, err
+	}
+
+	var groups []string
+	for _, entry := range sr.Entries {
+		a := entry.GetAttributeValues("memberOf")
+		println(a)
```
manirajv06, line 184: debug statement, not required.

manirajv06 (changes requested):
Overall looks good.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+			s := strings.Split(a[i], ",")
+			newgroup := strings.Split(s[0], "CN=")
+
+			groups = append(groups, newgroup[1])
+		}
+	}
+	return groups, nil
+}
+
+var LDAPConn_Bind = func(userName string) (*ldap.SearchResult, error) {
```
pbacsko: I'd refactor this code for better testability.
Try to hide as much LDAP-specific calls as possible, thus making the main business logic more testable.

For example, an interface could be:
```
type LdapAccess interface {
  DialURL(..) (*Conn, error)
  Bind(...) error
  Search(...) (*SearchResult, error)
}
```

Then we can go for two implementations, a true `LdapAccessImpl` and `LdapAccessMock`.
As usual, write a helper function which creates `LdapAccessImpl`, eg.

```
func newLdapAccessImpl(LdapResolverConfig config) *LdapAccessImpl {
  ... create based on config entries ...
}
```

This way, you don't have to replace `LDAPConn_Bind` for testing, just call `newLdapAccessImpl()` in the production code and use a mock in the test code.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,229 @@
...
+		lookup:        LdapLookupUser,
+		lookupGroupID: LdapLookupGroupID,
+		groupIds:      LDAPLookupGroupIds,
+	}
+}
+
+// Default linux behaviour: a user is member of the primary group with the same name
+func LdapLookupUser(userName string) (*user.User, error) {
+	return &user.User{
+		Uid:      "1211",
```
pbacsko: Magic value, please extract constant with a name.

pkg/common/security/usergroup_ldap_resolver_test.go
```diff
@@ -0,0 +1,309 @@
...
+	assert.Assert(t, common.DefaultLdapPort == ldapConf.Port, "DefaultLdapPort should be equal to ldapConf.Port")
+	assert.Assert(t, common.DefaultLdapBaseDN == ldapConf.BaseDN, "DefaultLdapBaseDN should be equal to ldapConf.BaseDN")
+	assert.Assert(t, common.DefaultLdapFilter == ldapConf.Filter, "DefaultLdapFilter should be equal to ldapConf.Filter")
+	assert.Assert(t, common.DefaultLdapGroupAttr == ldapConf.GroupAttr, "DefaultLdapGroupAttr should be equal to ldapConf.GroupAttr")
+	assert.Assert(t, strings.Join(common.DefaultLdapReturnAttr, ",") == strings.Join(ldapConf.ReturnAttr, ","), "DefaultLdapReturnAttr should be equal to ldapConf.ReturnAttr")
+	assert.Assert(t, common.DefaultLdapBindUser == ldapConf.BindUser, "DefaultLdapBindUser should be equal to ldapConf.BindUser")
+	assert.Assert(t, common.DefaultLdapBindPassword == ldapConf.BindPassword, "DefaultLdapBindPassword should be equal to ldapConf.BindPassword")
+	assert.Assert(t, common.DefaultLdapInsecure == ldapConf.Insecure, "DefaultLdapInsecure should be equal to ldapConf.Insecure")
+	assert.Assert(t, common.DefaultLdapSSL == ldapConf.SSL, "DefaultLdapSSL should be equal to ldapConf.SSL")
```
pbacsko: use `assert.Equal()` everywhere for consistency

pkg/common/security/usergroup_ldap_resolver_test.go
```diff
@@ -0,0 +1,309 @@
...
+	assert.Equal(t, common.DefaultLdapHost, ldapConf.Host)
+	assert.Equal(t, common.DefaultLdapPort, ldapConf.Port)
+	assert.Equal(t, common.DefaultLdapBaseDN, ldapConf.BaseDN)
+	assert.Equal(t, common.DefaultLdapFilter, ldapConf.Filter)
+	assert.Equal(t, common.DefaultLdapGroupAttr, ldapConf.GroupAttr)
+	assert.Assert(t, strings.Join(common.DefaultLdapReturnAttr, ",") == strings.Join(ldapConf.ReturnAttr, ","))
+	assert.Assert(t, common.DefaultLdapBindUser == ldapConf.BindUser)
+	assert.Assert(t, common.DefaultLdapBindPassword == ldapConf.BindPassword)
+	assert.Assert(t, common.DefaultLdapInsecure == ldapConf.Insecure)
+	assert.Assert(t, common.DefaultLdapSSL == ldapConf.SSL)
```
pbacsko: use assert.Equal() everywhere for consistency

pkg/common/security/usergroup_ldap_resolver_test.go
```diff
@@ -0,0 +1,309 @@
...
+	// Assert that rest of ldap conf is set to default values
+	assert.Equal(t, common.DefaultLdapHost, ldapConf.Host)
+	assert.Equal(t, common.DefaultLdapBaseDN, ldapConf.BaseDN)
+	assert.Equal(t, common.DefaultLdapFilter, ldapConf.Filter)
+	assert.Equal(t, common.DefaultLdapGroupAttr, ldapConf.GroupAttr)
+	assert.Assert(t, strings.Join(common.DefaultLdapReturnAttr, ",") == strings.Join(ldapConf.ReturnAttr, ","))
+	assert.Assert(t, common.DefaultLdapBindUser == ldapConf.BindUser)
+	assert.Assert(t, common.DefaultLdapBindPassword == ldapConf.BindPassword)
+	assert.Assert(t, common.DefaultLdapInsecure == ldapConf.Insecure)
+	assert.Assert(t, common.DefaultLdapSSL == ldapConf.SSL)
```
pbacsko: use assert.Equal() everywhere for consistency

pbacsko (changes requested):
I'll have a second round later on, now I'm adding comments based on my first impression.

pkg/common/security/usergroup_ldap_resolver.go
pbacsko: Also, rename `LDAPConn_Bind()` to something better like `LdapSearch()`. Then this function can receive the `LdapAccess` implementation in the first parameter:

```
func LdapSearch(ldapAccess *LdapAccess, userName string) (*ldap.SearchResult, error) {
   ...
}
```

Obviously a different path can be chosen if that's more appropriate.

mitdesai:
Thanks for the review @pbacsko  and @manirajv06. Valid concerns. I will update the PR today

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+	ReturnAttr:   common.DefaultLdapReturnAttr,
+	BindUser:     common.DefaultLdapBindUser,
+	BindPassword: common.DefaultLdapBindPassword,
+	Insecure:     common.DefaultLdapInsecure,
+	SSL:          common.DefaultLdapSSL,
+}
+
+// read secrets from the secrets directory
+// returns true if at least one secret was loaded and the configuration is valid, false otherwise
+var readSecrets = func() bool {
```
pbacsko: Just like in case of `LdapAccess`, I propose an interface for this, eg. `SecretReader` with two impementations: `SecretReaderImpl` and `SecretReaderMock`.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+
+	log.Log(log.Security).Info("Finished loading LDAP secrets",
+		zap.Int("numberOfSecretsLoaded", secretCount),
+		zap.Bool("configurationValid", isValid),
+		zap.Int("missingRequiredFields", len(missingFields)))
+
+	return secretCount > 0 && isValid && len(missingFields) == 0
+}
+
+func GetUserGroupCacheLdap() *UserGroupCache {
```
pbacsko: This function should take two arguments: one for `LdapAccess` and one for `SecretReader`. That way, we eliminate global state which can affect unit tests. See below for details.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+	secretsLoaded := readSecrets()
+
+	if !secretsLoaded {
+		// Log a FATAL level message - this is very prominent and will typically cause the application to exit
+		log.Log(log.Security).Fatal("LDAP configuration not found or invalid. No secrets were loaded from the secrets directory.",
```
pbacsko, line 253: Do we truly need this to be fatal? I'm wondering if we have to be that strict.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+
+func (l *LdapAccessImpl) Close(conn *ldap.Conn) {
+	conn.Close()
+}
+
+// ldapAccessFactory is a function type that creates LdapAccess instances
+type ldapAccessFactory func(config *LdapResolverConfig) LdapAccess
+
+// defaultLdapAccessFactory is the default factory function that creates real LdapAccessImpl instances
+var defaultLdapAccessFactory ldapAccessFactory = func(config *LdapResolverConfig) LdapAccess {
```
pbacsko: See comment below, store the `LdapAccess` as a variable in the struct, so such global variables are not needed.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+
+// newLdapAccessImpl creates a new LdapAccess instance using the current factory
+// This can be replaced in tests to return mock implementations
+var newLdapAccessImpl = defaultLdapAccessFactory
+
+// resetLdapAccessFactory resets the factory to the default implementation
+// This is used in tests to ensure the global state is restored
+func resetLdapAccessFactory() {
+	newLdapAccessImpl = defaultLdapAccessFactory
+}
```
pbacsko: Can be removed if we store `LdapAccess` as struct member.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+		// we could also panic here to ensure the application stops
+		panic("LDAP configuration not found or invalid")
+	}
+
+	return &UserGroupCache{
+		ugs:           map[string]*UserGroup{},
+		interval:      cleanerInterval * time.Second,
+		lookup:        LdapLookupUser,
+		lookupGroupID: LdapLookupGroupID,
+		groupIds:      LDAPLookupGroupIds,
```
pbacsko: These functions should be part of a type called `LdapLookup`. So this
```
return &UserGroupCache{
		ugs:           map[string]*UserGroup{},
		interval:      cleanerInterval * time.Second,
		lookup:        LdapLookupUser,
		lookupGroupID: LdapLookupGroupID,
		groupIds:      LDAPLookupGroupIds,
                stop:          make(chan struct{}),
	}
```
becomes
```
ldapLookup: &LdapLookup{secretReader: secretReader, ldapAccess: ldapAccess}

return &UserGroupCache{
		ugs:           map[string]*UserGroup{},
		interval:      cleanerInterval * time.Second,
		lookup:        ldapLookup.LdapLookupUser,
		lookupGroupID: ldapLookup.LdapLookupGroupID,
		groupIds:      ldapLookup.LDAPLookupGroupIds,
                stop:          make(chan struct{}),
	}
```

Which is better because we can store both `secretReader` and `ldapAccess` inside `ldapLookup`. That way, there's no need to always create an `LdapAccessImpl` every time we call `LDAPLookupGroupIds()`. Also makes unit testing completely stateless - there will be no need to setup/teardown stuff, defer calls, etc.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+
+func LdapLookupGroupID(gid string) (*user.Group, error) {
+	log.Log(log.Security).Debug("Looking up LDAP group ID",
+		zap.String("groupID", gid))
+	group := user.Group{Gid: gid}
+	group.Name = gid
+	return &group, nil
+}
+
+func LDAPLookupGroupIds(osUser *user.User) ([]string, error) {
```
pbacsko: nit: LDAP vs Ldap, use consistent capitalization ("Ldap" is preferred)

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+	Host         string
+	Port         int
+	BaseDN       string
+	Filter       string
+	GroupAttr    string
+	ReturnAttr   []string
+	BindUser     string
+	BindPassword string
+	Insecure     bool
+	SSL          bool
```
pbacsko: nit: "UseSSL" instead of "SSL" to make it less ambiguous

pkg/common/security/usergroup_test.go
```diff
@@ -27,10 +27,67 @@ import (
...
+			newLdapAccessImpl = factory
+		}
+	}
+
+	if originalRead, ok := originalFunctions["readSecrets"]; ok {
+		if readFunc, ok := originalRead.(func() bool); ok {
+			readSecrets = readFunc
+		}
+	}
+}
```
pbacsko: With the proposed changes I outlined in `usergroup_ldap_resolver.go`, this can be all eliminated.

pkg/common/security/usergroup_ldap_resolver.go
mitdesai: My initial implementation did not have it. But while addressing the review comments I received from you and Mani, I through that if LDAP is actually a requirement, and there is something wrong with the secrets, logging a warning and moving with defaults could result in situations which could be undesirable especially in prod like environments.
The resolver initialization happens in the very beginning of scheduler startup and we would never come back to read the secrets. So it is very easy to have that one warning message to get lost among all the other logs.

pkg/common/security/usergroup_ldap_resolver.go
```diff
@@ -0,0 +1,375 @@
...
+	Port:         common.DefaultLdapPort,
+	BaseDN:       common.DefaultLdapBaseDN,
+	Filter:       common.DefaultLdapFilter,
+	GroupAttr:    common.DefaultLdapGroupAttr,
+	ReturnAttr:   common.DefaultLdapReturnAttr,
+	BindUser:     common.DefaultLdapBindUser,
+	BindPassword: common.DefaultLdapBindPassword,
+	Insecure:     common.DefaultLdapInsecure,
+	SSL:          common.DefaultLdapSSL,
+}
```
pbacsko: We should consider moving this into the suggested new type `LdapLookup`. This is yet another global variable we should try to get rid of.

pkg/common/security/usergroup_ldap_resolver.go
pbacsko: Yes, I was thinking about the same thing, ie. undesired side-effects. I'm OK with leaving it in the code, but we should definitely document it.

Two questions:
1) Is the Ldap mount always expected to exist? Are we always supposed to have this directory with files inside?
2) By taking a deeper look at `readSecrets()`, it looks like we're reading configuration entries, not really secrets. Is this proper naming in the context of LDAP?

pbacsko: BTW sorry for the slightly fragmented feedback, it takes some time to understand this PR, it's a slightly bigger one +3500 LOC :)

pkg/common/security/usergroup_ldap_resolver.go
pbacsko: @mitdesai see question from Mani. I'd like to know the same thing.

pkg/common/security/usergroup_ldap_resolver.go
mitdesai: No problem. I actually appreciate you taking interest in reviewing such PRs :-)

For your questions:
1. Yes, because the the secret that I am expecting will be created like this:
```
kubectl -n yunikorn create secret generic ldap-secret \
    --from-literal=Host=$host \
    --from-literal=Port=$port \
    --from-literal=BaseDN=$baseDN \
    --from-literal=Filter=$filter \
    --from-literal=GroupAttr=$groupAttr \
    --from-literal=ReturnAttr=$returnAttr \
    --from-literal=BindUser=$bindUser \
    --from-literal=BindPassword=$bindPassword \
    --from-literal=Insecure=$insecure \
    --from-literal=SSL=$ssl \
    --dry-run=client -o yaml | kubectl apply -f -
```
When ldap-secret is mounted, I get as many files as there are keys in the secret.

2. It is called `readSecrets()` because we actually attempt to load and read the kubernetes secret. You are right that most of these are configurations. except the `BindPassword`. In order to keep all the LDAP related settings together, I put them inside the ldap-secret.

Part of the reason is also that I had moved on from reading the ldap details from configmap to reading it from secret. At that time, it was a lift and shift effort in order to minimize code changes.

pbacsko: Alright, got it. We definitely need to extend the documentation with this.

pkg/common/security/usergroup_ldap_resolver.go
craigcondit: With regards to configuration and secrets management, I'm pretty strongly against having everything in the secret file. At most username and password should be there; everything else ought to be handled via the normal configuration framework the rest of YuniKorn uses. Otherwise the LDAP resolver becomes this weird "add-on" that doesn't integrate well with the rest of YuniKorn. This would also solve the question Mani is asking -- if the feature is disabled, we skip loading the secrets entirely. If enabled, then we load secrets. If not present or empty we assume anonymous access (yes, people still do this). I think we should also use the K8s secret API directly rather than file access as we do elsewhere in the code. The secrets can then be watched via an Informer so that credentials can be hot reloaded (as should the rest of the config like we do elsewhere in YuniKorn.

mitdesai: Noted! Let me check how we are handling the secrets elsewhere. Rest of the discussion about the secrets I had with @pbacsko is in a separate thread few comments below.

Failing on startup if secrets cannot be loaded is now part of the PR.

pbacsko: +1 for the informer + secrets API. Not sure why I haven't thought of it, but that seems to be the optimal solution.

mitdesai: @craigcondit @pbacsko : to read the secrets using the informers + secrets API, core will get dependent on k8s libraries. I believe, this is something we are trying to avoid. Is that correct?

craigcondit: The configuration is read in the shim and passed into the core. This will be no different. Updates will be needed to both.

pbacsko:
Note: I merged master & pushed a refactoring commit.

I did pretty much everything I recommended, plus:
* Unexported some types
* Renamed methods, variables
* Removed unnecessary code
* Tidied up function access a bit eg. `(l *LdapAccessImpl)` --> `(LdapAccessImpl)`
* Fixed linter `staticcheck` complaints

Regarding the config: that's one more step, so it still reads from a mounted file. But so far it's enough change for review, so let's do that as a follow-up.

manirajv06 (approved):
looks good, pending conflict resolution.

Thanks @pbacsko and @mitdesai for taking this bigger one pending for very long time in.

PR #1022, opened 2025-06-05
[YUNIKORN-3084] Fix Inconsistency in Allocation Removal from sortedRequests

mitdesai:
### What is this PR for?
The binary search in the remove method used the LessThan() comparison function, which compares allocations based on priority and creation time. The comparison should be by allocation key.

When multiple allocations have the same priority and creation time, it is possible that the binary search will find an allocation that is different than the one we want to remove.

### What type of PR is it?
* [X] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3084

Create a Job with 1000 executors. Delete all the executors. Atleast one of the executors will have issue getting removed.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/sorted_asks_test.go
+++ b/pkg/scheduler/objects/sorted_asks_test.go
@@ -98,6 +98,65 @@ func TestInsertRemove(t *testing.T) {
 	assert.Equal(t, 99, len(sorted))
 }
 
+func TestRemoveWithSamePriorityAndTime(t *testing.T) {
+	// Create a sorted requests list
+	sorted := sortedRequests{}
+
+	// Create allocations with same priority and creation time
+	// but different allocation keys
+	baseTime := time.Now()
+	alloc1 := &Allocation{
+		createTime:    baseTime,
+		priority:      10,
+		allocationKey: "alloc-1",
+	}
+	alloc2 := &Allocation{
+		createTime:    baseTime,
+		priority:      10,
+		allocationKey: "alloc-2",
+	}
+	alloc3 := &Allocation{
+		createTime:    baseTime,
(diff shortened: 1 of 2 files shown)


mitdesai:
I don't think the e2e test failure is due to this change.

chenyulin0719 (approved):
+1 LGTM.

A minor feedbcak,  I think the second test `TestRemoveWithSamePriorityAndTimeButDifferentInstances`, might be a bit redundant. Maybe we can remove it?  Does anyone else have any comments on it?

pkg/scheduler/objects/sorted_asks_test.go
```diff
@@ -98,6 +98,106 @@ func TestInsertRemove(t *testing.T) {
...
+		allocationKey: "alloc-2",
+	}
+
+	// Try to remove using the copy
+	sorted.remove(alloc2Copy)
+
+	// Verify alloc2 is removed
+	assert.Equal(t, 1, len(sorted))
+	assert.Assert(t, !askPresent(alloc2, sorted), "alloc2 should be removed")
+	assert.Assert(t, askPresent(alloc1, sorted), "alloc1 should still be present")
```
pbacsko: This can be merged to the previous test as Yu-Lin said.

mitdesai:
The test in question was not adding any additional value. So I have removed it.

wilfred-s (approved):

pbacsko (approved):

PR #1025, opened 2025-07-07
[YUNIKORN-3092] Reservations can permanently block nodes, leading to …

manirajv06:
…preemption failure and a stuck scheduler state

### What is this PR for?

Consider even the nodes which has non daemon set reservations and reservation (allocation) not triggered preemption earlier as candidate to increase the chances of preemption. But cancel those reservations only if its priority is lesser than the preemptor priority and waiting in reservation for more than 1 hour.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3092

### Questions:

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -47,7 +47,9 @@ import (
 )
 
 var (
-	reservationDelay          = 2 * time.Second
+	reservationDelay = 2 * time.Second
+	// Make it configurable
+	reservationWaitTimeout    = 60 * time.Minute
 	completingTimeout         = 30 * time.Second
 	terminatedTimeout         = 3 * 24 * time.Hour
 	defaultPlaceholderTimeout = 15 * time.Minute
@@ -202,6 +204,13 @@ func SetReservationDelay(delay time.Duration) {
 	log.Log(log.SchedApplication).Debug("Set reservation delay",
 		zap.Duration("delay", delay))
 	reservationDelay = delay
+}
+
+// SetReservationWaitTimeout How long reservation should wait in queue?
+func SetReservationWaitTimeout(waitTimeout time.Duration) {
+	log.Log(log.SchedApplication).Debug("Set reservation wait timeout",
+		zap.Duration("wait timeout", waitTimeout))
+	reservationWaitTimeout = waitTimeout
 }
 
 // Return the current state or a checked specific state for the application.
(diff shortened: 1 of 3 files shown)


jdaciuk (commented):
@manirajv06 Thanks for taking up this issue and this LGTM. Looking forward to upgrading to the eventual release. 

Personally we'd like to also cancel reservations of equal priority, but I understand that wouldn't appeal to everyone. Left some more thoughts in the JIRA ticket since they are less relevant to this code.

pkg/scheduler/objects/application.go
```diff
@@ -204,6 +206,13 @@ func SetReservationDelay(delay time.Duration) {
 	reservationDelay = delay
 }
 
+// SetReservationWaitTimeout How long reservation should wait in queue?
+func SetReservationWaitTimeout(waitTimeout time.Duration) {
```
pbacsko: Do we need this method? It's called from the same package and it's only used by tests. I think it's fine to manipulate it directly.

pkg/scheduler/objects/preemption.go
```diff
@@ -161,7 +161,35 @@ func (p *Preemptor) initWorkingState() {
...
 	p.iterator.ForEachNode(func(node *Node) bool {
-		if !node.IsSchedulable() || (node.IsReserved() && !node.isReservedForAllocation(p.ask.GetAllocationKey())) || !node.FitInNode(p.ask.GetAllocatedResource()) {
+		hasOtherReservations := false
+		if node.IsReserved() && !node.isReservedForAllocation(p.ask.GetAllocationKey()) {
+			hasOtherReservations = true
+			for _, res := range node.GetReservations() {
+				// Is Allocation daemon set?
+				// Has this allocation already triggered preemption?
+				if res.alloc.requiredNode != "" || res.alloc.HasTriggeredPreemption() {
+					continue
```
pbacsko, line 171: Can we add coverage for this?

manirajv06: Taken care. Also added new common setup methods for the newly added tests.

pkg/scheduler/objects/application.go
manirajv06: Removed as anyway we don't have plans to expose this.

pbacsko (approved):

PR #1026, opened 2025-08-09
[YUNIKORN-3111] Limit event size to 1024

kousei47747:
### What is this PR for?
Truncate event message to 1024 characters for k8s compatibility.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3111

### DIFF
--- a/pkg/events/event_system_test.go
+++ b/pkg/events/event_system_test.go
@@ -226,7 +226,7 @@ func TestTruncateEventMessage(t *testing.T) {
 		{
 			name:     "message length greater than 1024 characters",
 			message:  getTestString(1100),
-			expected: getTestString(1021) + "...",
+			expected: getTestString(1024-3) + "...",
 		},
 	}
 


pbacsko (changes requested):
@kousei47747 please add a short unit test - it doesn't have to be one for `AddEvent()`, it can be one for `truncateEventMessage()`.

kousei47747:
Hi @pbacsko, thanks for reviewing! I've added the unit test.

pkg/events/event_system_test.go
```diff
@@ -205,3 +205,34 @@ func TestRingBufferCapacity(t *testing.T) {
...
+			message:        "Lorem ipsum",
+			expectedLength: 11,
+		},
+		{
+			name:           "message length exactly 1024 characters",
+			message:        "Lorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos. Lorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus duis convallis. Tempus leo eu aenean sed diam urna tempor. Pulvinar vivamus fringilla lacus nec metus bibendum egestas. Iaculis massa nisl malesuada lacinia integer nunc posuere. Ut hendrerit semper vel class aptent taciti sociosqu. Ad litora torquent per conubia nostra inceptos himenaeos. Lorem ipsum dolor sit amet consectetur adipiscing elit. Quisque faucibus ex sapien vitae pellentesque sem placerat. In id cursus mi pretium tellus d",
```
pbacsko: I suggest a more simplistic approach. Just use a helper function which generates a string (eg. "xxxxxx") with the specifed number of characters (`getTestString(5)` --> "xxxxx"). Also, compare the contents as well, not just the length.

pbacsko (changes requested):
Avoid long literals, use a helper function.

pkg/events/event_system_test.go
```diff
@@ -205,3 +206,34 @@ func TestRingBufferCapacity(t *testing.T) {
 	capacity = getRingBufferCapacity()
 	assert.Equal(t, uint64(configs.DefaultEventRingBufferCapacity), capacity)
 }
+
+func TestTruncateEventMessage(t *testing.T) {
+	testCases := []struct {
+		name           string
+		message        string
+		expectedLength int
```
pbacsko: Final comment: instead of the length, just construct the expected string as well

kousei47747: @pbacsko sorry I missed your last comment in event_system_test.go. I’ll update the PR shortly.

PR #1029, opened 2025-08-22
[YUNIKORN-3113] Resource-wise preemption

manirajv06:
### What is this PR for?

Increase the chances of preemption by looking at both the victim and ask queue left over (remaining) guaranteed resource from ask resource requirements perspective too, not only viewing it plainly. Victim queue with OG res types may have UG res types and Ask queue with UG res types may have OG res types. Trigger preemption during these situations when ask (preemptor) resource requirements starving for res types (supposed to get) but has nothing to do with UG res type in victim queue and OG res types in its own (ask) queue itself.

Introduced a Net remaining guaranteed resource variable to contain the latest value of modified resource type value for the given queue with potential victim eviction. This would help us in understanding the impact on guaranteed resources from victim eviction perspective. As long as relevant (impacted) res types is still above or equals guaranteed values, then it is good to ahead with actual preemption process.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3113

### Questions:

### DIFF
--- a/pkg/common/resources/resources.go
+++ b/pkg/common/resources/resources.go
@@ -1156,3 +1156,26 @@ func (r *Resource) DominantResourceType(capacity *Resource) string {
 	}
 	return dominant
 }
+
+// ExtractLatestIfModified Extract the latest value for every resource type only if value has been modified
+// If both resources are nil, return nil resource
+// If no resource type matches, return nil resource
+// If resource type exists and none modified, return nil resource
+func ExtractLatestIfModified(old *Resource, new *Resource) *Resource {
+	if old == nil || new == nil {
+		return nil
+	}
+	latest := NewResource()
+	for k, v := range old.Resources {
+		if newValue, exists := new.Resources[k]; exists {
+			if v != newValue {
+				latest.Resources[k] = newValue
+			}
+		}
+	}
+	if len(latest.Resources) > 0 {
+		return latest
+	} else {
+		return nil
+	}
+}
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -219,6 +219,7 @@ func (p *Preemptor) checkPreemptionQueueGuarantees() bool {
 		return false
 	}
 
+	oldRemaining := currentQueue.GetRemainingGuaranteedResource()
 	currentQueue.AddAllocation(p.ask.GetAllocatedResource())
 
 	// remove each allocation in turn, validating that at some point we free enough resources to allow this ask to fit
@@ -226,7 +227,12 @@ func (p *Preemptor) checkPreemptionQueueGuarantees() bool {
 		for _, alloc := range snapshot.PotentialVictims {
 			snapshot.RemoveAllocation(alloc.GetAllocatedResource())
 			remaining := currentQueue.GetRemainingGuaranteedResource()
-			if remaining != nil && resources.StrictlyGreaterThanOrEquals(remaining, resources.Zero) {
+
+			// Net remaining guaranteed is nothing but the latest remaining guaranteed value resulted after
+			// the removal of victim on ask queue. In case of res type not used in victim but defined in ask queue's guaranteed,
+			// Net remaining value ensures it has only RELEVANT res types and not res types not being touched at all from victim context.
+			netRemaining := resources.ExtractLatestIfModified(oldRemaining, remaining)
+			if remaining != nil && resources.StrictlyGreaterThanOrEquals(netRemaining, resources.Zero) {
 				return true
 			}
 		}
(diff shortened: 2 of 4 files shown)


pkg/scheduler/objects/preemption.go
```diff
@@ -219,14 +219,20 @@ func (p *Preemptor) checkPreemptionQueueGuarantees() bool {
 		return false
 	}
 
+	oldRemaining := currentQueue.GetRemainingGuaranteedResource()
```
wilfred-s, line 221: We can directly return here if the ask fits in the oldRemaining.
```
if oldRemaining.FitInMaxUndef(ask) {
  retrun true
}
```
examples:
oldRemaining 5 cpu, -10 GPU -100 memory
- ask 1 cpu -> pass
- ask 1 cpu, 1 GPU -> fail
- ask 1 cpu, 1 storage -> pass

pkg/common/resources/resources.go
```diff
@@ -1156,3 +1156,26 @@ func (r *Resource) DominantResourceType(capacity *Resource) string {
 	}
 	return dominant
 }
+
+// ExtractLatestIfModified Extract the latest value for every resource type only if value has been modified
+// If both resources are nil, return nil resource
+// If no resource type matches, return nil resource
+// If resource type exists and none modified, return nil resource
+func ExtractLatestIfModified(old *Resource, new *Resource) *Resource {
```
wilfred-s: This is so specific to just the preemption it would be better to add it in the preemption code

pkg/scheduler/objects/preemption.go
```diff
@@ -219,14 +219,20 @@ func (p *Preemptor) checkPreemptionQueueGuarantees() bool {
...
 		for _, alloc := range snapshot.PotentialVictims {
 			snapshot.RemoveAllocation(alloc.GetAllocatedResource())
 			remaining := currentQueue.GetRemainingGuaranteedResource()
-			if remaining != nil && resources.StrictlyGreaterThanOrEquals(remaining, resources.Zero) {
+
+			// Net remaining guaranteed is nothing but the latest remaining guaranteed value resulted after
+			// the removal of victim on ask queue. In case of res type not used in victim but defined in ask queue's guaranteed,
+			// Net remaining value ensures it has only RELEVANT res types and not res types not being touched at all from victim context.
+			netRemaining := resources.ExtractLatestIfModified(oldRemaining, remaining)
+			if remaining != nil && resources.StrictlyGreaterThanOrEquals(netRemaining, resources.Zero) {
```
wilfred-s: This breaks if a type changes but stays over guaranteed and that type is not requested by the ask:

- alloc 1 cpu, 2 GPU
- oldRemaining -1 cpu, -10 GPU -100 memory
- newRemaining 0 cpu, -8 GPU

in the pre-check we use a fitin setup we need to do something like it here just based on the types that are in the ask and ignore everything else

pkg/scheduler/objects/preemption.go
```diff
@@ -272,21 +278,32 @@ func (p *Preemptor) calculateVictimsByNode(nodeAvailable *resources.Resource, po
...
+					(netRemaining == nil || resources.StrictlyGreaterThanOrEquals(resources.Zero, netRemaining)) {
 					// add the current victim into the ask queue
+					askQueueOldRemaining := askQueue.GetRemainingGuaranteedResource()
 					askQueue.AddAllocation(victim.GetAllocatedResource())
 					askQueueNewRemaining := askQueue.GetRemainingGuaranteedResource()
 
+					// Net remaining guaranteed is nothing but the latest remaining guaranteed value resulted after
+					// the removal of victim on ask queue. In case of res type not used in victim but defined in ask queue's guaranteed,
+					// Net remaining value ensures it has only RELEVANT res types and not res types not being touched at all from victim context.
+					askQueueNetRemaining := resources.ExtractLatestIfModified(askQueueOldRemaining, askQueueNewRemaining)
```
wilfred-s: Same comment as above. ask and it types need to drive this check

pkg/scheduler/objects/preemption.go
manirajv06: Taken care.

pkg/common/resources/resources.go
manirajv06: Taken care.

pkg/scheduler/objects/preemption.go
manirajv06: Taken care.

pkg/scheduler/objects/preemption.go
manirajv06: Taken care.

manirajv06:
Introduced FitInActual() method to decide whether the ask would fit into the ask queue or not instead of using strict based methods to avoid problems due to (from ask resource requirement perspective) unrelated res types with -ve values.

Introduced two more methods in Preemptor to decide ask queue is under guaranteed and victim is over guaranteed or not from ask resource requirement perspective. Replace the strict based in all appropriate places to simplify the decision making. 

Added tests for the same.

PR #1030, opened 2025-09-18
[YUNIKORN-3125] Show allocation name and resource usage in the core when handling non-YK pods

pbacsko:
### What is this PR for?
Enhance printouts when deadling with foreign pods.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3125

### Questions:

### DIFF
--- a/pkg/scheduler/objects/allocation.go
+++ b/pkg/scheduler/objects/allocation.go
@@ -37,7 +37,6 @@ import (
 
 type Allocation struct {
 	// Read-only fields
-	allocationName    string
 	allocationKey     string
 	applicationID     string
 	taskGroupName     string    // task group this allocation belongs to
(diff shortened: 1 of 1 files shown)


pkg/scheduler/objects/allocation.go
```diff
@@ -37,6 +37,7 @@ import (
 
 type Allocation struct {
 	// Read-only fields
+	allocationName    string
```
manirajv06: Do we really need new field to expose the `PodName`? We have already exposed the allocation tags.

pbacsko: I removed the variable, but kept the method for simpler code on the caller side.

PR #1041, opened 2025-10-30
[YUNIKORN-3119] Add Metrics for Monitoring Applications and Nodes Attempted in Each Scheduling Cycle

mitdesai:
### What is this PR for?
Added additional metrics for monitoring nodes and applications attempted during scheduling cycle.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
Jira https://issues.apache.org/jira/browse/YUNIKORN-3119

### Questions:


pbacsko:
@mitdesai please rebase this PR. The unit test failure is unrelated.

manirajv06 (commented):
Do we need to include other types of allocations in scheduling cycles? PH, Reservation etc

pkg/scheduler/partition.go
```diff
@@ -853,6 +853,11 @@ func (pc *PartitionContext) tryPlaceholderAllocate() *objects.AllocationResult {
 		// nothing to do just return
 		return nil
 	}
+
+	// Reset the tryAllocate call counter at the beginning of each scheduling cycle
+	pc.root.ResetApplicationsTried()
+	pc.root.ResetNodesTried()
+
```
pbacsko: This code should not be here. Reset in a separate method and call that from `ClusterContext.schedule()`. The intent is much clearer that way.

pkg/scheduler/objects/queue.go
```diff
@@ -1485,13 +1487,32 @@ func (sq *Queue) TryAllocate(iterator func() NodeIterator, fullIterator func() N
...
+			sq.addNodesTried(countAfter - countBefore)
+
 			if result != nil {
 				log.Log(log.SchedQueue).Info("allocation found on queue",
 					zap.String("queueName", sq.QueuePath),
 					zap.String("appID", app.ApplicationID),
 					zap.Stringer("resultType", result.ResultType),
-					zap.Stringer("allocation", result.Request))
+					zap.Stringer("allocation", result.Request),
+					zap.Int64("applicationsTried:", sq.GetApplicationsTried()),
```
pbacsko: Nit: ":" not needed

pkg/scheduler/objects/queue.go
```diff
@@ -92,6 +92,8 @@ type Queue struct {
 	quotaChangePreemptionDelay         uint64
 	hasTriggerredQuotaChangePreemption bool
 	isQuotaChangePreemptionRunning     bool
+	applicationsTried                  int64 // number of applications tried per scheduling cycle
+	nodesTried                         int64 // number of nodes tried per scheduling cycle
```
pbacsko: These two fields are not related to a queue, but more like a per-scheduling cycle specific info. We must return this to the caller in `PartitionContext` and record the metrics there.

pbacsko (changes requested):
I definitely think this solution needs some re-work. Current approach is a bit hard to understand.

1. Code should be not communicating through metrics. Eg. in `Queue.tryAllocate()`, we call `GetTryNode()` twice to get a difference. Why not just return this from `app.tryAllocate()`? That would be much simpler. Then instead of calling `Inc()` every time from the app, you can just call `Add()` with the number of nodes which was tried.

2. We're storing transient information specifically in the root queue, which involves constant queue walking. It's not the speed that bothers me, but it's just weird. This information is not specific to a queue. Similarly to #1, this data (number of apps tried) should propagate back to a higher-level caller which does the necessary processing. You can easily add this to `AllocationResult` and record the metrics in `PartitionContext.tryAllocate()` after `pc.root.TryAllocate(...)` returns.

pbacsko (approved):
LGTM

wilfred-s (changes requested):
We add an extra return value for the application methods:
* tryAllocate
*  tryReservedAllocate
*  tryPlaceholderAllocate

However all callers ignore the new value returned and read the `NodesTried` out of the result object. We should not add that extra return value to the signatures.
It reduces the change size: no changes to applicatio_test.go and preemption_test.go, half the changes to the queue.go file.

For the two left over methods:
* tryNodes
* tryNodesNoReserve

Setting the `NodesTried` in the not nil result inside the method is unneeded. If the result is not nil we always override the `NodesTried` in the caller. We cannot assume the first call to these methods is a success. Both are called inside a loop and thus the caller is the only one that can track the `NodesTried` correctly and it does that.

PR #1042, opened 2025-10-30
[YUNIKORN-3121] Add REST Endpoint for Scheduling Order Visibility

mitdesai:
### What is this PR for?
Added a new Rest Endpoint for getting the scheduling order - These come in handy when debugging issues in scheduling.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3121

### Questions:


manirajv06 (commented):
Yes, it helps. 

I propose `orderLog` or something like that similar to `allocationLog` to capture the order of events. We can come up with list of constant or message to be logged for each important step. As and when each step pass through, log the corresponding constant/message to the `orderLog` from the originating place as opposed to simulating the code given in this PR as it does not require us to update both the places. Finally, it can be dumped in the state dump. Not really sure this is something needs to be exposed through REST API.

In addition, we can also capture [these](https://github.com/apache/yunikorn-core/pull/1046) kind of cases as well going forward.

pkg/scheduler/partition.go
```diff
@@ -1715,3 +1715,31 @@ func (pc *PartitionContext) getReservationCount() int {
...
+
+// GetOrderLog returns a snapshot of applications with pending requests grouped by queue
+func (pc *PartitionContext) GetOrderLog() []*dao.OrderLogEntry {
+	pc.RLock()
+	defer pc.RUnlock()
+
+	// Build a map of queue -> applications with pending requests
+	queueAppMap := make(map[string][]string)
+
+	for _, app := range pc.applications {
```
manirajv06, line 1727: Map containing applications does not guarantee specific order. Since we are looking for scheduling order at any given moment, I would rather log into data structure which strictly maintains order based on the insertion order so that it can consumed as is in the state dump. I can see `Queue.TryAllocate` as the best place to push the application into the above mentioned slice.

manirajv06 (changes requested):

pkg/scheduler/partition.go
mitdesai: Ah, Good Catch. Not sure how I missed that. Let me fix it.

corleyma:
Just wanted to echo my agreement with @mitdesai on the utility of exposing this via REST interface; I also already use the REST endpoints for selective debugging over the massive state dumps.  

Fwiw, in my experience genai agents also do better with interfaces that let them grab specific information vs trying to deal with something too big for them to fit in context.  I think as the utility of agents continues to increase it makes sense to keep agent ergonomics in mind too.

PR #1068, opened 2026-02-19
[YUNIKORN-3228] Assess the impact of restart on scheduled preemption …

manirajv06:
…and propose a solution

### What is this PR for?  

In case of quota has been lowered and delay set followed up by restart, accommodate the existing allocations only upto the quota's newly max resources and drop the remaining allocations as part of recovery process provided the other settings are also set as expected.

Refactored recovery setup, allocation creation, allocation releases etc process from all tests into common methods and reused the same all over the places. Cleaned up some other stuff too.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3228

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -533,7 +533,7 @@ func (sq *Queue) ShouldApplyQuotaPreemption() bool {
 	}
 	sq.RLock()
 	defer sq.RUnlock()
-	return sq.isManaged && sq.quotaPreemptionDelay != 0 && !sq.maxResource.StrictlyGreaterThanOrEqualsOnlyExisting(sq.allocatedResource)
+	return sq.isManaged && sq.quotaPreemptionDelay != 0 && !resources.IsZero(sq.maxResource) && !sq.maxResource.StrictlyGreaterThanOrEqualsOnlyExisting(sq.allocatedResource)
 }
 
 // setQuotaPreemptionState set or clear the running state for quota preemption. When done the start time is also cleared
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -519,6 +519,23 @@ func (sq *Queue) shouldTriggerPreemption() bool {
 	return !sq.quotaPreemptionStartTime.IsZero() && time.Now().After(sq.quotaPreemptionStartTime)
 }
 
+// ShouldApplyQuotaPreemption returns true if quota preemption should be enforced based on the settings and the
+// delay.
+func (sq *Queue) ShouldApplyQuotaPreemption() bool {
+	if sq == nil {
+		return true
+	}
+	if sq.parent != nil {
+		apply := sq.parent.ShouldApplyQuotaPreemption()
+		if apply {
+			return true
+		}
+	}
+	sq.RLock()
+	defer sq.RUnlock()
+	return sq.isManaged && sq.quotaPreemptionDelay != 0 && !sq.maxResource.StrictlyGreaterThanOrEqualsOnlyExisting(sq.allocatedResource)
+}
+
 // setQuotaPreemptionState set or clear the running state for quota preemption. When done the start time is also cleared
 // out preventing a re-run based on the same change
 func (sq *Queue) setQuotaPreemptionState(run bool) {
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -496,17 +496,6 @@ func (sq *Queue) setPreemptionTime(oldMaxResource *resources.Resource, oldDelay 
 	}
 }
 
-// OverridePreemptionTime Override the earlier set quota preemption time with the passed time.
-// This function MUST be called only from the scheduler recovery process.
-// Configured delay won't be able to adjusted based on the lost time during restart. Hence, overriding
-// preemption time set earlier through recovery config processing with the passed time would trigger preemption immediately
-// in the very first scheduling cycle itself. Flow remains same and works as usual from that point onwards.
-func (sq *Queue) OverridePreemptionTime() {
-	sq.Lock()
-	defer sq.Unlock()
-	sq.quotaPreemptionStartTime = time.Now().Add(-1 * time.Minute)
-}
-
 // ResetPreemptionTime reset the quota preemption variables
 // Only for testing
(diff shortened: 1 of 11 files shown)


wilfred-s (changes requested):
We cannot reject the allocation during recovery that would break badly. The order in which allocations come is is unknown. Old and new are mixed and the allocations that have a node assigned are running and must be accounted for in the app, queue and node. No rejections allowed.

We also need to make sure that we hanlde the VPA case correctly. We do not reject the change at the k8shim level and thus cannot reject it in the core. After we apply the change the queue might go over quota. We need to allow and accept that.

pkg/scheduler/objects/queue.go
```diff
@@ -496,6 +496,26 @@ func (sq *Queue) setPreemptionTime(oldMaxResource *resources.Resource, oldDelay
...
 
+// OverridePreemptionTime Override the earlier set quota preemption time with the passed time.
+// This function MUST be called only from the scheduler recovery process.
+// Configured delay won't be able to adjusted based on the lost time during restart. Hence, overriding
+// preemption time set earlier through recovery config processing with the passed time would trigger preemption immediately
+// in the very first scheduling cycle itself. Flow remains same and works as usual from that point onwards.
+func (sq *Queue) OverridePreemptionTime() {
+	sq.Lock()
+	defer sq.Unlock()
+	sq.quotaPreemptionStartTime = time.Now().Add(-1 * time.Minute)
```
wilfred-s: This would mean that a restart will immediately trigger quota preemption.
Not what I would expect as a behaviour. I would expect that the delay set on the queue would be applied.
Since this is also called for other cases it could trigger immediate preemption for the HPA (horizontal pod autoscaling) case, again unexpected...

pkg/scheduler/objects/queue.go
```diff
@@ -519,6 +539,30 @@ func (sq *Queue) shouldTriggerPreemption() bool {
 	return !sq.quotaPreemptionStartTime.IsZero() && time.Now().After(sq.quotaPreemptionStartTime)
 }
 
+// ShouldApplyQuotaPreemption returns true if quota preemption should be enforced based on the settings and the
+// delay.
+func (sq *Queue) ShouldApplyQuotaPreemption() bool {
```
wilfred-s: During the recovery and update of already assigned allocations the partition calls `IncAllocatedResource()` That allows adding existing allocations to go over the set quota. Inside of that increase the check for quota preemption should be added. The hierarchy is walked and the queue gets locked already.

pkg/scheduler/partition.go
```diff
@@ -1236,6 +1236,12 @@ func (pc *PartitionContext) UpdateAllocation(alloc *objects.Allocation) (request
 			zap.String("appID", applicationID),
 			zap.String("allocationKey", allocationKey))
 
+		// Though Quota preemption is set, allow allocations as usual for now as handling here would lead to unexpected behaviour.
+		// But override (only once) the preemption time set earlier through from config of the queue with passed time so that
+		// preemption would be triggerred immediately during the very first scheduling cycle itself.
+		if pc.IsQuotaPreemptionEnabled() && queue.IsPreemptionScheduled() && queue.ShouldApplyQuotaPreemption() {
```
wilfred-s: Pass `pc.IsQuotaPreemptionEnabled()` in with the call to `queue.IncAllocatedResource()` and hide all the checks there

pkg/scheduler/partition.go
wilfred-s: This could cause over quota also, needs to be covered same as the other two cases

wilfred-s: This point would cover the HPA case in which we increase a pod size and trigger over quota. The one thing we need is the `pc.IsQuotaPreemptionEnabled()` flag to be passed in.

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
```diff
@@ -519,6 +528,15 @@ func (sq *Queue) shouldTriggerPreemption() bool {
 	return !sq.quotaPreemptionStartTime.IsZero() && time.Now().After(sq.quotaPreemptionStartTime)
 }
 
+// ShouldApplyQuotaPreemption returns true if quota preemption should be enforced based on the settings and the
```
wilfred-s: move this into the `IncAllocatedResource()`, no need to expose this outside of the queue anyway.

pkg/scheduler/objects/queue.go
```diff
@@ -1259,19 +1277,31 @@ func (sq *Queue) TryIncAllocatedResource(alloc *resources.Resource) error {
...
+
+	// Override the earlier set quota preemption time with the configured delay.
+	// Delay clock ticking from now.
+	if isQuotaPreemptionEnabled && sq.quotaPreemptionStartTime.IsZero() && sq.ShouldApplyQuotaPreemption() {
+		sq.quotaPreemptionStartTime = time.Now().Add(sq.quotaPreemptionDelay)
+		log.Log(log.SchedQueue).Info("Overridden quota preemption time",
+			zap.String("queue", sq.QueuePath),
+			zap.Duration("delay", sq.quotaPreemptionDelay),
+			zap.Time("quotaPreemptionStartTime", sq.quotaPreemptionStartTime))
+	}
```
wilfred-s: Simplify flow:
```
	sq.updateAllocatedResourceMetrics()
	if !isQuotaPreemptionEnabled ||
		!sq.isManaged ||
		sq.quotaPreemptionDelay == 0 ||
		resources.IsZero(sq.maxResource) || 
		sq.maxResource.StrictlyGreaterThanOrEqualsOnlyExisting(sq.allocatedResource) {
		return
	}

	sq.quotaPreemptionStartTime = time.Now().Add(sq.quotaPreemptionDelay)
	log.Log(log.SchedQueue).Info("Overridden quota preemption time",
		zap.String("queue", sq.QueuePath),
		zap.Duration("delay", sq.quotaPreemptionDelay),
		zap.Time("quotaPreemptionStartTime", sq.quotaPreemptionStartTime))
```

wilfred-s (changes requested):
Two minor issues, overall looks good now

wilfred-s (approved):
LGTM

PR #1079, opened 2026-03-30
[YUNIKORN-3246] Summarize the preemption results and send event to the queue

manirajv06:
### What is this PR for?

Log the summarised preemption results and send event to the queue. 

Since there could be difference between the planned resources and resources claimed later, above info would be useful especially for parent queue to understand the whole preemption process

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3246

### Questions:

### DIFF
--- a/pkg/scheduler/objects/events/queue_events.go
+++ b/pkg/scheduler/objects/events/queue_events.go
@@ -104,6 +104,15 @@ func (q *QueueEvents) SendTypeChangedEvent(queuePath string, isLeaf bool) {
 	q.eventSystem.AddEvent(event)
 }
 
+func (q *QueueEvents) SendQuotaPreemptionEvent(queuePath string, results string, maxResource *resources.Resource) {
+	if !q.eventSystem.IsEventTrackingEnabled() {
+		return
+	}
+	event := events.CreateQueueEventRecord(queuePath, results, common.Empty, si.EventRecord_SET,
+		si.EventRecord_QUEUE_MAX, maxResource)
+	q.eventSystem.AddEvent(event)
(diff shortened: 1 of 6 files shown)


pkg/scheduler/objects/events/queue_events.go
```diff
@@ -104,6 +104,15 @@ func (q *QueueEvents) SendTypeChangedEvent(queuePath string, isLeaf bool) {
 	q.eventSystem.AddEvent(event)
 }
 
+func (q *QueueEvents) SendQuotaPreemptionEvent(queuePath string, results string, maxResource *resources.Resource) {
+	if !q.eventSystem.IsEventTrackingEnabled() {
+		return
+	}
+	event := events.CreateQueueEventRecord(queuePath, results, common.Empty, si.EventRecord_SET,
+		si.EventRecord_QUEUE_MAX, maxResource)
```
wilfred-s: The max change event should have been triggered when the config was changed. Sending it a second time without any corresponding link to the config change does not seem logical.
We should introduce a new event type: `EventRecord_QUEUE_PREEMPTION` (407)

For all victims we should already send out a `EventRecord_ALLOC_PREEMPT` (501) when the allocation is removed. That gives us both pieces: start of the quota preemption and all victims that are killed

wilfred-s (changes requested):

wilfred-s (approved):
LGTM, need a passing CI before we can commit this.
Change itself is looking good.

PR #1080, opened 2026-04-01
[YUNIKORN-3249] move quota preemption out from TryAllocate

adityadtu5:
### What is this PR for?

Moved quota preemption out from `tryallocate` scheduling cycle so that preemption would be triggered if there is any planned schedule even when there is no pending requests.

### What type of PR is it?
* [ ] - Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3249

### Questions:


pkg/scheduler/objects/queue.go
```diff
@@ -1694,6 +1676,26 @@ func (sq *Queue) TryAllocate(iterator func() NodeIterator, fullIterator func() N
...
+		log.Log(log.SchedQueue).Info("Preconditions has passed trigger preemption to enforce new max resources",
+			zap.String("queueName", sq.GetQueuePath()),
+			zap.Stringer("maxResource", sq.cloneMaxResource()))
+		preemptor := NewQuotaPreemptor(sq)
+		preemptor.tryQuotaPreemption()
+		// if quota preemption is running for this queue we do not want to trigger for any of the children.
+		// we do a top-down approach: parent first and when done we check the children
+		// there could be a child quota preemption running already
+		if !sq.getQuotaPreemptionRunning() && !sq.IsLeafQueue() {
+			for _, child := range sq.sortQueues() {
```
manirajv06: `sortQueues` returns queues for which pending requests are there. So, we cannot use it as it defeats the purpose of this pr. We can simply get children using `sq.GetCopyOfChildren()` and ensure each child is running and active.

Also, We are supposed to set `quotaPreemption` flag to false while working on the parent. Otherwise, `sq.getQuotaPreemptionRunning` condition does not work as expected. Just follow the old code especially conditional checks as is.

pkg/scheduler/objects/quota_preemptor_test.go
```diff
@@ -203,7 +203,7 @@ func TestQuotaChangeTryPreemption(t *testing.T) {
 			leaf.guaranteedResource = tc.guaranteed
 			preemptor := NewQuotaPreemptor(tc.queue)
 			preemptor.allocations = asks
-			preemptor.tryPreemption()
+			preemptor.tryQuotaPreemption()
```
manirajv06: We are already inside `Quota Preemptor`. So, I don't really see a need for renaming.

pkg/scheduler/scheduler.go
```diff
@@ -106,6 +107,21 @@ func (s *Scheduler) internalInspectOutstandingRequests() {
...
 }
 
+func (s *Scheduler) internalQuotaPreemption() {
+	for {
+		select {
+		case <-s.stop:
+			return
+		case <-time.After(time.Second):
+			if s.triggerQuotaPreemption() {
+				log.Log(log.Scheduler).Info("Quota preemption completed")
```
manirajv06: Bit misleading log statement as we are firing the method and simply returning `true` from  `triggerQuotaPreemption()` irrespective of whether the quota preemption has been really carried out or not.  If flag is enabled at partition level and none of the queue has really carried out preemption, this would flood the logs for every second.

I don't see a need for this logging as we do extensive logging inside preemptor only if there is a queue for which preemption needs to be done.

pkg/scheduler/scheduler.go
```diff
@@ -162,6 +178,19 @@ func (s *Scheduler) registerActivity() {
 	}
 }
 
+func (s *Scheduler) triggerQuotaPreemption() bool {
+	quotaPreemptionTried := false
+	for _, psc := range s.clusterContext.GetPartitionMapClone() {
+		if psc.quotaPreemptionEnabled {
```
manirajv06: We should use corresponding getter method as it read the value by acquiring a lock.

manirajv06 (changes requested):

pkg/scheduler/objects/quota_preemptor_test.go
adityadtu5: done

pkg/scheduler/scheduler.go
adityadtu5: removed the logs

pkg/scheduler/scheduler.go
adityadtu5: thanks, using getter method now.

pkg/scheduler/scheduler.go
```diff
@@ -181,7 +177,7 @@ func (s *Scheduler) registerActivity() {
 func (s *Scheduler) triggerQuotaPreemption() bool {
```
manirajv06: Do we need to return? I don't see a need.

manirajv06 (changes requested):

pkg/scheduler/scheduler.go
wilfred-s: don't think that is needed

pkg/scheduler/objects/queue.go
```diff
@@ -1694,6 +1676,28 @@ func (sq *Queue) TryAllocate(iterator func() NodeIterator, fullIterator func() N
 	return nil
 }
 
+func (sq *Queue) TryQuotaPreemption() {
```
wilfred-s, line 1710: The existing functionality allowed multiple quota preemption routines to run in parallel.
If quota preemption is started on the queue we do not check any of its children and return. We can keep checking the siblings for that queue.

Setup would look like:
```
if shouldTriggerPreemption() {
	go func() for preemption
	return // stop checking children etc
}
if isQuotaPreemptionRunning {
	return // already running stop here
}
if !sq.IsLeafQueue() {
	process children
}
```

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
wilfred-s: We need to make sure that the queue is marked correctly as preemption running, even if it is a parent queue.
Looking through the code the setting of the flag is not consistent for leafs and parents. That was hidden until now as we updated the copy of the global flag we passed down when walking the hierarchy.

That has been removed so we should set the flag after we create the go routine to process that queue, and not leave it to the `tryPreemptionInternal()` call much later. The `tryPreemptionInternal()` function btw is only called for leafs and not for parents. That is a bug that we need to fix anyway...

pkg/scheduler/objects/queue.go
adityadtu5: thanks, made changes as per suggestions

pkg/scheduler/objects/queue.go
adityadtu5: Instead of setting flag for all child, we can rely on the fact that quota preemption goes from root to leaf. So if at any time if the flag is set for a queue, we will never explore anything lower in the same tree branch.
As part of the recent changes, we set the flag during initial check itself. So there is no race condition between checking the flag and using the flag.

Please have a look again at the changes.

wilfred-s: Looking good, kicked of the workflow. Looks like this is ready to move out of draft.

pkg/scheduler/scheduler_test.go
```diff
@@ -91,3 +94,96 @@ func TestInspectOutstandingRequests(t *testing.T) {
 	assert.Equal(t, 0, noRequests)
 	assert.Assert(t, resources.IsZero(totalResources), "total resource is not zero: %v", totalResources)
 }
+
+func TestTriggerQuotaPreemption_NoPartitions(t *testing.T) {
```
manirajv06: I would prefer having single test method using t.run() approach to asserts each of these tests. Please refer other places where t.run() used.

manirajv06: I think tests like TestTriggerQuotaPreemption_EnabledButWithinQuota and TestTriggerQuotaPreemption_EnabledOverQuota_FlagSetAndCleared needs to be covered in queue_test.go and I believe it has been already covered. Please check and take it forward

pkg/scheduler/objects/quota_preemptor.go
```diff
@@ -126,8 +128,8 @@ func (qpc *QuotaPreemptionContext) tryPreemptionInternal() {
 		zap.Stringer("actual allocated resource", qpc.allocatedResource),
 		zap.Stringer("preemptable resource distribution", qpc.preemptableResource),
 	)
-	// quota change preemption has started, so mark the flag
-	qpc.queue.setQuotaPreemptionState(true)
+	// isQuotaPreemptionRunning is already set atomically by tryAcquirePreemption
+	// before this function is called; do not set it again here.
```
manirajv06: Comments not required here as it may be required in place where state is being set

pkg/scheduler/objects/queue_test.go
```diff
@@ -3268,6 +3317,512 @@ func TestQueue_setPreemptionTime(t *testing.T) {
...
+	app := newApplication(appID, "default", leaf.QueuePath)
+	app.SetQueue(leaf)
+	leaf.AddApplication(app)
+	ask := newAllocationAsk(askKey, appID, res)
+	err := app.AddAllocationAsk(ask)
+	assert.NilError(t, err)
+	return app
+}
+
+func TestTryAllocate_LeafNoApps(t *testing.T) {
```
manirajv06: I would prefer having single test method using t.run() approach to asserts each of these tests rather than having separate method for each one. Please refer other places where t.run() used.

pkg/scheduler/objects/queue.go
```diff
@@ -496,36 +496,38 @@ func (sq *Queue) setPreemptionTime(oldMaxResource *resources.Resource, oldDelay
...
+	if sq.quotaPreemptionStartTime.IsZero() || time.Now().Before(sq.quotaPreemptionStartTime) {
+		return false
+	}
+	sq.isQuotaPreemptionRunning = true
+	return true
+}
+
+// ResetPreemptionTime reset the quota preemption variables
+// Only for testing
+func (sq *Queue) ResetPreemptionTime() {
```
manirajv06: This method has been added as part of another pr. please do rebase and clean up.

manirajv06 (changes requested):

pkg/scheduler/objects/queue.go
adityadtu5: done

pkg/scheduler/scheduler_test.go
adityadtu5: Removed these for now, will add them later if required.

pkg/scheduler/objects/quota_preemptor.go
adityadtu5: this is for informing the caller to use `tryAcquirePreemption` before calling this method. I will add it at the top of the function to make it more explicit and clear.

pkg/scheduler/objects/queue_test.go
adityadtu5: I have removed these tests for now as there are no new changes in `TryAllocate` as part of this PR.
If needed I will add them later

pkg/scheduler/objects/queue_test.go
```diff
@@ -2451,18 +2451,22 @@ func TestQuotaPreemptionSettings(t *testing.T) {
 			parent.UpdateQueueProperties(oldMax)
 			// Wait till delay expires to let trigger preemption automatically
 			time.Sleep(parent.quotaPreemptionDelay + 50*time.Millisecond)
-			assert.Equal(t, parent.shouldTriggerPreemption(), tc.timeChange, "preemption should get trigger for set delay")
-			parent.TryAllocate(nil, nil, nil, false, true)
+			triggered := parent.tryAcquirePreemption()
+			assert.Equal(t, triggered, tc.timeChange, "preemption should get trigger for set delay")
+			// Simulate preemption completing: clears isQuotaPreemptionRunning and quotaPreemptionStartTime.
+			if triggered {
+				parent.setQuotaPreemptionState(false)
```
manirajv06: We already have a method to reset the values. Can we use it here?

pkg/scheduler/objects/queue_test.go
```diff
@@ -2531,18 +2553,44 @@ func TestShouldTriggerPreemption(t *testing.T) {
...
+		{"nil max resource with usage and past start time", nilMaxQueue, true},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			assert.Equal(t, tc.queue.shouldTriggerPreemption(), tc.preconditionResult)
+			result := tc.queue.tryAcquirePreemption()
+			assert.Equal(t, result, tc.preconditionResult)
+			// Reset the flag if acquired so it doesn't affect other test cases.
+			if result {
+				tc.queue.setQuotaPreemptionState(false)
```
manirajv06: Same comment

pkg/scheduler/objects/quota_preemptor.go
```diff
@@ -137,9 +139,7 @@ func (qpc *QuotaPreemptionContext) tryPreemptionInternal() {
 
 	// Preempt the victims
 	qpc.preemptVictims()
-
-	// quota change preemption has ended, so mark the flag
-	qpc.queue.setQuotaPreemptionState(false)
+	// isQuotaPreemptionRunning is cleared by the caller (via defer setQuotaPreemptionState(false)).
```
manirajv06: Comments not required

pkg/scheduler/objects/quota_preemptor_test.go
```diff
@@ -607,6 +608,190 @@ func TestQuotaChangeTryPreemptionForParentQueue(t *testing.T) {
 	}
 }
 
+// TestTryPreemptionInternal tests tryPreemptionInternal which filters, sorts, and preempts victims
+// based on the preemptable resource that must be set before calling this function.
+// nolint:funlen
+func TestTryPreemptionInternal(t *testing.T) {
```
manirajv06: We do have a test called `TestQuotaChangeTryPreemption..` in 3 different vairants. I think some of these different test cases should go into `TestQuotaChangeTryPreemption`. Where as others are already covered in `TestQuotaChangeFilterVictims`. Please check and decide accordingly.

pkg/scheduler/scheduler_test.go
```diff
@@ -91,3 +95,167 @@ func TestInspectOutstandingRequests(t *testing.T) {
...
+	for _, event := range testHandler.GetEvents() {
+		if _, ok := event.(*rmevent.RMReleaseAllocationEvent); ok {
+			t.Fatal("unexpected release event when quota preemption is disabled")
+		}
+	}
+}
+
+// TestTriggerQuotaPreemption_QuotaPreemptionEnabled verifies that triggerQuotaPreemption fires
+// preemption and produces release events when the queue is over max and the preemption delay has elapsed.
+func TestTriggerQuotaPreemption_QuotaPreemptionEnabled(t *testing.T) {
```
manirajv06: Can this and above test methods be clubbed together using t.run() approach?

pkg/scheduler/scheduler_test.go
```diff
@@ -91,3 +95,167 @@ func TestInspectOutstandingRequests(t *testing.T) {
...
+// partition map cleanly without panicking.
+func TestTriggerQuotaPreemption_NoPartitions(t *testing.T) {
+	scheduler := NewScheduler()
+	// clusterContext has no partitions; call should be a no-op
+	scheduler.triggerQuotaPreemption()
+}
+
+// TestTriggerQuotaPreemption_UsageBelowMax verifies that triggerQuotaPreemption does not trigger
+// preemption when the queue usage is within the configured max resource limit.
+func TestTriggerQuotaPreemption_UsageBelowMax(t *testing.T) {
```
manirajv06: I believe this MUST be already covered in queue_test.go as part of testing `tryAcquirePreemption()`

pkg/scheduler/scheduler_test.go
```diff
@@ -91,3 +95,167 @@ func TestInspectOutstandingRequests(t *testing.T) {
...
+	for _, event := range testHandler.GetEvents() {
+		if _, ok := event.(*rmevent.RMReleaseAllocationEvent); ok {
+			t.Fatal("unexpected release event when usage is below max")
+		}
+	}
+}
+
+// TestTriggerQuotaPreemption_PreemptionDelayNotElapsed verifies that no preemption occurs when
+// the queue is over quota but the configured preemption delay has not elapsed yet.
+func TestTriggerQuotaPreemption_PreemptionDelayNotElapsed(t *testing.T) {
```
manirajv06: I believe this MUST be already covered in queue_test.go as part of testing `tryAcquirePreemption()`

manirajv06 (changes requested):

pkg/scheduler/objects/queue_test.go
adityadtu5: Actually this is not needed, I will remove this.

pkg/scheduler/objects/queue_test.go
adityadtu5: which method you are refering to?

pkg/scheduler/objects/quota_preemptor.go
adityadtu5: I think it is good to explicitly call out the requirements for calling this function.
However, I think it is better to place it as a method comment so moving it to the top.

pkg/scheduler/scheduler_test.go
adityadtu5: this is to cover new lines added in scheduler.go.
Also in `tryAcquirePreemption()` we are not asserting for events.

pkg/scheduler/scheduler_test.go
adityadtu5: same as above

pkg/scheduler/objects/quota_preemptor_test.go
adityadtu5: I have removed these test cases, and improved the existing ones by asserting on exact allocations preempted (not just count)

pkg/scheduler/scheduler_test.go
adityadtu5: These all tests have different setup and different assertions. Trying to generalize them in a common struct which can be used as test table will be very messy I think.

pkg/scheduler/objects/queue_test.go
manirajv06: I meant `resetPreemption`

pkg/scheduler/objects/quota_preemptor.go
manirajv06: Just because we are moving the state change from here to the caller in this PR, we are thinking that commenting here would help us in walking through the code down the road or few months later. In fact, code walkthrough starts from the caller. Setting the state to true happens in `tryAcquire`. So, may be question raises at least over there "where do we reset it back to false?" but not definitely for any fresh eyes.

pkg/scheduler/objects/quota_preemptor_test.go
manirajv06: Would it be better to improve the existing methods to assert exact allocations in addition to the count as intention is to keep the tests simple, clean and intact? If not now, may be through a new jira later.

pkg/scheduler/scheduler_test.go
manirajv06: I would focus only on two things here:

1. Enabled at partition level
2. Disabled at partition level

Not digging too much on functionality of the methods called internally from `triggerQuotaPreemption`. Covering `psc.root.TryQuotaPreemption()` functionality here in detail is something can be avoided.

pkg/scheduler/scheduler_test.go
manirajv06: I believe events are already covered as part of https://github.com/apache/yunikorn-core/pull/1079.

Please check and see if you can add on top of it over there for any missing functionalities

pkg/scheduler/scheduler_test.go
manirajv06: I think this is already covered as part of https://github.com/apache/yunikorn-core/pull/1080/changes#diff-5dffc593356154a4418edb17409dc0c820d13c7d72c74208cadb3d996b7e8722R2551

Isn't it?

manirajv06 (changes requested):

pkg/scheduler/objects/queue_test.go
adityadtu5: I could not find method with this name

pkg/scheduler/objects/quota_preemptor.go
adityadtu5: comments are there at the calling place as well. 
Over commenting is fine than under commenting.

pkg/scheduler/scheduler_test.go
adityadtu5: Intention here is only to verify that `TryQuotaPreemption` was called (and not to check further internal implementation). This is the simplest way I could think without using mocks (which default testing package does not support) to verify this.

pkg/scheduler/scheduler_test.go
adityadtu5: Here focus is not about events. Focus is to check whether Quota preemption was triggered or not. Checking events is the proxy to do that.

adityadtu5: Removing this as this should be covered as part of `tryAcquirePreemption`

pkg/scheduler/scheduler_test.go
adityadtu5: Removing this as this should be covered as part of `tryAcquirePreemption`

wilfred-s: Does not matter if we have overlapping tests.
We have seen cases before that the unit tests pass but then a higher level test for the same scenario fails. Often the higher level tests surface possible race conditions earlier.

wilfred-s (approved):
LGTM, leaving finall approval and commit to @manirajv06

pkg/scheduler/objects/queue_test.go
manirajv06: `ResetPreemptionTime`

pkg/scheduler/scheduler_test.go
manirajv06: Exactly.

Lets have only TestTriggerQuotaPreemption method instead of 2 different methods.

Two test cases to be asserted:

1. Enabled
2.Disabled

Execute these two cases with different partiton config - one with ON, another with OFF. Assert release events only. events should be > zero for first and equlas zero for later one.

manirajv06:
@adityadtu5 Thanks for being patience with the reviews. We are almost there. It has taken an good shape now. Few more minor changes.

manirajv06 (changes requested):

pkg/scheduler/objects/queue_test.go
adityadtu5: that is for resetting preemption time. It does not reset the preemptionrunning flag.

manirajv06: Method has been added to reset the preemption variables. Only for testing purpose. You can reset running flag or any other preemption variables.

adityadtu5: done

pkg/scheduler/scheduler_test.go
adityadtu5: done

manirajv06 (approved):
looks good, pending tests.

fix lint warnings

PR #1081, opened 2026-04-02
[YUNIKORN-3250] Fix UserGroupCache entries never expiring due to uninitialized clock

mitdesai:
### What is this PR for?
The `UserGroupCache` contained a package-level variable `var now time.Time` that was
declared but never assigned, causing `now.Unix()` to always return the Go zero-time
constant (-62135596800). This made every TTL eviction condition in `cleanUpCache()`
permanently false, so cache entries — including stale LDAP group memberships — lived
forever and could only be cleared by a full scheduler restart.

A second bug allowed `GetUserGroup` to return an expired positive cache entry without
any age check, relying entirely on the background cleaner to evict it first (up to 60s
overage). This also meant that when a stale entry fell through to re-resolution, its
existing `Groups` slice was reused, causing group duplication.

This PR fixes both issues and adds regression tests that fail against the unfixed code.

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-3250
 
1. `go test ./pkg/common/security/... -v` — all existing tests must pass
2. `TestCleanUpCacheUsesRealTime` — verifies `cleanUpCache` evicts entries stamped with a real past timestamp; fails without the `now.Unix()` fix
3. `TestPositiveCacheHitExpiryTriggersRefresh` — verifies `GetUserGroup` re-resolves an expired positive entry and returns a fresh `resolved` timestamp with no duplicate groups; fails without the inline TTL check
4. Integration: configure the LDAP resolver, resolve a user, wait >5 minutes, re-submit a job — confirm a fresh LDAP lookup occurs without a scheduler restart

### Questions:

### DIFF
--- a/pkg/common/security/usergroup.go
+++ b/pkg/common/security/usergroup.go
@@ -41,7 +41,6 @@ const (
 )
 
 // global variables
-var now time.Time            // One clock to access
 var instance *UserGroupCache // The instance of the cache
 var once = &sync.Once{}      // Make sure we can only create the cache once
 var stopped atomic.Bool      // whether UserGroupCache is stopped (needed for multiple partitions)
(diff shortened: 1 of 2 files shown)


manirajv06 (approved):
looks good

PR #1083, opened 2026-04-27
[YUNIKORN-3257] Stop e2e test for plugin mode

manirajv06:
### What is this PR for?

Clean up plugin code in 
https://github.com/apache/yunikorn-core/blob/master/.github/workflows/pre-commit.yml
e2e test should only be run for standard deployment.

### What type of PR is it?
* [ ] - Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3257

### Questions:

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -49,7 +49,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        k8s: [v1.32.2]
+        k8s: [v1.34.0]
     steps:
       - name: Checkout yunikorn-k8shim source code
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


wilfred-s (approved):
Change is good, first step for plugin removal
However this is not the weekly run.

This should be tracked under YUNIKORN-3257. The weekly run is purely managed from the k8shim repository.

wilfred-s:
@manirajv06 Updated the comment and description. Please make sure we the commit points to the correct jira.

.github/workflows/pre-commit.yml
```diff
@@ -89,10 +88,9 @@ jobs:
         run: cd k8shim && ./scripts/run-e2e-tests.sh -a "test" -n "yk8s" -v "kindest/node:${KIND_NODE_IMAGE}" ${KIND_EXTRA_ARGS}
```
adityadtu5: shall we remove `KIND_EXTRA_ARGS` from here as well.

adityadtu5 (commented):
LG, one comment, please check

.github/workflows/pre-commit.yml
```diff
@@ -50,7 +50,6 @@ jobs:
       fail-fast: false
       matrix:
         k8s: [v1.32.2]
```
wilfred-s: This should be 1.34 not 1.32

wilfred-s (changes requested):

wilfred-s (approved):
LGTM

PR #1086, opened 2026-05-15
[YUNIKORN-3245] Reset queue properties on config update

ayubpathan:
### What is this PR for?
Reset queue fields derived from properties before applying the current effective property map during queue config updates. This lets removed properties return to defaults while preserving inherited parent properties for existing child queues on reload.

### What type of PR is it?
* [x] - Bug Fix

### Todos
* [x] - Tests

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3245

* make license-check
* make pseudo
* make check_scripts
* make lint
* make test

N/A

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -421,8 +421,12 @@ func (sq *Queue) applyConf(conf configs.QueueConfig, silence bool) (*resources.R
 // This function MUST be called holding the lock for the queue.
 func (sq *Queue) setPreemptionTime(oldMaxResource *resources.Resource, oldDelay time.Duration) {
 	// if quota preemption is running nothing we do not have an influence
+	if sq.isQuotaPreemptionRunning {
+		return
+	}
 	// if the delay for the queue is not set do not trigger preemption
-	if sq.isQuotaPreemptionRunning || sq.quotaPreemptionDelay == 0 {
+	if sq.quotaPreemptionDelay == 0 {
+		sq.quotaPreemptionStartTime = time.Time{}
 		return
 	}
 	// if no current limit we should not preempt even if it was set earlier, clear the start time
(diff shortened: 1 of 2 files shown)


wilfred-s:
@ayubpathan this seems to be a different solution for the same issue as is fixed in #1088. The merge code for the properties should handle the reset also.
Can you please check that?
If soo we can handle the cleanup of the mergeProperties here (see this [comment](https://github.com/apache/yunikorn-core/pull/1088#discussion_r3316169266) to make that reset nice and clean.

ayubpathan:
Thanks @wilfred-s, I checked #1088. The inherited-property reload part of my previous patch overlaps with YUNIKORN-3285, so I removed the ApplyConf merge change and the related inherited-property test from this PR.

This PR is now narrowed back to YUNIKORN-3245: resetting runtime fields derived from the effective queue properties in UpdateQueueProperties(), so removed properties revert to their defaults on config update.

I kept the mergeProperties cleanup out of this revision to avoid duplicating #1088. If you prefer that cleanup to be handled here instead, I can adjust.

pkg/scheduler/objects/queue.go
```diff
@@ -662,6 +662,18 @@ func (sq *Queue) UpdateQueueProperties(oldMaxResource *resources.Resource) {
 		sq.sortType = policies.FifoSortPolicy
 		return
 	}
+	// Reset fields derived from properties so removed properties revert to defaults.
```
wilfred-s: Can we move setting all these defaults into a single function: `sq.resetProperties()` ?

pkg/scheduler/objects/queue.go
```diff
@@ -727,20 +739,26 @@ func (sq *Queue) UpdateQueueProperties(oldMaxResource *resources.Resource) {
...
 				zap.String("value", value))
 		}
 	}
+	if quotaPreemptionDelaySet || oldDelay != sq.quotaPreemptionDelay {
+		if sq.quotaPreemptionDelay == 0 {
+			sq.quotaPreemptionStartTime = time.Time{}
+			return
+		}
+		sq.setPreemptionTime(oldMaxResource, oldDelay)
+	}
```
wilfred-s: This can be hidden inside the `sq.setPreemptionTime(oldMaxResource, oldDelay)` call and always call that method with a small change inside that method:
* if preemption is running the start time will be cleaned up automatically
* if preemption is not running and delay is not set now clear out the start time
* all other cases let it flow as it was

Removes the need to track `quotaPreemptionDelaySet`

wilfred-s (changes requested):

ayubpathan:
Addressed the review comments:

* Moved queue property defaults into resetProperties().
* Simplified quota preemption handling so UpdateQueueProperties() always calls setPreemptionTime() and the zero-delay cleanup is handled inside that method.
* Added coverage that removing quota.preemption.delay clears the queued preemption start time.
* Rebased on current master, which includes #1088 and #1091.

Validation:
* go test ./pkg/scheduler/objects -run TestUpdateQueuePropertiesReset|TestQuotaPreemptionSettings|TestQueue_setPreemptionTime|TestQueueBackoffProperties
* make lint
* make test

wilfred-s (approved):
LGTM +1

PR #1088, opened 2026-05-20
[YUNIKORN-3285] Child queue  is not inheriting parent queue propertie…

adityadtu5:
…s during queue update

### What is this PR for?
- Merging parent queue property to child queue during queue update
- Added UT for it   

### What type of PR is it?
* [x] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-3285

UTs are part of this PR

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -244,12 +244,12 @@ func (sq *Queue) applyTemplate(childTemplate *template.Template) {
 // getProperties returns a copy of the properties for this queue
 // Will never return a nil, can return an empty map.
 func (sq *Queue) getProperties() map[string]string {
+	props := make(map[string]string)
 	if sq == nil {
-		return make(map[string]string)
+		return props
 	}
 	sq.RLock()
 	defer sq.RUnlock()
-	props := make(map[string]string)
 	for key, value := range sq.properties {
 		props[key] = value
 	}
@@ -267,12 +267,11 @@ func (sq *Queue) GetProperties() map[string]string {
 // config reload to re-apply inherited properties from the parent.
 // Lock protected.
 func (sq *Queue) MergeParentProperties(config map[string]string) {
-	if sq.parent != nil {
-		parentProps := sq.parent.getProperties()
-		sq.Lock()
-		defer sq.Unlock()
-		sq.mergeProperties(parentProps, config)
-	}
+	// get parent properties outside of the lock to avoid potential deadlocks with parent queue lock.
+	parentProps := sq.parent.getProperties()
+	sq.Lock()
+	defer sq.Unlock()
+	sq.mergeProperties(parentProps, config)
 }
 
 // mergeProperties merges the properties from the parent queue and the config in the set from new queue
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -253,6 +253,22 @@ func (sq *Queue) getProperties() map[string]string {
 	return props
 }
 
+// GetProperties returns a copy of the properties for this queue.
+// Will never return nil; can return an empty map.
+func (sq *Queue) GetProperties() map[string]string {
+	return sq.getProperties()
+}
+
+// MergeProperties merges the parent queue properties with config properties into this queue.
+// Config properties override parent properties. This should be called after ApplyConf during
+// config reload to re-apply inherited properties from the parent.
+// Lock protected.
+func (sq *Queue) MergeProperties(parent, config map[string]string) {
+	sq.Lock()
+	defer sq.Unlock()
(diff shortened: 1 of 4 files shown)


pkg/scheduler/objects/queue.go
wilfred-s, line 286: We should optimise this as the passed in list of parent properties is a clean copy there is no reason to create a new copy. the filtering we can then do directly on the map entries inside the merge code.
File a jira for that, please

pkg/scheduler/objects/queue.go
```diff
@@ -253,6 +253,22 @@ func (sq *Queue) getProperties() map[string]string {
...
+// Will never return nil; can return an empty map.
+func (sq *Queue) GetProperties() map[string]string {
+	return sq.getProperties()
+}
+
+// MergeProperties merges the parent queue properties with config properties into this queue.
+// Config properties override parent properties. This should be called after ApplyConf during
+// config reload to re-apply inherited properties from the parent.
+// Lock protected.
+func (sq *Queue) MergeProperties(parent, config map[string]string) {
```
wilfred-s: proposal for a different signature and implementation:
```
func (sq *Queue) MergeProperties(config map[string]string) {
	parentProp := sq.parent.getProperties()
	sq.Lock()
	defer sq.Unlock()
	sq.mergeProperties(parentProp, config)
}
```
This saves us exposing the get properties in an exported function

wilfred-s (changes requested):

pkg/scheduler/objects/queue.go
adityadtu5: Updated the code as per suggestion

pkg/scheduler/objects/queue.go
```diff
@@ -253,6 +253,25 @@ func (sq *Queue) getProperties() map[string]string {
...
+	return sq.getProperties()
+}
+
+// MergeParentProperties merges the parent queue properties with config properties into this queue.
+// Config properties override parent properties. This should be called after ApplyConf during
+// config reload to re-apply inherited properties from the parent.
+// Lock protected.
+func (sq *Queue) MergeParentProperties(config map[string]string) {
+	sq.Lock()
+	defer sq.Unlock()
```
wilfred-s, line 273: This is dangerous! In other locations we have the order of locks in the opposite direction: parent first then child queue. This could cause a deadlock.
Solution: fix `getProperties()` by making it nil safe and use a read lock in the function (currently a write lock). Move the `sq.parent.getProperties()` call outside of the local queue lock.
Brings it inline with the way we doo it for other parent info retrieval.

pkg/scheduler/objects/queue_test.go
```diff
@@ -44,6 +44,167 @@ import (
...
+func TestMergeParentPropertiesNilParent(t *testing.T) {
+	// root queue has no parent, MergeParentProperties should be a no-op
+	root, err := createRootQueue(nil)
+	assert.NilError(t, err, "root queue create failed")
+	root.properties = map[string]string{"key": "value"}
+
+	root.MergeParentProperties(map[string]string{"other": "other-value"})
+
+	// properties should remain unchanged since root has no parent to merge from
+	props := root.GetProperties()
```
wilfred-s: Should not use the exported function (repeated all through the file)

wilfred-s (changes requested):

wilfred-s (approved):
LGTM

pkg/scheduler/objects/queue_test.go
```diff
@@ -44,6 +44,167 @@ import (
...
+	// root queue has no parent, MergeParentProperties should be a no-op
+	root, err := createRootQueue(nil)
+	assert.NilError(t, err, "root queue create failed")
+	root.properties = map[string]string{"key": "value"}
+
+	root.MergeParentProperties(map[string]string{"other": "other-value"})
+
+	// properties should remain unchanged since root has no parent to merge from
+	props := root.getProperties()
+	assert.Equal(t, "value", props["key"], "root queue properties should not be modified by MergeParentProperties")
```
wilfred-s: I overlooked one mistake in this test: line 53 replaces the properties via the config on the root queue. The assumption inline 55 is thus not correct as the replacement is not triggered by the parent but by the new config value passed in.
We should have the same map content passed in as the config value to leave the current value on the root queue. Otherwise we should expect the newly passed in list from the config to be present

wilfred-s (changes requested):

wilfred-s (approved):
LGTM +1
Waiting for build to finish before commit

PR #1091, opened 2026-06-08
[YUNIKORN-3291] Remove redundant config passing in mergeProperties

venkateshwaracholan:
### What is this PR for?

Remove redundant config passing in mergeProperties.

ApplyConf already stores a queue's configuration in sq.properties. This change makes mergeProperties use those existing properties directly instead of receiving the same configuration map as a parameter.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [x] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3291

Unit tests are included in this PR.

### Questions:

### DIFF
--- a/pkg/scheduler/objects/queue.go
+++ b/pkg/scheduler/objects/queue.go
@@ -267,11 +267,8 @@ func (sq *Queue) GetProperties() map[string]string {
 // This should be called after ApplyConf during config reload to re-apply inherited properties
 // from the parent. Lock protected.
 func (sq *Queue) MergeParentProperties() {
-	var parentProps map[string]string
-	if sq.parent != nil {
-		// get parent properties outside of the lock to avoid potential deadlocks with parent queue lock
-		parentProps = sq.parent.getProperties()
-	}
+	// get parent properties outside of the lock to avoid potential deadlocks with parent queue lock
+	parentProps := sq.parent.getProperties()
 	sq.Lock()
 	defer sq.Unlock()
 	sq.mergeProperties(parentProps)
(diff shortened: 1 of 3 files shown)


pkg/scheduler/objects/queue.go
```diff
@@ -262,21 +262,26 @@ func (sq *Queue) GetProperties() map[string]string {
...
+// MergeParentProperties merges the parent queue properties with this queue's config properties.
+// Config properties already set on the queue (from ApplyConf) override parent properties.
+// This should be called after ApplyConf during config reload to re-apply inherited properties
+// from the parent. Lock protected.
+func (sq *Queue) MergeParentProperties() {
+	var parentProps map[string]string
+	if sq.parent != nil {
+		// get parent properties outside of the lock to avoid potential deadlocks with parent queue lock
+		parentProps = sq.parent.getProperties()
+	}
```
wilfred-s: `getProperties()` is nil safe no need to test here or initialise the map, that is all handled already.

pkg/scheduler/objects/queue.go
```diff
@@ -262,21 +262,26 @@ func (sq *Queue) GetProperties() map[string]string {
...
 
-// mergeProperties merges the properties from the parent queue and the config in the set from new queue
+// mergeProperties merges filtered parent properties with this queue's config properties.
+// Config properties already set on the queue (from applyConf) override parent properties.
 // lock free call
-func (sq *Queue) mergeProperties(parent, config map[string]string) {
+func (sq *Queue) mergeProperties(parent map[string]string) {
+	config := sq.properties
 	// clean out all existing values (handles update case)
 	sq.properties = make(map[string]string)
```
wilfred-s: The map returned from `getProperties()` is a clean copy that we can use.
Store the results in the parent map. It will become the "resulting" map, no new map needed etc.

pkg/scheduler/objects/queue.go
wilfred-s: let range handle a nil or empty map

wilfred-s: let range handle the nil or empty map

wilfred-s: store everything in the passed in `parent` map
replace `sq.properties` in a final action with `parent`

wilfred-s (changes requested):

venkateshwaracholan:
Thanks for the review. I have addressed the comments and pushed an update.

wilfred-s (approved):
LGTM +1

wilfred-s:
kicking off workflow will commit when the tests pass.

wilfred-s:
Filed follow up jira for the e2e test failure. The change should not affect that test

PR #1093, opened 2026-06-10
YUNIKORN-3296: Fix race between AddEvent and Stop

venkateshwaracholan:
What is this PR for?

Fix a data race in EventSystemImpl.AddEvent() during shutdown.

Stop() closes and clears ec.channel while holding the event system mutex. AddEvent() previously accessed ec.channel without synchronization. During TestSchedulerRecoveryQuotaPreemption, scheduler goroutines could still emit events while StopAll() shut down the event system, resulting in a race detected by go test -race.

The reported race was between:

EventSystemImpl.AddEvent()
EventSystemImpl.Stop() (close(ec.channel) / ec.channel = nil)

This race could also lead to a send on closed channel panic.

What type of PR is this?
Bug Fix
What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-3296

How should this be tested?

Run:

go test -race ./pkg/events/...
go test -race -count=20 -run TestSchedulerRecoveryQuotaPreemption ./pkg/scheduler/tests

A new regression test, TestAddEventConcurrentStop, exercises concurrent calls to AddEvent() and Stop().

What does this PR do?

Synchronize AddEvent() with Stop() using the same mutex. If the event system has already been stopped or the channel has been cleared, the event is not sent and is counted as not channeled.

### Questions:

### DIFF
--- a/pkg/events/event_system.go
+++ b/pkg/events/event_system.go
@@ -220,6 +220,10 @@ func (ec *EventSystemImpl) Stop() {
 	ec.stopped = true
 }
 
+func (ec *EventSystemImpl) isStopped() bool {
+	return ec.stopped || ec.channel == nil
+}
+
 // AddEvent adds an event record to the event system. See the interface for details.
 func (ec *EventSystemImpl) AddEvent(event *si.EventRecord) {
 	if event != nil {
@@ -228,10 +232,10 @@ func (ec *EventSystemImpl) AddEvent(event *si.EventRecord) {
 
 	metrics.GetEventMetrics().IncEventsCreated()
 
-	ec.Lock()
-	defer ec.Unlock()
+	ec.RLock()
+	defer ec.RUnlock()
 
-	if ec.stopped || ec.channel == nil {
+	if ec.isStopped() {
 		metrics.GetEventMetrics().IncEventsNotChanneled()
 		return
 	}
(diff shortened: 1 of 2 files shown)


pkg/events/event_system.go
```diff
@@ -227,6 +227,15 @@ func (ec *EventSystemImpl) AddEvent(event *si.EventRecord) {
...
 
 	metrics.GetEventMetrics().IncEventsCreated()
+
+	ec.Lock()
+	defer ec.Unlock()
+
+	if ec.stopped || ec.channel == nil {
+		metrics.GetEventMetrics().IncEventsNotChanneled()
+		return
+	}
```
wilfred-s: We should not need more than a read lock.

Factor this out into a method:
```
func (ec *EventSystemImpl) isStopped() bool {
	return ec.stopped || ec.channel == nil
}
```
call
```
	if ec.isStopped() {
		metrics.GetEventMetrics().IncEventsNotChanneled()
		return
	}
```

wilfred-s (changes requested):

pkg/events/event_system.go
venkateshwaracholan: Thanks for the suggestion. Switched AddEvent() to use RLock/RUnlock and factored the state check into isStopped(). Re-ran the race tests and they pass.

wilfred-s (approved):
LGTM +1

PR #1095, opened 2026-06-17
[YUNIKORN-3239] handling infra events and allocation events separately

adityadtu5:
### What is this PR for?
In this PR, we are making changes to handle infra events, node events and allocation events separately. 
This will prevent node and allocation events to be getting blocked during spike in infra relate events.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3239

Test cases are added to the PR

### Questions:

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -32,7 +32,7 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
-// The Scheduler service that starts the needed sub services
+// Scheduler service that starts the needed sub services
 type Scheduler struct {
 	clusterContext     *ClusterContext  // main context
 	pendingAllocEvents chan interface{} // queue for allocation and application events
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -32,7 +32,7 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
-// Main Scheduler service that starts the needed sub services
+// The Scheduler service that starts the needed sub services
 type Scheduler struct {
 	clusterContext     *ClusterContext  // main context
 	pendingAllocEvents chan interface{} // queue for allocation and application events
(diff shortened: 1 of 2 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -120,7 +123,12 @@ func (s *Scheduler) internalQuotaPreemption() {
 
 // Implement methods for Scheduler events
 func (s *Scheduler) HandleEvent(ev interface{}) {
```
wilfred-s: fix this comment also,
can we add why the split is made here why?

pkg/scheduler/scheduler.go
```diff
@@ -32,32 +32,35 @@ import (
...
+	healthChecker      *HealthChecker
+	nodesMonitor       *nodesResourceUsageMonitor
 }
 
 func NewScheduler() *Scheduler {
 	m := &Scheduler{}
 	m.clusterContext = newClusterContext()
-	m.pendingEvents = make(chan interface{}, 1024*1024)
+	m.pendingAllocEvents = make(chan interface{}, 1024*1024)
+	m.pendingInfraEvents = make(chan interface{}, 1024*1024)
```
wilfred-s: Do we need 1 Million events to be buffered in each channel?

wilfred-s (changes requested):
Why not split the node events from the rest of infra into its own channel? Config updates can be large and take a time when 1000s of queues are used. Add some comments on why split like you did.

pbacsko:
Typo in commit "sepately"

pkg/scheduler/scheduler.go
adityadtu5: Updated the capacity as per our last discussion

pkg/scheduler/scheduler.go
adityadtu5: fixed it in the latest commit

adityadtu5:
> Why not split the node events from the rest of infra into its own channel? Config updates can be large and take a time when 1000s of queues are used. Add some comments on why split like you did.

Updated the PR with the changes we discussed.

adityadtu5:
> Typo in commit "sepately"

Thanks, fixed it.

wilfred-s (approved):
LGTM +1
waiting for the build to finish before committing

PR #1096, opened 2026-06-19
[YUNIKORN-3300] Use feasible nodes returned from K8s PreFilter plugin…

manirajv06:
### Description

Unlike before, two API's needs to be called to run the predicate checks to decide the node is feasible to run the pod or not. 

1. Predicates Pre Filter
2. Predicates Filter

K8s Pre filter API for the given pod returns feasible nodes with an error (if any). Used the nodes list as preventive measure to decide whether node should be used further to call the next filter API or not in all scheduling cycles like Allocate, Reserved Allocate, Placeholder, Required Node. In case of any errors, return immediately without even iterating the node iterator in some of these cycle flow. Node being traversed as part of traversing node iterator would be skipped if it is not available in feasible nodes list. Empty feasible nodes with nil errors means all nodes are eligible to run the pod.

Modified existing tests based on these changes. Added more new tests to test the scheduling cycles.

### Type of change

- [ ] Bug Fix
- [X] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3300

- [ ] The Jira ID is part of the title of this pull request.

### AI Tooling
If an AI tool was used:

### How has this been tested?

### Questions:


pbacsko:
@manirajv06 is this PR reviewable? Please make the SI changes so that it compiles.

manirajv06:
> @manirajv06 is this PR reviewable? Please make the SI changes so that it compiles.

[SI changes](https://github.com/apache/yunikorn-scheduler-interface/pull/161) are already up for review. Please take a look. Once SI change merges, this pr should pass.

manirajv06:
Updated SI pseudo version. Waiting for the test runs.

manirajv06:
Codecov warnings could be ignored as those missed lines in mocks are covered indirectly from application and other related tests.

pkg/scheduler/objects/application.go
```diff
@@ -1321,10 +1332,22 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator,
 			if !node.preAllocateCheck(reqFit.GetAllocatedResource(), resKey) {
 				return true
 			}
+
+			// Is this node suitable to run the pod?
+			if len(feasibleNodes) > 0 {
+				if _, ok := feasibleNodes[node.NodeID]; !ok {
+					log.Log(log.SchedApplication).Debug("skipping node as it is not feasible to run the pod",
+						zap.String("allocationKey", resKey),
+						zap.String("node", node.NodeID))
```
pbacsko: I wouldn't log this even on DEBUG level. It can generate too much output on a bigger cluster.

pkg/scheduler/objects/application.go
```diff
@@ -1525,6 +1558,17 @@ func (sa *Application) tryNodesNoReserve(ask *Allocation, iterator NodeIterator,
 		if !node.FitInNode(ask.GetAllocatedResource()) || node.NodeID == reservedNode {
 			return true
 		}
+
+		// Is this node suitable to run the pod?
+		if len(feasibleNodes) > 0 {
+			if _, ok := feasibleNodes[node.NodeID]; !ok {
+				log.Log(log.SchedApplication).Debug("skipping node as it is not feasible to run the pod",
+					zap.String("allocationKey", ask.GetAllocationKey()),
+					zap.String("node", node.NodeID))
```
pbacsko, line 1644: I wouldn't log this even on DEBUG level. It can generate too much output on a bigger cluster.

pkg/scheduler/objects/application.go
```diff
@@ -1564,41 +1618,56 @@ func (sa *Application) tryNodes(ask *Allocation, iterator NodeIterator) *Allocat
...
+
+		// Is there any pod predicate errors? No node would be picked up for allocation in case of any errors
+		// and better to get into process of picking up a node for reservation right away
+		if len(podPredicateErrors) == 0 {
+			// Is this node suitable to run the pod?
+			if len(feasibleNodes) > 0 {
+				if _, ok := feasibleNodes[node.NodeID]; !ok {
+					log.Log(log.SchedApplication).Debug("skipping node as it is not feasible to run the pod",
+						zap.String("allocationKey", allocKey),
+						zap.String("node", node.NodeID))
```
pbacsko: I wouldn't log this even on DEBUG level. It can generate too much output on a bigger cluster.

pkg/scheduler/objects/node.go
```diff
@@ -487,6 +488,23 @@ func (sn *Node) preAllocateConditions(ask *Allocation) error {
...
+	if err != nil {
+		preErrors := make(map[string]int, 1)
+		preErrors[err.Error()]++
+		ask.SendPredicatesFailedEvent(preErrors)
+		return err
+	}
+	// Is this node suitable to run the pod?
+	if len(feasibleNodes) > 0 {
+		if _, ok := feasibleNodes[sn.NodeID]; !ok {
+			log.Log(log.SchedApplication).Debug("skipping node as it is not feasible to run the pod",
```
pbacsko: I have the same concern like in `application.go`. If this is called from a loop (node iterator), it can generate way too much output.

pkg/scheduler/partition_test.go
```diff
@@ -1803,6 +1803,7 @@ func TestRequiredNodeCancelOtherReservations(t *testing.T) {
 
 	// the second one should be reserved as the 2nd node is not scheduling
 	result = partition.tryAllocate()
+	println(result.String())
```
pbacsko: debugging statement, delete

pbacsko (changes requested):
As far as I can see, you added `preAllocateConditions()` to `tryNodes()`, `tryNodesNoReserve()`, `tryPlaceholderAllocate()` and `Preemptor.tryNodes()`. But `tryRequiredNode()` and `tryReservedAllocate()` are missing.

Please check my comment under https://github.com/apache/yunikorn-k8shim/pull/1043#pullrequestreview-4639600122 for more details.

Also, on DEBUG level might end up logging too verbosely, causing frequent log rotations.

pkg/mock/predicate_plugin.go
```diff
@@ -61,9 +75,10 @@ func (f *PredicatePlugin) Predicates(args *si.PredicatesArgs) error {
 // mustFail will cause the predicate check to always fail
 // nodes allows specifying which node to fail for which check using the nodeID:
 // possible values: -1 fail reserve, 0 fail always, 1 fail alloc (defaults to always)
-func NewPredicatePlugin(mustFail bool, nodes map[string]int) *PredicatePlugin {
+func NewPredicatePlugin(mustPreFilterFail bool, mustFilterFail bool, nodes map[string]int) *PredicatePlugin {
```
wilfred-s: Comments do not line up with the signature anymore.

pkg/mock/predicate_plugin.go
```diff
@@ -29,30 +29,44 @@ import (
...
-			return fmt.Errorf("fake predicate plugin failed")
-		}
-		if !args.Allocate && fail <= 0 {
-			log.Log(log.Test).Info("fake predicate plugin node reserve fail",
-				zap.String("node", args.NodeID),
-				zap.Int("fail mode", fail))
-			return fmt.Errorf("fake predicate plugin failed")
+	for k, v := range f.nodes {
+		if args.Allocate {
+			if v > 0 && v < 200 {
```
wilfred-s: needs explanation as it is not clear what this does. is a tristate (-1,0,1) not enough to handle what is needed?

pkg/scheduler/objects/node.go
```diff
@@ -487,6 +488,23 @@ func (sn *Node) preAllocateConditions(ask *Allocation) error {
 
 // Checking pre-conditions in the shim for a reservation.
 func (sn *Node) preReserveConditions(ask *Allocation) error {
+	// run predicates for this pod before in hand and fetch feasible nodes
```
wilfred-s, line 491: Since we run the pre-filter predicates earlier I do not think we need to call it here.
`preReserveConditions()` is called from `tryPlaceholderAllocate()` and from `tryNodes()` 
Specially the case in the `tryNodes()` we reserve a node that has been found in the iteration over all the nodes. If the node was not feasible it would not have been chosen.
For the `tryPlaceholderAllocate()` case we might want to even special case that call and just pass in a single node into the check to lighten the load.

pkg/scheduler/objects/preemption.go
```diff
@@ -557,7 +557,23 @@ func (p *Preemptor) tryNodes() (string, []*Allocation, bool) {
 	// calculate victim list for each node
 	predicateChecks := make([]*si.PreemptionPredicatesArgs, 0)
 	victimsByNode := make(map[string][]*Allocation)
+
+	// run predicates for this pod before in hand and fetch feasible nodes
+	feasibleNodes, err := p.ask.preAllocateConditions(true)
```
wilfred-s: I think we need to do the same here as for the `tryPlaceholderAllocate()` we only want to look at a single node. We do not care about all nodes in the cluster. Need to think about this call.

pkg/scheduler/partition_test.go
wilfred-s: no print statements in production or test code unless specifically required (as in the e2e tests)

wilfred-s (changes requested):
two main comments:
* why are we not returning a PB message from the call wrapping the result we have. That means we have a slightly more flexible setup 
* the node list returned is a complex object: not sure I understand why as the plugin in K8s only passes back a set of NodeID. Create a message like this:
```
message PreFilterPredicatesResponse{
    // whether or not pre-filter fails
    bool success = 1;
    // map of nodes that are considered feasible
    map<string, Empty> FeasibleNodes = 2
}

message Empty {}
```

pkg/scheduler/partition_test.go
manirajv06: Taken care

manirajv06:
> As far as I can see, you added `preAllocateConditions()` to `tryNodes()`, `tryNodesNoReserve()`, `tryPlaceholderAllocate()` and `Preemptor.tryNodes()`. But `tryRequiredNode()` and `tryReservedAllocate()` are missing.

Ignored it as I thought of not doing predicate checks and have to make it at any cost. But then going through the flow again, felt that predicate checks are still required too and makes sense. So, made changes.

@pbacsko @wilfred-s 

I am working on the other review comments. Will cover it up in next commit.

Q: As of now, tryPreemption() for required node does not carry out preemption predicate checks unlike Intra queue preemption does it. I think we need to do even here. Thoughts?

pkg/scheduler/objects/node.go
manirajv06: > Specially the case in the tryNodes() we reserve a node that has been found in the iteration over all the nodes. If the node was not feasible it would not have been chosen.

First PreFilter call was made for "Allocate" attempt and second one is for "reserve" attempt. In addition, in case of any errors in first PreFilter call we don' t do any feasible nodes checks at all. Since K8s pre filter plugins used for "Allocate"  and "Reservation" process differs, result also might differ. Also, Traversing all the nodes itself in earlier node iterator in very large clusters setup consumption time also need to be account and things could change very quickly in rapidly changing environments. So, I think we still need to make a second call.

> For the tryPlaceholderAllocate() case we might want to even special case that call and just pass in a single node into the check to lighten the load.

PreFilter request doesn't required `nodeID` as an input at all. I am just managing with the existing `*si.PredicateArgs` itself. Ideally, we need to define a new one to not include `nodeID` as an input. Need to file a follow up for this.

Yes, it is single node only but still we need to ensure it is available in feasible nodes or not. Again, duration between ph allocation and real allocation replacement may take upto 15 minutes and things could  change very quickly in rapidly changing environments.

pkg/scheduler/objects/preemption.go
manirajv06: Same answers as above.

In addition, it is not really a single node need to be processed. `nodeAvailableMap` could have multiple nodes. So, Similar to node checks in `nodeIterator` processing, node checks while traversing `nodeAvailableMap` is also required to avoid unnecessary checks much later in the flow.

manirajv06:
> 
> * why are we not returning a PB message from the call wrapping the result we have. That means we have a slightly more flexible setup
> * the node list returned is a complex object: not sure I understand why as the plugin in K8s only passes back a set of NodeID. Create a message like this:

Introduced new API's in SI to take care of this.

pkg/scheduler/objects/node.go
manirajv06: > PreFilter request doesn't required nodeID as an input at all. I am just managing with the existing *si.PredicateArgs itself. Ideally, we need to define a new one to not include nodeID as an input. Need to file a follow up for this.

Introduced new API in SI to take care of this.

pkg/scheduler/objects/application.go
```diff
@@ -1085,8 +1085,29 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
 		request.setHeadroomCheckPassed(sa.queuePath)
 
 		requiredNode := request.GetRequiredNode()
+
+		// run predicates for this pod before in hand and fetch feasible nodes
+		feasibleNodes, predicatesResult := request.preAllocateConditions(true)
+
 		// does request have any constraint to run on specific node?
 		if requiredNode != "" {
```
wilfred-s: The only node that is possible if `requiredNode != ""` is the node set, no reason to run any pre-filter or anything as the only node possible is already known.

Run the predicate check for the generic case after we filter out required node allocations.

pkg/scheduler/objects/application.go
```diff
@@ -1085,8 +1085,29 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
 		request.setHeadroomCheckPassed(sa.queuePath)
 
 		requiredNode := request.GetRequiredNode()
+
+		// run predicates for this pod before in hand and fetch feasible nodes
+		feasibleNodes, predicatesResult := request.preAllocateConditions(true)
```
wilfred-s: We do not want to skip the `tryRequiredNode()` call. If the node has not been reserved yet we need to do that. The trigger is in `tryRequiredNode()`. This check could stop that reservation. If the predicates block the node the node is still reserved for this required node pod, i.e. daemonset pod.

pkg/scheduler/objects/application.go
```diff
@@ -1302,16 +1322,24 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator,
...
 	// pick the first fit and try all nodes if that fails give up
 	var allocResult *AllocationResult
 	if phFit != nil && reqFit != nil {
 		resKey := reqFit.GetAllocationKey()
+
+		// run predicates for this pod before in hand and fetch feasible nodes
+		feasibleNodes, predicatesResult := reqFit.preAllocateConditions(true)
+		if !predicatesResult {
+			return nil
+		}
```
wilfred-s, line 1394: We run this check again in line 1364 if it passed here. We need to be smarter about this and pick up the results from the first run we do within the loop. We might never need to run this if we do not pass the earlier checks.

pkg/scheduler/objects/application.go
```diff
@@ -1448,7 +1487,24 @@ func (sa *Application) tryReservedAllocate(headRoom *resources.Resource, nodeIte
...
+			}
+		} else {
+			skipReservedNode = true
+		}
+		if skipReservedNode {
+			getRateLimitedAppLog().Info("skipping reserved node as it is not feasible to run the pod",
+				zap.String("allocationKey", ask.GetAllocationKey()),
+				zap.String("reserved node", reserve.node.NodeID))
+			continue
+		}
```
wilfred-s, line 1559: This check is already run inside the `tryNode()` just below. This is not required here and should be only run inside the `tryNodesNoReserve()`

pkg/scheduler/objects/application.go
```diff
@@ -1510,6 +1566,13 @@ func (sa *Application) tryPreemption(headRoom *resources.Resource, preemptionDel
 // This should never result in a reservation as the allocation is already reserved
 func (sa *Application) tryNodesNoReserve(ask *Allocation, iterator NodeIterator, reservedNode string) *AllocationResult {
 	var allocResult *AllocationResult
+
+	// run predicates for this pod before in hand and fetch feasible nodes
+	feasibleNodes, predicatesResult := ask.preAllocateConditions(true)
+	if !predicatesResult {
+		return nil
+	}
```
wilfred-s, line 1625: We need to be smarter about this and pick up the results from the first run we do within the loop. [See previous comment](https://github.com/apache/yunikorn-core/pull/1096/changes#r3619341858)

wilfred-s (changes requested):

pkg/scheduler/objects/application.go
manirajv06: No, thats not the case. This one does PreFilter checks where as line 1364 does filter checks.

pkg/scheduler/objects/application.go
manirajv06: Earlier Prefilter run was based on `reservation` filters before doing the reservation in `tryAllocate` cycle. Now, here in `tryReservedAllocate` cycle, we are processing the reservations happened (`tryAllocate` cycle) earlier to take it forward. So, we are running PreFilter based on `allocation` filters this time. Also, we are not sure about the delay between these two steps. We cannot be so sure that nothing has changed from the node resource usage point of view in the meantime.

Since `tryNodesNoReserve()` is about trying "other" nodes, we run prefilter checks. so, yes correct

pkg/scheduler/objects/application.go
manirajv06: (Answering both comments for the same lines together)

Like we discussed offline, I do agree that we need to run this pod on this node at any cost. That is the intent. But there could be chance that node might not even fit with basic eligibility rules itself like affinity rules etc. So, we need to run the predicate checks to get the bare minimum checks passed in the first place.

Inaddition, PreFilter is required as whole filter plugins works based on `cycleState` persisted during former happens. During this PreFilter call, `cyclestate` would be persisted and used when the filter predicate run happens later through `tryNode` inside `tryRequiredNode`

wilfred-s: It does not matter if the predicates fail at this point in time. The pod cannot run anywhere else. If current pods are a problem we need to handle that via the preemption if needed. If we do not want to do that we leave it pending but reserve the node.

pkg/scheduler/objects/application.go
wilfred-s: Filter in the shim must run pre-filter first and it does. That is needed to make sure the cycle state is initialised. Have a look at `predicateManagerImpl.podFitsNode()`

manirajv06: `Have a look at predicateManagerImpl.podFitsNode()`

Yes, used to be. but not anymore. It does only filter checks. Newly introduced API predicatesPreFilter does the prefilter checks which is required to be called from the core side to ensure prefilter checks are passed. Please check [Shim PR](https://github.com/apache/yunikorn-k8shim/pull/1043).

pkg/scheduler/objects/application.go
manirajv06: `tryRequiredNode` calls `tryNode`. `tryNode` does `Filter` checks using `Predicates` API. So, `PreFilter` call has to be made before this for sure so that `cycleState` is generated for later use in `Filter`  checks.

Only thing we can do here is, ignore the `PreFilter` results only but not the "call" and move on by not skipping the required node. Hope this aligns with your thoughts.

manirajv06: > If current pods are a problem we need to handle that via the preemption if needed.

BTW, YUNIKORN-3330 is all about running preemption predicate checks during preemption for daemon set pods through `tryReservedAllocate` cycle. If at all for some reason check fails, then we don't have any other option waiting for the situation to change on its own. Up until that point, pods would be in "reserve" state only.

manirajv06: Predicate checks won't run for daemon set pods. Reserve that specific node.

pbacsko:
@manirajv06 why were predicate checks removed for DaemonSet pods?

I did compare this new behavior to that of the default scheduler and now the following checks are not performed by us:

- NodeResourcesFit (not a problem, we check natively)
- NodePorts
- VolumeBinding / NodeVolumeLimits / VolumeZone
- TaintToleration (not a problem because DS controller already pre-checks them)
- NodeAffinity (redundant, so it's ok)
- Inter-pod affinity / topology spread

With this change, we modify the semantics of DS pod scheduling, ie. we no longer enforce certain constraints. What's the reason behind this?

manirajv06:
> why were predicate checks removed for DaemonSet pods?

We had a detailed look on this again and decided to reserve the node at any cost in the first go as we don't have any other node to run the pod. Once we reserve the node, we decided to run the predicate checks while trying out preemption for the daemon set pods during tryReservedAllocate scheduling cycle. We had filed a jira for [this](https://issues.apache.org/jira/browse/YUNIKORN-3330). Yes, predicate checks may fail over there but the node is reserved. In the worst case, we might be keep trying the same pod over and over.

As of now, we ignore the predicate results and go ahead. We would like to retain the same behaviour as is.

pkg/scheduler/objects/node.go
wilfred-s: We can only do this if the existing cycle state that we build earlier is for the reserve is deleted and replaced with this new cycle state for allocate. If we do not do that we could have issues. Depending on the correct handling in the k8shim I am OK with this.

wilfred-s (approved):
LGTM
one remark outstanding: we need to do the correct thing on the K8shim side if that is done, see my reviews there, I am OK.

wilfred-s:
e2e tests are failing due to chicken and egg:; we need the k8shim commit to pass the e2e test and we need the core commit to change the k8shim dependency.
@manirajv06 Please run the e2e tests locally with an updated K8shim before committing this

manirajv06:
Ran E2E tests with updated k8shim. Tests passed. Will commit the PR.

@wilfred-s @pbacsko Thanks.

PR #1103, opened 2026-07-10
[YUNIKORN-3318] Deadlock in preemption init un-reserving apps

wilfred-s:
### Description
If an application A holds a reservation R on a node N the preemption initialisation will deadlock.

During the initWorkingState() call reservations are released from a node if they are older than the timeout, not a required node allocation, and the priority is lower than the allocation we are trying to preempt for.

Fix time calculation to use the reservation creation time for accuracy. Track the number of reservations released.
Correctly determine if a node is still reserved after releases.

fix unit tests

### Type of change
- [X] Bug Fix

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3318

### AI Tooling
If an AI tool was used:

### How has this been tested?

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -133,10 +133,11 @@ func (p *Preemptor) initQueueSnapshots() {
 }
 
 // initWorkingState builds helper data structures required to compute a solution
-func (p *Preemptor) initWorkingState() {
+// returns the number of reservations released while preparing the data structures
+func (p *Preemptor) initWorkingState() int {
 	// return if we have already run
 	if p.nodeAvailableMap != nil {
-		return
+		return 0
 	}
 
 	// ensure queue snapshots are populated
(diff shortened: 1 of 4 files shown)


pkg/scheduler/objects/preemption.go
```diff
@@ -133,10 +133,11 @@ func (p *Preemptor) initQueueSnapshots() {
 }
 
 // initWorkingState builds helper data structures required to compute a solution
-func (p *Preemptor) initWorkingState() {
+// returns the number of reservations released while preparing the data structures
+func (p *Preemptor) initWorkingState() int {
```
pbacsko, line 137: Am I missing something? I can't see the related change where this int value is consumed.

pbacsko (changes requested):
Make sure the return value of `initWorkingState()` is processed on the call site.

pkg/scheduler/objects/preemption.go
wilfred-s: You are not missing anything. Not yet you are correct, I have logged a separate jira for that as that change is bigger.[YUNIKORN-3319](https://issues.apache.org/jira/browse/YUNIKORN-3319)

The released reservations are not accounted for correctly. The partition tracks the number of reservations but this preemption code does not pass back the released reservations and does not update that counter. That should be fixed to make sure we do not run the reservation schedulling loop unneeded.

pbacsko (approved):
+1 lgtm

pkg/scheduler/objects/preemption.go
```diff
@@ -158,38 +159,48 @@ func (p *Preemptor) initWorkingState() {
...
-						zap.Int("reservations count", num))
-					hasOtherReservations = false
+				if res.alloc.GetPriority() < p.ask.priority && time.Since(res.createTime) > reservationWaitTimeout {
+					log.Log(log.SchedPreemption).Info("Cancelling reservation to consider node for preemption",
+						zap.String("triggeringAppID", p.application.ApplicationID),
+						zap.String("triggeringAllocationKey", p.ask.allocationKey),
+						zap.String("reservingAppID", res.appID),
+						zap.String("reservingAllocationKey", res.allocKey),
+						zap.String("node", node.NodeID))
+					num := 0
```
adityadtu5, line 184: sorry for the late review, it would be good to have comment explaining why this is done

wilfred-s: [YUNIKORN-3319 ](https://issues.apache.org/jira/browse/YUNIKORN-3319) has the details

PR #1104, opened 2026-07-16
[YUNIKORN-3316] Handle AssumePod failures in core

adityadtu5:
### Description
Core changes to handle updates of tasks with termination type `SCHEDULING_FAILED_ON_RM `

Remove allocation from node, application,  update and clear resources on queue, update partition allocation count.

Scheduler Interface PR - https://github.com/apache/yunikorn-scheduler-interface/pull/162
Shim changes - https://github.com/apache/yunikorn-k8shim/pull/1052

### Type of change

- [x] Improvement

### Jira issue
https://issues.apache.org/jira/browse/YUNIKORN-3316

### AI Tooling
If an AI tool was used:

### How has this been tested?

### Questions:


pkg/scheduler/objects/application.go
```diff
@@ -799,6 +810,51 @@ func (sa *Application) DeallocateAsk(allocKey string) (*resources.Resource, erro
...
+	}
+
+	ask := sa.requests[allocKey]
+	if ask == nil {
+		return nil, fmt.Errorf("failed to locate ask with key %s for rollback", allocKey)
+	}
+	alloc := sa.allocations[allocKey]
+	if alloc == nil {
+		return nil, fmt.Errorf("failed to locate allocation with key %s for rollback", allocKey)
+	}
```
wilfred-s: The same reference (pointer) is used in both maps as the value.
This should always be true: `ask == alloc`
allocations must always be a subset of requests checking allocations should be enough

pkg/scheduler/partition.go
```diff
@@ -1504,6 +1504,45 @@ func (pc *PartitionContext) removeAllocation(release *si.AllocationRelease) ([]*
 		return nil, nil
 	}
 
+	// Handle shim-initiated scheduling failure: roll the allocation back to a pending ask
```
wilfred-s, line 1507: Can we factor out this code in its own method?

pkg/scheduler/partition.go
```diff
@@ -1504,6 +1504,45 @@ func (pc *PartitionContext) removeAllocation(release *si.AllocationRelease) ([]*
...
+				zap.String("nodeID", nodeID))
+		}
+		if err := queue.DecAllocatedResource(res); err != nil {
+			log.Log(log.SchedPartition).Warn("failed to release resources from queue during rollback",
+				zap.String("appID", appID),
+				zap.String("allocationKey", allocationKey),
+				zap.Error(err))
+		}
+		pc.updateAllocationCount(-1)
+		metrics.GetQueueMetrics(queue.GetQueuePath()).AddReleasedContainers(1)
```
wilfred-s: Two points:
* are we sure that the container was added to the allocated containers metric already? The metric gets increased when `UpdateAllocation()` is called after confirming the allocation. I think that has not happened yet.
* if above is not true: why not use `IncReleasedContainer()` ? less overhead

wilfred-s (changes requested):

pkg/scheduler/partition.go
wilfred-s: We should do the same for the code in if if clause for `TerminationType_PLACEHOLDER_REPLACED` in the same function

pkg/scheduler/partition.go
adityadtu5: you are correct, allocated container metrics was not updated till this point.

I think it is good to remove this call to make sure that sum of allocated and released containers matched with the total submitted containers.

wilfred-s:
@adityadtu5 change looks good now, please fix the build issue. After that I'll have a final look at the change.

wilfred-s (approved):
LGTM +1

PR #1108, opened 2026-07-26
[YUNIKORN-3333] Fix flaky unit test TestQuotaChangeTryPreemption

PoiBlackTea:
### Description
TestQuotaChangeTryPreemption shares victim slices across table-driven sub-tests. Preemption modifies these slices in place (sorting and truncating), so earlier sub-tests corrupt the state for later ones, causing non-deterministic failures.

Replace the shared pre-built victim slices with a prepareVictims factory function called inside each sub-test to create fresh allocations per run.

### Type of change

- [X] Bug Fix
- [ ] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3333

### AI Tooling
If an AI tool was used:
- [X] The PR includes the phrase "Generated by Antigravity IDE", where Antigravity IDE is the name of the AI tool used.
- [X] My use of AI contributions follows the ASF legal policy.

### How has this been tested?

### Questions:

N/A

### DIFF
--- a/pkg/scheduler/objects/quota_preemptor_test.go
+++ b/pkg/scheduler/objects/quota_preemptor_test.go
@@ -156,29 +156,40 @@ func TestQuotaChangeTryPreemption(t *testing.T) {
 		},
 	})
 
-	suitableVictims := make([]*Allocation, 0)
-	notSuitableVictims := make([]*Allocation, 0)
-	oversizedVictims := make([]*Allocation, 0)
-	overflowVictims := make([]*Allocation, 0)
-	shortfallVictims := make([]*Allocation, 0)
-
-	suitableVictims = append(suitableVictims, createVictim(t, "ask1", node, 5, resources.NewResourceFromMap(map[string]resources.Quantity{"first": 10})))
-	// ask2 uses {first:8} (smaller than ask1's {first:10}) so it is deterministically sorted first and preempted
-	suitableVictims = append(suitableVictims, createVictim(t, "ask2", node, 4, resources.NewResourceFromMap(map[string]resources.Quantity{"first": 8})))
-
-	oversizedVictims = append(oversizedVictims, createVictim(t, "ask21", node, 4, resources.NewResourceFromMap(map[string]resources.Quantity{"first": 9})))
-	oversizedVictims = append(oversizedVictims, createVictim(t, "ask3", node, 3, resources.NewResourceFromMap(map[string]resources.Quantity{"first": 11})))
-
-	overflowVictims = append(overflowVictims, createVictim(t, "ask4", node, 3, resources.NewResourceFromMap(map[string]resources.Quantity{"first": 5})))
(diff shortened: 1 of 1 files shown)


manirajv06:
@PoiBlackTea 

As a human :), I am happy that AI pointed out in place edits of the variables etc could be the possible reason of this flakiness. Of course, Our brain too guessed the same and also "possible" causes and reasons around the same lines of "Could be because of variables not being reset? or Could be because same input has been passed to different tests" or something like that. 

I am happy that it pointed out but the amount of code generates is really a good signal for us to make a decision whether to raise a pr with the code as is or not. As I said earlier, passing same input could possibly be the reason and since it is related to tests, committer or contributor or whoever developed the code earlier might not have paid greater attention and thus caused this issue. All it requires here is "Pass correct input to get the expected output". So, I would prefer passing a correct input rather passing the one meant for different test case. or If it is going to be the same, clone and pass it. Not definitely a new method with lot of switch cases. Code maintenance would become nightmare and burden us very soon.

Yes, I am up for AI tooling but definitely not going with the code it generates as is blindly. We should use it and make the call judiciously.

PoiBlackTea:
> @PoiBlackTea
> 
> As a human :), I am happy that AI pointed out in place edits of the variables etc could be the possible reason of this flakiness. Of course, Our brain too guessed the same and also "possible" causes and reasons around the same lines of "Could be because of variables not being reset? or Could be because same input has been passed to different tests" or something like that.

Thanks for the feedback! I've updated the PR to remove the factory method. We now clone the victim allocations inside t.Run() to keep the test clean and  declarative without extra abstractions.

manirajv06:
Is cloning required for all test inputs? I believe problem was with the last two tests inputs. Those are the ones used for more than 1 tests. Isn't it? Please validate my points against the test failures. If correct, clone those with a new variable and use the same for failed tests.

PoiBlackTea:
You're correct, only `overflowVictims` and `shortfallVictims` need distinct variables, since they are the only inputs shared across multiple subtests.

Root cause: `preemptVictims()` calls `victim.MarkPreempted()` which sets `preempted = true` on the `*Allocation` pointer, then sends the event to the Event System's channel (`ec.channel <- event`) for background goroutine processing. Since both subtests reference the same heap pointer, the next subtest's `filterAllocations()` hits `alloc.IsPreempted() == true` at [quota_preemptor.go#L285](https://github.com/apache/yunikorn-core/blob/master/pkg/scheduler/objects/quota_preemptor.go#L285) and skips the allocation, producing wrong victim counts.

This issue is easily reproducible simply by removing time.Sleep(500 * time.Millisecond). Furthermore, printing the memory pointer addresses (%p) of the allocation objects confirms that the subtests are sharing the exact same pointer address in memory, leading to state pollution across runs.

wilfred-s:
kicked off the workflow again. Restart test for event system sometimes has less goroutines after the restart than before, should change the check to `before >= after` instead of equal.

manirajv06 (approved):
looks good.

PR #1110, opened 2026-07-27
[YUNIKORN-3338] Add golden decision-trace regression tests for the scheduler

tigerquoll:
### Description

Test-only. **No product code is touched.**

The scheduler's existing tests assert individual outcomes — "this ask was allocated", "this queue is
over its limit". None assert its output *as a whole*: who gets what, on which node, in what order,
and what is handed back for release. Most interesting scheduler changes are ordering changes, and a
change to ask ordering, node scoring, headroom accounting, victim selection or placeholder
replacement can leave every existing assertion green while changing where work lands.

This adds two tests that make that class of change visible. Both drive a fixed workload through the
real `RMProxy`/partition API and pin the exact, ordered sequence of decisions the RM observes via
`UpdateAllocation` — `(applicationID, allocationKey, nodeID)` per allocation, `(applicationID,
allocationKey, terminationType)` per release, `(applicationID, allocationKey)` per rejected
allocation — compared verbatim against a golden JSON file.

#### What's added

| File | |
|---|---|
| `golden_trace_test.go` | `TestGoldenDecisionTrace` — one mixed workload + shared helpers |
| `golden_trace_scenarios_test.go` | `TestGoldenDecisionTraceScenarios` — 6 isolated sub-tests |
| `testdata/golden_trace*.json` | 7 goldens (14 entries for the mixed workload, 3–4 each for scenarios) |
| `mock_rm_callback_test.go` | additive: `traceEntry`, an opt-in `trace` field, `enableTrace()`, `getTrace()`, `getNodeAllocations()` |

Capture is opt-in — the mock is shared with the whole package, including `BenchmarkScheduling`, so
capturing unconditionally would put an append inside the callback's write lock on a measured path.

Two words are used throughout, for two different things that both happen to number six:

* **Phases** — the six *sequential* stages inside `TestGoldenDecisionTrace`. One cluster, each phase
  building on the state the last left, one 14-entry golden for all of them.
* **Scenarios** — the six *independent* sub-tests of `TestGoldenDecisionTraceScenarios`. Each gets a
  fresh cluster, its own config and its own small golden.

They are not the same six. `gang` and `preemption` name both a phase and a scenario, deliberately —
see the duplication note below.

**The mixed workload's phases** are: priority ordering within an app, pending-ask removal, gang
placeholder replacement, restart + recovery, app failure with pending asks, and priority preemption.
The whole pinned trace is short enough to read:

```
p-high, p-mid, p-low            -> golden-node-main:1
r-1, r-3, r-4                   -> golden-node-main:1     (r-2 removed while pending)
g-placeholder -> main:1 ; g-placeholder :: PLACEHOLDER_REPLACED ; g-real -> main:1
p-recover-high                  -> golden-node-main:1     (p-recover-low does not fit)
low-1, low-2 -> preempt:1 ; low-2 :: PREEMPTED_BY_SCHEDULER ; high-1 -> preempt:1
```

Three of those six phases only pin anything because the workload is sized to make them decide
something:

* **Pending-ask removal** — four asks at four distinct priorities, no scheduling cycle before the
  middle one (`r-2`) is removed. The golden pins that the survivors are still served in priority
  order and that the removal never reaches the RM.
* **Restart and recovery** — recovered usage 55/7 against a 105/10 node leaves 50/3 free; the two
  post-recovery asks are 30/1 each, so exactly one fits. Lost usage allocates both, double-counted
  usage neither.
* **Failed-application ask cleanup** — the node still has 20/2 free and the failed app's asks are
  10/1 each, so they *would* be allocated had cleanup not happened.

Those figures aren't comments: `node-main`'s capacity is computed from the per-phase ask sizes and
each phase's headroom is asserted, so resizing an earlier ask fails with a message naming what to
recompute rather than quietly hollowing out a later phase. Those headroom checks — and the set of
allocations the restart replays — are *preconditions*, so unlike the trace-length barriers they fail
hard even under `UPDATE_GOLDEN`: a workload edit that hollows out a phase must not be able to record
a golden that looks like a behaviour change.

**The scenarios** each isolate a single decision:

| Scenario | What it pins |
|---|---|
| `app-ordering-fifo` / `app-ordering-fair` | Two apps tie on priority for one remaining slot; the policy's second key decides (submission time vs usage ratio). Pre-loaded so the policies pick different winners. |
| `node-binpacking` / `node-fair` | Four equal asks, two identical nodes. Binpacking packs, fair alternates. Traces differ **only** in the node column. |
| `gang` | Placeholder → real handshake on an otherwise empty cluster. |
| `preemption` | Victim selection, release, and reschedule, on an otherwise empty cluster. |

Policy pairs share an identical workload, so the diff between a pair *is* the policy's effect.

The `gang` and `preemption` scenarios repeat choreography the mixed workload also runs, rather than
sharing a helper with it. That duplication is the point: a shared helper would move both traces
together on a mistake, and "the shared trace broke but the scenario didn't, so it's prior-state
interaction" would stop being a conclusion anyone could draw.

#### Determinism

* Fixed app/node IDs, pinned strictly-increasing creation times on every ask.
* Scheduling advanced by explicit `MultiStepSchedule`, never the background loop.
* Trace lengths asserted **exactly**, not "at least": a `>= n` barrier is satisfied early by a
  surplus decision, which then drifts into a later wait and presents as a timeout instead of a diff.
* Every assertion is preceded by a quiescence round trip (`traceProbe.sync`), because polling a
  length cannot prove *absence*: where no new decision is expected, the first evaluation already
  succeeds. Costs nothing measurable — 7.9s → 7.7s.
* WARN-level logging, re-applied after each `ms.Init` (which resets it) and restored from a snapshot
  on cleanup, since it is process-global.
* Candidate lists come from Go map iteration, so each scenario avoids ties in the comparator that
  decides it (distinct submission times *and* usage for app-sort; node IDs that order unambiguously
  under `nodeRef.Less`'s `NodeID` fallback).

One exception: creation times are **not** pinned across the restart — `NewSIFromAllocation` copies
neither `AllocationTags` nor `Priority`, so replayed allocations get `time.Now()` and priority 0.
Nothing pinned depends on their order; they are already allocated, and `tryAllocate` skips those.

#### Scope and limitations

These pin decision *sequences* for the behaviours above on small clusters. They are **not** a general
conformance suite — a green run does not mean "scheduling is unchanged". Not covered:

* **Reservations lifecycle** — reservation happens internally during preemption, but reserve/unreserve
  is never pinned and reservation timeout is never reached.
* **Ordering of recovered allocations among themselves** — the recovery phase pins that recovered
  *usage* is accounted for, not that recovered requests sort like live ones; the recovery path loses
  the createTime and priority that would make such an assertion meaningful.
* **Resource accounting on failure cleanup** — the failed-application phase pins that the abandoned
  asks are discarded and that no decision about them reaches the RM. It does not pin the accounting
  behind that: `cleanupAsks` clears the request map and the sorted list only, so the application's
  own pending resource and `askMaxPriority` still carry the discarded asks afterwards. Nothing in the
  workload asserts either way, and no later phase depends on them.
* **Rejection reasons** — rejected allocations are captured, so a change that turns an allocation
  into a rejection shows up as a changed row rather than a missing one, but only
  `(applicationID, allocationKey)` is recorded. `si.RejectedAllocation.Reason` is free text built
  from whichever error the partition returned, so pinning it would make rewording an error message a
  failed decision-regression test.
* **The partition column** — single-partition workloads make it a constant carrying the normalised
  name, which would write the RM ID into every golden line.
* **User and group quota limits** — no config sets `limits:`; all apps run as one user, no groups.
* **Required-node / daemonset asks** — no ask carries a required-node tag.
* **Unschedulable-ask backoff** — `backoffDeadline` is never armed; no queue sets a backoff threshold.
* **Placeholder timeout and gang failure paths** — only the successful placeholder→real replacement.
* **Guaranteed-resource-driven preemption and preemption fencing** — both preemption scenarios are
  priority-driven; no queue uses `FencePreemptionPolicy`.
* **Nested queue hierarchies** — every config is `root` plus one level of leaves.
* **Multiple partitions** — everything runs on the single `default` partition.
* **Custom resource types** — only `memory` and `vcore`.
* **Foreign allocations** — none registered.
* **Queue placement rules** — apps name their target queue directly.
* **The placeholder-replacement revert** in `Application.tryPlaceholderAllocate` — reaching it needs
  two goroutines racing on one placeholder within a cycle, which a step-driven harness can't script
  without the timing nondeterminism a golden test must not have.

#### Running and regenerating

```
go test ./pkg/scheduler/tests/ -run '^TestGoldenDecisionTrace' -count=1   # both tests
```

Raise `-count` to check determinism. A mismatch prints a side-by-side listing naming the golden file,
marking differing rows and, for the mixed workload, attributing each row to the phase that produced
it. A phase that produces no decisions owns no row, so it is named at its boundary instead.

Regenerate only when a change *intentionally* alters scheduling decisions:

```
# all seven goldens
UPDATE_GOLDEN=1 go test ./pkg/scheduler/tests/ -run '^TestGoldenDecisionTrace' -count=1

# a single scenario
UPDATE_GOLDEN=1 go test ./pkg/scheduler/tests/ -run '^TestGoldenDecisionTraceScenarios$/^node-fair$' -count=1
```

Under `UPDATE_GOLDEN` the length barriers, and the workload's observations of the behaviour being
replaced, report their divergence and carry on rather than failing ahead of the rewrite, so the run's
log doubles as a summary of what changed. The workload's own *preconditions* still fail hard, because
a broken precondition does not mean the scheduler decided something new — it means the run set up a
different situation than the phase is named for, and a golden recorded from it would pin the wrong
experiment while looking green. The value is parsed as a boolean; an unparseable one fails rather than
silently verifying.

**The run always ends RED** — a green one would be indistinguishable from a run that verified
something. So:

1. `git diff pkg/scheduler/tests/testdata/` — that diff **is** the behaviour change.
2. Account for every changed line; one you can't explain is a side-effect, not noise.
3. Update the trace-length arguments the run reported, so the barriers pin the new behaviour.
4. Re-run **without** `UPDATE_GOLDEN` and confirm green — the regenerating run says nothing about
   correctness.
5. Include the golden diff in the PR, so the new sequence is reviewed as a behaviour change.

Only goldens matched by `-run` are rewritten, and on an unmodified tree they come back byte-identical
— so a dirty `git status` afterwards always means a real delta.

### Type of change

- [x] Improvement

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3338

### AI Tooling

If an AI tool was used:

### How has this been tested?

- [ ] A pull request will be opened for new e2e tests (apache/yunikorn-k8shim repository). — not
  applicable: no product code changes, so there is no new behaviour for e2e to cover.

Beyond `make test_all`:

| Check | Result |
|---|---|
| `git diff origin/master --name-only` | 10 files, test and testdata only |
| `-run '^TestGoldenDecisionTrace' -count=10` | ok, 74.2s |
| `go test ./pkg/scheduler/tests/ -count=1` | ok, 48.3s |
| `go test ./pkg/scheduler/tests/ -race -count=1` | ok, 49.7s |
| `-bench BenchmarkScheduling -benchtime=1x` | ok, unaffected by trace capture |

Adds ~7.3s to a package run: 41.0s without these tests, 48.3s with them.

A golden test that never fails is decoration, so each perturbation below was applied, observed, and
reverted in full. Where a run stops at a precondition or an observation rather than at the trace
comparison, that is the intended order: the workload notices that its own setup or its own stated
expectation broke before it gets as far as diffing a golden.

1. **Recovery replay skipped** → phase 4 stops at the headroom precondition: `recovery left
   map[memory:105 vcore:10] free on golden-node-main:1, the capacity ledger says 50 memory / 3 vcore`.
   Hard in both modes, so `UPDATE_GOLDEN` cannot record a golden from the hollowed-out phase.
2. **`LessThan` priority flipped** → phase 4 allocates `p-recover-low` instead of `p-recover-high`,
   so the phase's own observation fires: `p-recover-low should still be waiting`.
3. **`LessThan` createTime flipped** → 5 of 6 scenarios fail (`a1..a4` reversed in the node pair,
   `e-1`↔`e-2`, `low-1`↔`low-2`) and the mixed trace fails at phase 6; `gang` correctly stays green,
   never holding two comparable pending asks.
4. **`nodeRef.Less` by descending score** → the node pair fails *by swapping into each other's
   golden*: binpacking spreads, fair packs. Only those two scenarios move.
5. **`sortApplicationsByPriorityAndSubmissionTime` reversed** → only `app-ordering-fifo` fails
   (`e-3`→`l-2` on the contested slot); `app-ordering-fair` correctly stays green.
6. **Failure cleanup bypassed** → phase 5 fails on `failure cleanup should have cleared f-1`.
7. **Regeneration, both directions** → with (4) applied, `UPDATE_GOLDEN` rewrites exactly the two
   node-sort goldens and still ends red; the two files swap md5s (`b8746a0a…`↔`baad5fa5…`), the
   other five are untouched, and reverting (4) and regenerating again restores all seven
   byte-for-byte.

### Questions:

None apply: the change is test-only, adds no user-facing behaviour, and introduces no dependencies.

A mismatch names the golden file, marks differing rows with `!!`, and attributes each row to the
phase that produced it (a phase producing no decisions is named at its boundary instead):

```
scheduling decisions do not match the golden file.
  golden file: testdata/golden_trace.json (14 entries)
  this run:    14 entries

       6 phase 3 gang placeholder replacement  golden: alloc g-placeholder ...        got: alloc g-placeholder ...
  !!   7 phase 3 gang placeholder replacement  golden: release g-placeholder of  ...  got: release g-placeholder of golden-app-gang ...
       9 phase 4 restart and recovery          golden: alloc p-recover-high ...       got: alloc p-recover-high ...
      --- phase 5 failed application ask cleanup: no decisions ---
      10 phase 6 preemption                    golden: alloc low-1 ...                got: alloc low-1 ...
```

### DIFF
--- a/pkg/scheduler/tests/golden_trace_scenarios_test.go
+++ b/pkg/scheduler/tests/golden_trace_scenarios_test.go
@@ -1,0 +1,516 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package tests
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
+// ---------------------------------------------------------------------------------------------
+// TestGoldenDecisionTraceScenarios is the per-behaviour companion to TestGoldenDecisionTrace (see
+// golden_trace_test.go). That test scripts one long mixed workload on a single fixed configuration;
+// this one runs a set of short, independent workloads, each on its own configuration, so that a
+// single scheduling decision is the only thing the resulting trace can be reporting on.
+//
+// Two things motivate the split. First, configuration coverage: TestGoldenDecisionTrace pins the
+// default leaf application-sort policy and the default node-sort policy only. The sort policies are
+// user-facing, configurable contracts - "binpacking" is documented as packing work onto loaded nodes
+// and "fair" as spreading it - and nothing in the existing scheduler tests pins the observable,
+// end-to-end consequence of picking one over the other. A refactor of the comparators, of node
+// scoring, or of the order in which candidates are walked can leave the unit tests around those
+// comparators green while changing where work actually lands.
+//
+// Second, diagnosability: policy pairs are captured as two goldens over an IDENTICAL workload, so the
+// diff between the pair is precisely the policy's effect. When one member of a pair fails and the
+// other does not, the failure has already been localised to that policy's decision path.
+//
+// The gang and preemption scenarios are deliberately narrower re-runs of behaviour that
+// TestGoldenDecisionTrace also touches. There they are embedded in a long workload with accumulated
+// queue and node state; here each runs on a fresh, exactly-sized cluster. If the shared trace breaks
+// and these do not, the cause is interaction with prior state rather than the mechanism itself.
+//
+// Determinism follows the same rules as TestGoldenDecisionTrace: fixed strictly-increasing ask
+// createTimes rather than wall-clock times, no background scheduling loop, an observable barrier
+// between submitting asks and stepping the scheduler, and WARN-level logging.
+// The candidate lists being sorted here are built by iterating Go maps, so every scenario is
+// constructed to avoid ties in the comparator that decides it: application-sort ties are avoided by
+// giving the competing applications distinct submission times and distinct usage, and node-sort score
+// ties fall back to a NodeID comparison (nodeRef.Less), for which the node IDs below are chosen to
+// order unambiguously.
+//
+// As with TestGoldenDecisionTrace, a failure here means a scheduling decision changed. Regenerate a
+// golden with UPDATE_GOLDEN=1 only for a change that is intended, and call out the trace diff in
+// review.
+//
+// The gang and preemption scenarios repeat, on a fresh cluster, the choreography that phases 3 and 6
+// of TestGoldenDecisionTrace run against accumulated state. That repetition is not extracted into a
+// helper the two files share, because a shared helper would be a single point of failure for both
+// goldens: a mistake in it would move both traces together, and "the shared trace broke and the
+// scenario did not, so the cause is interaction with prior state" - the whole reason these scenarios
+// exist - would stop being a conclusion anyone could draw. What is shared between the two files is
+// infrastructure that does not decide anything (building SI messages, the trace probe, the golden
+// comparison), not the scripts themselves.
+// ---------------------------------------------------------------------------------------------
+
+// scenario application, queue and node identifiers, all fixed rather than time- or counter-derived so
+// the traces are stable.
+const (
+	scLeaf = "root.leaf"
+	// every scenario config carries this leaf, and no scenario ever submits work to it: it exists so
+	// that traceProbe.sync can park its barrier applications outside the queue whose decision the
+	// scenario is isolating. See sync() in golden_trace_test.go.
+	scBarrier = "root.barrier"
+
+	scAppEarly = "sc-app-early"
+	scAppLate  = "sc-app-late"
+	scAppPack  = "sc-app-pack"
+	scAppGang  = "sc-app-gang"
+	scAppLow   = "sc-app-low"
+	scAppHigh  = "sc-app-high"
+
+	// node IDs are chosen so that lexical order - the node-sort score tiebreak in nodeRef.Less - is
+	// unambiguous: a sorts before b.
+	scNode  = "sc-node:1"
+	scNodeA = "sc-node-a:1"
+	scNodeB = "sc-node-b:1"
+
+	scQueueLow  = "root.plow"
+	scQueueHigh = "root.phigh"
+)
+
+// goldenScenario is one behaviour: a scheduler configuration, a scripted workload, and the golden
+// file its resulting decision trace is pinned to. The trace is read from ms.mockRM after run returns.
+// barrierQueue names a leaf queue of config that traceProbe.sync can park its inert barrier
+// applications in, and which the scenario itself never schedules in.
+type goldenScenario struct {
+	name         string
+	goldenPath   string
+	config       string
+	barrierQueue string
+	run          func(t *testing.T, ms *mockScheduler, seq *createTimeSeq, probe *traceProbe)
+}
+
+// leafSortConfig builds a partition whose measured leaf uses the given application.sort.policy. The
+// barrier leaf alongside it deliberately carries no sort policy of its own: nothing is ever scheduled
+// there, and giving it one would suggest otherwise.
+func leafSortConfig(appSortPolicy string) string {
+	return `
+partitions:
+  - name: default
+    queues:
+      - name: root
+        submitacl: "*"
+        queues:
+          - name: leaf
+            properties:
+              application.sort.policy: ` + appSortPolicy + `
+          - name: barrier
+`
+}
+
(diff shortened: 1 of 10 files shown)


wilfred-s:
Assigned PR, added reviewers and kicked off the workflow.

pbacsko:
I definitely like the idea - it's similar the golden JSON file in Cadence that can be used for workflow validation.

Since we have a bunch of mocks in YK, so we can enable trace collection in multiple places, not just in `mockRMCallback`. Not sure, we can start with this and later can expand the coverage.

wilfred-s (approved):
LGTM no issues seen from my side.

@pbacsko any other remarks from your side?
Leaving open for others to also have a look.

pbacsko (approved):

PR #1111, opened 2026-07-27
[YUNIKORN-3336] event system shutdown leaks go routine or panics

wilfred-s:
### Description
Calling StartService for an existing stopped EventPublisher does not create a new stop channel. Event publisher will ignore start request when already running. Calling stop twice on the EventPublisher panics as it tries to close an already closed channel.

Move config callback setup for event system into the Init() allowing a start from config even when system start did not start the event system. Move stop channel creation into start and stop functions

Cleanup layout in event system and publisher
Remove exports from functions and objects where possible.

### Type of change
- [X] Bug Fix

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3336

### AI Tooling
If an AI tool was used:

### How has this been tested?

### Questions:

### DIFF
--- a/pkg/events/event_publisher.go
+++ b/pkg/events/event_publisher.go
@@ -50,13 +50,11 @@ func createShimPublisher(store *EventStore) *eventPublisher {
 func (sp *eventPublisher) start() {
 	log.Log(log.Events).Info("Starting event publisher")
 	// handle a restart correctly
-	if sp.stopped.Load() {
-		sp.stopped.Store(false)
-		sp.stopCh = make(chan struct{})
-	} else {
+	if !sp.stopped.CompareAndSwap(true, false) {
 		log.Log(log.Events).Info("Event publisher already running")
 		return
 	}
+	sp.stopCh = make(chan struct{})
 	go func() {
 		for {
 			select {
@@ -78,7 +76,7 @@ func (sp *eventPublisher) start() {
 }
 
 func (sp *eventPublisher) stop() {
-	if sp.stopped.Load() {
+	if !sp.stopped.CompareAndSwap(false, true) {
 		log.Log(log.Events).Info("Event publisher already stopped")
 		return
 	}
@@ -86,7 +84,6 @@ func (sp *eventPublisher) stop() {
 	sp.stopCh <- struct{}{}
 	close(sp.stopCh)
 	sp.stopCh = nil
-	sp.stopped.Store(true)
 }
 
 func (sp *eventPublisher) getEventStore() *EventStore {
--- a/pkg/events/event_publisher.go
+++ b/pkg/events/event_publisher.go
@@ -19,6 +19,7 @@
 package events
 
 import (
+	"sync/atomic"
 	"time"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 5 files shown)


pbacsko (changes requested):
There are some tricky race problems stil in the code. I performed a manual + AI review and following came out:

1) Events are dropped silently after a runtime restart. The root cause is identical in both versions (before / after) and is untouched by the PR: `Stop()` sets `ec.channel = nil`, and `StartServiceWithPublisher()` never recreates it. So after any `Stop()`→start cycle (i.e. `Restart()` / `restart()`), the input channel stays `nil` and events are dropped.
To be fair, the goroutine leak + panics are fixed, but we're still losing data.

2) **Regression**: config responsiveness is lost after the first restart. The callback is now registered in `Init()`, but `Stop()` still calls `RemoveConfigMapCallback()` and start no longer re-adds it. Since `restart()` = `Stop()` + `StartServiceWithPublisher()`, the callback is removed and never restored.

3) Non-atomic stopped gate. Both start and stop do `Load()` then `Store()` outside the lock instead of `CompareAndSwap`. Two concurrent `Stop()` calls (reachable: rapid config updates each spawn `go reloadConfig()` → `restart()` → `Stop()`) can both pass the gate; the second reaches `ec.stop <- struct{}{}` after the handler already exited and blocks forever holding `ec.Lock()` - this means deadlock. Using `stopped.CompareAndSwap(false, true)` / `CompareAndSwap(true, false)` closes this window. Very unlikely to happen, but who knows.

wilfred-s:
> There are some tricky race problems stil in the code. I performed a manual + AI review and following came out:
> 
> 1. Events are dropped silently after a runtime restart. The root cause is identical in both versions (before / after) and is untouched by the PR: `Stop()` sets `ec.channel = nil`, and `StartServiceWithPublisher()` never recreates it. So after any `Stop()`→start cycle (i.e. `Restart()` / `restart()`), the input channel stays `nil` and events are dropped.

I had not seen that one, fixed now. Added test for dropped events on restart.

>    To be fair, the goroutine leak + panics are fixed, but we're still losing data.

yep that was the focus of what I saw happen.

> 2. **Regression**: config responsiveness is lost after the first restart. The callback is now registered in `Init()`, but `Stop()` still calls `RemoveConfigMapCallback()` and start no longer re-adds it. Since `restart()` = `Stop()` + `StartServiceWithPublisher()`, the callback is removed and never restored.

The callback should not be removed, fixed now.

> 3. Non-atomic stopped gate. Both start and stop do `Load()` then `Store()` outside the lock instead of `CompareAndSwap`. Two concurrent `Stop()` calls (reachable: rapid config updates each spawn `go reloadConfig()` → `restart()` → `Stop()`) can both pass the gate; the second reaches `ec.stop <- struct{}{}` after the handler already exited and blocks forever holding `ec.Lock()` - this means deadlock. Using `stopped.CompareAndSwap(false, true)` / `CompareAndSwap(true, false)` closes this window. Very unlikely to happen, but who knows.

I had thought about that but left it as to small a chance. Changed now to `CompareAndSwap`

pbacsko (approved):
+1 lgtm

PR #1113, opened 2026-07-30
[YUNIKORN-3320] Add test for required node reservation not cancelled on timeout

kaijaytu:
### Description
Add test coverage for the fix introduced in YUNIKORN-3318 (commit 3dbbb5d) that prevents
required node reservations from being released when the reservationWaitTimeout expires.

The new test `TestTryReservedAllocateRequiredNodeNotCancelledOnTimeout` verifies that
required node (DaemonSet) reservations are NOT cancelled on timeout, complementing the
existing `TestTryAllocateWithReservedHeadRoomChecking` which verifies normal reservations
ARE cancelled.

### Type of change

- [x] Bug Fix

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3320

- [x] I have created a Jira issue for this pull request.
- [x] The Jira ID is part of the title of this pull request.

### AI Tooling
- [x] The PR includes the phrase "Generated by GitHub Copilot (Claude)", where the tool is the name of the AI tool used.
- [x] My use of AI contributions follows the ASF legal policy.

### How has this been tested?
- [x] New unit tests were added to cover new or changed code paths.
- [x] `make test_all` was run, and no failures reported.
- [ ] A pull request will be opened for new e2e tests (apache/yunikorn-k8shim repository).

### Questions:
- [ ] The change needs documentation, a pull request for apache/yunikorn-site repository will be created.
- [ ] There is breaking changes for older versions: jira is tagged with `release-notes` label.
- [ ] The licenses files needs to be updated.

### Screenshots or other details
Test output:
```
go test ./pkg/scheduler/objects/ -run TestTryReservedAllocateRequiredNodeNotCancelledOnTimeout -v
--- PASS: TestTryReservedAllocateRequiredNodeNotCancelledOnTimeout (0.00s)
PASS
```

PR #1114, opened 2026-07-30
[YUNIKORN-3321] Fix removeAsksInternal return ignored for reservation count

kaijaytu:
### Description

Fix `removeAsksInternal` return value being ignored in three callers, causing the partition's reservation counter to not be decremented correctly:

- `partitionContext.removeApplication`: capture return value of `RemoveAllocationAsk("")` and call `decReservationCount`
- `partitionContext.removeAllocation`: capture return value of `RemoveAllocationAsk(allocationKey)` and call `decReservationCount`
- `timeoutPlaceholderProcessing`: add `reservationReleasedCallback` (following the existing `terminatedCallback` pattern) so the async placeholder timeout can notify the partition to decrement the reservation counter

### Type of change

- [x] Bug Fix

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3321

### AI Tooling
- [x] The PR includes the phrase "Generated by GitHub Copilot (Claude)", where the tool is the name of the AI tool used.

### How has this been tested?

The fix ensures the partition reservation counter stays consistent when:
1. An application with active reservations is removed
2. An allocation with a reserved ask is released (non-timeout)
3. Placeholder timeout triggers removal of all pending asks that have reservations

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -121,11 +121,12 @@ type Application struct {
 	runnableByUserLimit  bool                        // whether the application is runnable/schedulable based on user/group quota. Default is true.
 	backoffDeadline      time.Time                   // no scheduling from this application until this deadline
 
-	rmEventHandler        handler.EventHandler
-	rmID                  string
-	terminatedCallback    func(appID string)
-	appEvents             *schedEvt.ApplicationEvents
-	sendStateChangeEvents bool // whether to send state-change events or not (simplifies testing)
+	rmEventHandler              handler.EventHandler
+	rmID                        string
+	terminatedCallback          func(appID string)
+	reservationReleasedCallback func(released int)
+	appEvents                   *schedEvt.ApplicationEvents
+	sendStateChangeEvents       bool // whether to send state-change events or not (simplifies testing)
 
 	locking.RWMutex
 }
@@ -483,7 +484,8 @@ func (sa *Application) timeoutPlaceholderProcessing() {
 			zap.Int("pending", len(pendingRelease)),
 			zap.Int("preempted", preempted),
 			zap.String("gang scheduling style", sa.gangSchedulingStyle))
-		sa.removeAsksInternal("", si.EventRecord_REQUEST_TIMEOUT)
+		released := sa.removeAsksInternal("", si.EventRecord_REQUEST_TIMEOUT)
+		sa.executeReservationReleasedCallback(released)
 		// trigger the release of the allocated placeholders: accounting updates when the release is done
 		sa.notifyRMAllocationReleased(toRelease, si.TerminationType_TIMEOUT, "releasing allocated placeholders on placeholder timeout")
 		// trigger the release of the pending placeholders: accounting has been done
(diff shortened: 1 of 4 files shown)


pkg/scheduler/objects/application_test.go
```diff
@@ -3340,6 +3340,52 @@ func TestTryAllocateWithReservedHeadRoomChecking(t *testing.T) {
...
 
+// TestTryReservedAllocateRequiredNodeNotCancelledOnTimeout verifies that required node reservations
+// are not cancelled even if they exceed the reservationWaitTimeout.
+func TestTryReservedAllocateRequiredNodeNotCancelledOnTimeout(t *testing.T) {
+	setupUGM()
+	defer func() {
+		if r := recover(); r != nil {
+			t.Fatalf("reserved headroom test regression: %v", r)
+		}
+	}()
```
pbacsko: Is this part needed? I don't think we need to recover from an assertion panic.

pbacsko:
@kaijaytu check the formatting issues

pkg/scheduler/objects/application_test.go
kaijaytu: Good point--removed. That recover block was copied from the adjacent TestTryAllocateWithReservedHeadRoomChecking and isn't needed. It appeared in this PR because the branch was originally stacked on top of YUNIKORN-3320. The PRs are now rebased independently, and the fix has been pushed to PR #1113 where the test lives.

pkg/scheduler/objects/application.go
```diff
@@ -2162,6 +2164,18 @@ func (sa *Application) executeTerminatedCallback() {
...
 
+func (sa *Application) SetReservationReleasedCallback(callback func(released int)) {
+	sa.Lock()
+	defer sa.Unlock()
+	sa.reservationReleasedCallback = callback
+}
+
+func (sa *Application) executeReservationReleasedCallback(released int) {
+	if released > 0 && sa.reservationReleasedCallback != nil {
+		go sa.reservationReleasedCallback(released)
```
pbacsko, line 2175: Why are we running this on a separate goroutine? Can't we just call `sa.reservationReleasedCallback(released)` synchronously?

kaijaytu: The goroutine is necessary to prevent a deadlock. Here's what I checked:

1. `decReservationCount` requires `pc.Lock()` (partition write lock):
   - partition.go L1793-1796

2. `timeoutPlaceholderProcessing` holds `sa.Lock()` (application write lock) for its entire duration:
   - application.go L407-409: `sa.Lock()` with `defer sa.Unlock()`

3. The reverse lock order exists in `AddApplication`:
   - partition.go L346: `pc.Lock()` with `defer pc.Unlock()`
   - partition.go L404: calls `app.SetTerminatedCallback()` while still holding `pc.Lock()`
   - application.go L2153-2156: `SetTerminatedCallback` internally takes `sa.Lock()`
   - So: `pc.Lock()` -> `sa.Lock()`

4. If the callback were synchronous, `timeoutPlaceholderProcessing` would create:
   - `sa.Lock()` -> `pc.Lock()` (via callback -> `decReservationCount`)

This is a classic lock-ordering inversion:
- Thread A (AddApplication): pc.Lock() -> sa.Lock()
- Thread B (timeoutPlaceholderProcessing): sa.Lock() -> pc.Lock()

The partition struct comment (partition.go L71-80) explicitly documents this constraint:
> "The partition write lock must not be held while manipulating an application... If the partition write lock is held while manipulating an application a deadlock could occur."

The `go` keyword breaks the nesting so `pc.Lock()` is never acquired while `sa.Lock()` is held. This is the same reason `executeTerminatedCallback` (application.go L2159-2161) uses `go` for `moveTerminatedApp`, which also takes `pc.Lock()`.

manirajv06 (approved):
looks good.

Will wait for @pbacsko before merge.

pkg/scheduler/objects/application.go
pbacsko: Thanks.

pbacsko (approved):

PR #1115, opened 2026-07-30
[YUNIKORN-3341] gzip middleware double-compresses responses that handlers already encoded

DrFaust92:
### Description

The gzip middleware on the REST router compresses any response whose buffered body exceeds `minCompressionSize`, without checking whether the wrapped handler had already encoded the body itself.

`/ws/v1/metrics` is served by `promhttp.Handler()`, which performs its **own** gzip content negotiation: when the client sends `Accept-Encoding: gzip` it compresses the payload and sets `Content-Encoding: gzip`. The middleware sees the same `Accept-Encoding`, buffers promhttp's already-gzipped bytes, finds them over the threshold, and compresses a second time. `switchToGzip` uses `Header().Set`, so the response still advertises a single `Content-Encoding: gzip` while carrying a doubly encoded body.

Clients decode once and get gzip framing instead of the payload. For Prometheus this means every scrape of `/ws/v1/metrics` fails to parse and the target is marked down — scheduler metrics are lost entirely for any scraper that advertises gzip support, which is the default.

Reproduced against a running scheduler:

```console
$ curl -s -D- -o body.bin http://scheduler:9080/ws/v1/metrics
200 OK, 69714 bytes, "# HELP go_gc_duration_seconds ..."

$ curl -s -D- -o body.gz -H 'Accept-Encoding: gzip' http://scheduler:9080/ws/v1/metrics
200 OK, Content-Encoding: gzip, 8839 bytes

$ gunzip -c body.gz | file -
/dev/stdin: gzip compressed data          # still gzip after one decode

$ gunzip -c body.gz | gunzip -c | head -1
# HELP go_gc_duration_seconds ...         # payload only after two
```

**Fix:** treat a `Content-Encoding` set by the handler as a signal that the body is already encoded and pass the response through untouched. The buffered bytes are flushed verbatim via a new `passThrough` helper, which `finalize` now reuses for the below-threshold path.

This keeps metrics responses compressed — promhttp still gzips them, just once — and leaves the behaviour of every handler that does not set `Content-Encoding` unchanged.

An alternative would be to pass `promhttp.HandlerOpts{DisableCompression: true}` in `getMetrics`, but that only patches the one endpoint that self-encodes today; the middleware would still corrupt any future handler that does the same. Fixing the middleware seemed like the right layer.

### Type of change

- [x] Bug Fix

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3341

### AI Tooling

Generated by Claude Code.

### How has this been tested?

`pkg/webservice/gzip_test.go` is new — the file did not exist before. `TestCompressResponseSelfEncodedHandler` is the regression test: it fails on current `master` with the extra gzip framing visible in the diff, and passes with the fix.

```
--- FAIL: TestCompressResponseSelfEncodedHandler (0.02s)
    gzip_test.go:109: assertion failed:
        --- payload
        +++ →
          []uint8{
        + 	0x1f, 0x8b, 0x08, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xff, ...
          	... // 16320 identical bytes
        + 	0x01, 0x00, 0x00, 0xff, 0xff, 0xdc, 0x26, 0x2b, 0x3c, 0x00, 0x40, 0x00, 0x00,
          }
```

The self-encoded fixture is deliberately high-entropy, and this matters: a compressible body gzips down *below* `minCompressionSize`, the middleware never commits to a second pass, and the test would pass against the very bug it is meant to catch. The helper asserts its own gzipped size exceeds the threshold so the fixture cannot silently regress into being vacuous.

The remaining tests cover the pre-existing paths — compress over threshold, pass through under threshold, no gzip requested, status-code preservation — and `clientAcceptsGzip`.

### Questions:

No documentation change: this restores the documented behaviour of the endpoint rather than altering it. Not a breaking change — responses stay gzip-encoded for clients that ask, just singly. No new dependencies, so no license file update.

`pkg/webservice` coverage is 90.4% after the change.

### DIFF
--- a/pkg/webservice/gzip.go
+++ b/pkg/webservice/gzip.go
@@ -86,10 +86,19 @@ func (d *deferredGzipResponseWriter) Write(b []byte) (int, error) {
 	return n, err
 }
 
-// passThrough commits to sending the response uncompressed: it flushes any buffered
-// bytes to the underlying writer and marks the decision as final.
+// passThrough commits to sending the body as the handler produced it: it flushes any
+// buffered bytes to the underlying writer and marks the decision as final.
+//
+// Vary is added only when the response actually carries an encoding. A handler that
+// compressed the body itself (promhttp) sets Content-Encoding but never Vary, and a
+// compressed response without Vary lets a shared cache serve those bytes to a client
+// that cannot decode them. A body this middleware simply chose not to compress carries
+// neither header, which is the contract Test_GzipMinCompressionSize asserts.
 func (d *deferredGzipResponseWriter) passThrough() {
 	d.decided = true
+	if d.ResponseWriter.Header().Get("Content-Encoding") != "" {
+		d.ResponseWriter.Header().Add("Vary", "Accept-Encoding")
+	}
 	if d.statusCode != 0 {
 		d.ResponseWriter.WriteHeader(d.statusCode)
 	}
--- a/pkg/webservice/gzip.go
+++ b/pkg/webservice/gzip.go
@@ -64,6 +64,14 @@ func (d *deferredGzipResponseWriter) WriteHeader(code int) {
 // commits to gzip streaming. After the decision is made, writes go directly to the
 // chosen writer.
 func (d *deferredGzipResponseWriter) Write(b []byte) (int, error) {
+	// A handler is allowed to encode the response body itself, in which case it has
+	// already set Content-Encoding by the time it writes. Compressing again would
+	// produce a doubly encoded body advertised by a single Content-Encoding, which no
+	// client can decode. Hand the response over untouched instead.
+	if !d.decided && d.ResponseWriter.Header().Get("Content-Encoding") != "" {
+		d.passThrough()
+	}
+
 	if d.decided {
 		if d.useGzip {
 			return d.gz.Write(b)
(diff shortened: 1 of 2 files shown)


pkg/webservice/gzip.go
```diff
@@ -78,6 +86,22 @@ func (d *deferredGzipResponseWriter) Write(b []byte) (int, error) {
 	return n, err
 }
 
+// passThrough commits to sending the response uncompressed: it flushes any buffered
+// bytes to the underlying writer and marks the decision as final.
+func (d *deferredGzipResponseWriter) passThrough() {
+	d.decided = true
```
pbacsko, line 98: Nit: shouldn't we add Vary:Accept-Encoding here for consistency?

pbacsko (commented):
LGTM, I just have a question

pkg/webservice/gzip.go
DrFaust92: Good catch, added — but scoped to the encoded case.

`Test_GzipMinCompressionSize` already asserts `Vary` must *not* be set when the response is not compressed, so adding it unconditionally in `passThrough` broke that test. The line I settled on: set `Vary` only when the response actually carries a `Content-Encoding`.

That still closes the real gap — promhttp sets `Content-Encoding: gzip` but never `Vary` (nothing in `prometheus/promhttp` sets it), so a passed-through metrics response was going out compressed with no `Vary` at all, which lets a shared cache hand those bytes to a client that cannot decode them. A body we simply chose not to compress keeps neither header, as before.

Happy to make it unconditional and update that assertion instead, if you think the existing expectation is the wrong one.

pbacsko: I'm okay with this change.

pbacsko:
cc @wilfred-s as he wrote the original code

pbacsko (approved):
+1 (pending build + tests)

wilfred-s (approved):
LGTM, agree with the Vary solution chosen

PR #1117, opened 2026-08-03
[YUNIKORN-3323] Replacing google btree with tidwall btree

adityadtu5:
### Description
As google has stopped supporting their btree library, as part of this PR moving to another library which is maintained actively.

### Type of change

- [ ] Bug Fix
- [ ] Improvement
- [ ] Feature
- [x] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3323

### AI Tooling
If an AI tool was used:

### How has this been tested?

### Questions:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,6 @@ go 1.25.0
 require (
 	github.com/apache/yunikorn-scheduler-interface v0.0.0-20260727092410-674338955bdf
 	github.com/go-ldap/ldap/v3 v3.4.13
-	github.com/google/btree v1.1.3
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/julienschmidt/httprouter v1.3.0
@@ -33,6 +32,7 @@ require (
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.67.5
 	github.com/sasha-s/go-deadlock v0.3.9
+	github.com/tidwall/btree v1.8.1
 	go.uber.org/zap v1.27.1
 	go.yaml.in/yaml/v3 v3.0.4
 	golang.org/x/exp v0.0.0-20260312153236-7ab1446f8b90
--- a/go.sum
+++ b/go.sum
@@ -20,8 +20,6 @@ github.com/go-logr/stdr v1.2.2 h1:hSWxHoqTgW2S2qGc0LTAI563KZ5YKYRhT3MFKZMbjag=
 github.com/go-logr/stdr v1.2.2/go.mod h1:mMo/vtBO5dYbehREoey6XUKy/eSumjCCveDpRre4VKE=
 github.com/golang/protobuf v1.5.4 h1:i7eJL8qZTpSEXOPTxNKhASYpMn+8e5Q6AdndVa1dWek=
 github.com/golang/protobuf v1.5.4/go.mod h1:lnTiLA8Wa4RWRcIUkrtSVa5nRhsEGBg48fD6rSs7xps=
-github.com/google/btree v1.1.3 h1:CVpQJjYgC4VbzxeGVHfvZrv1ctoYCAI8vbl07Fcxlyg=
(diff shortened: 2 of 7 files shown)


manirajv06:
Filed https://issues.apache.org/jira/browse/YUNIKORN-3342 to track the flaky test failures.

manirajv06 (commented):
looks good.

https://github.com/tidwall/btree-benchmark shows `Walk` is faster than `Scan`. Any reason not choosing the former approach?

wilfred-s:
LGTM, but before we commit Scan vs Walk should be checked.

Can we please add to the jira the output of `BenchmarkScheduling()` for:
* old (google based btree)
* new tidwall BtreeG with Scan
* new tidwall BtreeG with Walk

Walk might be "faster" in the benchmark but does it show for us, and if it shows is the code complexity needed to use it a good trade-off for the speed gain.

tigerquoll:
I would suggest some level of caution - Walk is faster for iteration because it directly returns the internal element slice of the b-tree node, bypassing the element by element iteration delivery function call - consider it a batch-oriented iteration call back that  primary saves time because you get called once per node rather then once per element. So the time savings are proportional to how much getting a single function call per leaf node saves versus the per-element Yunikorn processing overhead.  This needs to be weighed up against risk of logic changes to your processing code - as using walk means you are no longer proposing a "drop-in" replacement for the old tree implementation.  My understanding of Yunikorn processing loops is that they are reasonably heavy compared to function call overhead, thus the overall savings for this change would likely be minimal.

On the subject of being a "drop-in" replacement - I believe tidwall defaults to having its own internal lock (the read lock gets held for the entire scan/walk operation - write lock gets held for sortedNode access ), which should probably be turned off with the NoLocks: true option - Yunikorn-level locks should already adequately protect the structure from concurrent modification.

adityadtu5:
Performance testing shows negligible difference between google's ascend and tidwall's walk or scan.
```
amaheshwari@M23LJ97R47 ~/D/G/o/f/y/yunikorn-core (master)> benchstat tidwall_scan.txt tidwall_walk.txt google_ascend.txt
goos: darwin
goarch: arm64
pkg: github.com/apache/yunikorn-core/pkg/scheduler/tests
cpu: Apple M4 Pro
                                  │ tidwall_scan.txt │         tidwall_walk.txt          │         google_ascend.txt         │
                                  │      sec/op      │   sec/op    vs base               │   sec/op    vs base               │
Scheduling/500Nodes/10000Pods-14          1.220 ± 2%   1.205 ± 1%  -1.20% (p=0.004 n=10)   1.214 ± 1%  -0.44% (p=0.035 n=10)
Scheduling/1000Nodes/10000Pods-14         1.225 ± 1%   1.219 ± 0%       ~ (p=0.218 n=10)   1.214 ± 1%  -0.91% (p=0.002 n=10)
Scheduling/2000Nodes/10000Pods-14         1.226 ± 1%   1.217 ± 1%  -0.76% (p=0.023 n=10)   1.214 ± 1%  -0.98% (p=0.043 n=10)
Scheduling/5000Nodes/10000Pods-14         1.225 ± 1%   1.222 ± 1%       ~ (p=0.280 n=10)   1.235 ± 2%       ~ (p=0.165 n=10)
geomean                                   1.224        1.216       -0.68%                  1.219       -0.38%
```

pbacsko (approved):

manirajv06:
looks good

tigerquoll:
I still think this PR is worthwhile - the older btree module is no longer being supported.

wilfred-s:
> I still think this PR is worthwhile - the older btree module is no longer being supported.

We do not use github to commit PRs as that replaces the committer with the github bot. That removes tracking of who was responsible for the commit. When using the github UI the committer is shown but in the real logs the committer is set with a like this:
```
Commit:  GitHub <noreply@github.com>
```
We use the command line to keep the original details but that means GH UI always marks the PR as closed.

PR #1122, opened 2026-08-08
[YUNIKORN-3352] Replace the O(n) ask-priority rescan with a pending histogram

tigerquoll:
Part of YUNIKORN-3350, which splits one change into three independent levers.

YUNIKORN-3351 (hoist the per-ask backoff getter) has landed as 7dc1287, and this branch has been
rebased onto current master. **The diff is now this change only** — two commits, `application.go`
plus its tests, nothing borrowed from the dependency.

The third lever, YUNIKORN-3353 (keep only pending asks in sortedRequests), is deliberately held back
until these two are settled — it carries an ordering change that deserves to be reviewed on its own.

## What

`Application.updateAskMaxPriority` iterates the whole `sa.requests` map to recompute the highest
priority among asks that are not yet allocated. It is called from `allocateAsk` on every allocation
whose priority is at or above the current maximum, and again from `removeAsksInternal` on ask removal.
Each call is O(number of asks), so scheduling N asks costs O(N^2), and the cost grows as an
application gets larger.

It cannot stop early, because pending and allocated asks share one map — it has to test
`IsAllocated()` on every entry, and that takes a read lock each time. In a CPU profile of master the
function is **55% of `Application.tryAllocate`**, split 56% map iteration, 22% `IsAllocated`, 15%
`GetPriority`.

Replace it with a per-priority histogram of pending asks (`pendingPriorities map[int32]int`),
maintained incrementally as asks enter and leave the pending set, so `askMaxPriority` is available in
O(1). `updateAskMaxPriority` is deleted.

This PR touches `application.go` only. It changes no ordering semantics anywhere.

## Design notes

A map of counts rather than a set, because many asks share a priority and the structure has to know
when a priority band actually empties. Every operation is O(1) except one: when the current top bucket
empties, the new maximum is recomputed by scanning the map keys, which is O(distinct priority values
in use) — a handful in practice, not O(number of asks).

`setAskMaxPriority` propagates to the queue only when the value actually changes, so this also issues
strictly fewer `UpdateApplicationPriority` calls than the rescan it replaces.

The non-O(1) path is worth quantifying, since it is the obvious question. `askMaxPriority` is read by
all four application sort policies (`considerPriority` decides whether priority is the primary or
secondary key, not whether it is consulted), so reads sit on the scheduling loop and scale with
applications per queue. Recomputation happens only when the *top* band empties. Measured on
BenchmarkScheduling, 5000 nodes, 10000 asks, with both applications in one queue so the sort
comparator actually runs:

```
GetAskMaxPriority = 10,000    inc = 10,000    dec = 10,000
recomputations    = 2         map keys scanned across both = 0
```

Two recomputations, one per application, when its last pending ask is allocated — a 5,000:1 ratio of
reads to recomputations. Under an adversarial mix (the property fuzzer, 11 distinct priorities, heavy
add/allocate/deallocate churn) it rises to 10.5% of decrements, still averaging only 2.0 map keys
each. Even the pathological shape — a high-priority ask interleaved with lower ones so the top band
drains on every allocation — is O(asks x distinct priorities), against the O(asks^2) it replaces.

One guard comes with it. `deallocateAsk` returns an ask to the histogram only when it is still the
tracked object for its key. `removeAsksInternal("")` wipes `sa.requests` while `sa.allocations`
survives until the shim confirms the releases, and a release arriving in that window reaches
`RollbackAllocation`. The rescan derived the maximum by scanning `sa.requests`, so an ask absent from
it never influenced the converged result; the guard preserves that.

## Measured effect

`BenchmarkScheduling`, allocate phase, 10000 pods, alternating A/B, 3 samples, medians. Linux 6.8
aarch64, 6 CPUs, GOMAXPROCS=6. Cumulative figures — the stack up to and including this PR, against
master:

| Configuration | master (c/s) | this PR (c/s) | cumulative | this PR alone |
|---|---:|---:|---:|---:|
| 500 nodes  | 5,481 | 16,832 | 3.07x | 2.49x |
| 1000 nodes | 5,517 | 16,351 | 2.96x | 2.49x |
| 2000 nodes | 5,396 | 16,419 | 3.04x | 2.49x |
| 5000 nodes | 5,353 | 16,333 | 3.05x | 2.44x |

"This PR alone" is the ratio against the previous PR in the stack. Spread up to 7%.

The table above was measured on base 54fa576, before 3351, YUNIKORN-3323 (tidwall btree, #1117) and
YUNIKORN-3319 (#1120) landed.

Scope: mock resource manager, no real bind, so this is core scheduling headroom rather than a
real-cluster pods/second figure. The workload is 2 applications x 5,000 asks; the cost removed is
per-application and quadratic in that application's ask count, so this shape is favourable — note
master's throughput is flat across node counts, which is the per-app rescan dominating everything
else. A cluster of many small applications sees a proportionally smaller absolute gain.

## Behaviour

Unchanged. The golden decision-trace tests added in YUNIKORN-3338 reproduce byte-for-byte, goldens not
regenerated. They pin the ordered sequence of decisions the resource manager observes through
`UpdateAllocation`, so reproducing them is the claim.

One edge worth stating precisely: `cleanupAsks` now also resets the histogram and `askMaxPriority`.
That happens only on entering a terminal state (Completed/Failed), where the application is no longer
schedulable, so it changes no scheduling decision — but it does mean a terminal application no longer
retains a stale `askMaxPriority`.

A property fuzzer (`application_property_test.go`) drives an Application through long randomized
interleavings of the real entry points — `AddAllocationAsk` including the replace-an-existing-ask
branch, `AllocateAsk`, `DeallocateAsk`, `RemoveAllocationAsk`, `RecoverAllocationAsk`,
`AddAllocation`, `RollbackAllocation`, and FSM-driven cleanup — and after every single step
re-derives the expected histogram and `askMaxPriority` from an independent reference model. Coverage
high-water marks are asserted so an operation that becomes a silent no-op fails rather than passing
trivially.

## Concurrency

No new lock acquisition and no new ordering. `setAskMaxPriority` takes the queue lock under the
application lock exactly as the rescan it replaces did, and does so less often. All histogram
mutation happens under the application write lock; the helpers are unlocked internals whose callers
all hold it. Verified green under `-race`, and under go-deadlock with lock-order detection armed
(`DEADLOCK_DETECTION_ENABLED=true DEADLOCK_DISABLE_LOCK_ORDER=false`).

## Scope and rollback

No configuration, REST, scheduler-interface, or metrics changes. No new public API. No persisted or
serialized state — `pendingPriorities` is core-internal and never crosses a process or version
boundary, so there is nothing to migrate and no mixed-version concern during a rolling upgrade.
Rollback is a single revert.

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1084,15 +1084,17 @@ func (sa *Application) canReplace(request *Allocation) bool {
 func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption bool, preemptionDelay time.Duration, preemptAttemptsRemaining *int, nodeIterator func() NodeIterator, fullNodeIterator func() NodeIterator, getNodeFn func(string) *Node) *AllocationResult {
 	sa.Lock()
 	defer sa.Unlock()
-	if sa.sortedRequests == nil {
+	if len(sa.sortedRequests) == 0 {
 		return nil
 	}
 	// calculate the users' headroom, includes group check which requires the applicationID
 	userHeadroom := ugm.GetUserManager().Headroom(sa.queuePath, sa.ApplicationID, sa.user)
 	unschedulable := uint64(0)
+	// constant for the cycle: hoisted out of the loop below. Safe to call unconditionally here
+	// because the len check above guarantees at least one iteration would occur.
+	backoffThreshold := sa.queue.GetMaxAppUnschedAskBackoff()
 	// get all the requests from the app sorted in order
 	for _, request := range sa.sortedRequests {
-		backoffThreshold := sa.queue.GetMaxAppUnschedAskBackoff()
 		if backoffThreshold > 0 && unschedulable >= backoffThreshold {
 			log.Log(log.SchedApplication).Info("too many unschedulable asks in the application, waiting",
 				zap.String("application ID", sa.ApplicationID),
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -115,7 +115,8 @@ type Application struct {
 	rejectedMessage      string                      // If the application is rejected, save the rejected message
 	stateLog             []*StateLogEntry            // state log for this application
 	placeholderData      map[string]*PlaceholderData // track placeholder and gang related info
-	askMaxPriority       int32                       // highest priority value of outstanding asks
+	askMaxPriority       int32                       // highest priority value of outstanding (pending) asks
+	pendingPriorities    map[int32]int               // count of pending (non-allocated) asks per priority value
 	hasPlaceholderAlloc  bool                        // Whether there is at least one allocated placeholder
 	runnableInQueue      bool                        // whether the application is runnable/schedulable in the queue. Default is true.
 	runnableByUserLimit  bool                        // whether the application is runnable/schedulable based on user/group quota. Default is true.
@@ -155,6 +156,7 @@ func NewApplication(siApp *si.AddApplicationRequest, ugi security.UserGroup, eve
 		rejectedMessage:       "",
 		stateLog:              make([]*StateLogEntry, 0),
 		askMaxPriority:        configs.MinPriority,
+		pendingPriorities:     make(map[int32]int),
 		sortedRequests:        sortedRequests{},
 		sendStateChangeEvents: true,
 		runnableByUserLimit:   true,
@@ -597,6 +599,7 @@ func (sa *Application) removeAsksInternal(allocKey string, detail si.EventRecord
 		}
 		sa.requests = make(map[string]*Allocation)
 		sa.sortedRequests = sortedRequests{}
+		sa.pendingPriorities = make(map[int32]int)
 		sa.askMaxPriority = configs.MinPriority
 		sa.queue.UpdateApplicationPriority(sa.ApplicationID, sa.askMaxPriority)
 	} else {
@@ -612,13 +615,12 @@ func (sa *Application) removeAsksInternal(allocKey string, detail si.EventRecord
 				deltaPendingResource = ask.GetAllocatedResource()
 				sa.pending = resources.Sub(sa.pending, deltaPendingResource)
 				sa.pending.Prune()
+				// the removed ask was pending: drop it from the priority histogram
+				sa.removeFromPriorities(ask.GetPriority())
 			}
 			delete(sa.requests, allocKey)
 			sa.sortedRequests.remove(ask)
 			sa.appEvents.SendRemoveAskEvent(sa.ApplicationID, ask.allocationKey, ask.GetAllocatedResource(), detail)
-			if priority := ask.GetPriority(); priority >= sa.askMaxPriority {
-				sa.updateAskMaxPriority()
-			}
 		}
 	}
 	// clean up the queue pending resources
@@ -664,6 +666,9 @@ func (sa *Application) AddAllocationAsk(ask *Allocation) error {
 	var oldAskResource *resources.Resource = nil
 	if oldAsk := sa.requests[ask.GetAllocationKey()]; oldAsk != nil && !oldAsk.IsAllocated() {
 		oldAskResource = oldAsk.GetAllocatedResource().Clone()
+		// the old ask was pending and is being replaced: drop it from the priority histogram so the
+		// new ask's addAllocationAskInternal (via addToPriorities) nets correctly.
+		sa.removeFromPriorities(oldAsk.GetPriority())
 	}
 
 	// Check if we need to change state based on the ask added, there are two cases:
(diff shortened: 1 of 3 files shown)


pbacsko (approved):
+1 LGTM (pending rebase)

tigerquoll:
@pbacsko PR rebased - can you trigger re-running of tests?

pkg/scheduler/objects/application.go
```diff
@@ -664,6 +666,9 @@ func (sa *Application) AddAllocationAsk(ask *Allocation) error {
 	var oldAskResource *resources.Resource = nil
 	if oldAsk := sa.requests[ask.GetAllocationKey()]; oldAsk != nil && !oldAsk.IsAllocated() {
 		oldAskResource = oldAsk.GetAllocatedResource().Clone()
+		// the old ask was pending and is being replaced: remove it from the pending histogram so the
+		// new ask's addAllocationAskInternal (via incPendingPriority) nets correctly.
+		sa.decPendingPriority(oldAsk.GetPriority())
 	}
```
wilfred-s, line 672: I am just walking through the callers of `AddAllocationAsk()`. There is only one caller: `PartitionContext.UpdateAllocation()`. The call to add the ask is only triggered if the allocationKey is not found in the requests for the app. That means `oldAsk` will always be nil.

There could be a small exception of a race condition. This is however negated by the fact that the processing loop that triggers this update via `PartitionContext.UpdateAllocation()` is single threaded. We process changes in order from the informers as one update could require the previous one.

We can thus not have an add and remove at the same time processing.
Test code however relies on this check. We need a new jira to clean this up.

Two further points which need new jiras:
1) only in place pod vertical scaling can change the resources of a container or pod. For that the pod must be running. That means asks, pending allocations, can not change their resources only allocations can.
2) priority class is an immutable object, pods must reference a priority on creation and that value is also immutable. Priority can thus not change after creation.

pkg/scheduler/objects/application.go
```diff
@@ -778,15 +783,68 @@ func (sa *Application) RecoverAllocationAsk(alloc *Allocation) {
...
 }
 
+// incPendingPriority records that an ask at priority p just became pending (added, or
+// deallocated back to pending). Call with sa.Lock() held.
+func (sa *Application) incPendingPriority(p int32) {
```
wilfred-s: NIT: rename to `addToPriorities()`

pkg/scheduler/objects/application.go
```diff
@@ -778,15 +783,68 @@ func (sa *Application) RecoverAllocationAsk(alloc *Allocation) {
...
+	sa.pendingPriorities[p]++
+	if p > sa.askMaxPriority {
+		sa.setAskMaxPriority(p)
+	}
+}
+
+// decPendingPriority records that a pending ask at priority p just left the pending set
+// (allocated, or removed). Call with sa.Lock() held.
+func (sa *Application) decPendingPriority(p int32) {
```
wilfred-s: NIT: rename to `removeFromPriorities()`

pkg/scheduler/objects/application.go
```diff
@@ -877,11 +933,19 @@ func (sa *Application) deallocateAsk(ask *Allocation) (*resources.Resource, erro
...
+	// finds the entry in sa.allocations and deallocates it. The identity comparison (not just a
+	// presence check) also covers a stale ask object that has since been replaced by a new ask under
+	// the same key.
+	// This matches the converged behaviour of the full rescan this change replaced:
+	// updateAskMaxPriority derived the max by scanning sa.requests, so an ask absent from sa.requests
+	// never influenced it. Without the guard that pre-existing accounting drift would turn into a
+	// permanent leak in the incremental histogram instead.
+	if sa.requests[ask.GetAllocationKey()] == ask {
+		sa.incPendingPriority(ask.GetPriority())
 	}
```
wilfred-s, line 949: This is is not needed, `deallocateAsk()` is the unlocked version of `DeallocateAsk()` which already does this check. All other callers must have taken the application write lock before calling this and retrieved the ask reference from the application inside that lock. That means the ask has to exist in the requests map. This should be a straight call without the check.

pkg/scheduler/objects/application.go
```diff
@@ -2278,6 +2330,10 @@ func (sa *Application) GetAskMaxPriority() int32 {
 func (sa *Application) cleanupAsks() {
 	sa.requests = make(map[string]*Allocation)
 	sa.sortedRequests = nil
+	// a Failed app can still hold pending asks: reset the histogram or the consistency check would
+	// fire on terminal apps.
```
wilfred-s, line 2334: This is called after the application is removed from the queue. This cleans up the app state internally. It is not relevant anymore as the queue and the partition are no longer linked to the app. Nothing can reference these fields or trigger updates. The only reason to do this is to release memory. The comment is not correct.

wilfred-s (changes requested):

pkg/scheduler/objects/application.go
tigerquoll: You are right - oldAsk is always nil in production.  Only a single-threaded add from a new ask or a recovery option results in an insertion into sa.requests — and the two operations cannot overlap by basic design.

[YUNIKORN-3358](https://issues.apache.org/jira/browse/YUNIKORN-3358) added for the cleanup task

pkg/scheduler/objects/application.go
tigerquoll: Comment corrected

pkg/scheduler/objects/application.go
tigerquoll: @wilfred-s  Some overly-cautious checks here due to me not being completely familiar with the code base.

`DeallocateAsk` does the same check, and you're right that this guard is redundant because of that - it passes the object it just found, so identity holds by construction. Same for the two revert points in `tryPlaceholderAllocate`, which source from `sortedRequests`. The odd one out seems to be `RollbackAllocation`, which takes its ask from `sa.allocations` and never consults `sa.requests`, so the invariant isn't enforced there the way it is elsewhere - is that something we need to worry about?

I couldn't find a path in production code where the ask is still in `sa.allocations` when it has been removed from `sa.requests`. The gang-timeout route I traced is blocked: the outstanding placeholder keeps `allocatedPlaceholder` non-zero, which holds the app in Resuming, and `RollbackAllocation` rejects anything that isn't Accepted or Running.

The reason I added it: master bumps `askMaxPriority` up in `deallocateAsk` with no check either, but there it self-corrects - the value only ever moves up incrementally, and `updateAskMaxPriority` recomputes it wholesale from sa.requests whenever the maximum could drop. The histogram has no equivalent recompute, so anything it accumulates is permanent.

If you're sure this can't be an issue then I will remove the check.

wilfred-s: > The odd one out seems to be RollbackAllocation, which takes its ask from sa.allocations and never consults sa.requests, so the invariant isn't enforced there the way it is elsewhere - is that something we need to worry about?

we should check that the key is in requests in `RolbackAllocation`. The current behaviour is that as long as the pod exists on K8s the reference is maintained in `requests`. However we should not blindly assume that as it would cause the pod to be ignored in late scheduling cycles.
Filed a jira

tigerquoll: I may have found a path: 
a soft-gang app that is Running via a plain (non-task-group) allocation while none of its placeholders ever allocated. The placeholder timeout then takes the else branch - `ResumeApplication` is rejected from Running, so the app stays rollback-eligible — and `removeAsksInternal("")` wipes `sa.requests` while `sa.allocations` survives until the shim confirms the releases. A `SCHEDULING_FAILED_ON_RM` for the in-flight bind arriving in that window reaches RollbackAllocation, and deallocateAsk gets an ask that is no longer tracked.

#1127 adds `TestRollbackAllocationAskNotTracked` to pin this, and also puts the pending-resource re-add behind the same check: without it, an untracked rollback leaves phantom pending on the app and queue, so the app stops completing on its own and queue pending stays inflated until the app is removed. 

On the requests check in `RollbackAllocation`: #1127 effectively implements the safe half of that, one level down - `deallocateAsk` only returns an ask to the pending structures when it is still tracked, so an untracked rollback can no longer produce an ask that looks pending but never schedules, and that holds for every caller. A hard reject in `RollbackAllocation` itself would overshoot: on error the partition drops the release, so the node and queue would keep counting a dead allocation. What may still be worth doing under your jira is a warning log when the key is missing from requests.  If you do proceed with a JIRA, I will add a ref to #1127 about it.

wilfred-s: That case cannot happen for multiple reasons.

removeAsksInternal without an ID is only called on removal of the application, triggered from the k8shim on completion of the last application pod. It is a cleanup, not a simple pod exit.
There cannot be a running allocation in a gang setup if none of the placeholders ever got allocated. Does not matter if that is soft or hard gang scheduling. Timing out of the placeholders does not start until the first placeholder is allocated. At least one placeholder must be running for the timeout to start counting down. While the placeholder timeout runs and not all placeholders are allocated we cannot have non placeholders allocations.
We also keep the application in an accepted state until all placeholders are allocated.
If all placeholders are allocated the first non placeholder will be released by the k8shim. That might be a gang scheduled allocation but it does not have to be. Allocating that ask moves the state of the app to running.

Further comments will be added in #1127

wilfred-s (approved):
LGTM
will fix one comment on commit

pbacsko (approved):

PR #1124, opened 2026-08-11
[YUNIKORN-3357] Add goroutine leak detection to tests via goleak

tigerquoll:
### What is this PR for?

Adds goroutine leak detection to the test suite using [uber-go/goleak](https://github.com/uber-go/goleak), wired into **every test package** (20 of them) via a shared `pkg/common/leakcheck` helper:

- One `TestMain` per package; all shared exemptions live in a single documented list in `leakcheck.options()`; `VerifyTestMain` accepts per-package extra options so the shared list never needs widening
- `pkg/rmproxy` is the one deliberate exception: it has no test functions, so a hook there would guard nothing (documented in the package doc)

No existing test or production code is modified. `goleak` was already a transitive dependency (`go.sum` unchanged, re-verified after the rebase onto current master); it is promoted to a direct test-only dependency.

### The baseline exemption list is a burn-down list

Switching detection on surfaced seven pre-existing leaked-goroutine occurrences. Each has an exemption that names its JIRA, so this change lands green and blocks new *kinds* of leaks immediately; each exemption is removed by its own follow-up fix. Status after rebasing onto master (2026-08-30):

| # | Goroutine | Cause | JIRA | Status |
|---|---|---|---|---|
| 1 | `EventSystemImpl.StartServiceWithPublisher.func1` | tests start the event system without `Stop()`; core is not restartable in-process | YUNIKORN-3370 | open |
| 2 | `eventPublisher.start.func1` | same as 1 (the configmap-callback restart case was folded into 3370) | YUNIKORN-3370 | open |
| 3a | `EventStreaming.CreateEventStream.func1` | test never calls `RemoveStream` | YUNIKORN-3372 | **fixed** (#1135) |
| 3b | `EventStreaming.CreateEventStream.func1` | history-replay send at stream open is a bare `consumer <- event` | YUNIKORN-3364 | PR #1133 open |
| 3c | `EventStreaming.CreateEventStream.func1` | in-loop `consumer <- event` is also bare, so slow-consumer eviction cannot release the forwarder; `TestEventStreaming_SlowConsumer` still leaks after 3a | YUNIKORN-3436 | open, filed 2026-08-30 |
| 4 | `partitionManager.cleanRoot` | tests build a `ClusterContext` without `Stop()` | YUNIKORN-3370 | open |
| 5 | `partitionManager.cleanExpiredApps` | same as 4 | YUNIKORN-3370 | open |
| 6 | `security.UserGroupCache.run` | same as 4 — `ClusterContext.Stop()` already stops it | YUNIKORN-3370 | open |
| 7 | `ClusterContext.notifyRMNewAllocation` | shutdown race: `StopAll` stops the scheduler before the RM proxy, a queued allocation notify blocks forever on its unbuffered reply. Maintainers consider it benign at process exit | YUNIKORN-3365 | open, no fix scheduled |

Row 3c was described in the original version of this table but had no JIRA of its own; it does now. Row 7 reproduced in ~43% of `pkg/scheduler/tests` runs when this PR was opened and in 0 of 11 on current master; the exemption stays because the race is unchanged in the code, and removing a race guard only manufactures a flake.

Known limitation, stated in the code: the exemptions match on top stack frame, so the baseline is a ratchet against new leak *kinds*, not an instance count; three entries key on positional `.funcN` closure names. A dependency bump can also rename a frame out from under an exemption — this happened to the shim PR (#1061) when its core pin moved, so the caveat is now documented in both repos.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-3357

Rebased onto current master and re-verified:

- `go test ./pkg/... -race` with deadlock detection enabled: all 20 packages ok
- Every exemption re-probed by deleting it and confirming the check then fails — all seven are still live on master; none is dead weight
- `make lint`, `make license-check`, `go vet ./pkg/...`, `gofmt`, `go mod tidy` (no-op): clean

Note: `pkg/scheduler` and `pkg/scheduler/tests` fail under `go test -count=5` on unmodified master as well (`TestAllocReserveNewNode`, `TestApplicationHistoryTracking`, `TestPlaceholderBiggerThanReal`, `TestRequiredNodeCancelOtherReservations`, `TestScheduleRemoveReservedAsk`), with no goleak output — pre-existing non-idempotent tests over process-global state, unrelated to this change.

Generated by the Author with assistance from Claude Code

### DIFF
--- a/pkg/common/leakcheck/leakcheck.go
+++ b/pkg/common/leakcheck/leakcheck.go
@@ -56,28 +56,33 @@ import (
 //
 // The list is the baseline that was present when detection was switched on, and
 // it is meant to be burned down rather than extended. Each exemption names its
-// goroutine and the JIRA that tracks removing it.
+// goroutine and the JIRA that tracks it.
 func options() []goleak.Option {
 	return []goleak.Option{
-		// Event system handler (EventSystemImpl.StartServiceWithPublisher). See YUNIKORN-3366 (test cleanup) and YUNIKORN-3363 (restart-after-Stop).
+		// Event system handler (EventSystemImpl.StartServiceWithPublisher). See YUNIKORN-3370.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*EventSystemImpl).StartServiceWithPublisher.func1"),
 
-		// Event publisher (eventPublisher.start). See YUNIKORN-3363.
+		// Event publisher (eventPublisher.start). See YUNIKORN-3370.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*eventPublisher).start.func1"),
 
-		// Event stream forwarder (EventStreaming.CreateEventStream). See YUNIKORN-3364.
+		// Event stream forwarder (EventStreaming.CreateEventStream). Two bare
+		// sends: the history replay (YUNIKORN-3364, fix in #1133) and the in-loop
+		// consumer send hit by slow-consumer eviction (YUNIKORN-3436). The
+		// test-hygiene half landed as YUNIKORN-3372.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*EventStreaming).CreateEventStream.func1"),
 
-		// Partition queue cleaner (partitionManager.Run). See YUNIKORN-3366.
+		// Partition queue cleaner (partitionManager.Run). See YUNIKORN-3370.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/scheduler.(*partitionManager).cleanRoot"),
 
-		// Partition expired-application cleaner (partitionManager.Run). See YUNIKORN-3366.
+		// Partition expired-application cleaner (partitionManager.Run). See YUNIKORN-3370.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/scheduler.(*partitionManager).cleanExpiredApps"),
 
-		// User/group cache cleaner (UserGroupCache.run). See YUNIKORN-3366.
+		// User/group cache cleaner (UserGroupCache.run). See YUNIKORN-3370.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/common/security.(*UserGroupCache).run"),
 
-		// Scheduler allocation notify (ClusterContext.notifyRMNewAllocation). See YUNIKORN-3365.
+		// Scheduler allocation notify parked on an RM reply the proxy never
+		// drains at shutdown. Maintainers consider it benign at process exit; no
+		// fix scheduled. See YUNIKORN-3365.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/scheduler.(*ClusterContext).notifyRMNewAllocation"),
 	}
 }
--- a/pkg/common/leakcheck/leakcheck.go
+++ b/pkg/common/leakcheck/leakcheck.go
@@ -48,6 +48,12 @@ import (
 // the list stops new KINDS of leak from being added, it is not a proof that the
 // exempted counts stay put.
 //
+// Two matching caveats: the three ".func1" entries key on positional,
+// compiler-assigned closure names, so inserting an earlier closure in the same
+// method silently breaks the match (goleak v1.3.0 cannot match the creator
+// frame); and a test that catches one of these goroutines mid-body, rather than
+// parked in its select, sees a different top frame and can fail spuriously.
+//
 // The list is the baseline that was present when detection was switched on, and
 // it is meant to be burned down rather than extended. Each exemption names its
 // goroutine and the JIRA that tracks removing it.
--- a/pkg/common/leakcheck/leakcheck.go
+++ b/pkg/common/leakcheck/leakcheck.go
@@ -41,86 +41,37 @@ import (
 
 // options returns the goleak exemptions shared by all instrumented packages.
 //
-// What this list does and does not buy us, precisely:
+// goleak already filters out the goroutines that the testing, runtime and
+// tracing packages run themselves, so every entry below is a goroutine that
+// really does outlive the test binary today. IgnoreTopFunction matches on the
+// top stack frame only, so an entry can also mask a new leak of the same shape:
+// the list stops new KINDS of leak from being added, it is not a proof that the
+// exempted counts stay put.
 //
-//   - goleak already filters out the goroutines that the testing, runtime and
-//     tracing packages run themselves, so every entry below is a goroutine that
-//     really does outlive the test binary today.
-//   - IgnoreTopFunction matches on the top stack frame only. It is not bounded
-//     by count and not bounded by package. So the baseline stops new KINDS of
-//     leak from being added; it does not stop new instances of these seven
-//     shapes, and a package that has never leaked any of them still has them
-//     exempted because the list is shared. It is a ratchet against regression,
-//     not a proof that the exempted counts stay put.
-//   - Matching also assumes the goroutine is parked in its select when goleak
-//     takes the stack snapshot, which holds for the default tick intervals used
-//     in tests. A test that shortens an interval enough to catch one of these
-//     mid-body would see a different top frame and a spurious failure.
-//
-// The list is the baseline that was present when detection was switched on. It
-// is meant to be burned down, not extended: each entry names the goroutine,
-// what starts it, and what has to change before the entry can be deleted.
-//
-// Three entries key on compiler-assigned closure names (".func1"). Those names
-// are positional: adding another closure earlier in the same enclosing method
-// renumbers them and the exemption silently stops matching, turning the
-// affected package red. goleak v1.3.0 cannot match on the creator frame, so
-// there is no more robust spelling available here. The durable fix is to hoist
-// those three goroutine bodies into named methods, which is a production change
-// and belongs in the follow-up that fixes the leaks themselves.
+// The list is the baseline that was present when detection was switched on, and
+// it is meant to be burned down rather than extended. Each exemption names its
+// goroutine and the JIRA that tracks removing it.
 func options() []goleak.Option {
 	return []goleak.Option{
-		// Event system handler, started by EventSystemImpl.StartServiceWithPublisher.
-		// Tests across events, scheduler and objects call events.Init() followed by
-		// StartService()/StartServiceWithPublisher() without a matching Stop().
-		// Stoppable: delete this entry once those tests defer Stop().
+		// Event system handler (EventSystemImpl.StartServiceWithPublisher). See YUNIKORN-3366 (test cleanup) and YUNIKORN-3363 (restart-after-Stop).
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*EventSystemImpl).StartServiceWithPublisher.func1"),
 
-		// Shim event publisher, started by eventPublisher.start via StartService.
-		// Leaks together with the handler above, and additionally when a configmap
-		// update restarts an already stopped event system: Init() registers a
-		// configmap callback that Stop() never removes, so a reload after Stop()
-		// resurrects a system nobody holds a reference to. Suspected production
-		// defect; delete this entry once Stop() deregisters the callback.
+		// Event publisher (eventPublisher.start). See YUNIKORN-3363.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*eventPublisher).start.func1"),
 
-		// Event stream forwarder, started by EventStreaming.CreateEventStream.
-		// Two distinct causes share this top frame. One is a test that creates a
-		// stream and never calls RemoveStream. The other is a consumer that stops
-		// reading, which wedges the forwarder on the "consumer <- event" send
-		// outside its select, where neither stop channel can reach it; that one is
-		// a suspected production defect. Both must be fixed before this entry goes.
+		// Event stream forwarder (EventStreaming.CreateEventStream). See YUNIKORN-3364.
 		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*EventStreaming).CreateEventStream.func1"),
 
-		// Partition queue cleaner, started by partitionManager.Run when a partition
-		// is added. Tests that build a ClusterContext never call ClusterContext.Stop.
-		// Stoppable: delete this entry once those tests defer Stop().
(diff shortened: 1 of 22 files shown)


chenyulin0719 (approved):
+1 LGTM

- It's a good start to cleaning up goroutine leaks. Exemptions should be tracked later.
- funcN closure names should be tracked later too — positional naming is fragile.
- Comments in leakcheck.go are verbose but clear.

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+		// Stoppable: delete this entry once those tests defer Stop().
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*EventSystemImpl).StartServiceWithPublisher.func1"),
+
+		// Shim event publisher, started by eventPublisher.start via StartService.
+		// Leaks together with the handler above, and additionally when a configmap
+		// update restarts an already stopped event system: Init() registers a
+		// configmap callback that Stop() never removes, so a reload after Stop()
+		// resurrects a system nobody holds a reference to. Suspected production
+		// defect; delete this entry once Stop() deregisters the callback.
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*eventPublisher).start.func1"),
```
wilfred-s: This is done on purpose we need that routine to keep running after a stop to make sure the system can be turned on and off via the config update.
It is also a false positive as the function exits directly after calling. The function is registered but the call only happens on change, no leaks here.

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+		// Suspected production shutdown race: ServiceContext.StopAll stops the
+		// scheduler before the RM proxy, and handleAllocEvent's select can pick a
+		// queued allocation event over the just-closed stop channel. The resulting
+		// notify posts to the RM proxy and blocks on its reply, but the proxy's
+		// handleRMEvents returns on its own stop without draining pendingRMEvents,
+		// so the reply never comes. Intermittent: it needs that select to lose the
+		// race. Delete this entry once shutdown drains or abandons in-flight
+		// replies. Note notifyRMAllocationReleased has the same unbuffered-reply
+		// shape and would leak under a different top frame; it has not been seen
+		// yet, so it is deliberately not exempted here.
```
wilfred-s: Can be added as there is no reason to fix either of these: the events are dropped either way and the app is exiting so nothing to be handled.

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+		// configmap callback that Stop() never removes, so a reload after Stop()
+		// resurrects a system nobody holds a reference to. Suspected production
+		// defect; delete this entry once Stop() deregisters the callback.
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*eventPublisher).start.func1"),
+
+		// Event stream forwarder, started by EventStreaming.CreateEventStream.
+		// Two distinct causes share this top frame. One is a test that creates a
+		// stream and never calls RemoveStream. The other is a consumer that stops
+		// reading, which wedges the forwarder on the "consumer <- event" send
+		// outside its select, where neither stop channel can reach it; that one is
```
wilfred-s: The push outside of the select is the first batch of records during the opening of the stream. This would mean that the requestor closed the connection even before it was really started and they did request a non 0 history (not the default).
This is a defect and needs a jira to fix the push of the history.

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+// there is no more robust spelling available here. The durable fix is to hoist
+// those three goroutine bodies into named methods, which is a production change
+// and belongs in the follow-up that fixes the leaks themselves.
+func options() []goleak.Option {
+	return []goleak.Option{
+		// Event system handler, started by EventSystemImpl.StartServiceWithPublisher.
+		// Tests across events, scheduler and objects call events.Init() followed by
+		// StartService()/StartServiceWithPublisher() without a matching Stop().
+		// Stoppable: delete this entry once those tests defer Stop().
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*EventSystemImpl).StartServiceWithPublisher.func1"),
```
wilfred-s, line 63: This should be fixed on the core side before we turn this check on

pkg/common/leakcheck/leakcheck.go
wilfred-s, line 66: false positive, the function to update from config cannot be called after shutdown as the system is exiting.

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+		// defect; delete this entry once Stop() deregisters the callback.
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*eventPublisher).start.func1"),
+
+		// Event stream forwarder, started by EventStreaming.CreateEventStream.
+		// Two distinct causes share this top frame. One is a test that creates a
+		// stream and never calls RemoveStream. The other is a consumer that stops
+		// reading, which wedges the forwarder on the "consumer <- event" send
+		// outside its select, where neither stop channel can reach it; that one is
+		// a suspected production defect. Both must be fixed before this entry goes.
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*EventStreaming).CreateEventStream.func1"),
```
wilfred-s, line 72: This should be fixed on the core side before we turn this check on

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+
+		// Partition queue cleaner, started by partitionManager.Run when a partition
+		// is added. Tests that build a ClusterContext never call ClusterContext.Stop.
+		// Stoppable: delete this entry once those tests defer Stop().
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/scheduler.(*partitionManager).cleanRoot"),
+
+		// Partition expired application cleaner, the second goroutine started by
+		// partitionManager.Run. Same cause and same fix as cleanRoot above.
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/scheduler.(*partitionManager).cleanExpiredApps"),
```
wilfred-s, line 78: dito

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+
+		// User/group cache cleaner, started once by security.GetUserGroupCache when
+		// a partition resolves users. Stoppable: UserGroupCache.Stop() closes the
+		// cleaner and resets the singleton, and ClusterContext.Stop() calls it.
+		// Delete this entry once those tests defer Stop().
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/common/security.(*UserGroupCache).run"),
```
wilfred-s, line 81: dito

pkg/common/leakcheck/leakcheck.go
wilfred-s, line 63: this should get a follow up jira

pkg/common/leakcheck/leakcheck.go
```diff
@@ -0,0 +1,134 @@
...
+		// Leaks together with the handler above, and additionally when a configmap
+		// update restarts an already stopped event system: Init() registers a
+		// configmap callback that Stop() never removes, so a reload after Stop()
+		// resurrects a system nobody holds a reference to. Suspected production
+		// defect; delete this entry once Stop() deregisters the callback.
+		goleak.IgnoreTopFunction("github.com/apache/yunikorn-core/pkg/events.(*eventPublisher).start.func1"),
+
+		// Event stream forwarder, started by EventStreaming.CreateEventStream.
+		// Two distinct causes share this top frame. One is a test that creates a
+		// stream and never calls RemoveStream. The other is a consumer that stops
```
wilfred-s: that is a test breakage and needs a follow up jira

pkg/common/leakcheck/leakcheck.go
wilfred-s, line 81: one follow up jira for all 3 cases in one jira: a call to create a new cluster context should have a defer to `Stop()`

wilfred-s (changes requested):
Instead of the large comments for the exclusions, point to the jira logged for the fix and use the comment as the jira description.
if we're not fixing it we should have a comment.

Missing: the go.sum update

tigerquoll:
> Instead of the large comments for the exclusions, point to the jira logged for the fix and use the comment as the jira description. if we're not fixing it we should have a comment.
> 
> Missing: the go.sum update

@wilfred-s go.sum update not required - this module was previously brought in as a secondary dependency

tigerquoll:
JIRA Issues added to https://issues.apache.org/jira/browse/YUNIKORN-3357 , with comments updated to point to the JIRA

tigerquoll:
Rebased onto master and pushed. Taking the review points in turn.

**go.sum.** Still no change needed: `go mod tidy` on the rebased branch leaves `go.sum` byte-identical to master, because goleak is already pulled in transitively through zap's test dependency. The only conflict in the rebase was `go.mod` (zap/yaml/x-exp bumps on master).

**"Fix what we can before turning this on."** Three of the seven have moved since the review:
- the test-hygiene half of the stream leak is fixed (YUNIKORN-3372, #1135)
- the history-replay send is in review (YUNIKORN-3364, #1133)
- the remaining stream case - the in-loop `consumer <- event` that slow-consumer eviction cannot reach, which this PR's table already described as 3(b) - now has its own JIRA, YUNIKORN-3436. It is a production leak (one forwarder plus two 1000-entry buffers pinned per evicted slow client), and `TestEventStreaming_SlowConsumer` is a ready reproduction. It's the same one-line `select` as #1133 applied at the other send.

The `ClusterContext` cleaners and `UserGroupCache` (rows 4–6) are the `defer Stop()` fix you asked for. The shim PR (#1061) measured what happens when a test stops the core it started: intermittent failures at `-count>1`, because the services aren't restartable in-process. That's why the test-side `Stop()` work (3366) was folded into YUNIKORN-3370 by its assignee rather than done here - the restart problem has to go first.

**Exemptions we are not fixing.** The `notifyRMNewAllocation` entry now says so in the code, per your note that it's benign at process exit. It reproduced in ~43% of `pkg/scheduler/tests` runs when this was opened and 0 of 11 on current master, but the code path is unchanged so the guard stays.

**Comments.** All `See YUNIKORN-####` references point at open issues again (3363 and 3366 were closed as duplicates of 3370 after the last push).

I also re-probed every exemption on master by deleting it and confirming the check fails without it — all seven are still live, so nothing in the list is stale.

tigerquoll:
Rebased onto master (`aa59599`) and pushed. Two of the exemptions have been retired because their root causes landed upstream in the meantime.

**Retired**

- `EventStreaming.CreateEventStream.func1` — this covered the two bare `consumer <- event` sends. Both are now selectable on the stop channels after YUNIKORN-3364 (#1133) and YUNIKORN-3436 (#1141), both resolved for 1.10.0. @wilfred-s, this was your "should be fixed on the core side before we turn this check on" — the forwarder no longer outlives a test binary, so the entry is gone rather than argued down.
- `ClusterContext.notifyRMNewAllocation` — no longer observed at all. Removed so a recurrence fails the run instead of being silently ignored. YUNIKORN-3365 stays open for the underlying shutdown ordering.

**How that was established**

Rather than reading the diff and assuming, I removed *every* exemption and ran the full suite (`-race -tags deadlock`), then repeated the leak-prone packages. Tally of goroutines that actually outlive a test binary, over 6 runs:

| top function | occurrences |
| --- | --- |
| `EventSystemImpl.StartServiceWithPublisher.func1` | 83 |
| `partitionManager.cleanRoot` | 65 |
| `partitionManager.cleanExpiredApps` | 65 |
| `eventPublisher.start.func1` | 28 |
| `UserGroupCache.run` | 5 |
| `EventStreaming.CreateEventStream.func1` | 0 |
| `ClusterContext.notifyRMNewAllocation` | 0 |

With the five remaining entries the suite is green: 20 packages, exit 0. Notably `pkg/scheduler/tests` now passes with *zero* exemptions.

**Still exempted**

All five remaining entries are the "core services are not restartable in-process" family and are tracked by YUNIKORN-3370, which is the single follow-up JIRA you asked for. They are tests calling `StartService()` / `NewClusterContext()` without a matching `Stop()`, plus the two `partitionManager.Run()` goroutines. The list is meant to be burned down as 3370 progresses.

**CI**

`build` passes. `e2e-tests` fails on `*AsyncRMCallback does not implement ... (missing method PreFilterPredicates)` — a shim-side build break that is failing identically on every open core PR right now, unrelated to this change. `codecov/patch` reports 0% patch coverage, which is inherent to a `TestMain`-only change.

Generated by the Author with assistance from Claude Code.

pkg/common/leakcheck/leakcheck.go
tigerquoll: Fixed upstream and the exemption is now gone. Both bare sends in `CreateEventStream` became selects on the stop channels via YUNIKORN-3364 (#1133) and YUNIKORN-3436 (#1141), both resolved for 1.10.0. I confirmed empirically rather than by inspection: with every exemption stripped, this top frame appears 0 times across 6 full runs, and `pkg/events` / `pkg/webservice` pass with no entry for it.

pkg/common/leakcheck/leakcheck.go
tigerquoll: Agreed, and that JIRA is YUNIKORN-3364 — filed, fixed in #1133, resolved for 1.10.0. The history replay now selects on `e.stopCh` / `stop` and closes the consumer, so the disconnect-before-open case no longer wedges the forwarder.

pkg/common/leakcheck/leakcheck.go
tigerquoll: That was YUNIKORN-3372 — filed, fixed in #1135, resolved for 1.10.0. It is in master ahead of this branch, which is part of why the `CreateEventStream` exemption could be dropped.

pkg/common/leakcheck/leakcheck.go
tigerquoll: I have removed the `notifyRMNewAllocation` entry instead of widening it to `notifyRMAllocationReleased`. Neither top frame shows up any more once the exemptions are stripped, so exempting them would only mask a future regression of that shape. YUNIKORN-3365 stays open to track the shutdown ordering itself.

PR #1125, opened 2026-08-12
[YUNIKORN-3331] Run preemption preconditions before instantiating preemptor

stantheman0128:
### Description
`tryPreemption` built a `Preemptor` and then called `CheckPreconditions` on it. The checks read
only the ask and the preemption delay, so the preemptor was being allocated even for asks that were
about to be rejected.

This turns `CheckPreconditions` into a package level function taking the ask and the delay, and moves
`NewPreemptor` below the check. Asks that fail the checks no longer allocate a preemptor to find that
out.

Two notes:

- `NewPreemptor` is a plain struct literal with no side effects, so moving it after the check does not
  change behaviour. All six checks read `ask` and `preemptionDelay` only, which is what makes the
  extraction possible in the first place.
- `CheckPreconditions` changes from a method on `*Preemptor` to a package level function. It stays
  exported so the identifier is still reachable, but it is only used inside this package and its
  tests, so I am happy to make it unexported if you would rather narrow the surface.

On effect: the allocation avoided is one 112 byte `Preemptor` per call that fails the checks. That
path is not rare, because `preemptAttemptFrequency` gates a given ask to one real attempt every 15
seconds, so repeated scheduling cycles over a pending unschedulable ask return at the frequency check.
I did not run a macro benchmark for this. `BenchmarkScheduling` drives asks that get allocated, so it
does not walk the preemption path, and I would rather report nothing than report a number that does
not measure this change.

### Type of change

- [x] Improvement
- [x] Refactoring

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3331

### AI Tooling

### How has this been tested?

No new tests. `TestCheckPreconditions` already covered every branch of these checks, and it now calls
the function directly instead of through a preemptor, which is the point of the change.

`make test_all` on Linux, all green:

    running shellcheck
    checking license headers:
      all OK
    running golangci-lint
    0 issues.
    running unit tests
    ok  github.com/apache/yunikorn-core/pkg/scheduler          23.283s  coverage: 75.6% of statements
    ok  github.com/apache/yunikorn-core/pkg/scheduler/objects  11.851s  coverage: 89.6% of statements
    ok  github.com/apache/yunikorn-core/pkg/scheduler/tests    52.107s
    (every other package ok, output trimmed)

Also run locally, not part of CI:

    go test ./pkg/scheduler/objects/ -race -run 'TestCheckPreconditions|Preempt' -count=50
    ok  github.com/apache/yunikorn-core/pkg/scheduler/objects  389.748s

`gofmt -l pkg/` is clean and `go vet` passes.

### Questions:
- The change does not need documentation.

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -89,36 +89,36 @@ func NewPreemptor(application *Application, headRoom *resources.Resource, preemp
 
 // CheckPreconditions performs simple sanity checks designed to determine if preemption should be attempted
 // for an ask. If checks succeed, updates the ask preemption check time.
-func (p *Preemptor) CheckPreconditions() bool {
+func CheckPreconditions(ask *Allocation, preemptionDelay time.Duration) bool {
 	now := time.Now()
 
 	// skip if ask is not allowed to preempt other tasks
-	if !p.ask.IsAllowPreemptOther() {
+	if !ask.IsAllowPreemptOther() {
 		return false
 	}
 
 	// skip if ask has previously triggered preemption
-	if p.ask.HasTriggeredPreemption() {
+	if ask.HasTriggeredPreemption() {
 		return false
 	}
 
 	// skip if ask requires a specific node (this should be handled by required node preemption algorithm)
-	if p.ask.GetRequiredNode() != "" {
+	if ask.GetRequiredNode() != "" {
 		return false
 	}
 
 	// skip if preemption delay has not yet passed
-	if now.Before(p.ask.GetCreateTime().Add(p.preemptionDelay)) {
+	if now.Before(ask.GetCreateTime().Add(preemptionDelay)) {
 		return false
 	}
 
 	// skip if attempt frequency hasn't been reached again
-	if now.Before(p.ask.GetPreemptCheckTime().Add(preemptAttemptFrequency)) {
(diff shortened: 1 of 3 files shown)


manirajv06 (approved):
looks good

pkg/scheduler/objects/preemption.go
```diff
@@ -88,37 +88,39 @@ func NewPreemptor(application *Application, headRoom *resources.Resource, preemp
 }
 
 // CheckPreconditions performs simple sanity checks designed to determine if preemption should be attempted
-// for an ask. If checks succeed, updates the ask preemption check time.
-func (p *Preemptor) CheckPreconditions() bool {
+// for an ask. If checks succeed, updates the ask preemption check time. The checks only depend on the ask
+// and the preemption delay, so they run before a Preemptor is built to keep that allocation off the path
+// taken by asks that cannot preempt anyway.
```
manirajv06: Updated comment is verbose and don't think it is required. Older comment is good enough.

manirajv06 (changes requested):
Minor nit

pkg/scheduler/objects/preemption.go
stantheman0128: Reverted to the original comment. The only change left in that block is the signature line. Force-pushed as 4817d6e.

manirajv06 (approved):
looks good

PR #1127, opened 2026-08-13
[YUNIKORN-3353] Keep only pending asks in sortedRequests

tigerquoll:
Part of YUNIKORN-3350, which splits one change into three independent levers. This is the third and
last lever, the one held back from the earlier submissions because it carries the contested
decisions: pending-only membership, where a returning ask lands, and identity-based `remove`.

## What

`sortedRequests` is ordered by priority, then age. Whether an ask is allocated plays no part in that
ordering, so allocated and pending asks sit side by side in the list. `tryAllocate` walks it from the
front, skipping the allocated ones with `if request.IsAllocated() { continue }` — a read lock each. The
more asks an application has already had satisfied, the more there are to skip, so each scheduling
cycle costs more than the last. In a CPU profile of master that skip is 18% of
`Application.tryAllocate`.

Keep the slice pending-only instead: `allocateAsk` removes the ask, `deallocateAsk` puts it back. The
container, its comparator and its insert behaviour are unchanged — only membership changes, plus one
placement rule for the way back in (below). (`insert`'s slow path is shared with the new `reinsert`
since the two are byte-identical, so they cannot diverge.)

## Measured effect

`BenchmarkScheduling`, allocate phase, 10000 pods, alternating A/B, 3 samples, medians. Linux 6.8
aarch64, 6 CPUs, GOMAXPROCS=6. Cumulative is the whole stack against master; "this PR" is the ratio
against the previous PR in the stack.

| nodes | master (c/s) | this PR (c/s) | cumulative | this PR |
|---|---:|---:|---:|---:|
| 500  | 5,481 | 44,421 | 8.10x | 2.64x |
| 1000 | 5,517 | 42,452 | 7.70x | 2.60x |
| 2000 | 5,396 | 43,086 | 7.99x | 2.62x |
| 5000 | 5,353 | 40,577 | 7.58x | 2.48x |

Spread up to 7%. After the change both hot symbols are absent from the profile entirely and node
iteration becomes the dominant term, which is the correct shape.

Scope: mock resource manager, no real bind, so this is core scheduling headroom rather than a
real-cluster pods/second figure — on a real cluster the ceiling remains the Kubernetes API bind rate.
The workload is 2 applications x 5,000 asks, which is favourable to a per-application quadratic fix.

Reproduce with:

```
go test ./pkg/scheduler/tests/ -run '^$' -bench 'BenchmarkScheduling' -benchtime=1x -v
```

## What else reads `sortedRequests`

Pending-only membership changes what every reader sees, so all were checked. `sortedRequests` is
unexported, has no accessor, and appears in no DAO, REST handler, metric or proto conversion. Inside
`pkg/scheduler/objects` there are exactly three readers:

| reader | effect |
|---|---|
| `tryAllocate` | the target of this change |
| `tryPlaceholderAllocate` | already skips on `IsAllocated()`, so sees the same set |
| `getOutstandingRequests` | unchanged here, already skips on `IsAllocated()`, so sees the same set |

No caller uses `len(sortedRequests)` as a count of an application's asks, so its changed meaning is
not observed anywhere.

Insertion of new asks is untouched, so for a workload that only adds asks all three readers see the
slice exactly as master builds it. The one order change is the re-insert path (next section), and it
is visible to all three readers, including `getOutstandingRequests` — which consumes headroom as it
walks, so which asks land in the outstanding set can differ when headroom runs out inside a tie
group. That set feeds `UpdateContainerSchedulingState`, i.e. the pods reported to the resource
manager as unschedulable for cluster autoscaling.

## Where a returning ask lands

Nothing on master re-inserts into `sortedRequests`: an ask that becomes pending again (preemption
revert, placeholder-replacement revert, rollback after a failed bind) has been sitting in its
original position all along. Pending-only membership creates the question of where it goes back in,
and the existing machinery cannot answer it: `createTime` is whole-second resolution end to end — the
shim sends `pod.CreationTimestamp.Unix()`, the core does `time.Unix(secs, 0)`, and Kubernetes
`creationTimestamp` is second-resolution at source — so every pod of a burst ties on
`(priority, createTime)`, and `LessThan` reports both directions as less-than for a tie. Plain
`insert` would place a returning ask by slice contents: behind its tie-peers when the group sits at
the tail (the append fast path), in front of them otherwise.

So the placement is made a stated rule instead of a side effect: **a returning ask goes to the head
of its `(priority, createTime)` tie-group** — `sortedRequests.reinsert`, which is `insert` minus the
append fast path. The scheduler was already trying that ask ahead of its tie-peers when it allocated
it; putting it back at the head means it is retried before the others rather than sent to the back
of a group it had already cleared. The comparator is untouched, no tiebreaker is added, and ties
between newly *arriving* asks keep the existing position-dependent placement — no canonical order is
imposed where nothing needs one.

The trade-off is deliberate: an ask that keeps failing (say, a repeated bind failure) is re-attempted
first each cycle. That costs the group one attempt per cycle, not the cycle: when an attempt fails
`tryAllocate` walks on to the next pending ask, and the application-level unschedulable-ask backoff
bounds the walk as it always has. If retry-first proves wrong for that path specifically, sending the
returning ask to the back of its group instead is the same binary search with the opposite tie
handling — a scheduling-policy decision that can be made in its own change without touching this
structure.

## Three guards

The full rescan tolerated states that incremental membership does not:

- `AddAllocationAsk`'s replace branch drops the displaced ask from `sortedRequests`. The unconditional
  insert that follows would otherwise leave two entries for one key. The duplicate exists on master,
  masked because nothing depended on the slice holding each key once.
- `sortedRequests.remove` matches on **pointer identity** rather than `allocationKey`. On master
  `remove` has a single cold call site; this PR puts it on the allocation path, so "first entry with
  this key" is no longer good enough. Every direct call site passes a map- or slice-resolved object,
  so no outcome changes today. This is correctness by construction rather than a fix for a live
  hazard: the one path that could present a stale object — `AddAllocationAsk` replacing an ask that
  holds a reservation, whose captured `reserve.alloc` `tryReservedAllocate`'s second loop uses
  without re-resolving — is unreachable, because that function's only caller adds an ask solely when
  `GetAllocationAsk` returned nil, and both run on the single allocation-event goroutine. Identity
  matching means the invariant does not depend on that remaining true.
- `deallocateAsk` returns an ask to the pending structures — histogram, `sortedRequests`, `sa.pending`
  and queue pending — only when it is still the tracked object for its key, so all four sit behind one
  decision. Since YUNIKORN-3360 no production caller can present an untracked ask (`RollbackAllocation`
  rejects it first, `DeallocateAsk` resolves the ask from `sa.requests`, the placeholder reverts pass the
  ask they just allocated), so this is defence-in-depth rather than a fix. The returned delta is
  unchanged either way: callers use it to unwind node/queue *allocated* tracking. Pinned by
  `TestRollbackAllocationAskNotTracked`.

### Why not a different structure

- **Leave allocated asks in and keep skipping them.** That is master, and it is the cost being removed.
- **Tombstone and compact periodically.** Keeps positions stable, but adds a compaction policy and
  leaves the loop walking tombstones in between.
- **A second list, or an order-maintenance structure.** Position-preserving, so no re-insert rule
  would be needed — but it trades one ordering invariant for two structures that must agree, and
  gives up the binary-search insert the slice already provides. This is the alternative worth
  revisiting if the re-insert placement proves contentious.

### `remove` is O(n) — has the quadratic just moved?

No, and this was measured. `remove` → `removeAt` → `memmove` is **~30ms** of the post-change profile,
against the **~0.31s** of allocated-ask skipping this PR removes on the same workload. (That 0.31s is
the share of `IsAllocated` attributed to `tryAllocate`; the rest belongs to the rescan the previous PR
removes and is not claimed here.) Two reasons: `tryAllocate` allocates the first ask that fits, so
`remove`'s linear search almost always terminates at or near index 0; and moving pointers is far
cheaper per element than the scan it replaces, where each element cost a read lock. The cost also
falls across a run as the pending list shrinks, whereas the scan it replaces grew.

The one place a miss-scan was guaranteed is also closed: `removeAsksInternal` now skips the slice
removal for allocated asks, which left `sortedRequests` at `allocateAsk` time — without the skip,
every release of a satisfied allocation would scan the whole pending list to find nothing.

### Why the histogram is kept

Once the slice is pending-only it is ordered highest priority first, so `sortedRequests[0]` is the
pending maximum and the previous PR's histogram becomes derivable — the equivalence holds across the
whole property fuzzer. It is kept anyway, because deriving it would make queue-priority accounting
silently contingent on the ask sort policy, and on the pending-only invariant this PR introduces. The
two values answer different questions.

## Hardening

`tryAllocate` ranged over `sortedRequests` while paths inside call `allocateAsk`, which now removes from
that same slice. Every allocating path returns, so a range loop would still be correct, but it is
correct by convention only. It becomes an index loop that re-reads `len()` each iteration, so a
remove-and-continue is safe — which the one remaining guard uses: the `IsAllocated()` skip is dead
under the pending-only invariant, so it is upgraded to scream-then-repair. It `DPanic`s (hard failure
under core's development logger, i.e. unit tests; a logged line in a Kubernetes deployment, which
presets `Development: false`), drops the ghost entry and carries on with the next ask. An allocated
entry in the slice means the remove-on-allocate pairing broke and nothing else would clean it up.
It cannot fire against the current code; `TestTryAllocateAllocatedAskInSortedRequests` pins it.

A `slices.Clone` snapshot (as `tryPlaceholderAllocate` uses, where the revert path re-inserts
mid-loop) was measured instead of assumed: on `BenchmarkScheduling/1000Nodes/10000Pods`, 3 alternating
samples, medians, the index loop gives 38,604 pods/s at 203 MB/op and the clone 34,461 pods/s at
407 MB/op — an 11% cost, because the clone is O(pending) per call for a loop that normally exits at
index 0. `tryPlaceholderAllocate` keeps its clone: it needs one, and it is cold.

## Behaviour and testing

For workloads that never revert an allocation — the overwhelming majority of scheduling activity —
ordering is byte-for-byte what master produces, because the comparator and `insert` are untouched.
The behaviour change is confined to the revert paths, and the tests carry it:

- `TestReinsertHeadOfTieGroup` (sorted_asks_test.go) pins the `reinsert` contract: a returning ask
  heads its tie-group, whether the group sits at the tail or has lower-priority asks behind it, and
  an ask with nothing sorting after it returns to the end, not the front.
- `TestGetOutstandingRequests_PartialHeadroom` pins that outstanding-ask selection follows slice
  order; it holds before and after. The pre-existing cases at all three levels size headroom so that
  either all or none are selected, so they pass under any ordering.
- The previous PR's property fuzzer is extended to assert `sortedRequests` membership (pending keys
  exactly, no duplicates, nothing allocated) and comparator-validity of the order against its
  independent reference model, across every entry point including rollback and replace.
- `TestAddAllocationAskReplaceExistingPendingAsk` and `TestRollbackAllocationAskNotTracked` pin the
  two guards above at the application level.
- Full `pkg/scheduler/...` green, including under `-race`.

Concurrency: no new lock acquisition and no new ordering. All slice mutation happens under the
application write lock, where the previous PR's histogram maintenance already lives.

## Scope and rollback

No configuration, REST, scheduler-interface or metrics changes. No new public API. No persisted or
serialized state — nothing crosses a process or version boundary, so there is nothing to migrate and
no mixed-version concern during a rolling upgrade. Rollback is a single revert.

Release note: an ask whose allocation is reverted (preemption, placeholder replacement, failed-bind
rollback) is now retried before other pending asks of equal priority and creation time, instead of
holding its original arrival slot.

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -1175,12 +1175,26 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
 	// because the len check above guarantees at least one iteration would occur.
 	backoffThreshold := sa.queue.GetMaxAppUnschedAskBackoff()
 	// get all the requests from the app sorted in order
-	// LOAD-BEARING INVARIANT: sa.sortedRequests now holds only pending asks (item 2) and
-	// allocateAsk/deallocateAsk mutate it in place. This range is only safe to mutate mid-loop
-	// because every successful-allocation path below returns immediately instead of continuing the
-	// loop. Do NOT add a "continue" after a successful allocateAsk call in this loop - restructure
-	// to a snapshot copy (see tryPlaceholderAllocate) if that is ever needed.
-	for _, request := range sa.sortedRequests {
+	// LOAD-BEARING INVARIANT: sa.sortedRequests now holds only pending asks and
+	// allocateAsk/deallocateAsk mutate it in place. An index based loop is used deliberately: it
+	// re-reads len() and re-indexes the current slice on every iteration, so a mid-loop remove can
+	// never make it read the nil'd tail slot, and it stays correct if an insert reallocates the
+	// backing array. A "range" loop captures the slice header once and would panic in that case.
+	// On top of that the length is asserted below: every successful-allocation path returns
+	// immediately today, so a length change mid-loop means a future edit mutated and continued.
+	// Worst case without the assert is skipping a single ask for this cycle (self-correcting on the
+	// next cycle) rather than a crash. If a mutate-then-continue is ever really needed, restructure
+	// to a snapshot copy (see tryPlaceholderAllocate).
+	startLen := len(sa.sortedRequests)
+	for i := 0; i < len(sa.sortedRequests); i++ {
+		if len(sa.sortedRequests) != startLen {
+			log.Log(log.SchedApplication).DPanic("sortedRequests mutated during tryAllocate iteration",
+				zap.String("application ID", sa.ApplicationID),
+				zap.Int("length at loop start", startLen),
+				zap.Int("current length", len(sa.sortedRequests)))
+			return nil
+		}
+		request := sa.sortedRequests[i]
 		if backoffThreshold > 0 && unschedulable >= backoffThreshold {
 			log.Log(log.SchedApplication).Info("too many unschedulable asks in the application, waiting",
 				zap.String("application ID", sa.ApplicationID),
@@ -1190,7 +1204,17 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
 			return nil
 		}
 		if request.IsAllocated() {
-			continue
+			// sortedRequests holds pending asks only, so reaching an allocated one means the
+			// remove-on-allocate pairing has been broken and this entry is a ghost that nothing
+			// else will clean up: every other removal path skips allocated asks precisely because
+			// they cannot be in the slice. Scream, then repair - drop the ghost and end the cycle,
+			// so a broken pairing is a hard failure in tests and a single logged self-heal in
+			// production, rather than the same entry being re-detected every cycle forever.
+			log.Log(log.SchedApplication).DPanic("allocated ask found in pending-only sortedRequests",
+				zap.String("appID", sa.ApplicationID),
+				zap.String("allocationKey", request.GetAllocationKey()))
+			sa.sortedRequests.remove(request)
+			return nil
 		}
 		// check if there is a replacement possible
 		if sa.canReplace(request) {
@@ -1379,7 +1403,7 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator, 
 	// NOTE: iterate over a snapshot, not sa.sortedRequests directly. The revert path below calls
 	// sa.deallocateAsk(request), which re-inserts the ask into sa.sortedRequests mid-range - mutating
 	// the live slice while ranging over it would skip/revisit entries. allocateAsk/deallocateAsk keep
-	// only pending asks in sortedRequests (item 2).
+	// only pending asks in sortedRequests.
 	for _, request := range slices.Clone(sa.sortedRequests) {
 		// skip placeholders they follow standard allocation
 		// this should also be part of a task group just make sure it is
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -928,9 +928,9 @@ func (sa *Application) allocateAsk(ask *Allocation) (*resources.Resource, error)
 	// the ask just left the pending set
 	sa.removeFromPriorities(ask.GetPriority())
 	// keep sortedRequests limited to pending asks only: remove on allocate, re-insert on
-	// deallocate. This is safe here because every caller that ranges over sa.sortedRequests while
-	// calling allocateAsk either returns immediately on success (tryAllocate) or ranges over a
-	// snapshot (tryPlaceholderAllocate) - see the comments at those call sites.
+	// deallocate. This is safe here because every caller that iterates sa.sortedRequests while
+	// calling allocateAsk either uses an index loop that tolerates the removal (tryAllocate) or
+	// ranges over a snapshot (tryPlaceholderAllocate) - see the comments at those call sites.
 	sa.sortedRequests.remove(ask)
 
 	delta := ask.GetAllocatedResource()
@@ -948,16 +948,14 @@ func (sa *Application) deallocateAsk(ask *Allocation) (*resources.Resource, erro
 	}
 
 	// The ask returns to the pending set, but only if it still IS this application's ask: an ask that
-	// has already been dropped from sa.requests must not be resurrected. That is reachable today:
-	// removeAsksInternal("") wipes sa.requests while leaving sa.allocations intact until the shim
-	// confirms the releases, and a release arriving in that window reaches RollbackAllocation, which
-	// finds the entry in sa.allocations and deallocates it. The identity comparison (not just a
-	// presence check) also covers a stale ask object that has since been replaced by a new ask under
-	// the same key.
+	// has already been dropped from sa.requests must not be resurrected as a ghost that tryAllocate
+	// would then try to schedule. Since YUNIKORN-3360 no production caller can present one -
+	// RollbackAllocation rejects an untracked ask before it reaches here, the public DeallocateAsk
+	// resolves its ask out of sa.requests, and the tryPlaceholderAllocate reverts pass back the ask
+	// they just allocated - so this is defence-in-depth.
 	// This matches the converged behaviour of the full rescan this change replaced:
 	// updateAskMaxPriority derived the max by scanning sa.requests, so an ask absent from sa.requests
-	// never influenced it. Without the guard the pre-existing accounting drift above would be
-	// upgraded into a ghost entry in sortedRequests that tryAllocate would try to schedule.
+	// never influenced it.
 	delta := ask.GetAllocatedResource()
 	if sa.requests[ask.GetAllocationKey()] == ask {
 		sa.addToPriorities(ask.GetPriority())
@@ -965,14 +963,9 @@ func (sa *Application) deallocateAsk(ask *Allocation) (*resources.Resource, erro
 		// tie-group so the next cycle retries it before the peers it was already tried ahead of
 		sa.sortedRequests.reinsert(ask)
 		// The pending resource follows the same rule as the histogram and sortedRequests: it only
-		// counts asks the application still tracks. For an untracked ask this add would be a
-		// permanent leak - removeAsksInternal already zeroed sa.pending when it dropped the ask,
-		// and every later removeAsksInternal("") short-circuits on the empty sa.requests, so
-		// nothing would ever subtract it again. The leak is not cosmetic: it keeps
-		// resources.IsZero(sa.pending) false forever, which blocks the Completing transition in
-		// RemoveAllAllocations/removeAsksInternal, and inflates queue pending until app removal.
-		// Guarding it here (rather than in the caller) keeps all four pending-set structures -
-		// histogram, sortedRequests, sa.pending, queue pending - behind one decision.
+		// counts asks the application still tracks. Guarding it here (rather than in the caller)
+		// keeps all four pending-side structures - histogram, sortedRequests, sa.pending, queue
+		// pending - behind one decision instead of three guarded and one not.
 		sa.pending = resources.Add(sa.pending, delta)
 		// update the pending of the queue with the same delta
 		sa.queue.incPendingResource(delta)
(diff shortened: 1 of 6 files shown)


tigerquoll:
@wilfred-s this PR should be ready for review

pkg/scheduler/objects/application.go
```diff
@@ -1162,7 +1187,26 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
...
+	startLen := len(sa.sortedRequests)
+	for i := 0; i < len(sa.sortedRequests); i++ {
+		if len(sa.sortedRequests) != startLen {
+			log.Log(log.SchedApplication).DPanic("sortedRequests mutated during tryAllocate iteration",
+				zap.String("application ID", sa.ApplicationID),
+				zap.Int("length at loop start", startLen),
+				zap.Int("current length", len(sa.sortedRequests)))
+			return nil
+		}
+		request := sa.sortedRequests[i]
```
wilfred-s: This cannot happen. The application is write locked when the tryAllocate starts. Any remove or add of an ask MUST have a write lock also and will thus wait. Updating or accessing sortedRequest outside of locking is a race.

To keep the code simple we should range over a clone of the slice, we can then cleanup the real sortedRequests without an issue.

pkg/scheduler/objects/application.go
```diff
@@ -1358,7 +1412,11 @@ func (sa *Application) tryPlaceholderAllocate(nodeIterator func() NodeIterator,
 	var phFit *Allocation
 	var reqFit *Allocation
 	// get all the requests from the app sorted in order
-	for _, request := range sa.sortedRequests {
+	// NOTE: iterate over a snapshot, not sa.sortedRequests directly. The revert path below calls
+	// sa.deallocateAsk(request), which re-inserts the ask into sa.sortedRequests mid-range - mutating
+	// the live slice while ranging over it would skip/revisit entries. allocateAsk/deallocateAsk keep
+	// only pending asks in sortedRequests.
+	for _, request := range slices.Clone(sa.sortedRequests) {
```
wilfred-s, line 1404: This same structure should be used in tryAllocate

pkg/scheduler/objects/application.go
```diff
@@ -1172,7 +1216,17 @@ func (sa *Application) tryAllocate(headRoom *resources.Resource, allowPreemption
...
+			// remove-on-allocate pairing has been broken and this entry is a ghost that nothing
+			// else will clean up: every other removal path skips allocated asks precisely because
+			// they cannot be in the slice. Scream, then repair - drop the ghost and end the cycle,
+			// so a broken pairing is a hard failure in tests and a single logged self-heal in
+			// production, rather than the same entry being re-detected every cycle forever.
+			log.Log(log.SchedApplication).DPanic("allocated ask found in pending-only sortedRequests",
+				zap.String("appID", sa.ApplicationID),
+				zap.String("allocationKey", request.GetAllocationKey()))
+			sa.sortedRequests.remove(request)
+			return nil
```
wilfred-s: This is a behavioural change I do not think we should make. It short circuits the allocation cycle and moves on to the next app/queue etc.

Using a clone of the slice to iterate over would allow cleanup here without changing behaviour.

pkg/scheduler/objects/sorted_asks.go
```diff
@@ -49,9 +59,13 @@ func (s *sortedRequests) insertAt(index int, ask *Allocation) {
 	(*s)[index] = ask
 }
 
+// remove drops the entry that IS the passed ask, matching on pointer identity and not on
+// allocationKey. The slice tracks the pending asks the application holds in sa.requests, and every
+// caller passes the tracked object. Matching on key would remove the first entry that shares the
+// key, which is a different ask than the caller asked to remove as soon as two ever coexist.
```
wilfred-s: `allocationKey` is a UID that is generated by K8s for the pod. There cannot be two identical allocation keys for the same app ever.

Comment is incorrect

Pointer comparison is faster in this case so we can keep the code change just not the comment.

wilfred-s (changes requested):

pkg/scheduler/objects/application.go
tigerquoll: Agreed the length assert cannot fire - dropped it together with the comment block. One clarification for the record: the hazard the loop guarded was same-goroutine, not concurrent access - `allocateAsk` is called from inside the loop and now removes from the slice being iterated. Every allocating path returns, so that hazard is equally dead today.

I measured performance on the clone approach - `BenchmarkScheduling/1000Nodes/10000Pods`, 3 alternating samples, medians:

| loop | pods/s | B/op |
|---|---:|---:|
| index loop | 38,604 | 203 MB |
| `slices.Clone` | 34,461 | 407 MB |

The clone is O(pending) per call for a loop that normally exits at index 0, so it costs 11% of throughput and doubles allocation on this workload. Can I suggest we keep the plain index loop (it re-reads `len()`, so the ghost repair below can remove and continue) with a three-line comment and no assert. If you would still rather have the clone for uniformity it is a two-line change - but I would prefer not to hand back 11% of what this PR buys.

pkg/scheduler/objects/application.go
tigerquoll: I think `tryPlaceholderAllocate` should keep its clone: its revert path re-inserts into the slice mid-loop, which an index loop cannot tolerate. . For `tryAllocate` see the thread above - the clone measured at −11%, so I am suggesting it uses the index loop.

pkg/scheduler/objects/application.go
tigerquoll: Agreed — it now drops the ghost and continues with the next ask; the cycle is no longer cut short.

pkg/scheduler/objects/sorted_asks.go
tigerquoll: Fair — the comment is trimmed to the identity/cost rationale; code kept.

tigerquoll:
Rebased onto master now that #1122 and #1136 are in, so the branch is down to its own commits; the only conflicts were the YUNIKORN-3360 test alignment, resolved toward master. The review fixes are in the last commit (b0a2830).

While rebasing I found the description's claim that the `deallocateAsk` guard closes a live pending-resource leak was already stale - YUNIKORN-3360 rejects an untracked rollback before it gets there so this now a defence in depth change.

pkg/scheduler/objects/application.go
tigerquoll: To show the working on why the two loops differ:

`tryAllocate` never continues past a mutation: every path that allocates returns (`tryNode` always returns a result; preemption only reserves). The only remove-and-continue is the ghost repair, which an index loop handles by re-reading `len()`. The loop normally exits at index 0, so a clone there adds an O(pending) copy to an O(1) loop - that is the −11%.

`tryPlaceholderAllocate` does mutate and keep going: the revert path is `allocateAsk` (removes) → `SetReleased` fails → `deallocateAsk` (`reinsert`s at the head of the tie-group) → `continue`. An index loop would still be correct today, but only because `reinsert` always lands at or before the current index - a property of `LessThan`'s tie handling. A snapshot removes that dependency, which is what your advice is for. And this loop already scans the whole slice looking for task-group asks that match a placeholder, so the copy is a constant factor on a linear loop, not a new term - nothing to give back.

Same principle both places: pick the iteration form that is robust for what the loop does to the slice, and pay for a snapshot only where the loop walks the whole slice anyway.

tigerquoll: Correction to my note above, after re-checking against the code: `tryPlaceholderAllocate` does not scan the whole slice - it returns on the first successful replacement, so it exits early just as `tryAllocate` does, and the clone there is an O(pending) copy per call as well. What differs is the population (only applications currently holding placeholders, during replacement) and that the snapshot is what makes the mutate-and-continue revert correct by construction rather than by the tie-handling property. I have not measured that cost - `BenchmarkScheduling` has no gang workload. If it ever matters, an index loop is safe there today for the reason above, but where the loop genuinely mutates and continues I would rather keep the snapshot.

Also re-measured the `tryAllocate` comparison on the rebased head (b0a2830 against the same with `slices.Clone`), same method: index loop 38,041 pods/s / 205 MB/op, clone 33,431 pods/s / 410 MB/op — −12%. The earlier figures were taken on the pre-rebase base and stand.

manirajv06: From safety perspective, both approaches should not cause any issues as scheduling cycle itself is sequential in nature - not a parallel processing. As long as only one loop is process the `sortedRequests` sequentially, we should be good.

From performance perspective, Since there is extra clone() operation and most importantly the way each item has been accessed determines the outcome. I think performance numbers might change if we access the element directly with index in `for ... range abc.clone()` loop. Something to try out for.

PR #1132, opened 2026-08-18
[YUNIKORN-3378] Fix flaky TestSchedulerRecoveryQuotaPreemption

PoiBlackTea:
### Description
Remove redundant initial allocation phase and conflicting assertions in TestSchedulerRecoveryQuotaPreemption, aligning the test with its actual recovery quota preemption verification goals.

#### Root Cause Analysis (RCA):                                                                                                                                                         
- **Race Condition in Pre-restart Phase**: In `TestSchedulerRecoveryQuotaPreemption`, `ms.proxy.UpdateAllocation("allocation-key-01", ...)` was called asynchronously via the event channel. Right after the call, synchronous assertions (`assert.Equal(part.GetTotalAllocationCount(), 0)` and `part.GetNode(...).GetAllocatedResource() == 0`) were executed immediately without waiting for event processing.                                                                                                                                                                
    - When the background goroutine was scheduled slower than the main test thread, the assertions passed by chance because the node resource was still at its initial state (`0`).          
    - When the background goroutine finished processing faster (e.g., under CI load), `node-1:1234` was successfully allocated `1024` memory, causing the assertion to fail (`1024 != 0`).   
    - Adding `time.Sleep` or `waitForAllocations` 100% reproduces this assertion failure deterministically.                                                                                  
- **Inconsistent Pre-restart Assertion**: In `TestSchedulerRecoveryQuotaPreemption`, the application was already registered (`addApp = true`), meaning `allocation-key-01` would succeed once processed, contradicting the assertions that expected `TotalAllocationCount == 0` and node memory usage `== 0`.                                                                         
- **Substantive Test Intent**: The core purpose of `TestSchedulerRecoveryQuotaPreemption` is to verify the 6 table-driven recovery quota preemption behaviors under different configurations. The actual state injection is done via `queueA.IncAllocatedResource(tt.allocated, false)` after restart, and validated by `ms.mockRM.waitForAllocations(t, tt.addedAllocations, 1000)`. The initial allocation in Phase 1 was completely discarded after `ms.serviceContext.StopAll()` and never propagated to the restart phase.

### Type of change

- [ ] Bug Fix
- [X] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3378

### AI Tooling
If an AI tool was used:
- [X] The PR includes the phrase "Generated by Antigravity IDE", where Antigravity IDE is the name of the AI tool used.
- [X] My use of AI contributions follows the ASF legal policy.

### How has this been tested?

### Questions:

N/A

### DIFF
--- a/pkg/scheduler/tests/recovery_test.go
+++ b/pkg/scheduler/tests/recovery_test.go
@@ -511,8 +511,7 @@ func TestPlaceholderRecovery(t *testing.T) {
 	ms.mockRM.waitForApplicationState(t, appID1, "Completing", 1000)
 }
 
-// TestSchedulerRecoveryQuotaPreemption Test scheduler recovery with quota preemption when shim doesn't report existing application
-// but only include existing allocations of this app.
+// TestSchedulerRecoveryQuotaPreemption Test scheduler recovery with quota preemption
 func TestSchedulerRecoveryQuotaPreemption(t *testing.T) {
 	config := `
 partitions:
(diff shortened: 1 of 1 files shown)


manirajv06 (changes requested):
Yes, I do agree with the cleanup.

Is test really asserting the intended checks? Doesn't look like. It is more of ensuring IncAllocatedResource() correctness rather than checking the Quota Preemption related values. Isn't it?

Can you ensure the test is doing the work that it is intended to? It is supposed to ensure Quota Preemption feature correctness even after the restart.

https://github.com/apache/yunikorn-core/pull/1068 had introduced this test. May be you can take a look at it to understand the reason for introducing this test at the first place.

PoiBlackTea:
> Yes, I do agree with the cleanup.
> 
> Is test really asserting the intended checks? Doesn't look like. It is more of ensuring IncAllocatedResource() correctness rather than checking the Quota Preemption related values. Isn't it?
> 
> Can you ensure the test is doing the work that it is intended to? It is supposed to ensure Quota Preemption feature correctness even after the restart.
> 
> #1068 had introduced this test. May be you can take a look at it to understand the reason for introducing this test at the first place.

Thanks for the guidance @manirajv06 
I have updated the test to properly reflect the scheduler recovery and quota preemption verification goals:

- Real Recovery Simulation: Uses registerAllocations() instead of manual resource increments to trigger the actual PartitionContext.UpdateAllocation recovery flow.
- Accurate Completion Check: Asserts queueA.GetAllocatedResource() to confirm the recovered allocation is fully reflected in the queue.
- Direct Preemption State Validation: Directly asserts QuotaPreemptionStartTime from GetPartitionQueueDAOInfo(false) to verify whether the preemption delay timer is scheduled under each test condition, replacing the flawed waitForAllocations(0/1) checks.

Please take another look when you have time. Thanks

chenyulin0719 (approved):
LGTM, 

registerAllocations simulates the recovery, and "Quota preemption start time is scheduled" means quota preemption will be triggered later.

The assertions make sense to me.

chenyulin0719:
CI failed due to https://issues.apache.org/jira/browse/YUNIKORN-3401. It's fixed in master branch.
Merge ref is still stale unless a new commit is pushed

manirajv06 (approved):
looks good

PR #1130, opened 2026-08-17
[YUNIKORN-3360] Check allocation is present in requests map before ro…

adityadtu5:
…llback

### Description
- fixed broken UT
- Before rolling back checking for allocation in `requests`

### Type of change
Please delete options that are not relevant.

- [x] Bug Fix
- [ ] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3360

- [ ] I have created a Jira issue for this pull request.
- [x] The Jira ID is part of the title of this pull request.

### AI Tooling
If an AI tool was used:
- [ ] The PR includes the phrase "Generated by \<tool>", where \<tool> is the name of the AI tool used.
- [x] My use of AI contributions follows the ASF legal policy.

Check https://www.apache.org/legal/generative-tooling.html for details.

### How has this been tested?
- [x] `make test_all` was run, and no failures reported.

PR #1136, opened 2026-08-24
[YUNIKORN-3399] Align RollbackAllocation tests with the ask-presence guard from YUNIKORN-3360

tigerquoll:
### What is this PR for?

Fixes the two test failures on current master ([YUNIKORN-3398](https://issues.apache.org/jira/browse/YUNIKORN-3398)) caused by a semantic conflict between two independently merged PRs:

- [YUNIKORN-3360](https://issues.apache.org/jira/browse/YUNIKORN-3360) (#1130) added a guard to `RollbackAllocation` that rejects the rollback when the ask is missing from `sa.requests`.
- [YUNIKORN-3352](https://issues.apache.org/jira/browse/YUNIKORN-3352) (#1122) added tests that exercise exactly that "ghost" state (ask wiped from `sa.requests`, allocation still in `sa.allocations`) and expected the pre-guard behaviour where the rollback succeeds.

There was no textual conflict, so both merged cleanly and the combination broke `make test`:

- `TestRollbackAllocationAskNotTracked` — asserted the ghost rollback succeeds
- `TestApplicationPropertyFuzzHistogram` — asserted at least one successful ghost rollback across the seed set

This is a **test-only** change aligning both tests with the YUNIKORN-3360 semantics:

- `TestRollbackAllocationAskNotTracked` now pins the guard: the rollback of a confirmed allocation whose ask is no longer tracked must be **rejected**, and must leave the allocation, `allocatedResource` and the pending histogram untouched. Since #1130 added no test of its own, this becomes the guard's first direct coverage.
- The property fuzzer's rollback case now counts ghost rollback **attempts** and asserts each one is rejected; its reference model already treated a rejected rollback as a no-op, so the per-step invariant checks are unchanged.

One note kept out of scope here: the ghost state itself is reachable in production (`removeAsksInternal("")` leaves `sa.allocations` populated until the shim confirms the releases), and with the guard in place a `SCHEDULING_FAILED_ON_RM` release arriving in that window is now dropped without cleaning up the allocated side. Whether that needs proper handling is a question for YUNIKORN-3360, not for these tests.

### What type of PR is it?
* [x] Bug Fix (test only)

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-3399

- `go test -tags deadlock -race -run 'TestRollbackAllocationAskNotTracked|TestApplicationPropertyFuzzHistogram' ./pkg/scheduler/objects/` — both pass (fail on master before this change)
- full `./pkg/scheduler/objects/` package green with `-race -tags deadlock`

Generated by the Author with assistance from Claude Code.

### DIFF
--- a/pkg/scheduler/objects/application_property_test.go
+++ b/pkg/scheduler/objects/application_property_test.go
@@ -43,26 +43,29 @@ import (
 //
 // The point is coverage of state combinations rather than of entry points: TestMaxAskPriority pins
 // the handful of transitions that are easy to reason about by hand, while this test reaches the
-// interleavings that are not - a replaced ask whose priority differs from the one it displaces, a
-// rollback of an allocation whose ask has already been dropped, an allocate that empties the top
-// priority bucket while lower buckets still hold pending asks. Those are precisely the cases where
-// incremental bookkeeping and a full rescan can disagree.
+// interleavings that are not - a replaced ask whose priority differs from the one it displaces, an
+// attempted rollback of an allocation whose ask has already been dropped, an allocate that empties
+// the top priority bucket while lower buckets still hold pending asks. Those are precisely the cases
+// where incremental bookkeeping and a full rescan can disagree.
 func TestApplicationPropertyFuzzHistogram(t *testing.T) {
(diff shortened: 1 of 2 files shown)


manirajv06 (approved):
looks good

PR #1137, opened 2026-08-25
[YUNIKORN-3408] Adopt the vet-lock static lock analyser (annotations + build wiring)

tigerquoll:
### Description

Machine-checks the lock conventions that are today enforced only by review: which fields a lock guards, which functions require a lock held on entry, which must not be called with a lock already held, and the `String()`-takes-its-own-lock trap. Adds the [vet-lock](https://github.com/tigerquoll/vet-lock) static analyser (pinned in an isolated tools module, never in the main `go.mod`), wires it into `make vetlock` / `test_all` / the pre-commit workflow, and annotates the scheduler.

Behaviour-neutral. The annotations are comments; the only runtime code change is the `pkg/locking` named-field lock + forwarding methods (inlined, verified with `-gcflags=-m`) that let annotations resolve to the wrapper field.

- **Analysers gated**: `checklocks` (guarded-field, lock-requirement, exclusion and precondition annotations), `lockstringer` (`String()` methods that take their own receiver's lock), `lockblocking` (blocking under a classed lock).
- **Tool pinned**: `github.com/tigerquoll/vet-lock v0.8.0` in `scripts/vetlock`; binary keyed on tool + Go version; `-checklocks.inferred=false`.
- **Annotated**: `pkg/common/configs`, `pkg/common/resources`, `pkg/common/security`, `pkg/events`, `pkg/locking`, `pkg/metrics`, `pkg/plugins`, `pkg/rmproxy`, `pkg/scheduler` (+ `objects`, `placement`, `ugm`), `pkg/webservice`.
- **Canary** (`pkg/locking/checklocks_canary.go`): one known violation per analysis; `make vetlock` fails if an analyser silently stops covering what it should.
- The bare `+lockclass:<name>` declarations are kept because `lockblocking` consumes the class taxonomy; the lock-**ordering** relations belong to a separate PR.

**The suppression list is a burn-down list.** Where the code does not meet what the annotation states, the annotation is not weakened to fit: the site carries a one-line `// YUNIKORN-NNNN:` marker plus the analyser's ignore directive, so this lands green and blocks new violations immediately, and each marker is deleted by the fix PR of its own JIRA. The JIRAs carry the analysis; they are linked from YUNIKORN-3408. None of the issues is fixed here (the same split as the goleak adoption, YUNIKORN-3357). Suppressions that are by design (closure / interface-dispatch / construction-time boundaries) carry a plain explanatory comment and no JIRA.

Not in this PR: the lock-order analyser and runtime lock-class checker; the `lockgap` analyser. Each is a follow-up.

### Type of change

- [x] Improvement

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3408

### AI Tooling
Generated by the Author with assistance from Claude Code

### How has this been tested?
- [ ] New unit tests were added to cover new or changed code paths. (The canary fixture is the test of the gate: injecting an unguarded write to any annotated field fails `make vetlock`.)

### Questions:
- [ ] The change needs documentation, a pull request for apache/yunikorn-site repository will be created. (The developer-concurrency page from YUNIKORN-3349 documents the conventions; an "annotations" section follows once this merges.)

`make vetlock` produces no output on this branch. The canary stage of the same target proves each analyser still reports: it runs the fixture with one known violation per check and fails the build if any expected report is missing.


tigerquoll:
Updated (force-push `c1653b9` → `28b3a4b`):

- Rebased onto current master (YUNIKORN-3300, 3432, 3372). YUNIKORN-3432 already moved `cleanUpCache`/`resetCache` to the receiver's lock, so the YUNIKORN-3416 markers and their three ignores are gone; the analyser is clean on that file without them.
- A full pass over every annotation against the code (writers enumerated per guarded field, callers per precondition, each `+checklocksunguarded` checked for immutability). Corrections from that pass:
  - `Queue.queueEvents` was stored *after* `addChildQueue` published the queue while being marked unguarded; both constructors now create it before the queue is added to its parent (the one non-comment change in `pkg/scheduler/objects`, noted in the commit message).
  - `tryNodes` declared a read precondition but its closure needs the write lock (its only caller holds it); now `+checklocks:sa.RWMutex`, and the same write precondition is declared on `tryNodesNoReserve`, `Preemptor.initWorkingState`, `HandleApplicationEvent` and `HandleApplicationEventWithInfo`, closing the chain the FSM callback guards rely on. No ignore was added; the analyser confirms every caller holds the lock.
  - Marker text fixes: the `stateTime` marker (the writer has never had a production caller), the YUNIKORN-3410 marker removed from `processRMRegistrationEvent` (the re-entry is unreachable there), a YUNIKORN-3417 marker added on the one reply send that lacked it, and the `getOutStandingRequestsInternal` marker corrected to say application fields.
  - Stale prose removed where it contradicted the new annotations (`updateNodeSortingPolicy`, `getPlacementManager`, `mergeProperties`, `handleQueueEvent`, the `configs.go` var block).
  - CI: the tools cache key now includes `.go_version` so a Go bump gets a fresh cache entry.

All gates green locally (`make vetlock` incl. canary, lint, license-check, check_scripts, full `make test`). The pre-commit workflow run for this PR is waiting on approval; the same commits are green on the fork's CI.

PR #1145, opened 2026-09-03
[YUNIKORN-3410] Fix self-deadlock on ClusterContext lock during partition removal

hedger9487:
### Description
Fixes a self-deadlock on the `ClusterContext` write lock during partition removal:
1. During a configuration reload that drops or renames a partition (`updateSchedulerConfig`), and during RM partition removal events (`removePartitionsByRMID`), the caller stops the partition manager while holding the `ClusterContext` write lock (`cc.Lock()`).
2. `partitionManager.Stop()` calls `manager.remove()` synchronously (since YUNIKORN-2233). In the previous code, `manager.remove()` ended with `manager.cc.removePartition(name)`, which attempted to acquire the non-reentrant `cc.Lock()` again, resulting in an immediate self-deadlock that hung the scheduling loop, REST endpoints, and RM callbacks.

This PR:
- Drops the `manager.cc.removePartition()` call from `partitionManager.remove()`, keeping `remove()` focused solely on cleaning up internal partition resources (queues, applications, nodes).
- Removes the unused private `removePartition()` callback from `ClusterContext`.
- In `updateSchedulerConfig()` and `removePartitionsByRMID()`, deletes removed partitions directly from `cc.partitions` during iteration under the already-held `cc.Lock()`, eliminating intermediate slice/map allocations (per maintainer feedback).
- Wraps the `pc.stateTime` write in `handlePartitionEvent()` with `pc.Lock()` / `pc.Unlock()` to eliminate potential unprotected concurrent writes.
- Adds `TestContext_RemovePartition` in `context_test.go` covering config reload partition removal, RM partition removal events, and active partition removal with running applications and nodes without duplicate fixtures.

### Type of change
- [x] Bug Fix
- [ ] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3410

### AI Tooling
If an AI tool was used:
- [x] The PR includes the phrase "Generated by Antigravity", where Antigravity is the name of the AI tool used.

### How has this been tested?
- [x] Deadlock reproduction and active application/node cleanup verified in `TestContext_RemovePartition`.
- [x] Full test suite run under `-race` (`go test -race ./pkg/scheduler/...`), 100% passed.
- [x] Local linter passed with 0 issues in modified files (`golangci-lint run pkg/scheduler/...`).
- [x] Full repository regression suite (`go test ./...`), 100% passed with zero regressions.

### Questions:

*Generated by hedger9487 with assistance from Antigravity.*

### DIFF
--- a/pkg/scheduler/context.go
+++ b/pkg/scheduler/context.go
@@ -313,18 +313,13 @@ func (cc *ClusterContext) processNodes(request *si.NodeRequest) {
 func (cc *ClusterContext) removePartitionsByRMID(event *rmevent.RMPartitionsRemoveEvent) {
 	cc.Lock()
 	defer cc.Unlock()
-	partitionToRemove := make(map[string]bool)
 
 	// Just remove corresponding partitions
-	for k, partition := range cc.partitions {
-		if partition.RmID == event.RmID {
-			partition.partitionManager.Stop()
-			partitionToRemove[k] = true
+	for _, part := range cc.partitions {
+		if part.RmID == event.RmID {
+			part.partitionManager.Stop()
+			delete(cc.partitions, part.Name)
 		}
-	}
-
-	for partitionName := range partitionToRemove {
-		delete(cc.partitions, partitionName)
 	}
 	// Done, notify channel
 	event.Channel <- &rmevent.Result{
@@ -399,6 +394,7 @@ func (cc *ClusterContext) updateSchedulerConfig(conf *configs.SchedulerConfig, r
 	for _, part := range cc.partitions {
 		if !visited[part.Name] {
 			part.partitionManager.Stop()
+			delete(cc.partitions, part.Name)
 			log.Log(log.SchedContext).Info("marked partition for removal",
 				zap.String("partitionName", part.Name))
 		}
@@ -583,14 +579,6 @@ func (cc *ClusterContext) NeedPreemption() bool {
 	defer cc.RUnlock()
 
 	return cc.needPreemption
-}
-
-// Callback from the partition manager to finalise the removal of the partition
-func (cc *ClusterContext) removePartition(partitionName string) {
-	cc.Lock()
-	defer cc.Unlock()
-
-	delete(cc.partitions, partitionName)
 }
 
 // addNode adds a new node to the cluster enforcing just one unlimited node in the cluster.
--- a/pkg/scheduler/context_test.go
+++ b/pkg/scheduler/context_test.go
@@ -445,3 +445,103 @@ outer:
 
 	assert.Assert(t, checked, "Failed to find metric")
 }
+
+func TestContext_RemovePartition(t *testing.T) {
+	context := &ClusterContext{
+		partitions:     map[string]*PartitionContext{},
+		rmEventHandler: newMockEventHandler(),
+	}
+	defer context.Stop()
+
+	// Initial config with two partitions: default and gpu
+	configTwoPartitions := `
+partitions:
+  - name: default
+    queues:
+      - name: root
+        submitacl: '*'
+        queues:
+          - name: default
+            submitacl: '*'
+  - name: gpu
+    queues:
+      - name: root
+        submitacl: '*'
+        queues:
+          - name: default
+            submitacl: '*'
+`
+	conf1, err := configs.LoadSchedulerConfigFromByteArray([]byte(configTwoPartitions))
+	assert.NilError(t, err)
+
+	context.Lock()
+	err = context.updateSchedulerConfig(conf1, "rm:123")
+	assert.NilError(t, err)
+	context.Unlock()
+
+	assert.Equal(t, 2, len(context.GetPartitionMapClone()))
+	gpuPart := context.GetPartition("[rm:123]gpu")
+	assert.Assert(t, gpuPart != nil)
+
+	// Add node to gpu partition
+	nodeInfo := &si.NodeInfo{
+		NodeID: "node-gpu-1",
+		Action: si.NodeInfo_CREATE,
+		Attributes: map[string]string{
+			siCommon.NodePartition: "[rm:123]gpu",
+		},
+		SchedulableResource: &si.Resource{Resources: map[string]*si.Quantity{"first": {Value: 10}}},
+	}
+	err = context.addNode(nodeInfo, true)
+	assert.NilError(t, err)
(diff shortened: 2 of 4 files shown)


pkg/scheduler/context.go
```diff
@@ -396,13 +396,18 @@ func (cc *ClusterContext) updateSchedulerConfig(conf *configs.SchedulerConfig, r
...
 		if !visited[part.Name] {
 			part.partitionManager.Stop()
+			partitionsToRemove = append(partitionsToRemove, part.Name)
 			log.Log(log.SchedContext).Info("marked partition for removal",
 				zap.String("partitionName", part.Name))
 		}
 	}
+	for _, partitionName := range partitionsToRemove {
+		delete(cc.partitions, partitionName)
+	}
```
manirajv06: I think we can delete in earlier reading loop itself instead of building new one and reading it for delete. Should not cause any problems. Please check it out.

Also, Can we follow the same in other place too?

manirajv06 (changes requested):

pkg/scheduler/context.go
hedger9487: Updated! Deleted directly from cc.partitions during the first iteration in both updateSchedulerConfig() and removePartitionsByRMID(), eliminating intermediate slice/map allocations.

Also aligned both places to consistently use part.Name for clean, unified object semantics.

Additionally, ensured the third lifecycle caller ClusterContext.Stop() also clears cc.partitions under cc.Lock() after stopping background services, accompanied by a new unit test TestClusterContextStop. Thanks for the great suggestions!

pkg/scheduler/context.go
```diff
@@ -894,4 +882,7 @@ func (cc *ClusterContext) Stop() {
 		part.partitionManager.Stop()
 		part.userGroupCache.Stop()
 	}
+	cc.Lock()
+	defer cc.Unlock()
+	cc.partitions = make(map[string]*PartitionContext)
```
manirajv06: Is this required for unit tests?

pkg/scheduler/context_test.go
```diff
@@ -445,3 +445,202 @@ outer:
 
 	assert.Assert(t, checked, "Failed to find metric")
 }
+
+func TestUpdateSchedulerConfigRemovePartition(t *testing.T) {
```
manirajv06: Is it possible to wrap this, next and next to next and even last one into single test method using t.run() ?

Agree that these are the different situations needs to be tested but not each one requires a separate test method.

manirajv06 (changes requested):

pkg/scheduler/context.go
hedger9487: Hi @manirajv06, thanks for pointing this out! To provide some context on why this was added:

In the original master code, `ClusterContext.Stop()` called `part.partitionManager.Stop()`, which called `manager.remove()`, which indirectly invoked `manager.cc.removePartition(name)` to delete each partition from `cc.partitions`.

When we removed the `cc.removePartition()` callback to fix the self-deadlock in `updateSchedulerConfig` and `removePartitionsByRMID`, `ClusterContext.Stop()` also lost that indirect cleanup side-effect. To prevent a behavioral regression where `cc.partitions` would remain populated after `ClusterContext.Stop()`, we explicitly cleared `cc.partitions` under `cc.Lock()`, and added the test to verify shutdown behavior was preserved.

However, if the community prefers `ClusterContext.Stop()` to only stop background services without resetting `cc.partitions`, I'm completely happy to drop it. What do you think?

pkg/scheduler/context_test.go
hedger9487: Done! Consolidated all 4 test scenarios into `TestContext_RemovePartition` using `t.Run()` subtests:
- `remove via updateSchedulerConfig`
- `remove via removePartitionsByRMID`
- `remove with active apps and nodes`
- `stop clears partitions`

All subtests pass cleanly under the race detector.

manirajv06 (changes requested):
looks good. Minor nits. Fix lint warnings as well.

pkg/scheduler/context_test.go
manirajv06: Nice. Can we avoid repeating the same code for each run?

Have only single run and execute all code and finally assert each case expected output. There are lot of references out there in the codebase. Please take a look.

t.run() {

}

pkg/scheduler/context.go
manirajv06: > I'm completely happy to drop it. 

Yes, please

hedger9487: Dropped `ClusterContext.Stop()` changes and restored it back to original master.

pkg/scheduler/context_test.go
hedger9487: Updated! Refactored `TestContext_RemovePartition` into a single, clean test execution flow with a single setup. It tests `updateSchedulerConfig` partition removal (including active application and node teardown) and subsequently tests `removePartitionsByRMID`. This avoids code duplication and eliminates the `goconst` lint warning.

hedger9487:
Hi @manirajv06,

I have updated the PR to address all review comments:
1. **Reverted `ClusterContext.Stop()`**: Removed the mutex lock and map re-initialization, restoring `Stop()` back to its original implementation.
2. **Refactored `TestContext_RemovePartition`**: Consolidated into a single shared cluster context setup with clean `t.Run()` subtests (`remove via updateSchedulerConfig` and `remove via removePartitionsByRMID`), testing active node and application cleanup without duplicate setup code or duplicate YAML fixtures.
3. **Fixed Lint Warnings**: Resolved all duplicate string definitions (`goconst`), and verified that `golangci-lint run pkg/scheduler/...` reports 0 issues in the modified files.

All unit tests pass cleanly under `-race`. Thanks again for the review!

pkg/scheduler/context_test.go
hedger9487: Updated! Refactored `TestContext_RemovePartition` to use a single shared setup at the top, followed by `t.Run()` subtests for each case:
- `remove via updateSchedulerConfig`: tests partition removal and active app/node teardown without deadlock.
- `remove via removePartitionsByRMID`: tests remaining partition cleanup via RM event without deadlock.

PR #1146, opened 2026-09-03
[YUNIKORN-3358] Remove the unreachable replace-existing-ask branch

rjgoyln:
## Summary

`AddAllocationAsk` compensated for re-adding an allocation key the application already tracks and still has pending. That state cannot occur:

* `PartitionContext.UpdateAllocation` is the only caller, and it only calls in after `GetAllocationAsk` has returned nil for the key.
* All of those calls are processed by the single goroutine draining `pendingAllocEvents`.

The compensation was incomplete anyway. It rolled back `sa.pending` and the priority histogram but left the displaced ask in `sortedRequests`, which only removes one entry for a given key.

## Change

* The unreachable replacement branch is removed.
* `AddAllocationAsk` now rejects a tracked key with an error, ahead of any state mutation.

Removing the branch on its own would turn an impossible-but-protected state into a silently corrupting one, since a duplicate add would then double count `sa.pending` and the priority histogram and append another `sortedRequests` entry. `UpdateAllocation` never presents a tracked key, so the normal scheduler path is unchanged. `RecoverAllocationAsk` sits behind the same check but returns nothing and so cannot refuse, leaving it covered by the documented invariant alone.

## Tests

* `TestAddAllocationAskReplaceExistingPendingAsk` is replaced by `TestAddAllocationAskDuplicateKeyRejected`.
* The property fuzzer's case 8 now re-adds tracked keys and requires the add to be refused with nothing moved: tracked ask, `sa.pending`, `sortedRequests` and the task group's placeholder count all unchanged.
* Three fixtures that silently reused allocation keys are corrected with their assertions intact: the two placeholder tests, `TestOutStandingRequestMultipleChildrenWithMax` and `TestPredicateFailedEvents`.

Each new fuzz assertion was checked against a deliberately misplaced guard, so the case cannot pass vacuously. The existing invariant checks continue to cover the priority histogram and `askMaxPriority`.

## Relationship to #1127

#1127 repairs the same branch by adding the missing `sortedRequests.remove(oldAsk)`; this PR removes it instead. The two differ in implementation rather than intent, and whichever lands second should only need a small rebase.

### Type of change

- [x] Refactoring

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3358

### AI Tooling

Generated by Claude Code (Opus 5)

### How has this been tested?

### Questions:

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/scheduler/objects/application.go
+++ b/pkg/scheduler/objects/application.go
@@ -647,8 +647,9 @@ func (sa *Application) removeAsksInternal(allocKey string, detail si.EventRecord
 	return toRelease
 }
 
-// Add an allocation ask to this application
-// If the ask already exist update the existing info
+// Add an allocation ask to this application.
+// An ask under a key the application already holds is rejected: see addAllocationAskInternal for
+// the single adder invariant that requires it.
 func (sa *Application) AddAllocationAsk(ask *Allocation) error {
 	sa.Lock()
 	defer sa.Unlock()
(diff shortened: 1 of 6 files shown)


manirajv06 (approved):
looks good.

PR #1149, opened 2026-09-04
[YUNIKORN-2251] Add node instanceType to the node added event message

stantheman0128:
### Description
The node added event carried the fixed message "Node added to the scheduler", so anyone reading
the event stream had to look the node up separately to find out what kind of node had joined.

This adds the instance type to that message, using the same key and value style the file already
uses for the schedulable changed event:

    Node added to the scheduler, instanceType: <type>

Nodes without the `si/instance-type` attribute report `UNKNOWN`, which is what
`Node.GetInstanceType()` already returns in that case. The value is always present rather than
omitted when unknown, so the message keeps one shape and a missing attribute stays visible to
whoever reads the event.

Two notes on scope:

- This keeps the value in the existing `Message` string rather than adding a field to
  `EventRecord`, which is what the Jira title asks for and avoids a scheduler-interface change.
  Happy to move it to a structured field instead if you would rather have it typed.
- `Node.GetInstanceType()` is documented as a lock free call because attributes are read only, so
  calling it from `SendNodeAddedEvent` under the existing read lock does not change the locking.

### Type of change

- [x] Improvement

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-2251

### AI Tooling

### How has this been tested?

No new tests. The two existing assertions on this message were updated instead:
`TestSendNodeAddedEvent` covers the explicit value, and `TestApplicationHistoryTracking` covers the
`UNKNOWN` case through the real scheduler path, since the node in that test has no instance type
attribute.

They are not vacuous. Passing an empty string instead of `sn.GetInstanceType()` in `node.go` fails
the integration test:

    --- FAIL: TestApplicationHistoryTracking (0.54s)

`make test_all` on Linux, all green:

    running shellcheck
    checking license headers:
      all OK
    running golangci-lint
    0 issues.
    running unit tests
    ok  github.com/apache/yunikorn-core/pkg/scheduler                 26.951s  coverage: 75.6% of statements
    ok  github.com/apache/yunikorn-core/pkg/scheduler/objects         11.257s  coverage: 89.6% of statements
    ok  github.com/apache/yunikorn-core/pkg/scheduler/objects/events   1.246s  coverage: 100.0% of statements
    ok  github.com/apache/yunikorn-core/pkg/scheduler/tests           54.513s
    (every other package ok, output trimmed)

`gofmt -l pkg/` is clean and `go vet` passes.

### Questions:
- The site docs quote this message in a few places, so they will need a follow up. I can open a
  Jira against apache/yunikorn-site for that if you want it tracked.

### DIFF
--- a/pkg/scheduler/objects/events/node_events.go
+++ b/pkg/scheduler/objects/events/node_events.go
@@ -29,11 +29,11 @@ type NodeEvents struct {
 	eventSystem events.EventSystem
 }
 
-func (n *NodeEvents) SendNodeAddedEvent(nodeID string, capacity *resources.Resource) {
+func (n *NodeEvents) SendNodeAddedEvent(nodeID string, capacity *resources.Resource, instanceType string) {
 	if !n.eventSystem.IsEventTrackingEnabled() {
 		return
 	}
-	event := events.CreateNodeEventRecord(nodeID, "Node added to the scheduler", common.Empty, si.EventRecord_ADD,
+	event := events.CreateNodeEventRecord(nodeID, "Node added to the scheduler, instanceType: "+instanceType, common.Empty, si.EventRecord_ADD,
 		si.EventRecord_DETAILS_NONE, capacity)
 	n.eventSystem.AddEvent(event)
 }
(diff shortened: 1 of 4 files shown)


manirajv06 (changes requested):
looks good.

Can you add one test case for the valid instance type?

TestNodeEvents is one of the suitable place I can think of. Feel free to cover in any suitable place.

PR #1152, opened 2026-09-06
[YUNIKORN-3365] Shutdown ordering: scheduler allocation notify can block on an RM reply the proxy never drains

PoiBlackTea:
### Description
During scheduler shutdown (`ServiceContext.StopAll`), `Scheduler.handleAllocEvent` can select an in-flight allocation or release event concurrently with `s.stop` closing. This invokes `notifyRMNewAllocation` or `notifyRMAllocationReleased`, which posts an event to `RMProxy.pendingRMEvents` and synchronously waits on a reply channel. However, `RMProxy.handleRMEvents` previously exited immediately upon receiving `rmp.stop` without draining `pendingRMEvents`. Because no response was ever dispatched to the reply channel, the caller goroutine parked indefinitely, causing a goroutine leak on shutdown.

To resolve this:
1. **Drain on Stop**: `RMProxy.handleRMEvents` now calls `drainPendingEvents()` upon shutdown, consuming all remaining queued events and non-blockingly returning `Result{Succeeded: false, Reason: "RMProxy is stopping"}` to unblock pending callers.
2. **Buffered Reply Channels**: Converted notification reply channels in `pkg/scheduler/context.go` and `pkg/scheduler/objects/application.go` to buffered channels (`cap 1`).
3. **Unit Testing**: Added deterministic unit test `TestRMProxy_DrainPendingEvents` in `pkg/rmproxy/rmproxy_test.go` to verify graceful event draining on shutdown.

### Type of change

- [x] Bug Fix
- [ ] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3365

### AI Tooling
If an AI tool was used:
- [x] The PR includes the phrase "Generated by Antigravity CLI", where Antigravity CLI is the name of the AI tool used.

### How has this been tested?

### Questions:

### DIFF
--- a/pkg/rmproxy/rmproxy_test.go
+++ b/pkg/rmproxy/rmproxy_test.go
@@ -28,6 +28,27 @@ import (
 	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
 )
 
+func TestRMProxy_StopUnblocksWaitingCaller(t *testing.T) {
+	rmp := NewRMProxy(nil)
+	rmp.StartService()
+
+	c := make(chan *rmevent.Result, 1)
+	rmp.HandleEvent(&rmevent.RMReleaseAllocationEvent{
+		ReleasedAllocations: []*si.AllocationRelease{},
+		RmID:                "rm-test",
+		Channel:             c,
+	})
+
+	rmp.Stop() // exercises the case <-rmp.stop: drainPendingEvents() wiring
+
+	select {
+	case res := <-c:
+		assert.Assert(t, res != nil) // normal reply or drained — either is fine, point is no leak
+	case <-time.After(time.Second):
+		t.Fatal("caller leaked: no reply after Stop")
+	}
+}
+
 func TestRMProxy_DrainPendingEvents(t *testing.T) {
 	rmp := NewRMProxy(nil)
 
--- a/pkg/rmproxy/rmproxy_test.go
+++ b/pkg/rmproxy/rmproxy_test.go
@@ -1,0 +1,64 @@
+/*
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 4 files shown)


pkg/rmproxy/rmproxy_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"testing"
+	"time"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/apache/yunikorn-core/pkg/rmproxy/rmevent"
+	"github.com/apache/yunikorn-scheduler-interface/lib/go/si"
+)
+
```
HuangTing-Yao, line 30: The fix looks solid. One comment on the test: `TestRMProxy_DrainPendingEvents` calls `drainPendingEvents()` directly and never goes through `Stop()` / `handleRMEvents()`, so the `case <-rmp.stop: drainPendingEvents()` wiring is never exercised — if a refactor dropped that call and re-introduced this leak, the test would still pass. Worth adding one end-to-end test that reproduces the original scenario:
```suggestion
func TestRMProxy_StopUnblocksWaitingCaller(t *testing.T) {
	rmp := NewRMProxy(nil)
	rmp.StartService()

	c := make(chan *rmevent.Result, 1)
	rmp.HandleEvent(&rmevent.RMReleaseAllocationEvent{
		ReleasedAllocations: []*si.AllocationRelease{},
		RmID:                "rm-test",
		Channel:             c,
	})

	rmp.Stop() // exercises the case <-rmp.stop: drainPendingEvents() wiring

	select {
	case res := <-c:
		assert.Assert(t, res != nil) // normal reply or drained — either is fine, point is no leak
	case <-time.After(time.Second):
		t.Fatal("caller leaked: no reply after Stop")
	}
}
```
It complements rather than replaces the existing test (it can't deterministically assert the drained result), so I'd keep both.

HuangTing-Yao (changes requested):

pkg/rmproxy/rmproxy_test.go
PoiBlackTea: Thanks @HuangTing-Yao! Added `TestRMProxy_StopUnblocksWaitingCaller` as suggested to cover the stop wiring. Verified with `make test_all`.

HuangTing-Yao (approved):
LGTM

PR #1159, opened 2026-09-12
[YUNIKORN-3236] Inter Queue Preemption: Improve victim selection algorithm

PoiBlackTea:
### Description
This pull request integrates the improved victim selection process into Inter-Queue Preemption:

1. **Inter-Queue Preemption Algorithm Upgrade**: Replaces legacy `compareAllocationLess` sorting with `SortAllocationsBasedOnAsk` in `calculateAdditionalVictims`, passing `askQueue.GetMaxResource()` to properly evaluate hierarchical queue quotas.
2. **Deterministic Tie-Breaking**: Resolves non-deterministic sorting in `SortAllocationsBasedOnAsk` when candidate allocations have identical resource ratios (`comp == 0`) by introducing a cascading tie-breaker (creation time `After` for younger-first, followed by `allocationKey` lexicographical ordering).
3. **Cleanup & Test Stabilization**: Removes unused legacy helpers (`compareAllocationLess` and `scoreAllocation`) and stabilizes preemption tests with fixed baseline timestamps to prevent hour-boundary flakiness.

### Type of change
- [ ] Bug Fix
- [x] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3236

### AI Tooling
If an AI tool was used:
- [x] The PR includes the phrase "Generated by Antigravity CLI", where Antigravity CLI is the name of the AI tool used.

### How has this been tested?

### Questions:

### DIFF
--- a/pkg/scheduler/objects/preemption.go
+++ b/pkg/scheduler/objects/preemption.go
@@ -485,9 +485,7 @@ func (p *Preemptor) calculateAdditionalVictims(nodeVictims []*Allocation) ([]*Al
 			potentialVictims = append(potentialVictims, victim)
 		}
 	}
-	sort.SliceStable(potentialVictims, func(i, j int) bool {
-		return compareAllocationLess(potentialVictims[i], potentialVictims[j])
-	})
+	SortAllocationsBasedOnAsk(potentialVictims, askQueue.GetMaxResource(), p.ask.GetAllocatedResource())
 
 	// evaluate each potential victim in turn, stopping once sufficient resources have been freed
 	victims := make([]*Allocation, 0)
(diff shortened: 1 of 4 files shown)


pkg/scheduler/objects/preemption.go
```diff
@@ -485,9 +485,7 @@ func (p *Preemptor) calculateAdditionalVictims(nodeVictims []*Allocation) ([]*Al
 			potentialVictims = append(potentialVictims, victim)
 		}
 	}
-	sort.SliceStable(potentialVictims, func(i, j int) bool {
-		return compareAllocationLess(potentialVictims[i], potentialVictims[j])
-	})
+	SortAllocationsBasedOnAsk(potentialVictims, askQueue.GetMaxResource(), p.ask.GetAllocatedResource())
```
HuangTing-Yao, line 488: Would we get nil MaxResource in here?

HuangTing-Yao (changes requested):

pkg/scheduler/objects/preemption.go
PoiBlackTea: No, it won't be `nil`:
1. `QueuePreemptionSnapshot.GetMaxResource()` is guaranteed to return a non-nil `*resources.Resource` (the base case returns `resources.NewResource()`, so even unconfigured queues return an empty resource map rather than `nil`).
2. `SortAllocationsBasedOnAsk` is also nil-safe: `internalGetShares` explicitly handles `total == nil || total.Resources[k] == 0` by falling back to raw resource quantities, so there is no risk of panic.

HuangTing-Yao (approved):
LGTM

chenyulin0719:
One minor question for the time granularity, `SortAllocationsBasedOnAsk` is based on below ordering:
1. High allocation score first
2. Low priority first 
3. **Latest CreateTime first (In hour)** 
4. High resource usage ration first. 
5. **Latest CreateTime first (In sec)**。 (new,  come from the removed `compareAllocationLess` function. https://github.com/apache/yunikorn-core/blob/8bb5250e2b794fadcd4a7afba50a0727e064fcdc/pkg/scheduler/objects/preemption.go#L891) 


The original design of `in hour` comparison is from https://github.com/apache/yunikorn-core/pull/1072#issue-4023185205
To me, I'm good with adding the `in ms` comparison to improve the certainty, but also think it's might not be necessarily.

@manirajv06  What do you think?  Should we keep both granularity?

manirajv06:
I would prefer not to introduce those two new factors for now. Other than that, changes are good.

manirajv06 (changes requested):
Inline comments

PoiBlackTea:
> I would prefer not to introduce those two new factors for now. Other than that, changes are good.

Hi @manirajv06 
I understand your concern about the two new factors. Separately, I noticed comp == 0 → return true in SortAllocationsBasedOnAsk violates the antisymmetry property required for a strict weak ordering (used by sort.SliceStable). Since this also affects quota_preemptor.go beyond the scope of this PR, I'd appreciate your thoughts on how best to handle it.

manirajv06:
@PoiBlackTea Good to discuss about the properties and ensuring there is no violation.

Yes, we don't want to cause any violations. How about dropping the first one and keeping the second one? I see first one as more like a duplicate. That way, we ensure no such violations.

PR #1161, opened 2026-09-17
[YUNIKORN-3458] Premature victim loop termination in calculateVictimsByNode starves preemption

hedger9487:
### Description
In `calculateVictimsByNode()` (`pkg/scheduler/objects/preemption.go`), when a candidate victim exceeds `askQueueRemaining`, the loop executes `break`. Because candidate victims are sorted primarily by priority and creation time rather than resource size, an oversized candidate can appear before smaller, viable victims. The `break` prematurely aborts victim evaluation for the entire node, causing subsequent viable victims to be ignored and the preemptor ask to starve.

This PR changes `break` to `continue` so that subsequent viable victims on the node continue to be evaluated.

### Type of change
- [x] Bug Fix
- [ ] Improvement
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3458

### AI Tooling
If an AI tool was used:
- [x] The PR includes the phrase "Generated by Antigravity", where Antigravity is the name of the AI tool used.

### How has this been tested?
- [x] Added unit test `TestTryPreemption_PrematureVictimLoopTermination` in `preemption_test.go`.
- [x] Ran unit tests with race detection (`go test -race ./pkg/scheduler/objects/...`).

### Questions:

*Generated by hedger9487 with assistance from Antigravity.*


manirajv06:
Please fix the conflicts

> Because candidate victims are sorted primarily by priority and creation time rather than resource size, an oversized candidate can appear before smaller, viable victims.

Not anymore as https://github.com/apache/yunikorn-core/pull/1159/changes takes used resource also into account and improved the sorting really well. Still, it could happen. IIRC, Breaks are meant to stop and come out as and when we are done. So, Please ensure that we don't process victims all the way upto the end unnecessarily with the intent of searching "better, much better.." victim. There must be similar checks even in finding `extraVictims` too. Please check that as well.

hedger9487:
Hi @manirajv06,

Thanks for the review. Addressed your points below:

1. **Conflicts & Rebase**: Cleanly rebased onto master. 

2. **`break` Semantics in `calculateVictimsByNode`**:
   The `break` was inside the `else` branch when a candidate is too large to fit in `askQueueRemaining`:
   ```go
   if askQueueRemaining != nil && askQueueRemaining.FitInActual(...) {
       askQueue.AddAllocation(...)
   } else {
       queueSnapshot.AddAllocation(...)
       break // <-- triggered on fit failure, not on satisfaction
   }
   ```
   It did not exit when "done"; rather, it aborted on oversized victims and starved subsequent smaller, viable victims on the same node. Changing it to `continue` fixes this.

3. **Loop Termination**:
   The loop only processes allocations on the single node to build the candidate list (and set `StartIndex`). There is no unbounded search: batch predicate evaluation already stops early as soon as `bestResult.isSatisfactory()` is met (line 447).

4. **Identical check in `calculateAdditionalVictims`**:
   I will fix `calculateAdditionalVictims` in a follow-up Jira/PR right after this.

Please take another look. Thanks!

chenyulin0719:
This looks like a "perfect solution" vs. "good-enough solution" tradeoff rather than a clear-cut bug fix.

`continue` scans the entire potentialVictims list, trying to gather as many eligible victims as possible.
`break` stops as soon as it hits a victim that just barely exceeds the guaranteed resource limit.

When most of the pods are similar in size,  and node has many potential victims, `break` is better.  Conversely, `continue` is better.

manirajv06 (approved):
Since current preemption flow is about being generous in picking victims in the beginning and ensuring only required victims are getting killed at the later stage, I think it is ok to continue with "continue" approach itself :)

Victims are being collected based on the remaining guaranteed resource not really based on the exact ask requirement in the beginning stage. Ideally, it could be based on that as well but being very strict in this stage might not yield successful outcome as victim might get dropped during the course of action.

